PR #916, opened 2014-09-22
78911338 modify study report

emrojo:
Adds iframe comment to allocate the iframe when it's ready.
Adds cookie behaviour to distinguish between the newer (iframe) and the older version
for displaying the reports.
If the url has the param[:version] = new => iframe version
If the url has param[:version] = previous => previous reports.
In other cases, it will try to read from the session cookie study_report_testAB.
By default it will display the new iframe version.

Logs actual url, previous url and user agent per every access to the reports page, with a string
"StudyRequestReport",  the text NEW for the new iframe version, and DEPRECATED for the previous.


JamesGlover:
Keep the branch around, as we may want to re-use the tableau integration, but this branch has been deprecated.

PR #944, opened 2014-11-27
Changes column order name specification for mx tubes pool name

emrojo:


JamesGlover:
Mered into X10 branch

PR #963, opened 2015-01-19
Ss7 add submission validation options

JamesGlover:


JamesGlover:
Also contains a whole chenk of next release

PR #972, opened 2015-01-28
[#85886858] Fallback to empty hash for creation_options in the event of ...

JamesGlover:
...missing parent

I'm not certain about this fix, as it is a bit of a patch on bad data.
I'm going to be fixing the actual problem underlying out current issues,
which is the wrong purpose being assigned to MX library tubes from certain pipelines.

This fix does make things more pleasant for the users, but could result in more
serious underlying issues getting overlooked.

### DIFF
--- a/app/models/tube.rb
+++ b/app/models/tube.rb
@@ -78,7 +78,7 @@ class Tube < Aliquot::Receptacle
 
   class StandardMx < Tube::Purpose
     def created_with_request_options(tube)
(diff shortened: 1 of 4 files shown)


JamesGlover:
Had a bit of a chance to look at the various issues in the db at the moment, and it seems that we have a number of orphan mx library tubes which will benefit from these changes. While some are empty, some have been actively used in sequencing, as are probably worth preserving.

PR #1016, opened 2015-04-15
Ss263 equipment table

emrojo:
Is it correct with the barcode printing method added or should I remove it?

### DIFF
--- a/app/models/equipment.rb
+++ b/app/models/equipment.rb
@@ -21,8 +21,18 @@ class Equipment < ActiveRecord::Base
     Barcode.calculate_checksum(prefix, barcode_number)
   end
 
+  def printables
+    [PrintBarcode::Label.new({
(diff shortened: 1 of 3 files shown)


JamesGlover:
Couple of suggestions:

1) Take the barcode number from the barcode itself. You can use Barcode.number_to_human(ean13) to get that. Shouldn't make much of a difference, but might be useful if we start generating barcodes base on some other counter.
2) Pass in the name as the study, makes it easier to match the barcode up to what's being tagged.
3) Similar to 1, when you calculate the baroce in after_create, use ||= to set it, to allow us to set a custom one if needed.


Best Wishes,

James Glover

Senior Software Developer (PSD)
Ext. 4987
jg16@sanger.ac.uk

emrojo:
Ok, it's done

PR #1017, opened 2015-04-16
Ss253 cancel bulk submissions

emrojo:


app/models/plate_purpose/works_on_library_requests.rb
```diff
@@ -10,7 +10,8 @@ def self.included(base)
 
   def each_well_and_its_library_request(plate, &block)
     well_to_stock_id = Hash[plate.stock_wells.map { |well,stock_wells| [well.id, stock_wells.first.id] }]
-    requests         = Request::LibraryCreation.for_asset_id(well_to_stock_id.values).include_request_metadata.group_by(&:asset_id)
+    requests         = Request::LibraryCreation.for_asset_id(well_to_stock_id.values).excluding_states(['cancelled']).include_request_metadata.group_by(&:asset_id)
```
emrojo: Find where is this used

PR #1062, opened 2015-06-05
Fix add parent purpose to plate creators next release

emrojo:


emrojo:
This content will be included from the production pull request.

PR #1092, opened 2015-07-17
Changes tube manifest by a plate manifest for relevant RNA chip

emrojo:


emrojo:
This changes the RNA chip excel file with a plate version instead of the tube version. I'll merge it directly as it doesn't add any changes in functionality.

PR #1139, opened 2015-10-06
ss338 - Added new value 'Tissue' for DNA source

emrojo:


JamesGlover:
This will need to be done for the other manifests, (tube/library).

You'll also need to add the new type to the DNA_SOURCES array in sample.rb

PR #1175, opened 2015-11-24
Json pipeline serialization

emrojo:
Move all json serializers to lib/json_serializers


emrojo:
I'll close it in order to not hav pull requests to next_release

PR #1282, opened 2016-04-04
Ss432 ensure hiseqx request types are distinct

JamesGlover:


JamesGlover:
Pull directly into NR

PR #1298, opened 2016-04-12
Ss437 438 439 labwhere enhancement

Proskurina:


JamesGlover:
Included in: https://github.com/sanger/sequencescape/pull/1333 which also contains the fixes for my javascript.

PR #1299, opened 2016-04-12
Ss423 add program

emrojo:


emrojo:
This pull request is now updated with the following:
- Added tests for the creation and editing of programs
- Rake working:setup use programs to create studies
- has_many :through for looking for studies
- Destroy removed from programs

PR #1305, opened 2016-04-19
Bulk submission adds the assets mentioned in the csv to the study spe…

emrojo:
…cified in

it.
Test added to check the behaviour.


JamesGlover:
Sorry, the failing features on here are the result of my merging a broken branch into next release. Fixed in NR itself.

PR #1317, opened 2016-04-26
Ensure barcodes treated as strings

JamesGlover:
This is an old branch that I resurrected. I had thought it was already merged in.
MySQL is very inefficient when asked to search a string field with a numeric query.


JamesGlover:
Closed while I fix failing tests.

JamesGlover:
Okay, this branch was just dud.

PR #1322, opened 2016-05-05
Ss453 as lm want pull data dna rna from csv in drop folder

emrojo:


JamesGlover:
Discussed changes:
- Bring the new parser to the top of the parser_for method as it's likely to be the most common type
- Keep the update_values_for(asset) outside the parser class. (On plate/well) pass out concentration/volume/molarity/rin as hash, and perform appropriate lookup where methods are defined.
- Add in the support for volume/ RIN straight away.

emrojo:
Rin well_attribute creation is already in the pull request https://github.com/sanger/sequencescape/pull/1340, so I'm not adding it into this pull request

JamesGlover:
Grouped with other QE stuff on:
https://github.com/JamesGlover/sequencescape/pull/38
Once through UAT will merge in from there.

PR #1333, opened 2016-05-11
Ss437 438 439 labwhere enhancement

JamesGlover:
Yana's changes, pluss fixes for my javascript.


JamesGlover:
Merge following UAT

PR #1339, opened 2016-05-12
Changes order of setting of the buffer volume in tecan files

emrojo:


JamesGlover:
Feel free to merge in as soon as the tests pass

PR #1340, opened 2016-05-12
Ss452 as lb take stock and norm conc into sm qc reports

emrojo:


JamesGlover:
Discussed changes in https://github.com/JamesGlover/sequencescape/tree/ss452_as_lb_take_stock_and_norm_conc_into_sm_qc_reports merged into https://github.com/JamesGlover/sequencescape/tree/quant_essential_uat from which it will be deployed

PR #1351, opened 2016-06-09
Mysql57

JamesGlover:
Don't merge just yet, I'll get another release of our other bits out first.

This _should_ be backwards compatible with 5.6 and earlier.


JamesGlover:
We've just lost the dev database to the upgrade, so this will need to be merged in if we're going to be able to do anything.

app/models/request.rb
```diff
@@ -63,23 +65,25 @@ def validator_for(request_option)
           'INNER JOIN assets AS pw ON well_links.target_well_id=pw.id AND well_links.type="stock"',
         ]
       end
-    {
-      :select => 'uuids.external_id AS pool_id, GROUP_CONCAT(DISTINCT pw_location.description ORDER BY pw.map_id ASC SEPARATOR ",") AS pool_into, requests.*',
-      :joins => joins + [
+
+
+    select('uuids.external_id AS pool_id, GROUP_CONCAT(DISTINCT pw_location.description ORDER BY pw.map_id ASC SEPARATOR ",") AS pool_into, MIN(requests.id) AS id, MIN(requests.sti_type) AS sti_type, MIN(requests.submission_id) AS submission_id, MIN(requests.request_type_id) AS request_type_id').
```
emrojo: If the select is filtering to select the requests for the stock plate, what does it mean "MIN(requests.sti_type) AS sti_type"? Is it picking up the first name of the classes for the requests but in alphabetical order?

app/models/request.rb
```diff
@@ -89,27 +93,28 @@ def validator_for(request_option)
         ]
       end
 
-      select('min(uuids.external_id) AS group_id, GROUP_CONCAT(DISTINCT pw_location.description SEPARATOR ",") AS group_into, requests.*').
-      joins(joins + [
-        'INNER JOIN maps AS pw_location ON pw.map_id=pw_location.id',
+      select('min(uuids.external_id) AS group_id, GROUP_CONCAT(DISTINCT pw_location.description SEPARATOR ",") AS group_into, MIN(requests.id) AS id, MIN(requests.submission_id) AS submission_id, MIN(requests.request_type_id) AS request_type_id').
```
emrojo: This was in the previous code of Sequencescape, not in this branch.
About this: select('min(uuids.external_id) AS group_id
 I think after grouping the uuid is the same for all the group, and so min() will return the same uuid. However, if uuids.external_id is an UUID, are we looking for the oldest uuid?  Are UUIDs generated always in ascending order?

app/models/request.rb
JamesGlover: Essentially yes. Its a little arbitrary, as we SHOULD only have one value here, and could equally have used ANY_VALUE. We'll probably need to refactor this whole chunk soon, just didn't want to mess too much until after the switch.

app/models/request.rb
JamesGlover: In our case yes, we use uuid v1, which is timestamp based.

emrojo:
Ok, reviewed

PR #1355, opened 2016-06-22
Ss465 track i5 barcode oligo lots and do qc assay workflow

emrojo:


JamesGlover:
Actually forked this branch myself for UAT purposes. Allowed me to bring in the latest next_release and production changes. Those we could probably ignore, but I'm also trialling puma updates. If they work well, we can merge in from my branch.

PR #1363, opened 2016-07-11
changes in label printing (integration with print my barcode)

Proskurina:


app/controllers/barcode_printers_controller.rb
```diff
@@ -72,38 +72,38 @@ def destroy
...
-    rescue Savon::Error, Sanger::Barcode::Printing::BarcodeException => e
-      if e.kind_of? Savon::Error
-        flash[:warning] = "There is a problem with the selected printer. Please report it to Systems."
-      else
-        flash[:error] = "There was a problem with the printer. Select another and try again."
-      end
-      Rails.logger.error($!)
-
-      redirect_to failure_url
+  #   def print_asset_labels(succes_url, failure_url)
```
JamesGlover: Probably best to just delete code, rather then commenting it out, and remove it entirely. The old code still exists in Git, so can be retrieved if we need to restore it. You can still comment stuff out when actively working on it, just try and remove it before committing.

app/controllers/barcode_printers_controller.rb
```diff
@@ -72,38 +72,38 @@ def destroy
   end
   # This module define common behavior used by other controller to print things
   module Print
```
JamesGlover: This module is now empty. Remove it, and also any places where it gets included. Just makes it easier to follow things in future.

app/controllers/batches_controller.rb
```diff
@@ -432,113 +432,139 @@ def print_stock_multiplex_labels
...
+    print_job = LabelPrinter::PrintJob.new(params[:printer],
+                                        LabelPrinter::Label::BatchMultiplex,
+                                        count: params[:count], printable: params[:printable], batch: @batch)
+    if print_job.execute
+      flash[:notice] = print_job.success
+    else
+      flash[:error] = print_job.errors.full_messages.join('; ')
+    end
+
+    # printables = []
```
JamesGlover: Same as before. (Just assume the same goes for any block of commented code)

app/models/sample_manifest/core_behaviour.rb
```diff
@@ -37,5 +37,5 @@ def core_behaviour
     core = rapid_generation? ? 'RapidCore' : 'Core'
     @core_behaviour = "::SampleManifest::#{behaviour}::#{core}".constantize.new(self)
   end
```
JamesGlover: Be a bit wary of making public methods private. Instead think about what your actually wanting to expose and make that public. From looking through, it appears your mostly accessing the tubes/plates/samples. We'll step through how you might approach this one together, as I think it'll help clean up some of the code elsewhere as well.

app/models/sequenom_qc_plate.rb
```diff
@@ -77,6 +77,15 @@ def compute_and_set_name(input_plate_names)
     end
     true
   end
+
```
JamesGlover: Ahh, and good counterpoint to the earlier caution of making private stuff public. In this case its the right thing to do, as the print labels methods you removed quite rightly didn't belong here. So good.

PR #1365, opened 2016-07-21
Ss434 accession delayed studies

JamesGlover:


app/models/re_request_submission.rb
```diff
@@ -11,17 +11,4 @@ def is_asset_applicable_to_type?(request_type, asset)
   end
   private :is_asset_applicable_to_type?
 
-  def all_samples_have_accession_numbers?
```
JamesGlover: Because Multiplexed Libraries can be now registered via manifest, these assumptions are no longer valid.

JamesGlover:
Eeep, looks like I broke this. I'll fix it up tomorrow.

test/unit/sample_test.rb
```diff
@@ -11,6 +11,16 @@ def move(study_from, study_to, asset_group, new_assets_name, current_user, submi
   end
 end
 class SampleTest < ActiveSupport::TestCase
+    def assert_accession_service(type)
+      service = {
+        ega: EgaAccessionService,
+        ena: EraAccessionService,
```
emrojo: Is this a typo? ERA instead of ENA?

app/models/accession_service.rb
```diff
@@ -19,6 +19,14 @@ class AccessionedFile < File
     attr_accessor :original_filename
   end
 
+  class_attribute :priority, :instance_writer => false
+  class_attribute :no_study_accession_needed, :instance_writer => false
+  class_attribute :operational, :instance_writer => false
+
```
emrojo: Maybe we can add a comment for priority and operational to remember what they are used for. Priority is related with the priorities of other accession services and operational means that they are currently active, right?

app/models/asset_group.rb
```diff
@@ -14,23 +14,26 @@ class AssetGroup < ActiveRecord::Base
...
   def unaccessioned_samples
-    Sample.find(:all,
-      :joins => [
-        'INNER JOIN aliquots ON aliquots.sample_id = samples.id',
-        'INNER JOIN sample_metadata ON sample_metadata.sample_id = samples.id'
-        ],
-        :conditions => ['aliquots.receptacle_id IN (?) AND sample_ebi_accession_number IS NULL',assets.map(&:id)]
-    )
+    if new_record?
+      Sample.contained_in(assets).without_accession
```
emrojo: If asset_group_assets  are not stored, seems that assets and samples depend on it:
has_many :asset_group_assets
has_many :assets, :through => :asset_group_assets
has_many :samples, :through => :assets

What would happen in this case with Sample.contained_in(assets)?

JamesGlover:
Merge conflict resolved in:
https://github.com/sanger/sequencescape/pull/1378

JamesGlover:
Included with Steve's work.

PR #1368, opened 2016-07-27
Submission hotfixes

JamesGlover:


app/api/model_extensions/batch.rb
```diff
@@ -5,9 +5,6 @@
 module ModelExtensions::Batch
   def self.included(base)
     base.class_eval do
-      # These were in Batch but it makes more sense to keep them here for the moment
```
JamesGlover: Not really, no. I've moved them back onto batch.

app/models/aliquot.rb
```diff
@@ -9,6 +9,8 @@ class Aliquot < ActiveRecord::Base
   include Api::Messages::FlowcellIO::AliquotExtensions
   include AliquotIndexer::AliquotScopes
 
+  TAG_COUNT_NAMES = ['Untagged','Single','Dual']
+
```
JamesGlover: Tag number was confusing, as it was unclear if it applied to per sample, or per pool. I've opted for names instead, to make it more apparent.

app/controllers/bulk_submissions_controller.rb
```diff
@@ -12,10 +12,13 @@ def index
```
JamesGlover: Submission Templates have been added to the bulk submissions page as a list. I'll tweak this to be a before filter.

app/models/batch.rb
```diff
@@ -7,10 +7,28 @@
 require 'aasm'
 
 class Batch < ActiveRecord::Base
```
JamesGlover: Using proper rails relationships here.

app/models/role.rb
```diff
@@ -37,32 +37,21 @@ def self.included(base)
       base.extend(ClassMethods)
```
JamesGlover: Modernised the scopes. Rails wasn't able to order the old version properly as it couldn't generate the correct SQL

app/views/bulk_submissions/new.html.erb
```diff
@@ -12,7 +12,7 @@
       <%= form.input :spreadsheet, :as => :file, :label => 'Spreadsheet' %>
```
JamesGlover: Add list of templates, and prettyfy the page

app/views/pipelines/_requests.html.erb
```diff
@@ -19,14 +19,13 @@
         <th><%= link_to "Request ID", "javascript:void(0);" %></th>
         <th><%= link_to "Name", "javascript:void(0);" %></th>
         <th><%= link_to "Request comments", "javascript:void(0);" %></th>
-        <th><%= link_to "Asset type", "javascript:void(0);" %></th>
```
JamesGlover: Asset type is now redundant and misleading. I've removed it.

app/models/role.rb
```diff
@@ -37,32 +37,21 @@ def self.included(base)
...
-       scope :with_related_users_included, -> { includes(:roles => :users ) }
-       scope :of_interest_to, ->(user) {
-          {
-            :joins => joins_through_to_users,
-            :conditions => ['rj_u.id=?', user.id],
-            :group => "rj_r.authorizable_id"
-          }
+        scope :with_related_users_included, -> { includes(:roles => :users ) }
+        scope :of_interest_to, ->(user) {
+          joins(:users).where(users:{id:user}).uniq
```
emrojo: I think it should be 'user.id' like: joins(:users).where(users:{id:user.id}).uniq

emrojo: Ok, it is converted by rails automatically.

PR #1426, opened 2016-11-02
[ss537] - Add DynamicOption for run-time option generation

JamesGlover:
[ss537] - Ranges with dynamic options do not cache list and last columns

[ss537] - Add library type dropdown to manifests


Proskurina:
updated:
https://github.com/sanger/sequencescape/pull/1437

PR #1492, opened 2017-01-31
Ss611 stamping of stock

Proskurina:


app/views/robot_verifications/submission.html.erb
```diff
@@ -34,7 +34,7 @@
          <%   next_field = "bed_barcodes_#{bed_label.to_i + 1}" %>
         <% end %>
         <td><%= text_field_tag "plate_barcodes[#{plate_label}]", display_hash_value(params, :plate_barcodes, "#{plate_label}"), :onkeypress => "return disable_cr_and_change_focus(event,'plate_barcodes_#{plate_label}','#{next_field}')" %></td>
-        <td><%= select_tag("source_plate_types[#{plate_label}]", options_for_select(PlatePurpose.cherrypickable_as_source.map(&:name))) %></td>
+        <td><%= select_tag("source_plate_types[#{plate_label}]", options_for_select(PlateType.all.map(&:name))) %></td>
```
JamesGlover: PlateType.pluck(:name) may be better here. It has two advantages:
1) It generates SELECT name FROM plate_types;
2) It creates an array of names, without first instantiating plate types.
The benefits of this are mainly seen with larger numbers of more complicated objects, so won't have a huge impact here, but it can be a good habit to get into.
http://apidock.com/rails/ActiveRecord/Calculations/pluck

app/views/workflows/_fluidigm_template_batches.html.erb
```diff
@@ -58,7 +58,7 @@
         <table cellspacing='0' cellpadding='0'>
           <tr>
             <td><label for="plate_type">Source Plates Type</label>:</td>
-            <td><%= select_tag "plate_type", options_for_select(PlatePurpose.cherrypickable_as_source.map(&:name)) %></td>
+            <td><%= select_tag "plate_type", options_for_select(PlateType.all.map(&:name)) %></td>
           </tr>
```
JamesGlover: See previous column re. pluck

PR #1493, opened 2017-02-02
Ss618 hmdmc number is required to create managed study

Proskurina:


spec/models/study_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+require 'pry'
+
+describe Study do
+  let(:study) { Study.new(study_metadata_attributes: { data_release_strategy: "managed" }) }
+
+  it 'should not be valid without hmdmc number if managed' do
+    expect(study.valid?).to be false
+    expect(study.errors.messages.length).to eq 11
+    expect(study.errors.full_messages).to include "Study metadata hmdmc approval number can't be blank"
+  end
```
JamesGlover: I'd suggest adding a test 'it should be valid without hmdmc number if open'

PR #1570, opened 2017-06-07
Lcbm request types and submission templates

JamesGlover:
Begins with documentation improvements. I thought at this point it was done.

e156886 - In generic we auto trigger pooling. This doesn't work here as we actually pass through a normalization tube before reaching the final mx tube. The generic Lims stuff becomes its own class with this behaviour.

08d77fa - Add transfer request collections. Includes modifications to the API itself to support creation of resources through nested_attributes. (Essentially just removing the requirement that assets already exist.)

 f4c0998 - 8af56fb - Adding to api and exposing on plate
 
9f6e88d - Sets up eager loading. Including the target tube association isn't ideal, but avoids some cumbersome and slow means of handling some limitations in the api when it comes to polymorphic associations.


app/models/transfer_request.rb
```diff
@@ -59,12 +59,20 @@ class TransferRequest < SystemRequest
     end
   end
 
-  before_create(:add_request_type)
+  before_validation :add_request_type, :set_request_purpose
+
+  has_many :transfer_request_collection_transfer_requests
+  has_many :transfer_request_collections, through: :transfer_request_collection_transfer_requests
+
   def add_request_type
```
Proskurina: Create a concern to handle dynamic defaults and standardise the way defaults are created.

app/models/transfer_request.rb
```diff
@@ -59,12 +59,20 @@ class TransferRequest < SystemRequest
...
     self.request_type ||= RequestType.transfer
   end
   private :add_request_type
 
+  def set_request_purpose
+    self.request_purpose ||= request_type.request_purpose
+  end
+
   after_create(:perform_transfer_of_contents)
```
Proskurina: Move callbacks to the top

lib/limber/helper.rb
```diff
@@ -70,14 +70,6 @@ def build!
   end
 
   class TemplateConstructor
```
Proskurina: Use ActiveModel::Model here

PR #1571, opened 2017-06-07
Asset creators allow naming

JamesGlover:
Just on specific tube creators for the time being.
Accept nested attributes, rather than just name.


JamesGlover:
Wrong target branch

PR #1572, opened 2017-06-07
Asset creators allow naming

JamesGlover:
Just on specific tube creators for the time being.
Accept nested attributes, rather than just name.


app/models/specific_tube_creation.rb
```diff
@@ -16,6 +16,8 @@ class ChildPurpose < ActiveRecord::Base
 
   validates_presence_of :child_purposes
```
Proskurina: what is going on?

JamesGlover:
Closed as now merged in to next-release as hotfix.

PR #1577, opened 2017-06-15
Address travis build times

JamesGlover:
This pull request addresses the sever performance issues on travis. As of commit a701190 the execution time of the entire suite was half of that in next release. 453d773 addresses some issues with concurrency when running the javascript cukes, and thus may be expected to have minor performance impacts. (I'll update this once the tests have finished running)

Summary of changes:
- Switch the cucumber formatter to provide more diagnostic information about cuke performance.
- Add haveged to maintain a sufficient source of entropy. See: https://docs.travis-ci.com/user/languages/ruby/#JRuby%3A-Installation-takes-a-long-time-on-Standard-and-Trusty-environments and https://github.com/travis-ci/travis-ci/issues/6892
- Switch to trusty. I wasn't seeing the performance improvement in the precise environment.
- Enable sudo in travis.yml. The mysql service fails to start in the containerized trusty environments. This will probably be fixed shortly, as trusty will soon be the default environment.
- Strip out unneeded tests
- Fix issue that arrose with tag group names
- While inspecting the logs, the assert nil warnings indicated that the tecan tests were not functioning as intended. I switched to round to ensure consistent behaviour across Jruby and MRI, drastically simplifying the tests.
- Finally under the new environment the MRI builds were triggering concurrency issues with greater regulatity. This is a result of database connections being recycles while still in use for ajax requests. We're already using the transactional_capybara gem in rspec, so ported this across to cucumber. (Given I was fgixing performance issues, I wasn't about to switch to truncation)


JamesGlover:
Looks like I need to move transactional_capybara into another group. It works locally, but not on travis.

PR #1584, opened 2017-06-21
Mri development improvements

JamesGlover:


JamesGlover:
Setting rubocop to use 2.4 puts its standards ahead of the jruby support. Will hold back on this change untill we're sure we're happy.

PR #1602, opened 2017-07-18
Rails 5

Proskurina:


Proskurina:
Main changes:
- changed `uniq` to `distinct`
- As params do not inherit from Hash, added  `to_unsafe_h` or `to_h` in controllers
- As relations do not inherit from Array, add `to_a` in some places
- In controllers tests named parameters are used now
- Changed `redirect_to :back` to `redirect_back fallback_location: root_path`
- `alias_method_chain` deprecated, so used two `alias_method` 's for now
- `render :text` deprecated, used `render :plain ` or `render :xml ` depending on context
- if empty hash is passed from tests to controller as a parameter, it is received is nil, so added empty hash where required
- observers were not picked up properly, fixed CustomerRequestObserver, others need to be checked (if we use them)

Proskurina:
To do:
- have a look at how lib files are picked up
- check if observers work correctly

JamesGlover:
Merged in with NR and in UAT in separate branch

PR #1676, opened 2017-10-12
Seq 437 remove fields from sample manifest

stevieing:


stevieing:
This is not backward compatible. If a user tries to upload an old sample manifest it will throw a validation error which will say Columns IS SAMPLE A CONTROL?,IS RE-SUBMITTED SAMPLE?,SAMPLE VISIBILITY are not valid. SSRs are fine to remove them as needed. We can also ensure that we go live with when pending manifests are at a minimum.

PR #1696, opened 2017-11-01
Seq532 give warning for submission if samples do not have metadata

Proskurina:


app/models/order.rb
```diff
@@ -352,6 +352,11 @@ def study_is_active
     end
   end
 
+  # returns an array of samples, that potentially can not be included in submission
+  def unready_samples
```
Proskurina: pending

PR #1710, opened 2017-11-20
Seq 536 sc rna templates and request types plus rubocopper

JamesGlover:


JamesGlover:
This is just Rubocoppery. Not worth resolving merge conflicts.

PR #1729, opened 2017-12-22
Seq 568 metaprog sample

radome:


JamesGlover:
Closing as December Work is now in next_release

PR #1731, opened 2018-01-04
Sqs567 split transfer request from request

JamesGlover:


app/models/illumina_htp/requests.rb
```diff
@@ -4,62 +4,6 @@
 # authorship of this file.
 # Copyright (C) 2013,2014,2015,2016 Genome Research Ltd.
 
+# Mdule for namespacing ITHP requests
 module IlluminaHtp::Requests
-  class StdLibraryRequest < Request::LibraryCreation
```
JamesGlover: Moved this code into separate files

app/models/plate.rb
```diff
@@ -210,53 +221,14 @@ def iteration
     iter.zero? ? nil : iter # Maintains compatibility with legacy version
   end
 
-  class CommentsProxy
```
JamesGlover: THis has been moved. And a comment has been added to explain why this monstrosity exists.

app/controllers/api/assets_controller.rb
```diff
@@ -16,12 +16,4 @@ def parents
       format.json { render json: @object.parents.map(&:list_json) }
     end
   end
-
```
JamesGlover: Unused

app/controllers/assets_controller.rb
```diff
@@ -254,6 +254,7 @@ def print_assets
   end
 
   def show_plate
```
JamesGlover: Means we get a 404 if someone accesses a non-plate at show plate, rather than a 500.

app/controllers/workflows_controller.rb
```diff
@@ -14,20 +14,17 @@ class WorkflowsController < ApplicationController
```
JamesGlover: Associated with unused tasks. Corresponding tasks have all been removed from the database.

app/models/accession_service.rb
```diff
@@ -162,55 +162,6 @@ def private?
 
   private
```
JamesGlover: Unused.

app/models/aliquot/remover.rb
```diff
@@ -42,6 +45,9 @@ def remove_downstream_aliquots
   end
 
   def on_downstream_aliquots(aliquots_to_remove)
+    transfer_requests_as_source.each do |request|
+      request.target_asset.process_aliquots(aliquots_to_remove)
+    end
```
JamesGlover: Bit of duplication here, but one of the few places where requests and transfer requests get treated the same.

app/models/asset.rb
```diff
@@ -237,16 +237,6 @@ def self.find_from_machine_barcode(source_barcode)
     with_machine_barcode(source_barcode).first
   end
 
-  def self.find_by_human_barcode(barcode, location)
```
JamesGlover: Unused

app/models/assign_tags_to_wells_task.rb
```diff
@@ -1,219 +0,0 @@
-# This file is part of SEQUENCESCAPE; it is distributed under the terms of
```
JamesGlover: Unused

app/models/cherrypick/task/pick_by_micro_litre.rb
```diff
@@ -5,11 +5,6 @@
 # Copyright (C) 2007-2011,2012,2015 Genome Research Ltd.
 
 module Cherrypick::Task::PickByMicroLitre
-  def pick_by_micro_litre(*args)
```
JamesGlover: This code was only used by the cherrypick by pulldown pipeline, which has been deprecated, and the associated code removed. Same for the similar methods in the next few files.

app/models/cherrypick_request.rb
```diff
@@ -5,74 +5,35 @@
...
+    new_volume = asset.get_current_volume - subtracted_volume
+    asset.set_current_volume(new_volume)
   end
 
-  alias_method :on_cancelled, :on_failed
+  private
 
-  def perform_transfer_of_contents
-    transfer_aliquots # Ensures we set the study/project
+  # The transfer requests handle the actual transfer
```
JamesGlover: So the creation of transfer requests may seem a little redundant, but is a limitation in the way state is tracked in SS that we probably need to address. Essentially well state exists on the transfer requests into the well.This allows individual well failure, and I think its position on request, rather than the well (different wells on the same plate can have different states) itself was an attempt to consolidate state machines in one location, and to have everything be request driven. When Generic Lims was implimented there was a need to fail wells on the stock plate, before any work had been completed. At this point, cherypick requests became a subclass of transfer request. This isn't great, and we probably need to completely rethink how sll this gets modeled.

app/models/illumina_htp/plate_purposes.rb
```diff
@@ -85,21 +85,13 @@ module IlluminaHtp::PlatePurposes
...
 
-  PF_PLATE_PURPOSES_TO_REQUEST_CLASS_NAMES = [
-    ['PF Cherrypicked', 'PF Shear', :initial],
-    ['PF Shear', 'PF Post Shear'],
-    ['PF Post Shear', 'PF Post Shear XP'],
-    ['PF Post Shear XP', 'PF Lib XP'],
-    ['PF Lib XP', 'PF Lib XP2']
-  ]
-
   PLATE_PURPOSES_TO_REQUEST_CLASS_NAMES = [
```
JamesGlover: This is all stuff that would probably better belong in a yaml file with a corresponding factory, as its pretty much only used when seeding the databse. Part of the seeds project perhaps.

app/models/pac_bio_sample_prep_request.rb
```diff
@@ -26,14 +26,14 @@ def on_started
   end
```
JamesGlover: I'm aware of the slight change in behaviour. It allows the un-needed plate creation stage of legacy PacBio to be switched off without issue. The existing pipeline generate an RT ticket pretty much the same time as I was fixing the final method in this class, so I overhauled the whole thing.

app/models/pac_bio_sample_prep_request/initial.rb
```diff
@@ -1,8 +0,0 @@
-require_dependency 'pac_bio_sample_prep_request'
```
JamesGlover: Re-name spaced all the transfer request stuff.

app/models/plate.rb
```diff
@@ -146,10 +154,14 @@ def source_plate
       ]
     )
   }
+  scope :with_plate_purpose, ->(*purposes) { where(plate_purpose_id: purposes.flatten) }
 
   # About 10x faster than going through the wells
   def submission_ids
     @siat ||= container_associations
+              .joins('LEFT JOIN transfer_requests ON transfer_requests.target_asset_id = container_associations.content_id')
```
JamesGlover: I'm not sure this actually needs to be this complicated. This maintains existing behaviour, but we might be able to get away with something simpler.

PR #1741, opened 2018-02-06
Seq 581 add events

radome:
Added "Sequencing Complete" event to notify the Event Warehouse.


app/controllers/npg_actions/assets_controller.rb
```diff
@@ -19,50 +19,51 @@ class NpgActions::AssetsController < ApplicationController
...
+      create_method.call(params[:qc_information][:message] || 'No reason given')
+
+      request = @asset.source_request
+      batch = request.batch
+      raise ActiveRecord::RecordNotFound, 'Unable to find a batch for the Request' if (batch.nil?)
+
+      message = "#{state}ed manual QC".capitalize
+      send_method.call(request.id, '', message, '', 'npg', need_to_know_exceptions: true)
+
+      batch.npg_set_state if (state == 'pass')
```
radome: The `if` clause in this line doesn't seems right. However, removing the statement will fail the following test:
https://github.com/sanger/sequencescape/blob/879fec67464bf93c8e0cafefe8b62e9b4f2ae567/features/api/npg/5600990_npg_xml_asset_qc_state.feature#L6-L11

PR #1775, opened 2018-04-04
Seq623 locn report faculty sponsor

andrewsparkes:
Refactoring of location reports to allow selection by faculty sponsor (owner of the plates).

Moved view validations into a form object to better isolate them from the location report model.
Moved the barcode list entry box out of its own form across into the general selection form. Later (ticket 624) we will allow selection by LabWhere location and this will go into a separate form. You can still select plates from a list of barcodes but you can now combine this with other selection criteria if you choose to.


andrewsparkes:
This code is merged into the Seq624 pull request (see https://github.com/sanger/sequencescape/pull/1780#) and so this pull request does not need to be merged separately itself.

PR #1791, opened 2018-05-04
Seq636 manifests foreign bcs

andrewsparkes:
Added CGAP barcode format and the ability to upload foreign (CGAP) barcodes in both plate and tube sample manifests from the Tag Sequences upload view.

N.B. This code is folded into pull request 'Seq655 manifests overrides', which is in turn folded into pull request 'Seq656 manifests indexes'. All three are in the same UAT. Merge and deploy the Seq656 pull request when passed UAT, not this one.


app/sample_manifest_excel/sample_manifest_excel/upload/processor/base.rb
```diff
@@ -0,0 +1,98 @@
...
+          col_num = row.columns.find_column_or_null(:name, 'sanger_tube_id').number
+          return unless col_num.present? && col_num.positive?
+          curr_bc = row.data[col_num - 1]
+          if uniques.key?(curr_bc)
+            duplicates << curr_bc
+          else
+            uniques[curr_bc] = 1
+          end
+        end
+      end
```
stevieing: I think these two methods can be refactored to be much more concise as a single method find_duplicate_barcodes which would return true or false. This would use Array methods to construct an array of barcodes. If the barcode is nil don't add it if the barcode is already included in the Array of barcodes then return true if it gets to the end return false.
Happy to help on Tuesday if you get stuck.

PR #1841, opened 2018-06-15
A254 update materials

emrojo:


app/aker/factories/container.rb
```diff
@@ -23,12 +23,48 @@ def initialize(params)
 
       def create
         return unless valid?
-        @model = Aker::Container.find_or_create_by(barcode: barcode, address: address)
+        @model = Aker::Container.find_by(barcode: barcode, address: address)
```
stevieing: Did you know that find_or_create_by can accept a block?

app/models/aker/mapping.rb
```diff
@@ -0,0 +1,100 @@
...
+      end
+    end
+
+    def update!(attrs)
+      _each_model_and_setting_attrs_for(attrs) do |model, setting_attrs|
+        model.update!(setting_attrs)
+      end
+    end
+
+    def attributes
```
stevieing: You can use tap.

app/controllers/api/v2/aker/jobs_controller.rb
```diff
@@ -9,13 +9,22 @@ class JobsController < ApplicationController
...
 
         def create
-          @job = ::Aker::Factories::Job.new(params[:job].permit!)
-          if @job.valid?
-
-            @job.create
-            render json: @job, status: :created
-          else
-            render json: @job.errors, status: :unprocessable_entity
+          param_jobs = params[:data]
```
stevieing: I get the fact that you need to create multiple jobs but you are building a list of new jobs, then you are checking whether each of those jobs are valid, then you are creating them and if there is an error then you are looping through them again. How many jobs are you looking to create at a time? Will this create a performance issue. Maybe I am misunderstanding and I don't know what the answer is...

spec/aker/models/material_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+      it 'updates an attribute' do
+        expect(sample.sample_metadata.gender).to eq('Male')
+        mapping.update(gender: 'Female')
+        sample.sample_metadata.reload
+        expect(sample.sample_metadata.gender).to eq('Female')
+      end
+    end
+
+    context 'with private methods' do
```
stevieing: You are testing private methods rather than behaviour. Is there any reason for this?

app/models/aker/mapping.rb
emrojo: Ok, done

spec/aker/models/material_spec.rb
emrojo: Just to be sure that all methods do what they have to. Is it ok if I leave them or do you suggest to change it in some way?

PR #1869, opened 2018-07-24
Working dilution quant will now update parent plate. Downcased qc res…

stevieing:
…ult keys.

### DIFF
--- a/app/models/qc_result_factory.rb
+++ b/app/models/qc_result_factory.rb
@@ -27,7 +27,9 @@ class QcResultFactory
 
   def save
     return false unless valid?
-    resources.collect(&:save)
+    ActiveRecord::Base.transaction do
+      resources.collect(&:save)
+    end
     true
   end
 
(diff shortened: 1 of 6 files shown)


app/models/qc_result_factory.rb
```diff
@@ -86,18 +90,22 @@ def save
...
+      plate.plate_purpose.name == 'Working Dilution'
     end
 
     def concentration?
       key == 'concentration'
     end
 
+    def can_update_parent_well?
+      working_dilution? && concentration? && well_location.present? && parent_plate.dilution_factor.present?
+    end
```
JamesGlover: Dilution factor get set on the working dilution plate, not the stock plate.

app/models/qc_result_factory.rb
```diff
@@ -86,18 +90,22 @@ def save
     end
 
     def working_dilution?
-      assay_type == 'Working dilution'
+      return false if plate.blank?
+      plate.plate_purpose.name == 'Working Dilution'
```
JamesGlover: WorkingDilution plates actually have a different class, so this could be used to avoid the need for a magic string. (Which would also allow you just to use the :dilution_plate_purpose factory in the specs)

PR #1895, opened 2018-09-07
Bulk submission excel

JamesGlover:


app/models/order.rb
```diff
@@ -16,50 +31,12 @@ def complete_building
...
-        display_name: display_name,
-        key: key,
-        default_value: default,
-        kind: kind
-      }
-      values.update(selection: options) if selection?
-      FieldInfo.new(values)
-    end
-  end
-
```
JamesGlover: This behaviour has been integrated in field infos to allow them to be combined with &

PR #1915, opened 2018-10-03
Seq 786: Print Swipecard as Barcode

radome:
- Added `print_swipecard.js` to POST directly to Print My Barcode from the client
- Requires `swipecard_barcode_template` in Print My Barcode, the rake task for that is in `add_templates_and_printers_to_pmb.rake`
- Requires sequencescape domain added to Print My Barcode CORS


JamesGlover:
We'll wait until the process-tracker/asset audits stuff is done before including this to avoid confusion.

PR #1978, opened 2018-11-16
Seq 815 long read submission

stevieing:


app/models/request/long_read.rb
```diff
@@ -0,0 +1,27 @@
...
+  class LongRead < CustomerRequest
+    self.sequencing = true
+
+    has_metadata as: Request do
+      custom_attribute(:library_type, required: true, validator: true, selection: true)
+    end
+
+    validates :state, presence: true
+
+    destroy_aasm
```
JamesGlover: It might be unnecessary to do this in the end, but we can make that decision later.

PR #2013, opened 2018-12-18
Seq839 update manifest naming

JamesGlover:


JamesGlover:
Should probably UAT this with the other manifest work

PR #2016, opened 2018-12-19
Qc measurements update legacy tube columns

JamesGlover:


JamesGlover:
Tempted to suggest we just merge this one in, rather than holding off to the UAT. It'll sit there doing nothing until the limber side of things goes out.

PR #2017, opened 2018-12-19
Fix submission_pool for cancelled submissions

JamesGlover:
Cancelled submissions have no active requests
so can't find an outer request to get the plate count
from. Fortunately fo us, no requests means no plates.


JamesGlover:
Just doing a bit of spring cleaning and pulling in branches that got missed

JamesGlover:
Last failure was a test order issue, but fixed it up anyway.

PR #2075, opened 2019-02-18
Hotfix excessive join generation

JamesGlover:


app/models/container_association.rb
```diff
@@ -1,18 +1,6 @@
...
-  # Plate.first.wells.includes(:plate)
-  # generates invalid sql, where it attempts to eager_load the container associations. It tries to filter
-  # on assets.sti_type, but doesn't join on the assets table.
-  # Here we force the join, and ensure everything works as intended. It is also possible to switch to
-  # specifying the class_name: 'Asset' on each association, but this has the side effect of disrupting
-  # other eager loading as associations aren't specified on Asset.
-  # Ideally we should be avoiding using single table inheritance in this manner; this is far from the only
-  # place it has caused pain.
-  default_scope ->() { includes(:plate, :well) }
-
```
JamesGlover: This has been removed as I've moved the same scope onto the relationships in plate and well. This ensures we only load the relationship appropriate to the direction were traversing in. (Ie. wells only load the plate when finding the plate, rather than also re-loading themselves)

app/models/asset.rb
```diff
@@ -148,7 +148,7 @@ class VolumeError < StandardError
   # We accept not only an individual barcode but also an array of them.
   scope :with_barcode, ->(*barcodes) {
     db_barcodes = Barcode.extract_barcodes(barcodes)
-    includes(:barcodes).where(barcodes: { barcode: db_barcodes }).distinct
+    joins(:barcodes).where(barcodes: { barcode: db_barcodes }).distinct
   }
```
JamesGlover: This is the main fix, with a similar change elsewhere:
1) By default rails converts an include to a 'preload' which essentially results in separate queries.
2) However if there is a condition that covers the include, rails converts it to a join.
3) Unfortunately rails also converts ALL other includes to joins. Which is a problem if you have lots of aliquots and requests.
4) Perhaps unsurprisingly rails is actually in the process of adjusting this behaviour, forcing you to be a bit more explicit. I *think* the previous behaviour we had is actually deprecated in the latest version, although some of the discussion I could find was slightly contradictory.

PR #2090, opened 2019-02-26
Allow renaming of studies

JamesGlover:
- Add a study name field to the manage study page
- Validate uniqueness of study name on update and create

Note: This renders 22 Studies in the Sequencescape database invalid
as the have non-unique names. THis will need to be addressed
before this story can go live. SSRs have been emailed the following
(Study names and ids removed)

Hello all,

I’m currently adding the ability to rename studies via the ‘manage’ page in Sequencescape. As part of this I have discovered a number of studies which share identical names:

<snipped>List of ids and names of duplicate studies

I suspect many of these have been created in error, and will definitely be very difficult to work with in Sequencescape. Once we’ve added the ability to rename studies, you will also observe further issues, as Sequencescape will see the studies as invalid.

For each set of studies we have the following options:

1) Attempt to merge them together. In this case, please let us know which study we should merge things into.
	- We’ll keep the old study around, but add ‘(Legacy)’ to the end, a comment on the study will record what happened.
2) Differentiate them by renaming one or both studies. (Eg. Adding a year)
3) Abandon the plans to let you rename studies.

Leaving them unchanged , without taking option 3, is unfortunately not an option. If the studies are invalid it will end up having an impact on everything associated with them, as validation issues tend to cascade and affect anything associated with them, even sometimes indirectly. (eg. Submissions, samples, requests, plates, study reports, possibly even lanes associated with the study)

### DIFF
--- a/app/views/admin/studies/_edit.html.erb
+++ b/app/views/admin/studies/_edit.html.erb
@@ -6,32 +6,20 @@
       Study <%= study.id %> &middot; Created on <%= study.created_at.to_formatted_s(:long) %> &middot; <%= study.state.capitalize %>
     </div>
 
-    <div class='row'>
-      <div class='col-md-6'>
-        <%= panel(:info,title:'Study state') do %>
-          <%= form_group do %>
-            <%= form.label :state, "State", class: 'col-md-4' %>
-            <div class='col-md-8'>
-              <%= form.select :state, Study.aasm.states_for_select, {}, {class: 'custom-select'} %>
-            </div>
-          <% end %>
+    <div class='columnize'>
+      <%= panel(:info,title:'Study state') do %>
+        <%= form_group do %>
+          <%= form.label :state, "State", class: 'col-md-4' %>
+          <div class='col-md-8'>
+            <%= form.select :state, Study.aasm.states_for_select, {}, {class: 'custom-select'} %>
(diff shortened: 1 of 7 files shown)


JamesGlover:
SSRs have confirmed that studies can be renamed. NS has renamed them.

app/views/admin/studies/_edit.html.erb
```diff
@@ -6,33 +6,21 @@
...
-        <%= panel(:info,title:'Contacts') do %>
-            <%= link_to "View Study Contacts", collaborators_study_path(study) %>
-        <% end %>
-      </div>
+      <%= panel(:info,title:'Contacts') do %>
+          <%= link_to "View Study Contacts", collaborators_study_path(study) %>
+      <% end %>
     </div>
 
-    <%= panel(:info,title:'Study information') do %>
```
JamesGlover: This has been moved into the same area as the rest of the study options.

app/views/studies/_study.html.erb
```diff
@@ -27,12 +27,6 @@
   </div>
 <% end %>
 
-<%= fields_for(study) do |form| %>
```
JamesGlover: Moved into the shared form with the rest of the study properties.

PR #2095, opened 2019-03-01
Hotfix -Set empty supplier sample name flag

JamesGlover:
While we hope to be rid of this shortly, and it doesn't affect newly
generated manifests, older manifests still have the flag set
incorrectly on upload. This prevents samples from appearing
in reports, and some lab processes.


JamesGlover:
So not a problem with new manifests, as the samples are generated on upload and the flag defaults to false. But pre-existing manifests seem to set the flag to true on manifest generation, so it never gets unset.

Obviously this fix is only temporary, and can be removed once we've migrated historical data.

PR #2102, opened 2019-03-12
Seq 937 allow n tags

JamesGlover:
Draft pull request as am fixing further bugs:
- i5 only aliquots not flagged as 'library_index'

Also discovered a few other issues based on the use of the tagged? and untagged? methods.


JamesGlover:
Commit messages are quite detailed here, and govern each change.

PR #2107, opened 2019-03-15
Rails 5 2 upgrade

JamesGlover:
Was surprisingly simple. The main issues were:

1 - The capybara update was causing text matching to become sensitive to whitespace differences.
2- Scopes now detect potential security risks, and don't like raw sql getting passed in without clear flagging. This is currently a deprecation warning, I fixed up all issues in the tests, but we'll need to keep an eye on things in production.
3- Puma is getting a bit angry about our time traveling. Switching out time-cop in the specs relieved the problems, but they still seem to crop up in the cukes. I adjusted a few of the dates to relieve the issue, and hopefully it'll be fixed before it causes us problems again. Either way, this is a testing issue, not a production issue.
4- Minitest started trying to run after the cukes again.

I've had to monkey patch cucumber a tiny bit as well. The latest version fixes the issue, but we'll need to update the Transformation stuff to the replacement. Unfortunately there isn't a simple migration path for this, as the alternative gets added in the same version as they remove the old version. Just having a quick poke into this now to see how much work is involved.


radome:
Duplicate of #2117

PR #2157, opened 2019-05-24
Code cleanup and removal

JamesGlover:


JamesGlover:
Note: Migrations run against training. Took ~45m
Integration suite works fine

PR #2212, opened 2019-07-09
Gpl010 asset refactor continued

JamesGlover:
Tests passing in standard mode, and refactor mode (except for view contracts)


JamesGlover:
Extends on: https://github.com/sanger/sequencescape/pull/2192
Allow one release with those changes before merging this.

PR #2261, opened 2019-08-01
Sample manifest performance

JamesGlover:


JamesGlover:
Now part of https://github.com/sanger/sequencescape/pull/2264

Issue #2287, opened 2019-08-13
SEQ914 - Migrate all columns to UTF8


JamesGlover:
As a developer I would like to convert out database character encoding to utf8mb4 (utf8 isn't actually utf8 in mysql world) to increase supported user input, and move into this decade

"Caveats
Bit of experimentation reveals that his may not be entirely straightforward due to the way MSQL handles column limits. My quick and dirty exploratory approach suggests that MySQL mixes up byte limits and character limits, so heavily restrict column sizes for variable byte-length character sets. It was just a quick and dirty approach though, so might be wrong. or possibly it just affects indexed columns.

Acceptance: 
- All varchar/text etc. columns updated.
- Table and database defaults updated. 
- No data integrity issues post migration. (Need to validate on production like database)"

Tackle: https://github.com/sanger/sequencescape/issues/2286 to update high priority columns.

JamesGlover:
https://dev.mysql.com/doc/refman/5.7/en/charset-unicode-utf8mb4.html

JamesGlover:
Andrew code reviewed. Will hold off merge until I'm building the release, as see little reason to push the initial migrations out earlier now.

PR #2290, opened 2019-08-15
Seq914 migrate text columns to utf8

### LINKED ISSUE #2286: SEQ915 - UTF8 Encoding on TEXT columns
As a developer I would like to convert our database character encoding to utf8mb4 (utf 8 isn't actually utf8 in mysql world) on the key affected TEXT columns (mostly in study and sample metadata) to increase supported user input, and move into this decade, without incurring the additional costs of converting everything

A complete UTF8 migration may be tricky, as preliminary investigation suggests that it has an impact on column size restrictions.

"Caveats: May still run into the size issues above. 

Acceptance: 
- As many TEXT columns updated as possible.
- Defaults for new tables/columns updated (To avoid issues in future)
- No data integrity issues post migration. (Need to validate on production like database)"

JamesGlover:
Fixes #2286 
The restrictions in #2287 don't appear to be an issue anymore, although I'll run the migrations on one of the DBA managed servers to confirm. As best I can tell the index length issues I ran into when doing the preliminary research a while back are fixed in MySQL 5.7 when using update InnoDB filesystems (Barracuda), this all appears to be in place.

As a result I've actually managed to fix whole tables, including studies and samples. Main restrictions on updating the whole database are because the changes are locking.


JamesGlover:
This looks huge, but most of it is auto-generated migrations. I did one migration per table, as otherwise it becomes a massive pain to recover from if something goes wrong.

JamesGlover:
Hold off on merge, I'll co-ordinate it as part of the asset refactor.

PR #2305, opened 2019-08-23
Improve Interface

JamesGlover:
- Add mail icon to help button to better indicate what it does
- Add book icon to external help links to draw attention to them
- Switch yes/no select fields to radio buttons to reduce clicks and avoid implicit defaults
- Fix long standing duplicate text field on sample edit pages
- Modernise property tags
- Update required fields to use a clear * rather than » to make it more noticeable, and be more conventional. (Its a much clearer * than the one shown here)


JamesGlover:
Demoed to SSR

PR #2339, opened 2019-09-10
Remove legacy manual qc steps

JamesGlover:
As best I can tell, this code is not currently in use. While not
actually completely inaccessible, there are only two batches in a state
that would tigger the path, and both are in a state which results in a
number of exceptions.

Summary of removed behaviour
--------------------------------
1) Batches in the states qc_manual_in_progress or qc_manual would, when rendering tasks without specific templates, instead show specific qc pages to allow manual QC.
2) Released batches typically prevent access to the steps, and have done so for a while.
3) This would appear to render nested versions of the now defunct QC pipelines. (These were deprecated fairly recently, although I was unaware of this nested functionality when doing so).
4) I am not entirely sure how batches get into the qc_manual_in_progress or qc_manual states. It is possible that its through the rollback state functionality.
5) Attempting to trigger display of the pages using the two manual qc batches revealed several issues preventing the pages from rendering. Some of these have existed for a while. I've not seen the associated exceptions. I'm pretty sure at least one of these would trigger regardless of the status of a batch, and will have done for at least a year.


JamesGlover:
Spoke to QC teams, and they confirmed they don't use this functionality.

PR #2363, opened 2019-09-17
Gpl123 json api for extraction lims merged release 12.1.0

emrojo:


app/controllers/api/v2/sample_metadata_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # JSONAPI controller for {SamplesResource}
+    class SampleMetadataController < JSONAPI::ResourceController
```
JamesGlover: {SampleMetadataResource} not {SamplesResource}

emrojo: Thanks, I didnt see it

JamesGlover: Not a problem.
If you end up adding a new resource from scratch, there are also some scaffold that you might find useful: (Although not so useful here, where the resource already existed)

```bundle exec rails generate api_v2```

Might try and split it out into controllers and resources.

JamesGlover: Feel fre to merge this as soon as the tests pass.

Issue #2360, opened 2019-09-16
Rails 6 Upgrade


JamesGlover:
Here is an issue to track known issues blocking a Rails 6 upgrade.

### Gems
The following gems will need to be removed, replaced, or receive upgrades
- [x] actionpack-page_caching
- [x] jquery-tablesorter

### Other
Webpack is the default pipeline in Rails 6, but it will be easier if we switch first, especially if we consider yarn packages for some of the replacements above.
- [ ] Upgrade to webpack pipeline
- [ ] Auto-loading app constants in initialisers
- [x] Uniqueness validators should specify case-sensitivity
- [x] RSpec tests are not transactional (Disable 'transactional_capybara')
- [x] Media type deprecation warnings
- [x] without_report inheritance deprecation warning

JamesGlover:
actionpack-page_caching can just be removed. We don't use it any more.

JamesGlover:
JQuery table-sorter already interacts badly with other elements of Sequencescape.
Possible alternative: https://datatables.net (Only glanced briefly)

JamesGlover:
Datatables looks pretty awesome

JamesGlover:
Rails 6 needs Ruby 2.5, this is currently tricky to install on the metal.

JamesGlover:
On release, set:
Rails.application.config.action_dispatch.use_cookies_with_metadata to false.
For first deploy if we're going to have a rails5 accepting requests at the same time. It can be set to true once rails 5 versions are down.

JamesGlover:
DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison in Rails 6.1. To continue case sensitive comparison on the :login attribute in User model, pass `case_sensitive: true` option explicitly to the uniqueness validator.

This doesn't sound like something that will actually bother us, as generally case-insensitive uniqueness is preferred. So this one should probably be opting in to the new behaviour.

JamesGlover:
Early experimentation suggests that the transactional tests functionality is failing. I know that some of this functionality is now baked into Rails, so it might just be us needing to remove our changes.

JamesGlover:
transactiona_capybara is causing the issues with RSpec. Rails 6 actually adds a lot of this functionality natively, although there may be more work to do to enable it.

JamesGlover:
[...]

PR #2370, opened 2019-09-23
Remove tablestorter depedency

JamesGlover:


JamesGlover:
This is a pre-requisite of the Rails 6 upgrade #2360

PR #2371, opened 2019-09-23
Docs and deletions

JamesGlover:


JamesGlover:
This branch was sitting round on my laptop. Not sure why I never put a pull request in for it.

PR #2400, opened 2019-10-08
Added long read sample manifest.

stevieing:


db/schema.rb
```diff
@@ -10,17 +10,17 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190910103357) do
+ActiveRecord::Schema.define(version: 2019_10_01_140903) do
```
harrietc52: Is this new version number format ok? Is it becuase of the rails upgrate?

db/schema.rb
```diff
@@ -10,17 +10,17 @@
...
 
-  create_table "aker_containers", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC" do |t|
+  create_table "aker_containers", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC", force: :cascade do |t|
```
harrietc52: This rearrange might be becuase of the upgrate too?

harrietc52:
Just need to check the cucumber test, but all looks good

PR #2408, opened 2019-10-11
Create initial Issue templates

andrewsparkes:
Added bug, feature and research issue templates.


andrewsparkes:
This is currently on the Sequencescape repository branch. Better to put it at Sanger organisation level in a separate repository but would need agreement for that. Else need to also add to Limber and Sequencescape-Client-API repositories.

PR #2435, opened 2019-11-14
Try failing fast

pjvv:


pjvv:
Not sure how to do it with knapsack now. With plain rspec is was just: `bundle exec rspec --fail-fast`. What do you think?

KatyTaylor:
> Not sure how to do it with knapsack now. With plain rspec is was just: `bundle exec rspec --fail-fast`. What do you think?

This looks right - https://www.rubydoc.info/github/arturt/knapsack - see line "To do this with Knapsack you simply add your rspec arguments as parameters to the knapsack rake task."

PR #2440, opened 2019-11-21
Gpl 178 upload tube rack manifest tweaks

KatyTaylor:


lib/csv_parser_client.rb
```diff
@@ -23,7 +22,20 @@ def self.get_tube_rack_scan_results(tube_rack_barcode, object_to_add_errors_to)
...
+
+  def self.remove_no_read_results(tube_barcode_to_coordinate)
+    tube_barcode_to_coordinate&.reject! { |key| no_read?(key) } unless tube_barcode_to_coordinate.nil?
+    tube_barcode_to_coordinate
+  end
+
+  def self.no_read?(value_to_check)
+    return true if value_to_check.casecmp('no read').zero?
+
+    false
```
emrojo: We could write this, means the same:
value_to_check.casecmp('no read').zero?

emrojo: Because we will refactor, not worth changing :)

PR #2445, opened 2019-12-02
Split RSpec tests to make build quicker

KatyTaylor:


pjvv:
All looks fine. I would just add something to the README which describes a little what it does and when/how we should run it again.

PR #2452, opened 2019-12-05
Remove old apps

pjvv:


KatyTaylor:
What was 'crier' used for?

pjvv:
Not sure, will ask at stand up today..

pjvv:
It's an old app that isn't used anymore and the config in the deployment project already removed the links.

PR #2454, opened 2019-12-09
Gpl 216 edit study bugs

KatyTaylor:
Steps to reproduce the bug:
- Start with a study where the data release strategy is 'Not Applicable'
- Click Edit Study
- Change the data release strategy to 'Open' or 'Managed'
- Observe that nothing seems to happen on the page, and the following error appears in the JS console:
"An invalid form control with name='study[study_metadata_attributes][data_release_delay_approval]' is not focusable."

This change fixes the above bug. The key line that fixes it is the following:
```
// only enable input if it's not hidden - could be beneath a hidden element in the DOM
if(!$(this).is(':visible')) return;
```
The problem was that it was removing the disabled attribute from inputs that were nested under hidden divs, which meant that the validation fired on save but could not find the field to display the validation message against.

All the other changes are refactoring to try to make it more readable, because I found it quite hard to figure out what was going on!

This javascript file is rendered by the following line in _study.html.erb:
`<% metadata_fields.finalize_related_fields %>`

The other key file to look at to understand this form is form_builder.rb, which seems to be a custom form builder to replace Rails's default one.

P.S. The form still has major validation issues, so the steps to reproduce still produce an error - however, now it is a server side error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,36 +51,14 @@ jobs:
   - if: type != cron AND tag IS blank
     env:
     - RAILS_ENV=test
-    - CI_NODE_TOTAL=3
-    - CI_NODE_INDEX=0
+    # - CI_NODE_TOTAL=3
+    # - CI_NODE_INDEX=0
     before_script:
     # - RAILS_ENV=test bundle exec rails webdrivers:chromedriver:update -r webdrivers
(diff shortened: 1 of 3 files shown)


pjvv:
Thanks for the detail!

PR #2456, opened 2019-12-10
Gpl 216 edit study validation

KatyTaylor:
N.B. Most of the changes in this pull request will disappear once the following PR is merged in: https://github.com/sanger/sequencescape/pull/2454

Validation in this form is weird:
- radio buttons are always made required in the front end (form_builder.rb, radio_select method)
- other fields are never made required in the front end
- asterisks are added based on whether the field is required on the model, which doesn't work for some which are conditionally required (e.g. data access group)

This means that if you submitted the form without having filled out a (non-radio) field that is required on the model, it threw a server side error. This error should have been displayed in a nice Rails way, but -

`Rails.logger.warn "Failed to update attributes: #{@study.errors.map(&:to_s)}"`
The above line was throwing an exception due to the .map(&:to_s) syntax not working inside a 'rescue' block.
Changing it to the .map { } syntax allows it to get to the 'render' line and means the validation errors are displayed more nicely.


KatyTaylor:
Updated this with a change to the 'attach_option_updater' function in the related_fields file.

This fixes the following issue:
- The 3 'prevention' fields were not being saved to the database
- This is because of 'if: :never_release?' in the study model, which checks if the 'timing' field is set to 'never'
- The form allowed you to fill these fields in when 'timing' was not 'never'
- Looks like the form intended this to be prevented by restricting the value in 'timing' to only 'never', when 'strategy' was 'not applicable'
- But, this functionality was broken - now fixed

PR #2464, opened 2019-12-13
Attempt to fix build issue

KatyTaylor:
This is an attempt to fix the following build issue, which occurs erratically while executing the webdrivers:chromedriver:update command on the rspec or cucumber test jobs:

/home/travis/build/sanger/sequencescape/vendor/bundle/ruby/2.5.0/gems/ffi-1.11.3/lib/ffi/library.rb:112: 
[BUG] Illegal instruction at 0x00007f469b4da5c7
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-linux]

I have downgraded sassc to 2.1.0 from 2.2.1. Got this idea from this post: https://github.com/sass/sassc-ruby/issues/146

It seems to be to do with how it figures out the location of a library it is trying to load. 
Error occurs on line 112 of this ffi file - https://github.com/ffi/ffi/blob/1.11.3/lib/ffi/library.rb
That function gets passed a path to a file from line 11 of this file - https://github.com/sass/sassc-ruby/blob/v2.2.1/lib/sassc/native.rb (2.2.1) / https://github.com/sass/sassc-ruby/blob/v2.1.0/lib/sassc/native.rb (2.1.0)

Branch build has passed 5 times in a row, as opposed to only 1 out of 5 times before I made the change, so hopefully it is fixed?

P.S. Here is a link to a failing build (different branch), just for an easy reference to the error message - https://travis-ci.org/sanger/sequencescape/jobs/624172858

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,10 @@ env:
   global:
   - TZ=Europe/London
   - CUCUMBER_FORMAT=DebugFormatter
+branches:
+  only:
+  - master
+  - develop
 before_install:
 - mv config/aker.yml.example config/aker.yml
 - yarn install
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,6 @@ env:
   global:
   - TZ=Europe/London
   - CUCUMBER_FORMAT=DebugFormatter
-branches:
-  only:
-  - master
-  - develop
 before_install:
 - mv config/aker.yml.example config/aker.yml
 - yarn install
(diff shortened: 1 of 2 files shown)


pjvv:
How will be prevent it from updating though?

KatyTaylor:
> How will be prevent it from updating though?

Well, I suppose at least if Depfu suggests an upgrade to sassc then the build has quite a high chance of failing, so we'll notice and not merge it in. Hopefully they'll fix it at some point. Any ideas? Could we declare it explicitly in the Gemfile and leave a comment?

JamesGlover:
👍 The comment, git commit and this issue worked really well. 5 months later and I found this in a few seconds of noticing sassc was pinned.

PR #2466, opened 2019-12-16
Specify 'sassc' version explicitly in the Gemfile

KatyTaylor:
- this is to try to prevent it from being automatically updated
- we need it at this version because build failures happen when it's at 2.2.1 (see pull request https://github.com/sanger/sequencescape/pull/2464)


pjvv:
Yeah, let's give it a try!

PR #2483, opened 2020-01-08
Add alert to point to new Sequencescape URL

KatyTaylor:
display an alert under the navigation bar in the header if the base_url is anything other than http://sequencescape.psd.sanger.ac.uk

- the old_url method is in application_helper.rb, which is available on every page because it is included in application_controller.rb (along with all other helpers)

Putting it above the nav bar had more CSS issues so I've put it underneath so I don't have to fiddle with the CSS.
The placement also matches the 'app_info_box' (which looks like it used to have some styling before a massive rails / bootstrap ugrade (search here for 'app-info-box' if interested - https://github.com/sanger/sequencescape/commit/b5341ab34cdf2e620a351b5645148c564adadc16).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -259,7 +259,7 @@ module ApplicationHelper
   end
 
   def old_url
-    permitted_urls = ['http://sequencescape.psd.sanger.ac.uk']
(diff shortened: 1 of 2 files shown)


KatyTaylor:
P.S. I've tested this by adding 'http://localhost:3000' to the list of permitted_urls, and the banner goes away.

PR #2485, opened 2020-01-10
Gpl 192 sending failed messages

pjvv:


app/models/report_fail.rb
```diff
@@ -0,0 +1,74 @@
...
+    return false unless valid?
+
+    failed_labware.each do |labware|
+      labware.events.create_labware_failed!(failure_id, user.login)
+      BroadcastEvent::LabwareFailed.create!(
+        seed: labware,
+        user: user,
+        properties: { failure_reason: failure_options.key(failure_id) }
+      )
+    end
```
emrojo: If one of them fails, from what we talked, we catch the exception of the one that fails and put it in the screen for the user

PR #2507, opened 2020-01-22
Gpl192 add more failure options

emrojo:


config/locales/en.yml
```diff
@@ -48,6 +48,12 @@ en:
     logout: "Logout"
     help: "Help"
 
+  # Description for the keys published in the events generated for the report fails UI
+  report_fails:
+    fail_because_sample_integrity: "Sample integrity upon receipt"
```
pjvv: `fail_because_` is redundant here.

PR #2510, opened 2020-01-28
GPL-275 Phi X stock concentration

KatyTaylor:
make the Phi X concentration to 3 decimal places rather than 2


KatyTaylor:
Screenshot 2020-01-28 at 15 41 57

PR #2524, opened 2020-02-07
Gpl 273 duplex seq library type

KatyTaylor:
This code will create one of each of the following:
- Request Type ('Limber Duplex-Seq')
- Library Type ('Duplex-Seq')
- Submission Template ('Limber-Htp - Duplex-Seq')

- Product Catalogue - think this is obsolete, to do with Aker project, but it is required
- Order Role - think this is intended for printing on barcodes, but the existing order roles don't seem to match up to the pipelines very well, and the Adam and Sara didn't recognise them - think in reality the 'prefix' in Limber is used instead. Creating one here to be consistent with the rest of the code in this rake task.

### DIFF
--- a/lib/tasks/limber.rake
+++ b/lib/tasks/limber.rake
@@ -250,7 +250,8 @@ namespace :limber do
           'Ribozero RNA depletion',
           'Ribozero RNA-seq (Bacterial)',
           'Ribozero RNA-seq (HMR)',
-          'TraDIS'
+          'TraDIS',
+          'Chromium Visium'
         ],
         product_line: 'Bespoke',
         default_purposes: ['LBB Cherrypick'] # It requires default_purpose to accept an array.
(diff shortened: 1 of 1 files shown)


lib/tasks/limber.rake
```diff
@@ -155,6 +155,12 @@ namespace :limber do
       %w[WGS LCMB].each do |prefix|
         Limber::Helper::RequestTypeConstructor.new(prefix).build!
       end
+
+      Limber::Helper::RequestTypeConstructor.new(
+        'Duplex-Seq',
```
andrewsparkes: So this constructor prefixes the request type with 'Limber' right?

lib/tasks/limber.rake
```diff
@@ -244,7 +250,8 @@ namespace :limber do
           'Ribozero RNA depletion',
           'Ribozero RNA-seq (Bacterial)',
           'Ribozero RNA-seq (HMR)',
-          'TraDIS'
+          'TraDIS',
+          'Chromium Visium'
```
andrewsparkes: Is Visium here because it was missed out in the previous release?

KatyTaylor: yes

lib/tasks/limber.rake
KatyTaylor: yes

PR #2535, opened 2020-02-17
Fix for GPL273

andrewsparkes:
Adds LDS Stock and LDS Cherrypick plate purposes as defaults for the Duplex-Seq request type so it doesn't fall back to the default (LB Cherrypick)

### DIFF
--- a/lib/tasks/limber.rake
+++ b/lib/tasks/limber.rake
@@ -146,6 +146,34 @@ namespace :limber do
         size: 48
       )
     end
+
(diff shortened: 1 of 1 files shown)


KatyTaylor:
Script for plate purposes fix:
[plate_purposes_for_duplex_seq_script.txt](https://github.com/sanger/sequencescape/files/4214614/plate_purposes_for_duplex_seq_script.txt)

PR #2538, opened 2020-02-18
Gpl293 add uat actions for duplex seq integration suite test

andrewsparkes:
Adds additional UAT actions to support steps in the Duplex-Seq integration test.
- Generate Tag Layout Template - allows generation of tag layout templates
- Plate Information - returns well information given a plate barcode

### DIFF
--- a/app/uat_actions/uat_actions/generate_tag_layout_template.rb
+++ b/app/uat_actions/uat_actions/generate_tag_layout_template.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+# Will construct a tag layout template from existing tag groups
+class UatActions::GenerateTagLayoutTemplate < UatActions
+  DIRECTIONS = {
+    'InColumns (A1,B1,C1...)': 'TagLayout::InColumns',
+    'InRows (A1,A2,A3...)': 'TagLayout::InRows',
+    'InInverseColumns (H12,G12,F12...)': 'TagLayout::InInverseColumns',
+    'InInverseRows (H12,H11,H10...)': 'TagLayout::InInverseRows'
+  }.freeze
+
+  self.title = 'Generate Tag Layout Template'
+  self.description = 'Generates a tag layout template from one or two tag groups.'
+
+  form_field :name,
+             :text_field,
+             label: 'Tag Layout Template Name',
+             help: 'It will not create a tag group with a name that already exists.'
+  form_field :tag_group_id,
+             :select,
+             label: 'Tag Group',
+             help: 'Select tag group 1 for the template',
+             select_options: -> { TagGroup.visible.pluck(:name, :id) },
+             options: { include_blank: 'Select tag group...' }
+  form_field :tag2_group_id,
+             :select,
+             label: 'Tag2 Group',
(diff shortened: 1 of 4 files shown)


app/uat_actions/uat_actions/generate_tag_layout_template.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+# Will construct a tag layout template from existing tag groups
+class UatActions::GenerateTagLayoutTemplate < UatActions
+  DIRECTIONS = {
+    'InColumns (A1,B1,C1...)': 'TagLayout::InColumns',
+    'InRows (A1,A2,A3...)': 'TagLayout::InRows',
+    'InInverseColumns (H12,G12,F12...)': 'TagLayout::InInverseColumns',
+    'InInverseRows (H12,H11,H10...)': 'TagLayout::InInverseRows'
+  }.freeze
```
KatyTaylor: Did you consider referencing the constant in the TagLayoutTemplatesController class, to avoid duplicating it?

app/uat_actions/uat_actions/generate_tag_layout_template.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
```
KatyTaylor: It's a bit of a shame that the UAT page is an almost exact replica of the existing admin function (/tag_layout_templates/new?tag_group_id=), but I guess it's kind of necessary to fit in with the rest of the UAT / integration test stuff?

app/uat_actions/uat_actions/generate_tag_layout_template.rb
```diff
@@ -0,0 +1,77 @@
...
+    'InInverseRows (H12,H11,H10...)': 'TagLayout::InInverseRows'
+  }.freeze
+
+  self.title = 'Generate Tag Layout Template'
+  self.description = 'Generates a tag layout template from one or two tag groups.'
+
+  form_field :name,
+             :text_field,
+             label: 'Tag Layout Template Name',
+             help: 'It will not create a tag group with a name that already exists.'
```
KatyTaylor:
```suggestion
             help: 'It will not create a tag layout template with a name that already exists.'
```
?

app/uat_actions/uat_actions/plate_information.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
```
KatyTaylor: Stuff like this makes me wonder what's the difference between the UAT actions and the API? Should the integration tests be using the API, and we should be enhancing that?

app/uat_actions/uat_actions/generate_tag_layout_template.rb
andrewsparkes: Tried again and got it working now, so that's a small improvement.

app/uat_actions/uat_actions/generate_tag_layout_template.rb
andrewsparkes: Fixed thx

PR #2560, opened 2020-02-28
Bugfix local dev - Not printing plate barcodes

emrojo:
When printing barcodes, the string sent is in the format DN... instead of
the previous numeric version. After moving plate labels to code 39 we
can support alphanumeric, but not in ean 13. In local development, the
default label template is not code39, which it causes the label not to be
printed from development machines and appears as empty space in the label.


JamesGlover:
This has been migrated to record loader, and now has the same template setup as used in this PR.
https://github.com/sanger/sequencescape/blob/develop/config/default_records/barcode_printer_types/default_records.yml

PR #2568, opened 2020-03-11
Gpl 295 add fields to receptacles for duplex seq limber

KatyTaylor:


app/uat_actions/uat_actions/generate_plate_concentrations.rb
```diff
@@ -65,6 +65,16 @@ def conc_units
...
   end
 
+  def key
+    @key ||= decide_key
+  end
+
+  def decide_key
+    return 'molarity' if conc_units == 'nM'
+
+    'concentration'
```
harrietc52, line 75: Could possibly use rubys conditional operator?
` if conc_units == 'nM' ? 'molarity' : 'concentration' `

PR #2581, opened 2020-03-20
Create a Heron plate from a rack in Sequencescape plate creator UI

emrojo:
* Be able to create plates from stock plates by printing the barcode from the plate creator UI in Sequencescape

* *Bugfix* Coordinates from racks were not unpadded, which it created problems when creating plates from them.

* Support to create any plate purpose available in plate creation view


emrojo:
This also modifies sample name to be sanger_sample_id instead of supplier_sample_id as the names needs to be unique.

PR #2591, opened 2020-03-27
Remove issue templates

pjvv:
These are now available at: https://github.com/sanger/.github


pjvv:
Adding you @andrewsparkes since you did the original work of adding them.

Issue #2602, opened 2020-04-01
GPL-374 Re-print barcodes from plates/new page


KatyTaylor:
**User story**
As an SSR I would like to be able to come back to the plates/new page after creating plates using it, to print out the barcodes, in case they didn't print successfully the first time.

**Who the primary contacts are for this story**
Rich Clark & James Mack (& Danni Weldon)

**Acceptance criteria**
1. Way of printing barcodes in bulk on a group of plates created at one time using the plates/new page

**Additional context**
Don't think this is currently possible because there's nothing grouping those plates together - is there an alternative way of doing it?

JamesGlover:
With https://github.com/sanger/sequencescape/pull/2604 we could at least allow printing of replacement barcodes from the feedback page, although this would only work until someone navigated away from the page.

KatyTaylor:
Question - touched on by Rich C at some point I think -

What happens if the plates you've just created are the 'LHR RT' ones? If you print barcode from SS, presumably you get just the one barcode, but if you print them from Limber, you get 2 extra barcodes, for the PCR plates (this will go away when those PCR plates are tracked (GPL-375)).

Do we want this bulk print functionality from SS to print barcodes for just the RT plates, or somehow for those PCR plates as well?

KatyTaylor:
Should we have some sort of more generic 'bulk print labels' page, where you paste in barcodes?

KatyTaylor:
Got it working using Asset Groups:
<img width="1207" alt="Screenshot 2020-04-06 at 15 18 01" src="https://user-images.githubusercontent.com/10407809/78568618-43c9c180-781a-11ea-87f0-d787784530da.png">
<img width="1183" alt="Screenshot 2020-04-06 at 15 18 19" src="https://user-images.githubusercontent.com/10407809/78568623-46c4b200-781a-11ea-8fe5-78d430b05440.png">
<img width="1198" alt="Screenshot 2020-04-06 at 15 18 31" src="https://user-images.githubusercontent.com/10407809/78568627-49bfa280-781a-11ea-838e-acf7222a6a62.png">

KatyTaylor:
Alternative to above could be something like this (notice the URL):

<img width="1199" alt="Screenshot 2020-04-06 at 16 43 14" src="https://user-images.githubusercontent.com/10407809/78577346-d91e8300-7825-11ea-990c-e864f03f05ce.png">

[...]

PR #2604, opened 2020-04-02
Gpl370 visual feedback

emrojo:
Closes GPL370

* Display the barcodes of the source and destination in a table after clicking print

### DIFF
--- a/app/models/plate/creator.rb
+++ b/app/models/plate/creator.rb
@@ -16,8 +16,6 @@ class Plate::Creator < ApplicationRecord
 
   self.table_name = 'plate_creators'
 
-  attr_reader :created_plates
-
   # These are the plate purposes that will be created when this creator is used.
   has_many :plate_creator_purposes, class_name: 'Plate::Creator::PurposeRelationship', dependent: :destroy, foreign_key: :plate_creator_id
   has_many :plate_purposes, through: :plate_creator_purposes
@@ -26,6 +24,10 @@ class Plate::Creator < ApplicationRecord
   has_many :parent_plate_purposes, through: :parent_purpose_relationships, source: :plate_purpose
 
   serialize :valid_options
+
+  def created_plates
+    @created_plates ||=[]
+  end
 
   # Executes the plate creation so that the appropriate child plates are built.
   def execute(source_plate_barcodes, barcode_printer, scanned_user, creator_parameters = nil)
@@ -60,16 +62,13 @@ class Plate::Creator < ApplicationRecord
     end
     print_job = LabelPrinter::PrintJob.new(barcode_printer.name,
                                            LabelPrinter::Label::PlateCreator,
-                                           plates: created_plates,
+                                           plates: created_plates.pluck(:destinations).flatten.compact,
                                            plate_purpose: plate_purpose, user_login: scanned_user.login)
 
-    # unless print_job.execute
-    #   raise PlateCreationError, "Barcode labels failed to print."
-    # end
-
-    plate_factories.map do |pf|
-      pf.plate.human_barcode
+    unless print_job.execute
+      raise PlateCreationError, "Barcode labels failed to print."
     end
+    true
   end
 
   private
@@ -123,7 +122,7 @@ class Plate::Creator < ApplicationRecord
   end
 
   def add_created_plates(plates, source, destinations)
-    @created_plates.push({
+    created_plates.push({
       source: source,
       destinations: destinations
     })
--- a/app/models/plate/creator.rb
+++ b/app/models/plate/creator.rb
@@ -16,6 +16,8 @@ class Plate::Creator < ApplicationRecord
 
   self.table_name = 'plate_creators'
 
+  attr_reader :created_plates
+
   # These are the plate purposes that will be created when this creator is used.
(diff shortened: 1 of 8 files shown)


KatyTaylor:
Can we have a screenshot of how it looks please? :)

app/models/plate/creator.rb
```diff
@@ -42,19 +47,28 @@ def execute(source_plate_barcodes, barcode_printer, scanned_user, creator_parame
...
     print_job = LabelPrinter::PrintJob.new(barcode_printer.name,
                                            LabelPrinter::Label::PlateCreator,
-                                           plates: plate_factories.map(&:plate),
+                                           plates: created_plates.pluck(:destinations).flatten.compact,
                                            plate_purpose: plate_purpose, user_login: scanned_user.login)
-    return false unless print_job.execute
+
+    unless print_job.execute
+      raise PlateCreationError, 'Barcode labels failed to print.'
+    end
```
KatyTaylor, line 71: I added this in because I was changing the return type to be a list of plates rather than a Boolean, but since it's back to a Boolean, wonder whether we should take this exception out? I think it will prevent the plates from being created if the printing failed, which is changing the current behaviour. Rich C was just saying that the barcode printers are failing quite a lot, so it might be annoying if this prevents creation of the plates as well. What do you think?

JamesGlover: Agreed, especially given https://github.com/sanger/sequencescape/issues/2602

emrojo:
image

app/views/plates/creator/_printed_barcodes.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <% @creator.created_plates.each do |entry| %>
+      <tr>
+        <td><%= entry[:source].barcodes.first.barcode %> </td>
+        <% entry[:destinations].each do |destination| %>
+            <td><%= destination.barcodes.first.barcode %>  </td>
+        <% end %>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
```
KatyTaylor, line 20: Why is 'destinations' plural? How do you create multiple child plates from one source plate using this page?

app/models/plate/creator.rb
stevieing: I tried it locally and saw the message so it is working as expected.

app/views/plates/creator/_printed_barcodes.html.erb
emrojo: For some combinations of source plate purposes and destination purposes the creators can create more than one plate. For example, the added screenshot the source is the working dilution WD... barcode and the destination plate purposes is Pico Assay Plates selection

app/models/plate/creator.rb
```diff
@@ -103,11 +117,19 @@ def create_plates(source_plate_barcodes, current_user, creator_parameters = nil)
           raise PlateCreationError, "Scanned plate #{scanned} has a purpose #{plate.purpose.name} not valid for creating [#{plate_purposes.map(&:name).join(',')}]"
         end
 
-        plates.concat(create_child_plates_from(plate, current_user, creator_parameters))
+        add_created_plates(plates, plate, create_child_plates_from(plate, current_user, creator_parameters))
```
KatyTaylor, line 120:
```suggestion
        plates.concat(add_created_plates(plate, create_child_plates_from(plate, current_user, creator_parameters)))
```

find it confusing passing the `plates` parameter to the `add_created_plates` method because it's only used in the plate scenario rather than the tub rack scenario. Can you do this instead, and make `add_created_plates` return `destinations`?

app/views/plates/creator/_printed_barcodes.html.erb
KatyTaylor: Thanks. From looking at the table structure, it looks like all your source plates / racks have to be of the same 'purpose' - is that correct?

KatyTaylor:
Thanks for the screenshot! Very small and non-urgent comment - feel like it would be nicer as a green box rather than 'warning', and have a small title above the table that says 'Created plates:' or 'Created labware:' ?

app/views/plates/creator/_printed_barcodes.html.erb
emrojo: Good catch! yes, that could be a problem if we  print from TR48 along with TR96...
Destinations will be fine, but Sources may have different plate purpose.

app/models/plate/creator.rb
emrojo: ok

app/views/plates/creator/_printed_barcodes.html.erb
emrojo: So I'll change the table structure to show the plate purpose of every source

KatyTaylor: Yeah, could just call the first column 'Source' or something generic

emrojo:
image

After applying changes

app/models/plate/creator.rb
KatyTaylor: This is fine, the behaviour I saw was because when I checked it out, the whole method was in a transaction and therefore was rolling back the plate creation.

E's taken it out of the transaction so now behaviour will be unchanged.

PR #2610, opened 2020-04-09
gpl375 Deduplication of aliquots in transfer requests

emrojo:
Closes GPL375

* Added deduplication of aliquots when copying from multiple plates.
* TagClash error raised again by the API as it was not working.

Relates to Limber <https://github.com/sanger/limber/issues/410>

### DIFF
--- a/app/api/endpoints/transfer_request_collections.rb
+++ b/app/api/endpoints/transfer_request_collections.rb
@@ -1,7 +1,10 @@
 module Endpoints
   class TransferRequestCollections < ::Core::Endpoint::Base
     model do
-      action(:create, to: :standard_create!)
+      # action(:create, to: :standard_create!)
+      action(:create) do |request, response|
+        standard_create!(request, response)
+      end
     end
 
     instance do
--- a/app/api/endpoints/transfer_request_collections.rb
+++ b/app/api/endpoints/transfer_request_collections.rb
@@ -1,10 +1,7 @@
 module Endpoints
   class TransferRequestCollections < ::Core::Endpoint::Base
     model do
-      # action(:create, to: :standard_create!)
-      action(:create) do |request, response|
-        standard_create!(request, response)
-      end
+      action(:create, to: :standard_create!)
     end
 
     instance do
--- a/app/models/aliquot.rb
+++ b/app/models/aliquot.rb
@@ -24,7 +24,8 @@ class Aliquot < ApplicationRecord
 
   self.lazy_uuid_generation = true
 
-  TagClash = Class.new(ActiveRecord::RecordInvalid)
+  # TagClash = Class.new(ActiveRecord::RecordInvalid)
+  TagClash = Class.new(StandardError)
 
   # An aliquot can represent a library, which is a processed sample that has been fragmented.  In which case it
   # has a receptacle that held the library aliquot and has an insert size describing the fragment positions.
--- a/app/models/transfer_request.rb
+++ b/app/models/transfer_request.rb
@@ -7,6 +7,10 @@ class TransferRequest < ApplicationRecord
   include AASM
   extend Request::Statemachine::ClassMethods
 
+  # Determines if we attempt to filter out {Aliquot#equivalent? equivalent} aliquots
+  # before performing transfers.
+  attr_accessor :merge_equivalent_aliquots
+
   # States which are still considered to be processable (ie. not failed or cancelled)
   ACTIVE_STATES = %w[pending started passed qc_complete].freeze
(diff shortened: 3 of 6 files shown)


app/models/aliquot.rb
```diff
@@ -24,7 +24,8 @@ class Aliquot < ApplicationRecord
 
   self.lazy_uuid_generation = true
 
-  TagClash = Class.new(ActiveRecord::RecordInvalid)
+  # TagClash = Class.new(ActiveRecord::RecordInvalid)
+  TagClash = Class.new(StandardError)
```
KatyTaylor, line 28: What was the reason for changing the exception type it inherits from, out of interest?

app/models/transfer_request.rb
```diff
@@ -7,6 +7,10 @@ class TransferRequest < ApplicationRecord
   include AASM
   extend Request::Statemachine::ClassMethods
 
+  # Determines if we attempt to filter out {Aliquot#equivalent? equivalent} aliquots
+  # before performing transfers.
+  attr_accessor :merge_equivalent_aliquots
```
KatyTaylor, line 12: Where does `:merge_equivalent_aliquots` get set? By Limber's API call?

app/models/transfer_request.rb
```diff
@@ -170,17 +188,45 @@ def suggested_transition_to(target)
...
+    asset.aliquots.map do |aliquot|
+      aliquot.dup(aliquot_attributes(aliquot))
+    end
+  end
+
+  def duplicates_of_distinct_source_aliquots_only
+    duplicates_of_all_source_aliquots.reject do |candidate_aliquot|
+      target_asset.aliquots.any? { |existing_aliquot| existing_aliquot.equivalent?(candidate_aliquot) }
+    end
+  end
```
KatyTaylor, line 228: Little bit confused about what's happening here. Will this `perform_transfer_of_contents` method run once for PCR plate 1, at which point all the aliquots will be transferred, and then a second time for PCR plate 2, at which point none of the aliquots will be transferred, because it's checking against `target_asset.aliquots`?

app/api/endpoints/transfer_request_collections.rb
```diff
@@ -1,7 +1,10 @@
 module Endpoints
   class TransferRequestCollections < ::Core::Endpoint::Base
     model do
-      action(:create, to: :standard_create!)
+      # action(:create, to: :standard_create!)
+      action(:create) do |request, response|
+        standard_create!(request, response)
+      end
```
KatyTaylor, line 7: What difference does this change make? (Sorry for all the q's, not familiar with this API)

app/models/aliquot.rb
KatyTaylor: Ah, just seen James's comment on the issue, Rails was swallowing the RecordInvalid one?

emrojo: Yes, we think it was related with this: 
https://guides.rubyonrails.org/active_record_callbacks.html#halting-execution

app/models/transfer_request.rb
emrojo: By Limber API, so we had to define it in sequencescape-client-api:
https://github.com/sanger/sequencescape-client-api/commit/e934b023d186394e751b0834a93502f489ce3b04
And in Limber as well, when the pull request for limber is ready

app/api/endpoints/transfer_request_collections.rb
emrojo: I think no difference, we were changing some parts of code and i think we can roll back this to how it was.

app/models/aliquot.rb
KatyTaylor: Ah OK, so saving a record through the API is like calling `save` rather than `save!` ?

app/models/transfer_request.rb
emrojo: Limber creates a TransferRequestCollection in SS which contains all transfer request definitions coming from both PCR1 and PCR2; for every request we run perform_transfer_of_contents so in this list of transfer request creations, elements from PCR1 and PCR2 will go to the same destination well, the first one will create the request without problem, the second will need to check if it can happen by comparing with the first.

app/models/aliquot.rb
emrojo: Yes, we think; so because we were raising something like a ActiveRecord::RecordInvalid inside a callback, callbacks do not seem to re-raise it because that would break save() execution, so it must be because api v1 behaves the same as save.

PR #2612, opened 2020-04-14
Improved robustness of aliquot equivalence check

JamesGlover:
The previous version of the check compared hard-coded
attributes when checking equivalence. This resulted in
newer columns, such as primer_panel_id getting missed
from the check.

To avoid the need to remember that this method needs to
be updated when extending Aliquot, the columns to compare
are now generated dynamically from the database columns.
Columns which are permitted to change are now whitelisted.

This should substantially reduce the risk of future bugs
being introduced, and will ensure that the comparison
remains conservative, rather than permissive.

Closes #

*
*
* ...

### DIFF
--- a/app/models/aliquot.rb
+++ b/app/models/aliquot.rb
@@ -95,6 +95,13 @@ class Aliquot < ApplicationRecord
       .count
   end
 
+  # Returns a list of attributes which must be the same for two Aliquots to be considered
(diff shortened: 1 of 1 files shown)


emrojo:
Relates with: https://github.com/sanger/limber/issues/410

PR #2642, opened 2020-04-30
Gpl433 allow changing target plate type at verification

### LINKED ISSUE #2639: GPL-433 select the plate type and not have it pre-set for lighthouse process
**User story**
GPL-433 | As a scientist (Danni W) I would like to be able to select the plate type and not have it pre-set, exactly like I can for the source plates, currently having to edit the work list manually in word pad, this can be introduce errors and slows the process down.

**Who the primary contacts are for this story**
Danni W

**Acceptance criteria**
1. Change the plate type for the dest 1 (destination plate).

**Additional context**
Danni  W wrote Wednesday, 29 April 2020 at 17:47

Hi Rich,
 
I completed some test picks today for the light house samples and when generating the worklist in Sequencescape I noticed I couldn’t change the plate type for the dest 1 (destination plate). The light house samples are being picked into a different plate type (Not Abgene 0800). As a user I would like to be able to select the plate type and not have it pre-set, exactly like I can for the source plates. I asked the SSR’s and this is not set by them when creating a pick.
 
Currently as a user I am having to edit the work list manually in word pad, this can be introduce errors and slows the process down

JamesGlover:
Closes #2639

* Adds plate type drop down to destination plate
* Updates the plate with the selected type
* Populates the information in the TECAN file
* Fixes performance issue with table helper

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -168,6 +168,7 @@ group :test do
   # Rails performance tests
   gem 'rails-perftest'
   gem 'rspec-collection_matchers' # Provides matchers for dealing with arrays
+  gem 'rspec-longrun' # Extends scenario logging for more verbose tracking
   gem 'test-prof'
   # Provides json expectations for rspec. Makes test more readable,
   # and test failures more descriptive.
--- a/app/models/batch/tecan_behaviour.rb
+++ b/app/models/batch/tecan_behaviour.rb
@@ -7,7 +7,10 @@ module Batch::TecanBehaviour
       'destination' => {}
     }
 
-    requests.includes([{ asset: %i[plate map] }, { target_asset: [:map, :well_attribute, { plate: :barcodes }] }])
+    requests.includes([
+      { asset: [{ plate: [:barcodes, :labware_type] }, :map] },
+      { target_asset: [:map, :well_attribute, { plate: [:barcodes, :labware_type] }] }
+    ])
             .where(state: 'passed')
             .find_each do |request|
       target_plate = request.target_asset.plate
@@ -17,14 +20,15 @@ module Batch::TecanBehaviour
       full_source_barcode = request.asset.plate.machine_barcode
       full_destination_barcode = request.target_asset.plate.machine_barcode
 
-      source_plate_name = override_plate_type.presence || request.asset.plate.plate_type.tr('_', "\s")
-
       if data_object['source'][full_source_barcode].nil?
-        data_object['source'][full_source_barcode] = { 'name' => source_plate_name, 'plate_size' => request.asset.plate.size }
+        source_plate_type = override_plate_type.presence || request.asset.plate.plate_type.tr('_', "\s")
+        data_object['source'][full_source_barcode] = { 'name' => source_plate_type, 'plate_size' => request.asset.plate.size }
       end
+
       if data_object['destination'][full_destination_barcode].nil?
+        target_plate_type = (target_plate.plate_type || PlateType.cherrypickable_default_type).tr('_', "\s")
         data_object['destination'][full_destination_barcode] = {
-          'name' => PlateType.cherrypickable_default_type.tr('_', "\s"),
+          'name' => target_plate_type,
           'plate_size' => request.target_asset.plate.size
         }
       end
--- a/app/views/robot_verifications/submission.html.erb
+++ b/app/views/robot_verifications/submission.html.erb
@@ -3,6 +3,7 @@
 <%= form_for :robot_verifications, url: {action: "download"} do |f| %>
 
 <%= page_title "Scan Robot Beds And Plates", "Step 2 of 3" %>
+<% plate_types = PlateType.all.pluck(:name) %>
 
 <table class="bed_verification" id="source_beds">
   <thead>
@@ -30,7 +31,7 @@
          <%   next_field = "bed_barcodes_#{bed_label.to_i + 1}" %>
         <% end %>
         <td><%= text_field_tag "plate_barcodes[#{plate_label}]", params.dig(:plate_barcodes, plate_label), :onkeypress => "return disable_cr_and_change_focus(event,'plate_barcodes_#{plate_label}','#{next_field}')" %></td>
-        <td><%= select_tag("source_plate_types[#{plate_label}]", options_for_select(PlateType.pluck(:name))) %></td>
+        <td><%= select_tag("source_plate_types[#{plate_label}]", options_for_select(plate_types)) %></td>
       </tr>
     <% count = count + 1 %>
     <% end %>
@@ -49,7 +50,7 @@
        <%   next_field = "destination_bed_barcodes_#{bed_label.to_i +1 }" %>
       <% end %>
       <td><%= text_field_tag "destination_plate_barcodes[#{destination_label}]", params.dig(:destination_plate_barcodes, destination_label), :onkeypress => "return disable_cr_and_change_focus(event,'destination_plate_barcodes_#{destination_label}','#{next_field}')" %></td>
-      <td><%= PlateType.cherrypickable_default_type %></td>
+      <td><%= select_tag("destination_plate_types[#{destination_label}]", options_for_select(plate_types, PlateType.cherrypickable_default_type)) %></td>
     </tr>
     <% end %>
   <% end %>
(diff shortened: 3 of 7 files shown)


Gemfile
```diff
@@ -168,6 +168,7 @@ group :test do
   # Rails performance tests
   gem 'rails-perftest'
   gem 'rspec-collection_matchers' # Provides matchers for dealing with arrays
```
JamesGlover, line 170: Added because the test was running alarmingly slowly, and I wanted to find the bottleneck.

app/models/batch/tecan_behaviour.rb
```diff
@@ -7,7 +7,10 @@ def generate_tecan_data(target_barcode, override_plate_type = nil)
       'destination' => {}
     }
 
-    requests.includes([{ asset: %i[plate map] }, { target_asset: [:map, :well_attribute, { plate: :barcodes }] }])
+    requests.includes([
+      { asset: [{ plate: [:barcodes, :labware_type] }, :map] },
+      { target_asset: [:map, :well_attribute, { plate: [:barcodes, :labware_type] }] }
+    ])
             .where(state: 'passed')
```
JamesGlover, line 14: Added plate types to the eager loading, although as we cache things on a per plate basis this won't be a huge speed increase.

app/models/batch/tecan_behaviour.rb
```diff
@@ -17,14 +20,15 @@ def generate_tecan_data(target_barcode, override_plate_type = nil)
       full_source_barcode = request.asset.plate.machine_barcode
       full_destination_barcode = request.target_asset.plate.machine_barcode
 
-      source_plate_name = override_plate_type.presence || request.asset.plate.plate_type.tr('_', "\s")
-
```
JamesGlover, line 21: Move this inside the conditional that determines if we actually make use of it. (Most of this was happening while I was trying to track down why things were so slow)

JamesGlover: Also, renamed it to source_plate_type, because that's what it is.

app/models/batch/tecan_behaviour.rb
```diff
@@ -17,14 +20,15 @@ def generate_tecan_data(target_barcode, override_plate_type = nil)
...
       if data_object['source'][full_source_barcode].nil?
-        data_object['source'][full_source_barcode] = { 'name' => source_plate_name, 'plate_size' => request.asset.plate.size }
+        source_plate_type = override_plate_type.presence || request.asset.plate.plate_type.tr('_', "\s")
+        data_object['source'][full_source_barcode] = { 'name' => source_plate_type, 'plate_size' => request.asset.plate.size }
       end
+
       if data_object['destination'][full_destination_barcode].nil?
+        target_plate_type = (target_plate.plate_type || PlateType.cherrypickable_default_type).tr('_', "\s")
```
JamesGlover, line 29: Pull out the plate type for the destination in a similar way to the source. This uses the plate type which we recorded earlier. I've left the fallback to the default in cases where a type has not been set.

app/views/robot_verifications/submission.html.erb
```diff
@@ -3,6 +3,7 @@
 <%= form_for :robot_verifications, url: {action: "download"} do |f| %>
 
 <%= page_title "Scan Robot Beds And Plates", "Step 2 of 3" %>
+<% plate_types = PlateType.all.pluck(:name) %>
```
JamesGlover, line 6: Minor optimisation to only grab the plate types once.

app/views/robot_verifications/submission.html.erb
```diff
@@ -49,7 +50,7 @@
        <%   next_field = "destination_bed_barcodes_#{bed_label.to_i +1 }" %>
       <% end %>
       <td><%= text_field_tag "destination_plate_barcodes[#{destination_label}]", params.dig(:destination_plate_barcodes, destination_label), :onkeypress => "return disable_cr_and_change_focus(event,'destination_plate_barcodes_#{destination_label}','#{next_field}')" %></td>
-      <td><%= PlateType.cherrypickable_default_type %></td>
+      <td><%= select_tag("destination_plate_types[#{destination_label}]", options_for_select(plate_types, PlateType.cherrypickable_default_type)) %></td>
```
JamesGlover, line 53: Add the dropdown itself, defaulting to what it was previously hardcoded to. (Which will cover the vast majority of picks)

spec/features/pipelines/cherrypick/cherrypick_pipeline_spec.rb
```diff
@@ -22,30 +23,32 @@
...
   before do
-    assets = plates.each_with_object([]) do |plate, wells|
-      wells.concat(plate.wells)
-      plate.wells.each_with_index do |well, index|
-        well.well_attribute.update!(
-          measured_volume: 30 + (index % 30),
-          current_volume: 30 + (index % 30),
-          concentration: 205 + (index % 50)
-        )
+    step 'Before' do
```
JamesGlover, line 26: This looks like a lot of changes, but its mostly just grouping into steps for improved logging.

spec/features/pipelines/cherrypick/cherrypick_pipeline_spec.rb
```diff
@@ -83,122 +86,143 @@
...
-    expected_file_lines = expected_file.lines
-    # Shift off the comment lines
-    expected_file_lines.shift(2)
+    step 'Robot verification step 2' do
+      plate_type_list = PlateType.all.pluck(:name).join(' ')
+      table = [['Bed', 'Scanned robot beds', 'Plate ID', 'Scanned plates', 'Plate type'],
+               ['SCRC 1', '', 'DN1S', '', plate_type_list],
+               ['SCRC 2', '', 'DN10I', '', plate_type_list],
+               ['SCRC 3', '', 'DN5W', '', plate_type_list],
+               ['DEST 1', '', 'DN99999F', '', plate_type_list]]
```
JamesGlover, line 139: We now expect the list, not a single value

spec/features/pipelines/cherrypick/cherrypick_pipeline_spec.rb
```diff
@@ -83,122 +86,143 @@
...
+      fill_in('SCRC 1', with: '4880000001780')
+      fill_in('DN1S', with: 'DN1S')
+      fill_in('SCRC 2', with: '4880000002794')
+      fill_in('DN5W', with: 'DN5W')
+      fill_in('SCRC 3', with: '4880000003807')
+      fill_in('DN10I', with: 'DN10I')
+      fill_in('DEST 1', with: '4880000020729')
+      fill_in('DN99999F', with: 'DN99999F')
+
+      select 'Custom Type', from: 'destination_plate_types[DN99999F]'
```
JamesGlover, line 152: And we select it from a dropdown

spec/features/pipelines/cherrypick/cherrypick_pipeline_spec.rb
```diff
@@ -83,122 +86,143 @@
...
+      TECAN
+
+      expected_file_lines = expected_file.lines
+      # Shift off the comment lines
+      expected_file_lines.shift(2)
+
+      expect(generated_lines.length).to eq(expected_file_lines.length)
+
+      expected_file_lines.each_with_index do |expected_line, index|
+        expect(generated_lines[index]).to eq(expected_line), "Error on line #{index + 2} in #{expected_file}"
```
JamesGlover, line 224: This assertion was the wrong way round, which caused misleading error messages. I've also added spitting out the line number, and body of the response to make errors easier to diagnose.

spec/support/fetch_table.rb
```diff
@@ -4,11 +4,7 @@ module FetchTable
   def fetch_table(selector)
     find(selector).all('tr').map do |row|
       row.all('th,td').map do |cell|
-        if cell.all('option').present?
-          cell.all('option').collect(&:text).join(' ')
-        else
-          cell.text.squish
-        end
+        cell.text.squish
       end
```
JamesGlover, line 8: ```cell.all('option')``` would find all ```<option>``` elements within a cell, and then would go and extract the text from them, concatenating them together. However, if a cell didn't contain any option elements, it would wait until one appeared, or capybara timed out, before deciding that there weren't any. This meant that fetch table would end up waiting the whole capybara timeout per cell.

Meanwhile, cell.text already returns the concatenated values of any option tags within the cell. I can only assume that an older version of capybara or perhaps poltergeist behaved differently. Still, now the above step is entirely redundant.

Feature test:
Before: 223.03s
After: 6.87s

PR #2656, opened 2020-05-06
Bugfix not display created plates deleted after rollback

emrojo:
Closes RT 689589

* Not display rolled back plates when there is an error on creation
* Added test for this behaviour


emrojo:
Rebased to master with https://github.com/sanger/sequencescape/pull/2657 so I close this pull request

PR #2669, opened 2020-05-11
GPL-452 See if sample is control

KatyTaylor:
Converts existing 'control' field from boolean to enum, so it can hold more information.

Add 'control' field to samples show, plates show and tube_racks show views (screenshots attached).

Add 'control' field to API for use by Limber and Quanthub in coming stories.

To do:
* Wait for reply from David Jackson about whether NPG use existing field (although have now manipulated message so will keep sending consistent data)
* Make new story for backlog for editing this field, whether through UI or manifests

### DIFF
--- a/app/models/api/sample_io.rb
+++ b/app/models/api/sample_io.rb
@@ -30,7 +30,7 @@ class Api::SampleIO < Api::Base
     'control' => '1',
     'positive_control' => '1',
     'negative_control' => '1'
-  }
+  }.freeze
 
   renders_model(::Sample)
 
@@ -80,11 +80,7 @@ class Api::SampleIO < Api::Base
     if json_attributes['reference_genome'].blank?
       json_attributes['reference_genome'] = nil
     end
-    if CONTROL_DATA_MAPPING.has_key?(json_attributes['control'])
-      json_attributes['control'] = CONTROL_DATA_MAPPING[json_attributes['control']]
-    else
-      json_attributes['control'] = nil
-    end
+    json_attributes['control'] = CONTROL_DATA_MAPPING[json_attributes['control']] # sets to nil if not found in hash
   end
 
   # Whenever we create samples through the API we also need to register a sample tube too.  The user
--- a/app/models/api/sample_io.rb
+++ b/app/models/api/sample_io.rb
@@ -21,6 +21,17 @@ class Api::SampleIO < Api::Base
     end
   end
 
+  # Sequencescape field 'control' was changed to an enum May 2020
+  # MLWH field retained as a boolean
+  # because difficult to change the legacy warehouse which receives same message
+  # map all types of control to 1
+  CONTROL_DATA_MAPPING = {
+    'not_control' => '0',
+    'control' => '1',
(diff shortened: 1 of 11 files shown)


app/views/samples/_summary.html.erb
```diff
@@ -26,6 +26,10 @@
       <td class="item">Created at:</td>
       <td><strong><%=h @sample.created_at.to_formatted_s(:day_full_with_time) %></strong></td>
     </tr>
+    <tr>
+      <td class="item">Control:</td>
+      <td><%=h @sample.control_formatted %></td>
+    </tr>
```
KatyTaylor, line 32: Screenshot 2020-05-11 at 16 01 24

app/views/assets/samples_partials/_plate_samples.erb
```diff
@@ -39,6 +41,7 @@
               <td></td>
               <td></td>
               <td></td>
+              <td></td>
```
KatyTaylor, line 44: Screenshot 2020-05-11 at 16 05 56

app/models/api/sample_io.rb
```diff
@@ -69,6 +80,11 @@ def json_root
     if json_attributes['reference_genome'].blank?
       json_attributes['reference_genome'] = nil
     end
+    if CONTROL_DATA_MAPPING.has_key?(json_attributes['control'])
+      json_attributes['control'] = CONTROL_DATA_MAPPING[json_attributes['control']]
```
JamesGlover: I don't think you actually need to check the key here as missing keys return nil by default:
```
2.5.3 :005 >  CONTROL_DATA_MAPPING = {
2.5.3 :006 >         'not_control' => '0',
2.5.3 :007 >         'control' => '1',
2.5.3 :008 >         'positive_control' => '1',
2.5.3 :009 >         'negative_control' => '1'
2.5.3 :010?>     }
 => {"not_control"=>"0", "control"=>"1", "positive_control"=>"1", "negative_control"=>"1"}
2.5.3 :011 > CONTROL_DATA_MAPPING['control']
 => "1"
2.5.3 :012 > CONTROL_DATA_MAPPING['UNRECOGNIZED']
 => nil
2.5.3 :013 > CONTROL_DATA_MAPPING[nil]
 => nil
```

JamesGlover: Also tempted to suggest pulling the control value from the sample directly here, and removing the 'control' section from the standard attributes list above. (I think the sample is accessible via _object)

KatyTaylor: OK, wasn't sure if `map_attribute_to_json_attribute` was doing anything special or not

emrojo: another suggestion, to keep all enum keys in same file, if you get the value from sample directly, then maybe this conversion to boolean can happen in the sample class as well, so here could be like json_attributes['control'] = sample.is_boolean_control? and then in sample class make the conversion?

KatyTaylor:
Closing because we're adding a new field rather than changing the existing one

PR #2670, opened 2020-05-11
Fix - uat actions set study on aliquots

andrewsparkes:
Fixes an issue whereby the UAT actions for generating plates and tube racks were not linking their aliquots to the study. This has a knock on effect that they don't appear in things like the QC reports which select aliquots by study.

### DIFF
--- a/app/uat_actions/uat_actions/generate_plates.rb
+++ b/app/uat_actions/uat_actions/generate_plates.rb
@@ -50,6 +50,7 @@ class UatActions::GeneratePlates < UatActions
         report["plate_#{i}"] = plate.human_barcode
       end
     end
+    true
   end
 
   private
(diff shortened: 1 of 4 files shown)


app/uat_actions/uat_actions/generate_plates.rb
```diff
@@ -50,6 +50,7 @@ def perform
         report["plate_#{i}"] = plate.human_barcode
       end
     end
+    true
```
KatyTaylor, line 53: what effect did this missing return value have?

andrewsparkes: The perform in most of the uat actions returns a boolean (usually the return from the save of the new object). The plate and tube rack performs create multiple objects so were returning a counter number from the loop they were running, not very useful.
There were no tests for these uat actions, so I added some and needed to test the perform worked. Any better idea?

PR #2675, opened 2020-05-12
Add control type field to supplement control field

KatyTaylor:
- stores whether a control is positive or negative
- re-work of https://github.com/sanger/sequencescape/pull/2669

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -272,7 +272,7 @@ class Sample < ApplicationRecord
 
   validate :name_unchanged, if: :will_save_change_to_name?, on: :update
 
-  validate :control_is_true_if_control_type_set
+  validates :control_type, absence: { with: true, unless: :control?, message: 'should be blank if "control" is set to false' }
 
   enum control_type: {
     negative: 0,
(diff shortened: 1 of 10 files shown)


app/models/sample.rb
```diff
@@ -470,10 +477,31 @@ def can_be_included_in_submission?
...
   private
 
   def safe_to_destroy
     errors.add(:base, 'samples cannot be destroyed.')
     throw(:abort)
   end
+
+  def control_is_true_if_control_type_set
+    return unless control_type && control == false
+
```
JamesGlover, line 504: Finding this a little hard to juggle the logic in my head, I understand what its supposed to check (That control_type is only set for controls) but still took a while to unpick.

I think you can achieve something similar with basic rails validations:
```validates :control_type, absence: true, unless: :control?```
Although I'm not sure what the default error message for that is. You might want to provide your custom one (although it will begin with 'Control type' by default.)

File: app/models/sample.rb [hunk 2]
```diff
@@ -470,10 +477,31 @@ def can_be_included_in_submission?
...
+    case control_type
+    when 'positive'
+      'Yes (positive)'
+    when 'negative'
+      'Yes (negative)'
+    else
+      'Yes'
+    end
+  end
+
```
JamesGlover, line 494: Would
```
  def control_formatted
    return nil if control.nil?

    return 'No' if control == false
  "Yes (#{control_type||'Unspecified'})"
  end
```

be better? Its not quite the same behaviour, and perhaps I'm a little obsessed with removing branches.

app/models/sample.rb
KatyTaylor: Yeah we were wondering if we could use some snazzy rails validation. Will give it a go.

PR #2701, opened 2020-05-26
UAT action improvement - set supplier name

KatyTaylor:
get plate and rack UAT action creators to set supplier name
- needed by Samples Extraction to display the contents of the plates or racks
- doesn't slow down creation noticeably
- made to make testing GPL-476 and other Samples Extraction stories easier


KatyTaylor:
Incorporating this into a different pull request

Issue #2697, opened 2020-05-22
Maintenance - Dependency Updates


JamesGlover:
Focus on project Heron has meant some of the more general maintenance tasks have fallen by the wayside. This issue ties together these tasks. Roughly in order of priority:

- [ ] Update anything with identified vulnerabilities
- [ ] Update anything that is passing tests
- [ ] Minor code changes for updates
- [ ] Major updates (eg. Rails. Probably major ruby releases)

JamesGlover:
One of the vulnerabilities required a minor rails version update, which is resulting in a few changes.
1) Sprockets update introduces a manifest file, now updated
2) Seems to be slight changes to the validation on accessioning, currently investigating.

JamesGlover:
Seems to be a case of study not automatically validating the metadata. That's surprising behaviour to change in a minor release, so I suspect its something funky we're doing. (Metadata associations are a bit funky)

JamesGlover:
Yep, its our funky guards interacting with some changes to the way validating associations is performed. Essentially Rails isn't flagging the record as changed, so doesn't trigger the validations.

It appears that rails 4.1 added the ability to use custom validation contexts, which are correctly handled by the changes, and are EXACTLY what we want here.

JamesGlover:
This is obviously an ongoing task, but the particulars of this story are complete.

PR #2709, opened 2020-05-28
Update dependencies inc ruby

JamesGlover:
Progresses #2697

Major dependency updates. Including:
- Rails to 5.2.4.3
- Ruby to 2.6.6


app/api/core/service/content_filtering.rb
```diff
@@ -31,11 +31,7 @@ def process_request_body
     end
 
     def process_response_body
-      headers('Content-Type' => request_accepted)
-    end
-
-    def process_json_response_body
```
pjvv, line 37: Was this part of the upgrade?

app/models/metadata.rb
```diff
@@ -42,48 +42,26 @@ def #{association_name}
         build_#{association_name}
       end
 
-      def validating_ena_required_fields=(state)
```
pjvv, line 45: Were all these functions not required?

spec/factories/purpose_factories.rb
```diff
@@ -145,7 +145,7 @@
     target_type { 'Tube' }
   end
 
-  factory :tube_rack_purpose, class: TubeRack::Purpose do
+  factory :tube_rack_purpose, class: 'TubeRack::Purpose' do
```
pjvv, line 148: Why do these need to be quoted?

app/api/core/service/content_filtering.rb
JamesGlover: Yes. Rack changed its handling of arrays in Content-Type headers. The removal of process_json_response_body was just because it was unused code.

app/models/metadata.rb
JamesGlover: These were always a bit of an abuse of Rails validation, using temporary instance variables to toggle certain validations on or off in particular contexts. A bug-fix in 5.2.4 prevented this from working, as essentially it was trying to detect if the record had changed before validation, and instance variables were ignored in this.

Fortunately Rails 4 added custom validation contexts, which provided exactly the functionality we needed, without the need for any of the old cruft. So while the code was still being used prior to the update, post update it was no-longer working, so has been replaced with the 'official' way of doing things.

spec/factories/purpose_factories.rb
JamesGlover: A Rubocop suggestion. Essentially it allows the class to be lazy-loaded. This means you don't need to load in your entire app every time the factories are loaded. Makes for faster running of individual test files. At least in theory. In practice Sequencescape is such a spider's web that we probably end up loading in most of it anyway.

PR #2760, opened 2020-06-17
James glover/issue2758

JamesGlover:
- Adds metadata to asset audits


app/helpers/application_helper.rb
```diff
@@ -258,6 +258,39 @@ def help_email_link
     link_to admin_address.to_s, "mailto:#{admin_address}"
   end
 
+  #
+  # Handles rendering of JSON to a series of nested lists. Does the following:
+  # String: Rendered as-is
+  # Array: Unordered list (Strictly speaking arrays are ordered, but we probably don't care.)
+  # Object: Descriptive list
+  # Other: Calls to_s
+  # Processes each in turn and clled recursively
```
KatyTaylor, line 267: is `clled` meant to be `called`?

app/helpers/application_helper.rb
```diff
@@ -258,6 +258,39 @@ def help_email_link
...
+  # @param [Hash, String, Array,, #to_s] json The Object to render
+  #
+  # @return [String] HTML formatted for rendering
+  #
+  def render_parsed_json(json)
+    case json
+    when String; then json
+    when Array
+      tag.ul do
+        json.each { |elem, _string| concat tag.li(render_parsed_json(elem)) }
```
KatyTaylor, line 278: `tag.li(render_parsed_json(elem))` should return a list element, right, so what does the `concat` do?

spec/api/asset_audit_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+      let(:response_code) { 422 }
+
+      it 'prevents resource creation' do
+        api_request :post, resources, payload
+        expect(JSON.parse(response.body)).to include_json(JSON.parse(response_body))
+        expect(status).to eq(response_code)
+      end
+    end
+
+    context 'with vaild options' do
```
KatyTaylor, line 98:
```suggestion
    context 'with valid options' do
```

app/helpers/application_helper.rb
JamesGlover: tag.li generates the HTML as a string, concat actually pushes the tag to the page within the ul.

app/helpers/application_helper.rb
JamesGlover: Yep

JamesGlover:
Slight changes to improve formatting
Screenshot 2020-06-18 at 13 22 41

PR #2770, opened 2020-06-18
Fix worksheet on older batches

JamesGlover:
Fixes accessing worksheets on old batches


KatyTaylor:
Closing as this is old, will make a new one

PR #2790, opened 2020-06-25
GPL-539 split worksheets

KatyTaylor:


app/models/robot/pick_data.rb
```diff
@@ -6,18 +6,24 @@ class Robot::PickData
 
   delegate :requests, to: :batch
 
-  def initialize(batch, target_barcode, user: batch.user)
+  def initialize(batch, target_barcode, user: batch.user, max_beds: nil)
     @batch = batch
```
JamesGlover, line 10: The robots with no beds configured are passing in zero to this method, not nil (as nil.to_i is zero, not nil). Further down you treat nil as though it's a robot with infinite beds. If we go for that behaviour, we might be better setting the default to `Float::INFINITY` here, and that way we wouldn't even need to have the conditional in generate_picking_data_hash.

However, given almost everywhere this is used then expects the robots to have matching named beds configures, we're probably just punting the problem further down the line. It might be best just to keep the default max beds as zero, and throw an exception if we see it. At the moment I think we're at risk on an infinite loop.

Background: The robots without max_beds configured aren't Tecans/Hamiltons and haven't been used for cherrypicking. They are in the database purely for maintaining unique barcodes etc. Recently we added a scope to filter then out, so should probably make sure we use that in the dropdown list in the pipeline, and add validation in the robot verification process to make sure they can'r be used.

app/models/robot/pick_data.rb
```diff
@@ -57,6 +63,91 @@ def generate_picking_data
...
+
+      source_barcode = request.asset.plate.machine_barcode
+
+      # find if barcode already encountered
+      pick_to_use = source_barcode_to_pick_number[source_barcode]
+
+      unless pick_to_use
+        # if no max beds, default to all in one pick
+        if @max_beds.nil?
+          pick_to_use = 1
```
JamesGlover, line 100: See comment above. A default value of Float::INFINITY for eliminate the need for this check, but might not be the right behaviour anyway.

app/views/batches/_cherrypick_single_worksheet.html.erb
```diff
@@ -24,65 +30,70 @@
...
+            <% rowchar = (65+row).chr %>
+            <td><strong><%= rowchar %></strong></td>
+            <% (num_columns).times do |column| -%>
+              <% well = plate_wells[row*num_columns+column] -%>
+                <% if  well.present? -%>
+                  <% request = indexed_requests[well.id]  %>
+                  <% source_well = request&.asset %>
+                  <% if source_well.present? -%>
+                    <% source_barcode = source_well.plate.machine_barcode %>
+                    <% well_is_in_this_pick = ( source_layout.has_key?(source_barcode) || (control_layout && control_layout.has_key?(source_barcode)) ) -%>
```
JamesGlover, line 58: Tempted to suggest we still show the other wells, just with a white background and gray text. Means you get an overview of what to expect on the whole plate.

KatyTaylor: I went with this because 
a) if you do a pick with a partial plate, it doesn't seem to show the existing wells, just the ones you're picking, and
b) on the print preview, it doesn't show the colours on the wells

Do you know if the print preview not showing the colours is actually a bug?

app/models/robot/pick_data.rb
KatyTaylor: Thanks for the thoughts. 

I set it to nil and put this behaviour in as the easiest way to keep the existing `expected_layout` method functioning as before, with the expectation that we'd get rid of it and those existing methods once we'd done all the cherrypicking stories.

Can I let you change it as part of your story, as once that's done we can probably get rid of `expected_layout` and related methods?

P.S. I don't think we're at risk of an infinite loop - how?

PR #2793, opened 2020-06-26
Update cucumber

JamesGlover:
Closes #

*
*
* ...


.circleci/config.yml
```diff
@@ -154,7 +154,7 @@ jobs:
     <<: *defaults
     environment:
       RAILS_ENV: cucumber
-      CUCUMBER_FORMAT: DebugFormatter
+      CUCUMBER_FORMAT: summary
```
KatyTaylor, line 157: Where's the docs for what the possible formats are? Is this a cucumber thing or our thing?

JamesGlover: DebugFormatter was our thing, summary is theirs. It's almost identical, just lacks the timestamps (which were added when we were suffering timeouts, and I wanted to find a bottleneck).

This is the official documentation: https://cucumber.io/docs/cucumber/reporting/ but unfortunately its actually out of date, and the `#feature` method it mentions has been removed, which breaks the DebugFormatter.

I actually found the summary formatter while trying to find documentation for the new way of doing things:
https://www.rubydoc.info/github/cucumber/cucumber-ruby/Cucumber/Formatter/Summary

Not sure why its not listed on the docs page, perhaps it was added more recently.

KatyTaylor: Thanks for the explanation

PR #2801, opened 2020-07-01
Release GPL-553 - LHR Cherrypick

KatyTaylor:
Also many depfu updates


app/models/barcode/format_handlers.rb
```diff
@@ -126,7 +126,7 @@ def code128_barcode?
     end
 
     def code39_barcode?
-      %r{\A[A-Z0-9 \-\.$\/\+%]+\z}.match?(@human_barcode)
+      %r{\A[A-Z0-9 \-.$/+%]+\z}.match?(@human_barcode)
```
KatyTaylor, line 129: @JamesGlover what are the removal of the backslashes about in these regexes? Has one of the updates caused a change in ruby regex syntax?

PR #2802, opened 2020-07-01
Gpl 528 add priority to support cgap heron automated release

stevieing:
Closes #

* Add a priority to the samples table.
*
* ...


app/models/sample.rb
```diff
@@ -282,6 +282,13 @@ def release
     positive: 1
   }
 
+  enum priority: {
+    nopriority: 0,
+    backlog: 1,
+    surveillance: 2,
+    priority: 3
+  }
+
```
KatyTaylor, line 291: Looks good. Only small comment would be that no_priority is a bit easier to read than nopriority, so could go for that for the enum value instead?

config/sample_manifest_excel/columns.yml
```diff
@@ -870,3 +870,9 @@ control_type:
   conditional_formattings:
     empty_cell:
     is_error:
+priority:
+  heading: PRIORITY
+  unlocked: true
+  conditional_formattings:
+    empty_cell:
+    number_greater_than_zero:
```
KatyTaylor, line 878: Does this enforce it to be greater than zero in the excel file? Zero is a valid priority isn't it? (no priority)

PR #2804, opened 2020-07-01
Gpl552 control automatic

emrojo:
Closes GPL-552

* Add controls into the created plates in cherrypick


app/models/cherrypick_task.rb
```diff
@@ -126,25 +132,85 @@ def add_empty_well(wells)
...
+    end
+    requests_from_controls = destination_control_wells.map(&:requests_as_target).flatten
+    batch.requests << requests_from_controls
+
+    requests_from_controls
+  end
+
+  # Adds any consecutive list of control requests into the current_destination_plate
+  def add_any_consecutive_control_requests(control_positions, batch, control_assets, current_destination_plate)
+    while (asset_position = control_positions.find_index(current_destination_plate.content.length))
```
KatyTaylor, line 189: Should this be a comparison  == not assignment = ?

KatyTaylor: OK we think we understand it, it will stop when `control_positions.find_index(current_destination_plate.content.length)` returns nil, so asset_position is not truthy?

app/models/cherrypick_task.rb
```diff
@@ -126,25 +132,85 @@ def add_empty_well(wells)
...
-  def pick_new_plate(requests, template, robot, plate_purpose)
+  #
+  # Returns a list with the destination positions for the control wells distributed by
+  # using batch_id and num_plate as position generators.
+  def control_positions(batch_id, num_plate, num_free_wells, num_control_wells)
+    unique_number = batch_id
+
+    # Generation of the choice
+    positions = []
+    available_positions = (0..(num_free_wells - 1)).to_a
```
KatyTaylor, line 143:
```suggestion
    available_positions = (0...num_free_wells).to_a
```

emrojo: Didnt know that 👏

app/models/cherrypick_task.rb
emrojo: Yes, so it assigns and checks condition in same line. I thought it was more tidy than assigning and then checking the condition.

KatyTaylor: The whole method is very neat, but we did have some trouble understanding it :/

I wonder if something more verbose like this could be more readable? (Haven't tested it!)
```  
# Adds any consecutive list of control requests into the current_destination_plate
def add_any_consecutive_control_requests(control_positions, batch, control_assets, current_destination_plate)
  # find the index of the well we are filling right now
  current_well_index = current_destination_plate.content.length

  # check if this well should contain a control
  while control_positions.include?(current_well_index)
    control_asset = control_assets[control_positions.find_index(current_well_index)]
    add_control_request(batch, control_asset, current_destination_plate)
    # above adds to current_destination_plate, so current_well_index should be recalculated
    current_well_index = current_destination_plate.content.length
  end
end
```

Just a suggestion, see what you think. It is a lot longer

app/models/cherrypick_task.rb
```diff
@@ -153,18 +219,40 @@ def perform_pick(requests, robot)
...
+        num_plate += 1
+        control_positions = control_positions(batch.id, num_plate, current_destination_plate.size, control_assets.count)
+      end
     end
 
     plates_hash.each do |request_id, plate_barcode, well_location|
       source_plates << plate_barcode
+      if control_plate
+        # If we are adding control wells we will add everything that is together in this line (column or row)
+        # before passing control to add normal wells
```
KatyTaylor, line 244:
```suggestion
        # If we are automatically placing controls,
        # check if a control should be in the next well position (by looking at 'control_positions' array)
        # if so, add that control and any consecutive controls
        # then add the normal well and carry on
```

app/models/cherrypick_task.rb
```diff
@@ -153,18 +219,40 @@ def perform_pick(requests, robot)
...
+      num_plate = 0
+      batch = requests.first.batch
+      control_assets = control_plate.wells.joins(:samples)
+      control_positions = control_positions(batch.id, num_plate, current_destination_plate.size, control_assets.count)
+    end
+
     push_completed_plate = lambda do
       destination_plates << current_destination_plate.completed_view
       current_destination_plate = yield # reset to start picking to a fresh one
+      if control_plate
```
KatyTaylor, line 233:
```suggestion
      if auto_add_control_plate
```

We both got a bit confused with the name of this variable, cos it sounds like it's saying that the current source plate is a control plate. Can we change it to something like the above? Thanks

app/models/cherrypick_task.rb
KatyTaylor: Yay! I learned about .. and ... together and I can never remember which one's which, have to google it every time :(

app/models/cherrypick_task.rb
emrojo: Yes, is more descriptive and does the same thing. I'll update it.

app/models/cherrypick_task.rb
emrojo: Applied comment

app/models/cherrypick_task.rb
emrojo: Renaming with a better variable name

PR #2807, opened 2020-07-03
Gpl552 control automatic

emrojo:
Create of cp request without order
Rollback linear request graph change


spec/models/tasks/cherrypick_task_spec.rb
```diff
@@ -1,6 +1,150 @@
 require 'rails_helper'
 
 RSpec.describe CherrypickTask, type: :model do
+  let!(:plate) { create :plate_with_untagged_wells, sample_count: 4 }
+  let!(:control_plate) { create :plate_with_untagged_wells, sample_count: 2 }
```
andrewsparkes, line 5: You should be able to use that 'control_plate' factory that we made with 2 samples to to give you what you need instead of this 'plate_with_untagged_wells' which will not be the right plate class (ControlPlate). Can then remove the control sample updates from the before block below.

app/models/cherrypick_task.rb
```diff
@@ -91,8 +91,72 @@ def full?
...
+
+    def remaining_wells(control_positions)
+      remaining_controls = control_positions.select { |c| c > @wells.length }
+      remaining_used_wells = @used_wells.keys.select { |c| c > @wells.length }
+      remaining_controls.concat(remaining_used_wells).flatten
+    end
+
+    def push_with_controls(request_id, plate_barcode, well_location, control_positions, batch, control_assets)
+      @wells << [request_id, plate_barcode, well_location]
+      if control_positions
```
KatyTaylor, line 156: We think `if control_positions` will always return true, because an empty list is truthy

app/models/cherrypick_task.rb
```diff
@@ -91,8 +91,72 @@ def full?
...
+      end
+    end
+
     def push(request_id, plate_barcode, well_location)
       @wells << [request_id, plate_barcode, well_location]
+
+      add_any_wells_from_template_or_partial(@wells)
+      self
+    end
+
```
KatyTaylor, line 147:
```suggestion
# includes control wells and template / partial wells that are yet to be added
```

app/models/cherrypick_task.rb
```diff
@@ -91,8 +91,72 @@ def full?
...
+      remaining_controls = control_positions.select { |c| c > @wells.length }
+      remaining_used_wells = @used_wells.keys.select { |c| c > @wells.length }
+      remaining_controls.concat(remaining_used_wells).flatten
+    end
+
+    def push_with_controls(request_id, plate_barcode, well_location, control_positions, batch, control_assets)
+      @wells << [request_id, plate_barcode, well_location]
+      if control_positions
+        add_any_consecutive_control_requests(control_positions, batch, control_assets)
+        add_remaining_control_requests(control_positions, batch, control_assets) if (@wells.length + remaining_wells(control_positions).length) == @size
```
KatyTaylor, line 158: add comment `# This assumes that the template wells will fall at the end of the plate`

app/models/cherrypick_task.rb
KatyTaylor: add comment `# would be nil if no control plate selected`

spec/models/tasks/cherrypick_task_spec.rb
emrojo: Ok, changed

app/models/cherrypick_task.rb
emrojo: Added

app/models/cherrypick_task.rb
emrojo: Added

PR #2810, opened 2020-07-06
Added basic file content checks for pick files

andrewsparkes:
Adds a few pick file line checks for feature tests

### DIFF
--- a/spec/features/shared_examples/cherrypicking.rb
+++ b/spec/features/shared_examples/cherrypicking.rb
@@ -243,11 +243,14 @@ shared_examples 'a cherrypicking procedure' do
             case robot.generation_behaviour_property.value
             when 'Hamilton'
               # for Robot::Generator::Hamilton
+              # Hamilton files comprise a column headers row plus one row per transfer e.g.
+              # SourcePlateID,SourceWellID,SourcePlateType,SourcePlateVolume,DestinationPlateID,DestinationWellID,DestinationPlateType,DestinationPlateVolume,WaterVolume
+              # DN1000001A,A1,ABgene 0765,15.85,DN20000001B,A1,ABgene 0800,15.85,49.15
               generated_file = DownloadHelpers.downloaded_file("#{batch_id}_batch_#{destination_barcode}_#{pick_number_index}.csv", timeout: 10)
               generated_lines = generated_file.lines
 
-              # check generated lines match expected by calculation - headers row + one row per sample/control transfer
-              expected_num_lines = current_destination_plate.wells.map(&:transfer_requests_as_target).count + 1
+              # check generated lines match expected by calculation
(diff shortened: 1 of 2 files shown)


spec/features/shared_examples/cherrypicking.rb
```diff
@@ -237,14 +237,27 @@
...
             # robot file generation differs by generator
             case robot.generation_behaviour_property.value
             when 'Hamilton'
               # for Robot::Generator::Hamilton
               generated_file = DownloadHelpers.downloaded_file("#{batch_id}_batch_#{destination_barcode}_#{pick_number_index}.csv")
               generated_lines = generated_file.lines
 
-              # optionally if an expected file was supplied. compare it to the result
+              # check generated lines match expected by calculation - headers row + one row per sample/control transfer
+              expected_num_lines = current_destination_plate.wells.map(&:transfer_requests_as_target).count + 1
```
KatyTaylor, line 250: why do you only check num_lines for the Hamilton?

andrewsparkes: Because calculating it for the complex Tecan files is far from straight forward.
The Tecan has an additional 4 buffer lines (3 text one blank) for any destination wells that require buffer, which depends on what concentrations you've set in your test for each of the source samples.
Looking at it next but releasing what I've done so far for the UAT release.

andrewsparkes: Hamilton is much easier as the buffer is incorporated into the same sample row.

spec/features/shared_examples/cherrypicking.rb
```diff
@@ -260,6 +273,14 @@
               generated_file = DownloadHelpers.downloaded_file("#{batch_id}_batch_#{destination_barcode}_#{pick_number_index}.gwl")
               generated_lines = generated_file.lines
 
+              # check count of controls present in destination file lines is correct
+              # NB. Tecan file has additional plate barcode lines at bottom of file for beds so add 1
```
emrojo, line 277: Maybe we could define this +1 in a variable so the name describes why is added (like tecan_aditional_lines=1 and then +tecan_aditiona_lines in line 281)

spec/features/shared_examples/cherrypicking.rb
emrojo, line 250: Is that +1 the header?

andrewsparkes: Yes a single header row.
Tecan files have additional info at the top of the file e.g. user/date.
Hamilton files are just csv's with a column headers row at the top and one row per transfer..

spec/features/shared_examples/cherrypicking.rb
andrewsparkes: Ok will make this clearer.
Tecan has extra lines at the bottom of the file, one per plate, e.g. SCRC1 = DN1112124A
This is for the Tecan software on the robot to assign the beds I think.
Hamilton/Beckman files don't have this.

PR #2824, opened 2020-07-09
Gpl 565 select all bug

KatyTaylor:


KatyTaylor:
Triggering the change event means that the js code in show.html.erb picks up that the checkboxes have changed and updates the count.

These functions / classes are also used in print_labels.html.erb and print_plate_labels.html.erb, which are both links off the batch page, and I tested them both - they seem to still be fine (they don't have a count).

PR #2843, opened 2020-07-15
Gpl571 cherrypicking - limiting control locations

### LINKED ISSUE #2833: GPL-571 cherry-picking controls - tweek of the random control locations [RT #694235]
**User story**
GPL-571 | As library team representative (Ben Farr) I would like specific wells to be excluded from the (random) cherry picking control locations as they are required for downstream quantification control standards

Rich C
Iraad B

**Acceptance criteria**
- [ ] exclude locations below from the random cherry picking control locations algorithm

**Additional context**
Rich C wrote ( Mon Jul 13 15:35:12 2020)
"a post library quantification step in the 384 well process sacrifices some plate wells for quantitation standards. While its acceptable to sacrifices some library quants on a plate, sacrificing the control quants is not. Please can locations of cherrypicking controls exclude the following positions"

andrewsparkes:
Closes #2833

### DIFF
--- a/app/models/cherrypick_task.rb
+++ b/app/models/cherrypick_task.rb
@@ -11,6 +11,10 @@
 class CherrypickTask < Task
   EMPTY_WELL          = [0, 'Empty', ''].freeze
   TEMPLATE_EMPTY_WELL = [0, '---', ''].freeze
+  # Leaving the first three columns (3 x 8) of a 96-well plate free of controls so when stamped into 384-well
+  # there are no controls in the first 6 columns, as in QC steps standards go into some of these wells on the reader plate
+  CONTROL_START_INDX_96 = 24
+  CONTROL_START_INDX_OTHER = 0
 
   # An instance of this class represents the target plate being picked onto.  It can have a template
   # and be a partial plate, and so when wells are picked into it we need to ensure that we don't hit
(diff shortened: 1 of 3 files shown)


app/models/cherrypick_task.rb
```diff
@@ -217,11 +221,11 @@ def control_positions(batch_id, num_plate, num_free_wells, num_control_wells)
```
emrojo, line 221: Yes, sorry, my mistake. I think the name `num_free_wells` is not the right name.
The value is coming from PlateTemplate.size, which it is giving the size of the labware (96,384, etc...) not the number of free wells in the labware. 
I think we could rename it to something more correct (total_wells, num_wells, ...)

app/models/cherrypick_task.rb
```diff
@@ -11,6 +11,10 @@
 class CherrypickTask < Task
   EMPTY_WELL          = [0, 'Empty', ''].freeze
   TEMPLATE_EMPTY_WELL = [0, '---', ''].freeze
+  # Leaving the first three columns (3 x 8) of a 96-well plate free of controls so when stamped into 384-well
+  # there are no controls in the first 6 columns, as in QC steps standards go into some of these wells on the reader plate
+  CONTROL_START_INDX_96 = 24
+  CONTROL_START_INDX_OTHER = 0
```
emrojo, line 17: That comment is very useful. Also a good thing that is at the top of the file.

app/models/cherrypick_task.rb
```diff
@@ -114,23 +118,23 @@ def add_control_request(batch, control_asset)
     end
 
     # Adds any consecutive list of control requests into the current_destination_plate
-    def add_any_consecutive_control_requests(control_positions, batch, control_assets)
+    def add_any_consecutive_control_requests(control_posns, batch, control_assets)
```
emrojo, line 121: Thanks for this renaming. It was very confusing to have variables and methods using same word ```control_positions```.

PR #2858, opened 2020-07-24
Depfu/batch all/2020 07 18 + Race condition fix

JamesGlover:
Depfu changes plus fixing the race condition triggered in its last build.


KatyTaylor:
Possibly unrelated to your changes, but, I just noticed it's modifying the same file as we had some issues with recently. See pull request https://github.com/sanger/sequencescape/pull/2853/files, where a couple of changes were made to this file. They don't seem to be displaying in your pull request? (If you click the up arrow on the diff to show more of the file). Is this just Github's diff being buggy, or is there something wrong?

spec/support/download_helper.rb
```diff
@@ -42,6 +42,11 @@ def self.remove_downloads
 
   def self.create_directory
     PATH.parent.mkdir unless PATH.parent.exist?
-    PATH.mkdir
+    PATH.mkdir unless PATH.exist?
+  rescue Errno::EEXIST
+    # Saw this on travis, suggesting a race condition
+    # probably chrome creating the directory for the actual download.
+    # We retry, as we could potentially have failed creating the parent directory.
+    retry
```
KatyTaylor, line 50: So is this solving an issue where the directory does not exist at line 8, but by the time it's got to line 45 it does exist? 
When would Errno:EEXIST get thrown?

Thanks

JamesGlover: Yep, Errno:EEXIST gets thrown by PATH.mkdir (or PATH.parent.mkdir) if the directory already exists. Its actually even possible to get the exception thrown on line 44, as the line isn't atomic, so you could check for the existence, and find it missing, have chrome create the directory immediately after, and then try to create the directory yourself.

I've only seen the error the once, so it probably isn't a common occurrence, but I hate seemingly random test failures.

JamesGlover:
Regarding the diffs. I think this branch was forked off before your changes were made (I based it directly on depfu's branch) hence why the other changes aren't showing. It also wont show up in diffs as:
https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-comparing-branches-in-pull-requests
> By default, pull requests on GitHub show a three-dot diff, or a comparison between the most recent version of the topic branch and the commit where the topic branch was last synced with the base branch.
So the diff isn't agains HEAD.

KatyTaylor:
Ah that's useful to know, thank you. I think Bitbucket shows you the diff between the current base branch and the topic branch (https://blog.developer.atlassian.com/a-better-pull-request/) - so that's what I'm more used to.

PR #2861, opened 2020-07-28
Fix tag groups dropdown

JamesGlover:
- Fixes width of tag groups box (See screenshots)
- Fixes select 2 option on tag dropdowns
- Fixes performance issues for large lanes

Before
Screenshot 2020-07-28 at 10 15 40

After
Screenshot 2020-07-28 at 10 37 29

### DIFF
--- a/app/views/tag_substitutions/new.html.erb
+++ b/app/views/tag_substitutions/new.html.erb
@@ -94,7 +94,7 @@
     // Without this optimization select2 takes about 140s to update the DOM.
     // The majority of this is spent in the select2 _resolveWidth function, which
     // tries to match the width of the select2 element, to that of the original
-    // element. Providing a fixed withd overides this.
+    // element. Providing a fixed width overrides this.
     // It still takes about 30s to render, which isn't great, but somewhat more
     // tolerable.
     $(".tag-list").select2({
--- a/app/views/tag_substitutions/new.html.erb
+++ b/app/views/tag_substitutions/new.html.erb
@@ -37,15 +37,28 @@
         </tr>
       </thead>
       <tbody>
-        <% @tag_substitution.substitutions.each do |substitution| %>
+        <% @tag_substitution.substitutions.each_with_index do |substitution, index| %>
           <%= form.fields_for :substitutions, substitution, index: nil do |sub| %>
             <tr>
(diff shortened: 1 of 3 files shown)


app/views/tag_substitutions/new.html.erb
```diff
@@ -37,15 +37,28 @@
...
+              <td><%= sub.select :substitute_tag_id, tag_options_for(substitution.substitute_tag_id),
+                                 {},
+                                 class: 'tag-list', id: "ts_st_id_#{index}"  %></td>
+              <td>
+                <%= sub.hidden_field :original_tag2_id, id: "ts_ot2_id_#{index}" %>
+                <%= tag_name(substitution.original_tag2_id) %>
+              </td>
+              <td><%= sub.select :substitute_tag2_id, tag_options_for(substitution.substitute_tag2_id),
+                                 {},
+                                 class: 'tag-list' , id: "ts_st2_id_#{index}" %></td>
```
KatyTaylor, line 61: Why are you adding the ids? Doesn't look like they are used?

app/views/tag_substitutions/new.html.erb
```diff
@@ -75,6 +88,20 @@
       });
       $(".tag-list").append(optgroup);
     });
+
+    // We don't use the default .select2 binding here for performance reasons.
+    // With very high plexed lanes we end up rendering a LOT of select elements.
+    // Without this optimization select2 takes about 140s to update the DOM.
+    // The majority of this is spent in the select2 _resolveWidth function, which
+    // tries to match the width of the select2 element, to that of the original
+    // element. Providing a fixed withd overides this.
```
KatyTaylor, line 97:
```suggestion
    // element. Providing a fixed width overrides this.
```

app/views/tag_substitutions/new.html.erb
JamesGlover: Otherwise the rails form-tags will generate their own ids, and unfortunately they generate the same id for each set of substitution fields. Not only does this result in lots of warnings in the developer console, but also stops the select2 extension from properly binding to the dropdowns. (Select2 provides a search box inside the select field, which makes it much easier to find the relevant tag)

So really just adding an id to ensure they are unique, and make sure the HTML remains valid.

PR #2873, opened 2020-08-07
Modified ean13_barcode to human_barcode for label printer base_tube.

### LINKED ISSUE #2866: GPL-605 Unable to print barcode from multiplexed library manifests
**Describe the bug**
Checked a few others (and older) and the bug is for all **multiplexed library manifests** not specific to this instance
A NoMethodError occurred in sample_manifests#print_labels:
  undefined method `ean13_barcode' for #<Receptacle:0x00007f9867239658>
  lib/label_printer/label_printer/label/base_tube.rb:35:in `barcode'

**RT Ticket Number**
https://rt.sanger.ac.uk/Ticket/Display.html?id=695727

**To Reproduce**
http://sequencescape.psd.sanger.ac.uk/sdb/sample_manifests/15799
select printer g216bc (or other)
click "Reprint all labels"
=> "We're sorry, but something went wrong."

**Expected behaviour**
banner with labels printed to g216bc

**Additional context**
From slw (in RT ticket): "...  This isn't good for us at the moment as we are reliant on these reprint pages for barcodes as we are no longer on site to pick up barcodes as soon as they are printed. We need a working link for which ever SSR is going onsite to collect and sort barcodes for us. This is therefore quite an urgent one to be fixed if possible please?"

stevieing:
Closes #2866 

*
*
* ...


lib/label_printer/label_printer/label/base_tube.rb
```diff
@@ -32,7 +32,7 @@ def round_label_bottom_line(tube)
       end
 
       def barcode(tube)
-        tube.ean13_barcode
+        tube.human_barcode
```
KatyTaylor, line 35: Just wondering whether it would be more correct to use machine_barcode here (haven't tested it). 

Thinking back to a conversation about a similar problem I had with James in Slack on 9th July (pasted below!):

Katy  3:56 PM
Can someone who knows more about barcodes than me help with something?
- The stock_resource table in MLWH has a column 'labware_machine_barcode', which can't be null
- It's currently being populated with the ean13_barcode method on plate
- Foreign barcodes don't have an ean13 barcode
Option 1: Fill it with the human barcode if the ean13 barcode doesn't exist
Option 2: Change the MLWH column to allow nulls (edited) 

I'm thinking option 2, because there's another column that contains the human barcode already, but I guess making an existing column allow nulls might mess up some downstream processing?

James  3:58 PM
Use machine_barcode instead

It’ll mostly be the same as human_barcode now, but automatically falls back to the ean13_barcode for older plates

We should probably rebroadcast some of the plates created recently, as the intent is to provide a way of knowing what the plate will scan asin order to join onto tables from, eg. labwhere (edited) 

Katy  4:03 PM
OK so I just tested a Lighthouse-created plate and a sample manifest-created plate, in UAT, and machine_barcode gives me the human barcode even for the second plate

Is that correct

Whereas ean13_barcode gives me a different number

James  4:04 PM
Yep, we changed the format barcodes were printed in to code39, which means that the actual barcode contains the same number as the text printed on it

Katy  4:04 PM
OK great thanks

James  4:05 PM
If you do something plate Plate.first.machine_barcode you’ll see the ean13 (Theoretically at least, I wouldn’t be surprised if the first plate has no barcode)

stevieing: I took a punt. It was either machine_barcode or human_barcode guess I backed the wrong horse. Happy to change.

PR #2890, opened 2020-09-01
Gpl615 qcreport memory issues

### LINKED ISSUE #2877: GPL-615 BUG - Qc report failed with out of memory Error.
**Describe the bug**
Qc report created Thu, 13 Aug 2020 20:28:58 +0100 seems to have failed twice, with out of memory Error.
See journalctl at Aug 14 05:37:46 on Sequencescape worker 0

**RT Ticket Number**
If applicable

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**

JamesGlover:
Closes #2877

* Removes transaction added by aasm
* Resolves N+1 loading issues
* Adds object space profiler

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ group :default do
 
   # State machine
   gem 'aasm'
+  gem 'after_commit_everywhere', '~> 0.1', '>= 0.1.5' # aasm requirement
   gem 'configatron'
   gem 'formtastic'
   gem 'rest-client' # curb substitute.
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -33,7 +33,7 @@ GIT
 GEM
   remote: https://rubygems.org/
   specs:
-    aasm (5.0.8)
+    aasm (5.1.1)
       concurrent-ruby (~> 1.0)
     actioncable (5.2.4.3)
       actionpack (= 5.2.4.3)
@@ -96,6 +96,8 @@ GEM
       tzinfo (~> 1.1)
     addressable (2.7.0)
       public_suffix (>= 2.0.2, < 5.0)
+    after_commit_everywhere (0.1.5)
+      activerecord (>= 4.2)
     amq-protocol (2.3.2)
     arel (9.0.0)
     ast (2.4.1)
@@ -137,7 +139,7 @@ GEM
     choice (0.2.0)
     chronic (0.10.2)
     coderay (1.1.3)
-    concurrent-ruby (1.1.6)
+    concurrent-ruby (1.1.7)
     configatron (4.5.1)
     connection_pool (2.2.3)
     crack (0.4.3)
@@ -555,6 +557,7 @@ DEPENDENCIES
   activerecord-oracle_enhanced-adapter
   activeresource
   acts-as-dag!
+  after_commit_everywhere (~> 0.1, >= 0.1.5)
   axlsx!
   bootsnap
   bootstrap
--- a/app/models/aliquot.rb
+++ b/app/models/aliquot.rb
@@ -44,8 +44,8 @@ class Aliquot < ApplicationRecord
   # An aliquot is held within a receptacle
   belongs_to :receptacle, inverse_of: :aliquots
 
-  belongs_to :tag
-  belongs_to :tag2, class_name: 'Tag'
+  belongs_to :tag, optional: true
+  belongs_to :tag2, class_name: 'Tag', optional: true
 
   # An aliquot can belong to a study and a project.
   belongs_to :study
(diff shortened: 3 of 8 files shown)


Gemfile
```diff
@@ -8,6 +8,7 @@ group :default do
 
   # State machine
   gem 'aasm'
+  gem 'after_commit_everywhere', '~> 0.1', '>= 0.1.5' # aasm requirement
   gem 'configatron'
```
JamesGlover, line 12: Added as soft requirement for latest AASM to fix some race conditions.

Gemfile.lock
```diff
@@ -33,7 +33,7 @@ GIT
 GEM
   remote: https://rubygems.org/
   specs:
-    aasm (5.0.8)
+    aasm (5.1.1)
       concurrent-ruby (~> 1.0)
```
JamesGlover, line 37: Updated aasm as I was aware they'd changed their transaction handling, and didn't wanting to waste time with a fix which could break immediately.

app/models/aliquot.rb
```diff
@@ -44,8 +44,8 @@ class InsertSize < Range
   # An aliquot is held within a receptacle
   belongs_to :receptacle, inverse_of: :aliquots
 
-  belongs_to :tag
-  belongs_to :tag2, class_name: 'Tag'
+  belongs_to :tag, optional: true
+  belongs_to :tag2, class_name: 'Tag', optional: true
```
JamesGlover, line 49: This doesn't actually change the behaviour. By default more recent versions of rails make belongs_to required by default. We're enabling the legacy setting elsewhere, but this will make it easier to adopt the new defaults in future.

app/models/product_criteria/basic.rb
```diff
@@ -67,7 +67,7 @@ def well_location
   end
 
   def plate_barcode
-    @well_or_metric.plate.try(:human_barcode) || 'Unknown'
+    @well_or_metric.labware.try(:human_barcode) || 'Unknown'
   end
```
JamesGlover, line 71: Plate and labware are actually different associations, and labware is the one we're eager loading. Would be much nicer if rails let you alias an association. (As it is we can alias getters, but last I checked we couldn't alias for things like eager loading)

app/models/qc_report.rb
```diff
@@ -15,9 +15,13 @@ def available_states
...
         # When adding new states, please make sure you update the config/locals/en.yml file
-        # with decriptions.
+        # with descriptions.
 
-        aasm column: :state, whiny_persistence: true do
+        # We disable the transactions here as otherwise our entire job gets wrapped in a transaction.
+        # This causes problems for very large reports, as it causes objects with after transaction callbacks to
+        # be persisted for the entire job. We'd already chunked our transaction into batches, but this transaction
+        # was counteracting that.
+        aasm column: :state, whiny_persistence: true, use_transactions: false do
```
KatyTaylor, line 24: Just to clarify:
Would it do one transaction per state transition, so:
* One for `event :generate`
* One for `state :generating, after_enter: :generate_report`, which stretches on until the `generate_report` function has finished? ( which include the `generation_complete!` event )

i.e. the 'after transaction callbacks' you're referring to is `after_enter: :generate_report` in this case?

Thanks

lib/object_space_profile.rb
```diff
@@ -0,0 +1,55 @@
+# ObjectSpaceProfile provides a tool for assessing how many Ruby objects are in
+# memory. By taking measures over time you can watch the change in the number of
+# objects, and identify situations in which we're holding on to more objects
+# than intended. Sometimes it may be necessary to generate a report with each
+# cycle, especially if performance means you aren't getting to the end of the
+# process.
```
KatyTaylor, line 6: Not quite sure what you mean by `Sometimes it may be necessary to generate a report with each
# cycle, especially if performance means you aren't getting to the end of the
# process.`

Do you mean, necessary to generate the output of this profiler for one iteration of the loop in `study.each_well_for_qc_report_in_batches` (qc_report.rb)?

JamesGlover: Yes, I had to spit one out in each iteration, as it was dying before it even got past the third loop through. (Although this was partly because the initial version of the profiler was interacting very badly with configatron, and spawned new objects when it attempted to interrogate the settings there).

app/models/qc_report.rb
JamesGlover: I believe it wraps the entire event and the callbacks on the state transitions as well. There's an outline here:
https://github.com/aasm/aasm#lifecycle
So in this case it would begin at generate, and encompass everything through to the end of generate_report.

I'm not sure if it would open a 'nested' transaction for the 'generation_complete!' event, or if it's happy to use the existing one. I'd need to double check the aasm code.

JamesGlover: Ahh, actually, I think this behaviour is configurable. You can enforce it opening a new transaction per state transition. https://github.com/aasm/aasm#transaction-support

PR #2895, opened 2020-09-03
Fix error handling

JamesGlover:
Push errors out to the workflow_controller, not the workflow.
In a sane world we'd have the task return false and populate and error message, rather than  constantly making calls into the controller.

Closes #

*
*
* ...

### DIFF
--- a/app/models/cherrypick_task.rb
+++ b/app/models/cherrypick_task.rb
@@ -315,7 +315,7 @@ class CherrypickTask < Task
   def do_task(workflow_controller, params)
     workflow_controller.do_cherrypick_task(self, params)
   rescue Cherrypick::Error => e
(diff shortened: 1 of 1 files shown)


app/models/cherrypick_task.rb
```diff
@@ -315,7 +315,7 @@ def render_task(workflow_controller, params)
   def do_task(workflow_controller, params)
     workflow_controller.do_cherrypick_task(self, params)
   rescue Cherrypick::Error => e
-    workflow.send(:flash)[:error] = e.message
+    workflow_controller.send(:flash)[:error] = e.message
```
KatyTaylor, line 318: Ah I think this is my bad - I changed the parameter name a while back?

KatyTaylor (approved):

PR #2903, opened 2020-09-10
Montecarlo test controls

emrojo:
Add a random test


JamesGlover:
Looks like a version of this has already been merged in, and issue #2967 will make this redundant. So closing.

PR #2918, opened 2020-09-29
Heron Tailed fix

KatyTaylor:
Change the Order Role from 'Heron LTHR' to 'LTHR' so that the longer Purpose names fit on the barcode label


JamesGlover (approved):
This

PR #2925, opened 2020-10-01
GPL-662 plate picks sorting & GPL-663 shorter short names

KatyTaylor:


app/models/batch.rb
```diff
@@ -258,6 +258,10 @@ def output_plate_group
   def output_plates
     # We use re-order here as batch_requests applies a default sort order to
     # the relationship, which takes preference, even though we're has_many throughing
+    if output_labware.loaded?
+      return output_labware.sort_by(&:id)
+    end
+
```
KatyTaylor, line 264: @JamesGlover when would we expect to hit this condition (already loaded)? I tried a couple of things but couldn't make it go in here

JamesGlover: Looks like I was mistaken about us eager loading the required information for sorting a batch. I assumed we had done something like:

```
Batch.includes(:output_labware, :some_other_stuff).find...
```

However, it looks like we just loads the batch, and do all out eager loading when we extract the requests to build the data_object. (Which probably makes more sense, as we only need to handle the eager loading in one place)

So essentially we don't hit this here, and would only hit it is we did do something like the above.

JamesGlover (approved):

app/models/batch.rb
KatyTaylor: OK thanks

PR #2941, opened 2020-10-15
Gpl 679 p wgs sequencing template

emrojo:
Closes #

*
*
* ...


lib/tasks/limber.rake
```diff
@@ -572,8 +580,10 @@ namespace :limber do
         Limber::Helper::TemplateConstructor.new(prefix: prefix,
                                                 catalogue: catalogue,
                                                 sequencing_keys: params[:sequencing_list]).build!
-        Limber::Helper::LibraryOnlyTemplateConstructor.new(prefix: prefix, catalogue: catalogue).build!
-        Limber::Helper::LibraryAndMultiplexingTemplateConstructor.new(prefix: prefix, catalogue: catalogue).build!
+        unless params[:omit_library_templates]
+          Limber::Helper::LibraryOnlyTemplateConstructor.new(prefix: prefix, catalogue: catalogue).build!
+          Limber::Helper::LibraryAndMultiplexingTemplateConstructor.new(prefix: prefix, catalogue: catalogue).build!
+        end
```
KatyTaylor, line 586: Did you verify that they didn't want these 'library only' and 'library and multiplexing' templates, on the call?

lib/tasks/limber.rake
```diff
@@ -258,7 +258,7 @@ namespace :limber do
   task create_request_types: [:environment, :create_plates] do
     puts 'Creating request types...'
     ActiveRecord::Base.transaction do
-      %w[WGS LCMB].each do |prefix|
+      %w[WGS LCMB pWGS].each do |prefix|
```
KatyTaylor, line 261: Noticed you aren't adding a library type. 
* The Confluence page specified "Library type must be pWGS-384", 
* and I noticed there was an action in the email chain - "ACTION: Lesley will clarify parameters on the requred libarty type"

was this resolved for certain? NPG, the team downstream of us who sort out the sequencing data, tend to care about library types.

lib/tasks/limber.rake
Chris-Friend: Yes we verified they only wanted the 3 included submission templates, but I will check this

lib/tasks/limber.rake
Chris-Friend: I will check this

lib/tasks/limber.rake
Chris-Friend: It has been confirmed that the library only & library and multiplexing (Limber-Htp - pWGS and Limber-Htp - pWGS - Pool) submission templates will not be required

lib/tasks/limber.rake
Chris-Friend: Yep you were correct. I have updated to instead use the `pWGS-384` library type

KatyTaylor (approved):

Chris-Friend:
The name change from pWGS to pWGS-384 was required to avoid conflicting with existing PWGS submission templates that already exist in the UAT/Prod databases

JamesGlover:
With most 384 well variations we use the same request-type as the 96 well version, and just add another library type. This avoids the need for a separate submission template. This approach works fine too, but just thought I'd point it out here.

Chris-Friend:
There isn't a 96 well version of the pWGS pipeline configured in this rake task that I can see. And in the 384 pipeline we are using the `Standard` request_type, but a different library type: `pWGS-384`

JamesGlover:
Hmm, you're right. We do have a PWGS request type registered, but the vast majority of the templates associated with it have been deactivated. I've I'm remembering correctly we originally intended to have separate PWGS and MWGS mamalian) pipelines, and then because they were identical, decided to merge them together into a single WGS pipeline.

What I find a bit confusing is that the original PWGS request type was unused for three years, and then got used for the first time again yesterday, as someone reactivated it with the Limber-Htp - pWGS - NovaSeq 6000 Paired end sequencing.
This will be an entirely different pipleine to the normal PWGS.

JamesGlover:
But given the above, it probably does make sense to keep them separate from the WGS request types, as it doesn't sound like they'll be using the pipeline for Mamalian stuff.

PR #2945, opened 2020-10-20
Fix up rake tasks for integration suite

JamesGlover:
Fixes rake tasks for integration suite support.


JamesGlover:
Also see: https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/18

PR #2946, opened 2020-10-21
develop into master for duplex seq

andrewsparkes:
GPL295 Duplex Seq pipeline


db/schema.rb
```diff
@@ -1389,7 +1394,6 @@
     t.boolean "control"
     t.boolean "empty_supplier_sample_name", default: false
     t.boolean "updated_by_manifest", default: false
-    t.boolean "migrated_consent_withdrawn_to_metadata", default: false, null: false
```
KatyTaylor, line 1392: Bit weird that this is still hanging around. Seems legit to remove it though as it's not there on the table in prod.

KatyTaylor (approved):

PR #2964, opened 2020-11-26
issue2963/gpl 712 as a user when the destination plate is ready i would like to scan the barcode into the limber pipeline and decide whether it will be processed as a 384 or 96 well plate

JamesGlover:
Adds changes for GPL-712 with some code cleanup

- Upgrade Ruby to version 2.7.2
- Improve well generation speed
- Cleanup benchmark code
- All submissions can auto-detect their study/project
- Fix Sequencescape::Application warning
- Expose primer panel in orders API
- Allow fragment size to be flattened
- Expose state and timestamps on submission
- Fix some scope warnings
- Expose submissions on plate via API
- Improve primer panel validation
- Add record loaders
- Add missing order file
- Fix tag plate UAT action in quick succession
- Improve tag layout templates page
- Include primer panels loader rake task
- Rubocop


app/models/asset.rb
```diff
@@ -70,7 +70,6 @@ class Asset < ApplicationRecord
   scope :include_requests_as_source, -> { includes(:requests_as_source) }
 
   scope :sorted, ->() { order('map_id ASC') }
-  scope :for_summary, -> { includes(:map, :barcodes) }
```
KatyTaylor, line 73: Hey, I read your commit message for these scope removals. Can you explain it to me? In what sense is this re-defining a scope? Where is it being defined for the first time? Thanks

app/models/concerns/programs_validator.rb
```diff
@@ -20,6 +22,13 @@ def validate_each(record, attribute, value)
 
   private
 
+  def check_hash(record, attribute, value)
+    return true if value.is_a?(Hash)
+
+    record.errors.add attribute, "is a #{value.class.name}, not a hash"
+    false
+  end
+
```
KatyTaylor, line 31: What is a Program? Does it represent a method run on a robot or something?

app/models/labware.rb
```diff
@@ -29,7 +29,16 @@ class Labware < Asset
...
   has_many :transfer_requests_as_target, through: :receptacles
+
+  # Submissions in progress for the labware. Found by looking at those associated
+  # with the transfer requests into the receptacles. Be a little cautious using this,
+  # as it will not handle cross-submission pools, but is better for historical
+  # data.
   has_many :submissions, through: :receptacles
+
+  # Direct submissions are those made on the plate itself, and are found via
+  # the orders associated with the well.s
```
KatyTaylor, line 40:
```suggestion
  # the orders associated with the wells
```

app/models/labware.rb
```diff
@@ -29,7 +29,16 @@ class Labware < Asset
   has_many :requests_as_target, through: :receptacles
   has_many :transfer_requests_as_source, through: :receptacles
   has_many :transfer_requests_as_target, through: :receptacles
+
+  # Submissions in progress for the labware. Found by looking at those associated
+  # with the transfer requests into the receptacles. Be a little cautious using this,
+  # as it will not handle cross-submission pools, but is better for historical
+  # data.
```
KatyTaylor, line 36: Is it necessarily submissions 'in progress'? Is it filtering by state of the transfer request somewhere?

Why does it not include cross-submission pools? (Think I might need a diagram here!!)

app/models/labware.rb
```diff
@@ -29,7 +29,16 @@ class Labware < Asset
...
   has_many :submissions, through: :receptacles
+
+  # Direct submissions are those made on the plate itself, and are found via
+  # the orders associated with the well.s
+  has_many :direct_submissions, -> { distinct }, through: :receptacles
```
KatyTaylor, line 41: Sorry so many questions on this file...
It looks from the comment like this should be going via a different route through the data model, but looks identical to the above relationship apart from the `distinct` - how does that make it go through wells / orders rather than transfer requests?

app/models/plate.rb
```diff
@@ -37,13 +37,19 @@ class Plate < Labware
...
+        plate.maps.in_row_major_order.ids.map do |location_id|
+          { map_id: location_id }
+        end.tap do |wells|
+          plate.wells.import(wells)
+          ids = plate.wells.ids
+          well_type = Well.base_class.name
+          WellAttribute.import(ids.map { |well| { well_id: well } })
+          Uuid.import(ids.map { |well| { resource_id: well, resource_type: well_type, external_id: Uuid.generate_uuid } })
+          ids.each { |id| Warren::QueueBroadcastMessage.new(class_name: 'Well', id: id).queue(Warren.handler) }
+        end
```
KatyTaylor, line 52: Do you have to add all this stuff to do with WellAttribute, Uuid and Warren because this happened automatically when `create!` was used, but doesn't with `import` ?

app/models/labware.rb
KatyTaylor: Ah I see, it uses the new `direct_submissions` method on `receptacle` 👍

app/models/receptacle.rb
```diff
@@ -95,9 +95,14 @@ def any_barcode_matching?(other_barcode)
...
   has_many :orders, through: :submitted_assets
   has_many :ordered_studies, through: :orders, source: :study
+  has_many :direct_submissions, ->() { distinct }, through: :orders, source: :submission
 
   has_many :tags, through: :aliquots
 
+  # Historically this used to be the only way to jump up to the submission without
+  # going via stock-wells. However, newer aliquots are associated with the request.
+  # This newer approach better handles cross-submission pooling, such as that used
+  # in the GBS pipelines.
```
KatyTaylor, line 105: Did you want to reference the `direct_submissions` association in this comment, as the implementation of the 'newer approach' that you mention?

lib/bm_plate_creation.rb
```diff
@@ -0,0 +1,11 @@
...
+
+TIMES = 5
+
+purpose = PlatePurpose.first
+
+Rails.logger.level = :warn
+
+Benchmark.bmbm do |x|
+  x.report('construct!') { TIMES.times { purpose.create! } }
+end
```
KatyTaylor, line 11: Is this for testing performance? Can you add a little comment at the top explaining it? Thanks

lib/record_loader/primer_panel_loader.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# This file was automatically generated via `rails g record_loader`
+
+# RecordLoader handles automatic population and updating of database records
+# across different environments
+# @see https://rubydoc.info/github/sanger/record_loader/
+module RecordLoader
+  # Creates the specified plate types if they are not present
```
KatyTaylor, line 9: Copy paste error with comment above?
Actually looks like all the existing equivalent files reference 'plate types' in this comment (barcode_printer_type_loader, request_type_loader etc.).

spec/data/record_loader/primer_panels/two_entry_example.yml
```diff
@@ -0,0 +1,21 @@
+---
```
KatyTaylor, line 1: Did you want these example files hanging around?

lib/warren/queue_broadcast_message.rb
```diff
@@ -1,15 +1,28 @@
...
+      @class_name = record.class.name
+      @id = record.id
+    else
+      @class_name = class_name
+      @id = id
+    end
+  end
+
+  def queue(warren)
+    after_commit { warren << self }
```
KatyTaylor, line 18: I can see this is used in plate.rb, but not quite getting how it works and why it was not needed before? Thanks

app/models/asset.rb
JamesGlover: So in this case, this was actually the first definition of the scope, that then gets re-defined in Receptacle (with exactly the same content).

I decided to remove it from Asset, as its an abstract class, inherited by both Labware and Receptacle, and one that's only really still sticking around because we haven't had time to remove it. In this particular situation, for_summary:

- Is only used for receptacles
- Will only work for receptacles (as labware don't have map)

So it didn't really make sense to maintain the version on asset.

app/models/concerns/programs_validator.rb
JamesGlover: It represents a PCR program, which is a property of a primer panel. Essentially each primer panel has slightly different settings that it should be run under, which includes a name, and a duration. This gets displayed to the user when they are being asked to set-up the PCR reaction.

I've noticed that some of the more recent primer panels don't specify program names. That smells a bit wrong to me, as the PCR machine *will* have a program. I suspect it's partly because the program selected is already covered by the SOP, but might be other issues at play. (I know not all processes use the same two-stage PCR that the original did)

app/models/labware.rb
JamesGlover: I'll need to revisit the working here then, although I'm not quite sure how to make it clear, as its all quite complicated.
So, in this case in-progress submissions is intending to mean, those you're in the act of processing currently, instead of those that are associated with requests out of the plate. In this particular situation the information gets pulled from the submission associated with transfer request.
In cross submission pools the transfer_requests don't link submission, as they are associated with multiple submissions. At this point we added the direct link from aliquot to submission. Data wasn't back-filled due to the sheer size and complexity of the task, although I am increasingly regretting that decision.

app/models/labware.rb
JamesGlover: I'll wrap this into my doc changes to avoid triggering the CI too much.

app/models/plate.rb
JamesGlover: Yep, import doesn't trigger rails callbacks. Which is probably a good thing, as otherwise it would still end up generating 96 insert statements. But it does mean this gets a bit more complicated.

app/models/receptacle.rb
JamesGlover: Actually no, direct submissions is different, but I should probably make it clearer what the newer approach is. I'll also look at adding a story to backfill aliquot request data, as then we can simplify a lot of this.

lib/bm_plate_creation.rb
JamesGlover: No problem.

lib/record_loader/primer_panel_loader.rb
JamesGlover: Oooh... that's probably actually a record loader templating bug. I don't think I copy pasted anything here.

lib/warren/queue_broadcast_message.rb
JamesGlover: Previously the after save callback on well would use the after_commit on well to queue the well object for broadcast. However that depends on us instantiating and persisting the entire wells for the duration of the transaction, when all we need is the class name and the id. This lets us create a mini message consisting of just the class name and id, and adding that to the transaction instead.

I've actually popped a story in the backlog to look at doing this everywhere, as it should let us substantially reduce the memory requirements of very large transactions. (Aliquots in particular can get VERY expensive)

In this particular case the performance saving wasn't quite as great as I hoped when I started (I must have messed up on my 'quick and dirty version' as it looked like things got 10 times faster again. But I couldn't reproduce that even when I stopped creating wells completely) but left it in due to the memory saving.

spec/data/record_loader/primer_panels/two_entry_example.yml
JamesGlover: Yep, used by the tests. I'll rename the file though (future versions of record loaded already name them after the class) as its clearly confusing. (Andrew got similarly confused)

app/models/labware.rb
JamesGlover: Hmm, we have in_progress_submissions below, which is the same as this but with 'distinct' (probably what confused you below. I'll flag this one as deprecated as its really confusing, and tidy up the comments elsewhere

app/models/concerns/programs_validator.rb
KatyTaylor: OK thanks, yes I vaguely remember coming across these early in the Heron Limber pipeline development. I think I remember a conversation with R&D where they didn't care about the 'duration' stuff etc in this case, but can't remember why. 
Where is the interface where it gets displayed to the user? Is it just one of the plate screens in Limber?

JamesGlover: Yep, some of the plate presenters show the information, and we had an interstitial page on plate creation at one point, but I think that might have disappeared over the course of various pipeline changes.

app/models/labware.rb
KatyTaylor: OK thanks for explaining! 

So this returns submissions that are linked to any transfer requests into the labware's receptacles?
Still not sure why that means they are 'in progress' / 'you are in the act of processing currently' - wouldn't it pull submissions that were all done and finished with if you called it on an old labware?

Wonder if this area is worth a Confluence page (maybe there's a relevant one already?) because it's quite complicated, as you say! Happy to help with that, if I can.

KatyTaylor (approved):

app/models/labware.rb
JamesGlover: More inclined to suggest an md file in the sequencescape repo itself, can get yard to link the various sections of it then as relevant, and slightly less likely to get horrendously out of date.

You're right that the association will also return submissions that were done and dusted. The name originally came out of a requirement in limber to distinguish between work that was in progress, compared to work that was yet to start. Then name then cascaded across into sequencescape, where out of context it doesn't make as much sense.

That said, its probably worth auditing what's in place, and whether any of it can be removed. A lot of the logic these were originally added to support has been moved off into Limber, so it might be possible to strip some of it out.

PR #2965, opened 2020-11-26
Issue2962/gpl 755 beckman destination plate barcode is foreign

### LINKED ISSUE #2962: GPL-755	Beckman destination plate barcode is foreign
**User story**
As a dev team we need to make Beckman destination plate barcode native.

Steve I

**Acceptance criteria**
Place holder to capture requirement so is not lost.

- [ ] ...

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
Raised in rambling discussion on 25th Nov

JamesGlover:
Closes #2962 

* Add barcode format
* Adds barcode format to foreign barcodes list to allow auto-detection

### DIFF
--- a/app/models/barcode.rb
+++ b/app/models/barcode.rb
@@ -35,14 +35,15 @@ class Barcode < ApplicationRecord
     cgap_plate: 14,
     cgap_rack: 15,
     glasgow: 16,
-    cambridge_a_z: 17
+    cambridge_a_z: 17,
+    high_throughput_cherrypick: 18
   }
 
   # Barcode formats which may be submitted via sample manifests
   FOREIGN_BARCODE_FORMATS = %i[cgap fluidx_barcode fluidigm
                                uk_biocentre_v1 uk_biocentre_v2 uk_biocentre_unid
                                alderly_park_v1 alderly_park_v2 uk_biocentre_v3 cgap_plate cgap_rack
-                               glasgow cambridge_a_z].freeze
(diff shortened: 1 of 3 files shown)


app/models/barcode.rb
```diff
@@ -35,14 +35,15 @@ class Barcode < ApplicationRecord
...
+    cambridge_a_z: 17,
+    high_throughput_cherrypick: 18
   }
 
   # Barcode formats which may be submitted via sample manifests
   FOREIGN_BARCODE_FORMATS = %i[cgap fluidx_barcode fluidigm
                                uk_biocentre_v1 uk_biocentre_v2 uk_biocentre_unid
                                alderly_park_v1 alderly_park_v2 uk_biocentre_v3 cgap_plate cgap_rack
-                               glasgow cambridge_a_z].freeze
+                               glasgow cambridge_a_z high_throughput_cherrypick].freeze
```
stevieing, line 46: wonder if it might be better to call this heron_tailed as that is what HT stands for?

stevieing (changes requested):
All good. Just think it might be better to name it heron tailed as that is the only thing this prefix will be used for.

app/models/barcode.rb
JamesGlover: Ahh yes... that makes more sense.

stevieing (approved):

PR #2970, opened 2020-12-04
Issue2967/gpl 758 location of the heron negative controls is always between columns 9 and 16 of our 384 plates

### LINKED ISSUE #2967: GPL-758 location of the Heron negative controls is always between columns 9 and 16 of our 384 plates
**Description**
As R&D (Naomi & Louise A)  we noticed that the location of the Heron negative controls is always between columns 9 and 16 of our 384 plates, we would also like to see them in  1-8 or 17-14
 

Naomi P
Louise A

Other people that may have specific knowledge about this work or have a stake in how it is implemented. e.g. John Smith is an expert on x

Louise A wrote (06 November 2020 11:03)

Naomi and I noticed that the location of the Heron negative controls is always between columns 9 and 16 of our 384 plates.
 
We know that originally some wells were avoided on purpose in order to not end up where the QC standard would be. However as far as I understand the process, it shouldn’t prevent the controls to be randomly picked towards the right or left third of a plate (it that correct?). Do you know why the negative controls are never picked into col 1-8 or 17-14?
 
Beside the new tail process doesn’t require any cDNA quant QC so the constrains on avoiding standard well locations could be lifted to allow complete randomness when choosing controls location. Would it be possible put that in place? Is that a big job?
 
The positive controls seem to be  little bit more random but always gathered together on a 394 plate. Again they could be picked completely randomly like the negative controls now that there isn’t any cDNA QC.

JamesGlover:
Closes #2967

Probably easiest to look at this one on a per commit basis, as I've tried to keep refactors and behaviour changes distinct.

* Adjust layout of negative controls
* Refactor for clarity


app/models/cherrypick_task.rb
```diff
@@ -11,236 +11,41 @@
...
+      # position for each plate in that batch. If we have multiple controls on a
+      # plate, we try to space them evenly. This establishes the relative
+      # positions for each control. We then take the modulo against the number
+      # of available wells to set the final position on the plate itself
+      position_index = batch_id + # We advance the control position as the batch_id increases
+                       num_plate - # Then once per plate in the batch
+                       (control_well_index * control_separation) # Finally we distribute the controls evenly
+      # Finally we layout the controls on the actual available positions on our plate
+      wells_to_leave_free + (position_index % total_available_positions)
     end
```
emrojo, line 48: Current method spreads the controls better across the plate, however it has the drawback that the results have less variability, for example:

With this version:
```
irb(main):023:0> CherrypickTask.new.control_positions(0,0,96,3, wells_to_leave_free: 0)
=> [0, 64, 32]
irb(main):024:0> CherrypickTask.new.control_positions(96,0,96,3, wells_to_leave_free: 0)
=> [0, 64, 32]
irb(main):026:0> CherrypickTask.new.control_positions(192,0,96,3, wells_to_leave_free: 0)
=> [0, 64, 32]
```

If using same conditions, all controls will be placed in exactly the same position after every 96 batches.  The "cycle" now is in the relation between batch_id and num_of_wells.
In previous version clash would only happened after a full cycle of all controls across the plate has happened.
For example, with a 96 well plate that we want to put 3 controls but leaving the first 3 columns empty:
-  If the plate needs to have 3 columns empty, that leaves: 96 wells - 24 (3 columns) = 72.
- Allocating 3 controls means 72, 71 and 70 free places when putting them in position; so the clash happens every ```72*71*70``` = 357840 batches

With previous version:
```
irb(main):014:0> CherrypickTask.new.control_positions(0,0,96,3)
=> [24, 25, 26]
irb(main):039:0> CherrypickTask.new.control_positions(357840,0,96,3)
=> [24, 25, 26]
irb(main):040:0> CherrypickTask.new.control_positions(715680,0,96,3)
=> [24, 25, 26]
```

But all values in between are completely different. Even with that, clashes could still happen with some combinations of num_plate, but this is the best we could give for an even distribution of controls without using randomness. 
I like a lot your idea of splitting the plate in regions to place the controls with separation between each other and I was thinking, could this concept be merged with the distribution of batches? Could we split the the plate in 3 regions (if 3 controls) and use the modulus to place each individual control in their region but keeping the long distribution in place?

emrojo (changes requested):

app/models/cherrypick_task.rb
emrojo: That would mean, of course, that we will not be able to keep the value of 357840 when splitting in regions. But should be something around:
```
72 places / 3 controls = 24 locations
24*24*24=13824 batches
```

emrojo: Using your idea of placing controls in regions I think I found a way of generating the controls in different regions while keeping some variability in between the positions; the sequence starts with:
```
     1 [24, 48, 72]
     2 [25, 49, 73]
     3 [26, 50, 74]
     4 [27, 51, 75]
     5 [28, 52, 76]
...
```
and only repeats again after 13825:

```
...
 13824 [47, 70, 93]
 13825 [24, 48, 72]
 13826 [25, 49, 73]
 13827 [26, 50, 74]
 13828 [27, 51, 75]
...
```
I'll add this in another pull request as I cannot access your branch.

JamesGlover:
Close in favour of: https://github.com/sanger/sequencescape/pull/2975

PR #2972, opened 2020-12-04
Merge after 1st February 2021: Remove legacy sample reception

JamesGlover:
Deprecation notice will automatically hide the page on the 1st February,
this commit removes associated code.


lib/event_factory.rb
```diff
@@ -74,32 +74,6 @@ def self.project_refund_request(project, user, reference)
     # EventfulMailer.deliver_confirm_event(User.all_administrators_emails, event.eventful, event.message, event.content, "No Milestone")
   end
 
-  ###############################
```
JamesGlover, line 77: Found this when looking for registration associated stuff. Doesn't look like it was actually triggered there, and looks like whatever was calling it was removed a long time ago.

stevieing (approved):

PR #2973, opened 2020-12-04
Rt707158 missing poolings new relationships

JamesGlover:
Picked up some minor tweaks for RT#707158

- Poolings new wasn't creating asset-links, which made finding child tubes a bit tricky
- We only show direct parents and children on asset pages, this adds all ancestors and decendants
- Noticed a few performance issues which I decided to fix up.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -73,7 +73,9 @@ class Asset < ApplicationRecord
 
   scope :recent_first, -> { order(id: :desc) }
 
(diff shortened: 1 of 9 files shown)


emrojo (approved):
This applies to endpoint /poolings/new

emrojo:
In UAT waiting for validation from users (RT ticket updated)

KatyTaylor:
From Matt M through RT (number 708850):
"A few of us have had a play and it looks good to us. Let us know when you plan to go live with this."

Issue #2783, opened 2020-06-25
GPL-552 Cherrypicking - Automated layouts need to place controls in all destination plates [m]


andrewsparkes:
**User story**
GPL-552 As an SSR I would like cherrypicks with controls which are generated by using the automated layout functionality to add controls to all the destination plates. Furthermore for the same batch those control wells in the destination plates should be shifted on each subsequent plate so that plate swaps can be identified in the resulting data.

**Who are the primary contacts for this story**
James Mack
Deborah Plowman

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] large batches with multiple destination plates should be laid out with the controls on all plates and in different locations

**Dependencies**
None

**Additional context**
Current automated layout places the specified control plate wells only on one destination (as no way to specify multiple picks). Manual csv generated layouts are fine as far as we can tell. Automated route likely to be used for lighthouse large batches.
NB. They have also stated that they do not mind about the automated layout generation splitting wells from a single source into multiple destinations, as SM see this regularly anyway and can cope with it fine.

KatyTaylor:
A thought - maybe we should place the control on the first plate in a 'random' place, in case it's also useful to distinguish _between_ batches

emrojo:
![image](https://user-images.githubusercontent.com/5426355/86275180-ca4c3000-bbca-11ea-9161-5f7563d5a72b.png)

Image for controls positioned in next well in subsequent plates created for the same batch

emrojo:
Cherrypick from 6 plates x 96samples, partial plate with 2 wells at the beginning, template plate with 1 well at the end and applying controls:
Plates 1-3
![image](https://user-images.githubusercontent.com/5426355/86662796-f873aa80-bfe4-11ea-9b91-b1d8c1149e08.png)
Plates 4-6
![image](https://user-images.githubusercontent.com/5426355/86663044-3a9cec00-bfe5-11ea-831a-9c8e8f7d426a.png)

PR #2975, opened 2020-12-10
Issue2967/gpl 758 location of the heron negative controls is always between columns 9 and 16 of our 384 plates

emrojo:
Closes GPL758

*
*
* ...


app/models/cherrypick_task.rb
```diff
@@ -11,236 +11,46 @@
...
+
+    raise StandardError, 'More controls than free wells' if num_control_wells > total_available_positions
+
+    quotient = batch_id
+    size_region = (total_available_positions / num_control_wells)
+    regions = (wells_to_leave_free..(total_wells - 1)).each_slice(size_region).to_a
+
+    # Number of regions should equal number of controls, so sometimes last region is bigger than average
+    if regions.length > num_control_wells
+      last_region = regions.slice!(regions.length - 1, 1)
```
KatyTaylor, line 42:
```suggestion
      last_region = regions.pop
```

KatyTaylor: then following line could be the same, but take the `*` out

app/models/cherrypick_task.rb
```diff
@@ -11,236 +11,46 @@
...
+      regions[regions.length - 1].concat(*last_region)
+    end
+
+    position = 0
+    regions.each_with_index.map do |region, _num_region|
+      quotient, remain = quotient.divmod(region.length)
+      # Feeding the new position with the old position seems to add more
+      # variability between the controls positions
+      position = (remain + num_plate + position) % region.length
+      region[position]
```
KatyTaylor, line 52: Would be simpler as 
```
regions.each_with_index.map do |region, _num_region|
  remain = quotient % region.length 
  position = (remain + num_plate) % region.length
end
```

But you think changing the `quotient` each time, and adding the `position` adds something?

KatyTaylor: Actually is the above just the same as 
```
position = (quotient + num_plate) % region.length
```

app/models/cherrypick_task.rb
```diff
@@ -33,23 +33,23 @@ def control_positions(batch_id, num_plate, total_wells, num_control_wells, wells
 
     raise StandardError, 'More controls than free wells' if num_control_wells > total_available_positions
 
-    quotient = batch_id
-    size_region = (total_available_positions / num_control_wells)
-    regions = (wells_to_leave_free..(total_wells - 1)).each_slice(size_region).to_a
+    available_positions = (wells_to_leave_free..(total_wells - 1)).to_a
+    # If num plate is equal to the available positions, the cycle is going to be repeated.
```
JamesGlover, line 37: `(wells_to_leave_free..(total_wells - 1))` can just be `(wells_to_leave_free...total_wells)`

app/models/cherrypick_task.rb
```diff
@@ -33,23 +33,23 @@ def control_positions(batch_id, num_plate, total_wells, num_control_wells, wells
...
+    # To avoid it, every num_plate=available_positions we start a new cycle with a new seed.
+    seed = batch_id * ((num_plate / available_positions.length) + 1)
+    initial_positions = random_elements_from_list(available_positions, num_control_wells, seed)
```
JamesGlover, line 40: You could avoid the need to do this by moving `Random.new(seed)` out of `random_elements_from_list` and to the top of the `auto_add_control_plate` of `perform_pick`.
We then pass the PRG around in place of batch id, and don't even need to worry about passing in the plate number, the RNG will ensure we don't start cycling.
(Good call on the use Random.new and sample, was planning to do something similar when I glanced at your code review over the holiday)

app/models/cherrypick_task.rb
```diff
@@ -33,23 +33,23 @@ def control_positions(batch_id, num_plate, total_wells, num_control_wells, wells
...
 
-    position = 0
-    regions.each_with_index.map do |region, _num_region|
-      quotient, remain = quotient.divmod(region.length)
-      # Feeding the new position with the old position seems to add more
-      # variability between the controls positions
-      position = (remain + num_plate + position) % region.length
-      region[position]
+    initial_positions.map do |pos|
+      available_positions[(available_positions.index(pos) + num_plate) % available_positions.length]
```
JamesGlover, line 52: `available_positions[(available_positions.index(pos) + num_plate) % available_positions.length]`
Can now just become:
`available_positions[(available_positions.index(pos)]` as available positions is already randomized based on plate number. (This is still the case when you make the change I suggested above)

JamesGlover (changes requested):
Yep, use of Random.new(seed) is the right solution here, but spotted a few further simplifications that have dropped out of this. Plush a changes that avoids us needing to recalculate things on a per plate basis.

app/models/cherrypick_task.rb
JamesGlover: Hmm, thinking about it, this might even be a good chance to create a separate object to handle it... I'll mock one up

app/models/cherrypick_task.rb
JamesGlover: Actually, I'm not even sure you need that much, I think you can just use initial positions?

app/models/cherrypick_task.rb
JamesGlover:
```
class ControlLocator
  def initialize(seed, total_wells, number_controls, wells_to_leave_free: 0)
    @seed = Random.new(seed)
    @number_controls = number_controls
    @available_wells = (wells_to_leave_free...total_wells).to_a
  end

  # Returns the next set of control positions.
  # Subsequent calls will return a new set of positions 
  def control_positions!(number = number_controls)
    @available_wells.sample(number, random: @seed)
  end
end

# Usage
# def perform_pick(requests, robot, auto_add_control_plate, workflow_controller)
# ...
# control_assets = auto_add_control_plate&.wells&.joins(:samples) || 0
# control_locator = ControlLocator.new(batch.id, current_destination_plate.size, control_assets.count)
#
# Then each time you need them
# control_posns = control_locator.control_positions!

```

app/models/cherrypick_task.rb
emrojo: Ok, changing

app/models/cherrypick_task.rb
emrojo: Yes, that would make everything totally random for every plate, however that changes the actual behaviour so every plate in the same batch have a kind of predictable order in the controls positioning. The original story was:

https://github.com/sanger/sequencescape/issues/2783

There has been some changes in requirements but I will check with the SSRs what they think about this, so ask them between the 2 options:

- Option A: We generate random position for first plate, but next plates have a relation with this position by increasing 1, so they can use it to identify plate swaps, if that happens.
- Option B: We generate random positions for every batch and every single plate, independently of the position, so it has more variability

If they want to identify plate swaps, still in both options the controls have a predictable position, but in Option B could be easier to check.

app/models/cherrypick_task.rb
emrojo: Yes, that was to apply the logic of increasing with num plate, so depends on how each plate is generated, but if it is random, then it can be initial_positions (although then that variable name will need rewording)

KatyTaylor (approved):

app/models/cherrypick_task.rb
JamesGlover: Mentioned in the slack, but I misunderstood exactly what was happening here. Hadn't realised that we only generate a new seed every x plates.

JamesGlover (approved):

PR #2978, opened 2020-12-16
Add gh actions

JamesGlover:
Migrated existing CI to CH actions, and added yarn linting and tests.

Note: Behaviour isn't entirely identical, as arbitrary tags will no longer trigger release builds. This should be possible to fix up if necessary, but shouldn't be essential.


.circleci/config.yml
```diff
@@ -214,17 +214,6 @@ jobs:
...
-    <<: *defaults
-    steps:
-      - attach_workspace:
-          at: ~/repo
-      - run:
-          name: Upload coverage results to Code Climate
-          command: |
-            ls ./tmp
-            ./tmp/cc-test-reporter sum-coverage tmp/codeclimate.*.json -p 5 -o tmp/codeclimate.total.json
-            ./tmp/cc-test-reporter upload-coverage -i tmp/codeclimate.total.json
```
JamesGlover, line 217: Removed from circle-ci to avoid conflicts

.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
+name: Build Project
+on:
+  release:
+    types: published
+
```
JamesGlover, line 5: Triggers on release creation

app/javascript/plate-picks/components/PicksSectionBatchPick.vue
```diff
@@ -38,7 +38,10 @@
         }"
       >
         <i :class="`fas fa-${plate.scanned ? 'check' : ''}`" />
-        <i class="fas fa-vial" v-if="plate.control" />
+        <i
+          v-if="plate.control"
+          class="fas fa-vial"
+        />
         {{ plate.barcode }}
```
JamesGlover, line 45: This was yarn lint. Which apparently hadn't been run recently.

Gemfile
```diff
@@ -194,8 +201,7 @@ group :test, :cucumber do
   gem 'nokogiri', require: false
   gem 'shoulda-context', require: false
   gem 'shoulda-matchers', require: false
-  # Pin due to Code climate compatibility issues. Should be able to unpin soon.
-  gem 'simplecov', '~>0.17.0', require: false
+  gem 'simplecov', require: false
   gem 'timecop', require: false
```
JamesGlover, line 205: This was fixed just the other day!

.circleci/config.yml
```diff
@@ -57,21 +57,13 @@ jobs:
           name: reinstall mysql2 temporary
```
JamesGlover, line 57: Remove the coverage steps from Circle-CI to avoid confusing things when it receives multiple files per run. Can remove this completely once merged to master and we can disable circle CI completely.

.rubocop.yml
```diff
@@ -19,7 +19,7 @@ AllCops:
     - db/schema.rb
```
JamesGlover, line 19: Later versions of rubocop let you specify a cache folder. So we update rubocop as well to take advantage. It makes linting nice and fast if the rubocop yaml files haven't been touched.

.github/workflows/js_test.yml
```diff
@@ -0,0 +1,20 @@
...
+  eslint:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Install
+      uses: CultureHQ/actions-yarn@master
+      with:
+        args: install
+    - name: Lint
```
KatyTaylor, line 17: Not important, but think this should be called 'Test', not 'Lint' (linting is in another file, right)

JamesGlover: Whoops, you are correct.

.github/workflows/js_test.yml
```diff
@@ -0,0 +1,20 @@
...
+on:
+  - push
+  - pull_request
+
+jobs:
+  eslint:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
```
KatyTaylor, line 12: any reason for not using @v2 rather than @master ? (just reading the docs - https://github.com/actions/checkout - and seems like v2 is the one referenced)

.github/workflows/js_test.yml
```diff
@@ -0,0 +1,20 @@
+name: Javascript testing
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  eslint:
```
KatyTaylor, line 8: and eslint isn't relevant on the testing job is it? Just used for the linting?

.github/workflows/js_test.yml
JamesGlover: No... it looks like I used the configuration in https://github.com/marketplace/actions/yarn-cli but @v2 makes *much* more sense. I'll make the change.

.github/workflows/js_test.yml
JamesGlover: Ugh, teach me to copy paste templates.

.github/workflows/js_test.yml
JamesGlover: Oh... just noticed the deprecation warning on that

JamesGlover: Will switch to the suggested version.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,70 @@
...
+        # cache, then finally any cache for this repo.
+        # Github looks for the newest cache beginning with the first entry, before
+        # falling back the the second if none is present.
+        restore-keys: |
+          ${{ runner.os }}-rubocop-
+          ${{ runner.os }}-
+    # Install only the gems needed for linting, keep things nice and fast
+    # Keep an eye on https://github.com/rubygems/bundler-features/issues/59
+    # in case bundler add an only flag
+    # We also set the install path to vendor/bundle to assist with out caching
```
KatyTaylor, line 47:
```suggestion
    # We also set the install path to vendor/bundle to assist with our caching
```

?

.github/workflows/js_test.yml
```diff
@@ -0,0 +1,21 @@
...
+  - push
+  - pull_request
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    # These two commands essentially ensure yarn is setup and run yarn install and yarn lint respectively
```
KatyTaylor, line 13:
```suggestion
    # These two commands essentially ensure yarn is setup and run yarn install and yarn test respectively
```
:)

.github/workflows/ruby_test.yml
```diff
@@ -0,0 +1,218 @@
+name: Ruby Test
```
KatyTaylor, line 1: There's quite a lot of repetition in here isn't there? I wonder if we could refactor any of it out into a github actions repository or something

KatyTaylor (approved):

.github/workflows/ruby_test.yml
JamesGlover: We could possibly create our own actions to use. Haven't really looked into exactly what they look like, or indeed how-powerful they are.
Another possibility, which is a little simpler, would just be to use some of the inheritance mark-up built into yaml; although I did spot a suggestion online that the GH actions don't support some of the fancier yaml syntax.

KatyTaylor:
Looks like Ruby Test is completing in around 14 mins, so if the other jobs run in parallel I think that's quite a big improvement on Travis?

JamesGlover:
I'm not sure how many parallel jobs GH actions makes available, but in testing I didn't notice much queueing.

To be fair to travis I've optimised the install and setup process a little more than happens there,

.github/workflows/ruby_test.yml
JamesGlover: Yep, just quickly tested it, and yaml anchors aren't supported, so can't simplify that way.

PR #2982, opened 2020-12-18
Gem updates

JamesGlover:
We haven't really done any standard gem maintenance since August. Here I've run bundle update on everything, except Rubocop (I'm handling that in a separate branch). A few in code fix ups to sort out some test failures too.


app/models/product.rb
```diff
@@ -6,7 +6,7 @@ class Product < ApplicationRecord
   validates :name, uniqueness: { scope: :deprecated_at, case_sensitive: false }
   has_many :product_product_catalogues, dependent: :destroy
   has_many :product_catalogues, through: :product_product_catalogues
-  has_many :submission_templates, inverse_of: :product, through: :product_catalogues
+  has_many :submission_templates, inverse_of: :products, through: :product_catalogues
```
KatyTaylor, line 9: What happened here? Would it have been working before?

KatyTaylor (approved):

app/models/product.rb
JamesGlover: It was an update of one of the gems (shoulda-matchers I think) that changes the way 'should have_many' validation worked to actually check that the configuration of the association was valid (not sure if by design or accident). I'm assuming we didn't notice up until now as we aren't ever actually loading `product#submission_templates` and instead tend to go in the opposite direction.

If I try it in dev:
```
irb(main):023:0> Product.last.submission_templates
  Product Load (7.9ms)  SELECT  `products`.* FROM `products` ORDER BY `products`.`id` DESC LIMIT 1
Traceback (most recent call last):
        1: from (irb):23
ActiveRecord::InverseOfAssociationNotFoundError (Could not find the inverse association for submission_templates (:product in SubmissionTemplate))
```

PR #2985, opened 2020-12-22
Bug destination plate event no user

KatyTaylor:
- Allow user_identifier to be set by a value in the properties hash, rather than using a user id
- So that calls from lighthouse from Beckman destination plate creation endpoint work
- Delete a duplicated method

### DIFF
--- a/app/models/broadcast_event/plate_cherrypicked.rb
+++ b/app/models/broadcast_event/plate_cherrypicked.rb
@@ -34,9 +34,7 @@ class BroadcastEvent::PlateCherrypicked < BroadcastEvent
 
   def user_identifier
     # allows for user identifiers that aren't in the SS db
-    return properties[:user_identifier] if properties[:user_identifier]
-
-    super
+    properties[:user_identifier].presence || super
   end
 
   def robot_present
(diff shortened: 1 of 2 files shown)


JamesGlover (approved):

app/models/broadcast_event/plate_cherrypicked.rb
```diff
@@ -32,6 +32,13 @@ def initialize(args)
     _add_destination_plate
   end
 
+  def user_identifier
+    # allows for user identifiers that aren't in the SS db
+    return properties[:user_identifier] if properties[:user_identifier]
+
+    super
+  end
```
JamesGlover, line 40: Actually, thinking about it, this may be better as:

properties[:user_identifier].presence || super

As that would fallback to UNKNOWN if user_identifier was ''

JamesGlover: https://guides.rubyonrails.org/active_support_core_extensions.html#presence

KatyTaylor: Good spot, have changed as per your suggestion.

PR #2994, opened 2021-01-05
Issue2942/gpl 693 bug mlwh study user table not updating

### LINKED ISSUE #2942: GPL-693 BUG - MLWH study_user table not updating
When a contact/collaborator is removed from a study in Sequencescape the entry in the mlwh persists

**RT Ticket Number**
702503

**To Reproduce**
http://sequencescape.psd.sanger.ac.uk/studies/8/collaborators
add yourself as follower
go to mlwh.study_user 
find id_study_temp = 6
note users
in Sequencescape remove yourself
in mlwh.study_user user persists

**Expected behaviour**
in mlwh.study_user user is removed

**Additional context**
James G added
So based on past fixes:
1) Removal of the role_user entry wasn’t updating the study message in the WH. I had to uses **touch** to ensure that the timestamp on the study was updated, as otherwise the message wouldn’t have a more recent timestamp than the existing one. ( I think this took a couple of passes, as I don’t think I updated the timestamp first time round)

2)I think I might have needed to update the WH itself at some point to ensure existing relationships were purged. Its possible that this was just done on warehouse_two, and the unified WH may still be using the older code. Not sure though, it was a few years ago and is only a vague memory.

JamesGlover:
Closes #2942

* Fixes updating study timestamps on user removal
* Updates role management page to make selecting users easier
* Fixes blue button colour overriding custom colours
* Improves look of role management page.

Before
Screenshot 2021-01-05 at 16 31 24

After
Screenshot 2021-01-05 at 16 51 44


lib/rails-authorization-plugin/lib/publishare/object_roles_table.rb
```diff
@@ -94,30 +94,15 @@ def authorizables_for(authorizable_class)
...
-              authorizable_id: authorizable_obj.id
-            ).first
-          else
-            Role.where(
-              name: role_name,
-              authorizable_type: nil,
-              authorizable_id: nil
-            ).first
-          end
+          Role.find_by(
```
JamesGlover, line 97: Rails handles this logic itself now

pjvv (approved):
It's been a while since I've visited Rails-land but it looks straight forward and OK.

PR #2998, opened 2021-01-06
Increase offset between plates

JamesGlover:
Implements Naomi's requested increase in offset.

In order to avoid cycling round before all available well positions have been used, an offset (which isn't 1) can't share a common factor with the number of available wells (other than 1). In order to ensure this we:

1) Provide a list of primes
2) Detect the first one that isn't a factor of the available well length
3) Fall back to 1 in the extremely unlikely case we end up dealing with a plate of 3127 wells (or multiples thereof)

In practice the offset will be 53.


app/models/cherrypick_task.rb
```diff
@@ -33,7 +45,7 @@ def control_positions(batch_id, num_plate, total_wells, num_control_wells, wells
 
     raise StandardError, 'More controls than free wells' if num_control_wells > total_available_positions
 
-    available_positions = (wells_to_leave_free..(total_wells - 1)).to_a
+    available_positions = (wells_to_leave_free...total_wells).to_a
```
KatyTaylor, line 48: 👍

KatyTaylor:
Seems good to me. I'm wondering whether we should extract all the control-positioning logic into a specific helper file, to avoid cluttering up the main file with algorithms that should hopefully now stay untouched forever more?

KatyTaylor (approved):

JamesGlover:
Yeah, was considering the same. Will do.

KatyTaylor (approved):

PR #2999, opened 2021-01-06
Improve setup instructions

JamesGlover:
- We have an automatic set-up script, lets direct people to it
- Rbenv install can pull the version straight from ruby-version if its
  formatted correctly

Closes #

*
*
* ...


.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.7.2
+2.7.2
```
JamesGlover, line 1: While rbenv handles this fine when setting the environment, it can't use it to trigger install with `rbenv install`. This change simplifies the setup process. This lets us tell the user to use `rbenv install` rathern than needing to get them to `cat .ruby-version` or to keep a version number up to date in the documentation.

README.md
```diff
@@ -228,7 +244,7 @@ node module. To install it, make sure you have install the dev dependencies from
 the table of contents, run:
 
 ```shell
-./node_modules/.bin/markdown-toc -i README.md --bullets "*"
+yarn markdown-toc -i README.md --bullets "*"
```
pjvv, line 247: I found a cool way to run this without it needing to be installed: `npx markdown-toc -i README.md`

pjvv: We also have to decide how we want bullets to prevents back and forth between `-`s and `*`s 😆

pjvv (approved):

README.md
JamesGlover: That's cleaner, means we could remove it from the dependency list.

Been thinking about adding in some yaml linters for a while to keep things consistent. I've got one running in VS Code, but obviously thats only marginally useful if we all end up running with different parameters.

I do wish git-hooks had a good way of modifying files in the staging area. The only solutions I've seen tend to be a bit hacky, and don't play well with the use of `git add -p`.

PR #3001, opened 2021-01-08
Drastically prune back library

JamesGlover:
We barely use any of the functionality out the legacy
plugin, so we strip it back to the bare essentials here.

In practice it just comes down to a handful of is_<role>[_of]
method_missing handlers and the has_role method. (And the recently
patched has_no_role) As well as some of the rails associations.

This is all trivial stuff to re-implement in basic rails.


JamesGlover:
No real urgency to pull this in at this stage. Just want to keep code review in small chunks.
This strips out unused methods and code paths.

lib/rails-authorization-plugin/lib/publishare/identity.rb
```diff
@@ -15,88 +13,31 @@
...
-          if method_name =~ Regexp.new(is_either_regex + '_what$')
-            role_name = $3 || $6
-            has_role_for_objects(role_name, authorizable_object)
-          elsif method_name =~ Regexp.new(is_not_either_regex + '\?$')
-            role_name = $3 || $6
-            not is_role?(role_name, authorizable_object)
-          elsif method_name =~ Regexp.new(is_either_regex + '\?$')
+          # matches is_role? and is_role_of?
+          if method_name =~ Regexp.new(is_either_regex + '\?$')
             role_name = $3 || $6
```
emrojo, line 31: This is from previous code, but maybe can help in other changes to replace $3 and $6 with a variable name with
the description of the content, or maybe a coment explaining what it is referring in the regexp.

emrojo (approved):

lib/rails-authorization-plugin/lib/publishare/identity.rb
JamesGlover: This code is almost ready for complete removal, which is why I didn't touch it more (I was very tempted). I've already got the replacement methods working. But yes, agree with the suggestion in principal.

PR #3006, opened 2021-01-12
Issue3005/bug lh beckman cp destination created event use sanger barcode as friendly name

### LINKED ISSUE #3005: BUG - lh_beckman_cp_destination_created event use sanger barcode as friendly name
**Describe the bug**
lh_beckman_cp_destination_created list the destination plate with the sanger barcode, not the primary barcoe.

**RT Ticket Number**
If applicable

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**

JamesGlover:
Closes #3005

* Use the standard subject renderer to ensure the correct friendly name is used


JamesGlover:
Changes I chose not to make:
1) We're currently persisting the properties as a serialised instance of ActionController::Parameters rather than a plain hash. This could cause issues with potential future Rails updates. Definitely need to fix this, but it probably should be a different story.
2) BroadcastEvent::Helpers::ExternalSubjects should probably merge its subjects into the existing subjects, rather than completely over-writing them. This would let use just do ` seed_subject :cherrypicking_destination_labware` to handle the destination plate. However it would remove the ability to override the  cherrypicking_destination_labware via the properties. Probably worth holding off on this one until we have more   events with ExternalSubjects as at the moment its just my gut as to which course of action is best.

emrojo (approved):

PR #3007, opened 2021-01-13
Adjust Code Coverage

JamesGlover:
Code coverage upload was failing with an exit code of 0, so the actual step was showing as successful.
Series of commits to try and diagnose and fix the problem, now working.

Squash merge this, as the intermediate commits aren't useful.

### DIFF
--- a/.github/workflows/ruby_test.yml
+++ b/.github/workflows/ruby_test.yml
@@ -208,5 +208,7 @@ jobs:
     - name: Publish code coverage
       uses: paambaati/codeclimate-action@v2.7.5
       with:
+        workingDirectory: ./sequencescape
+        prefix: ${{github.workspace}}
         coverageLocations: |
           ${{github.workspace}}/tmp/codeclimate-*:simplecov
--- a/.github/workflows/ruby_test.yml
+++ b/.github/workflows/ruby_test.yml
@@ -201,6 +201,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [rake_tests, rspec_tests, cucumber_tests]
     steps:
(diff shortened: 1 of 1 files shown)


JamesGlover:
Code coverage was broken. This should be a squash merge, as there were a few issues that needed to be fixed.

.github/workflows/ruby_test.yml
```diff
@@ -209,9 +210,4 @@ jobs:
       uses: paambaati/codeclimate-action@v2.7.5
       with:
         coverageLocations: |
-          tmp/codeclimate-rake_tests-
-          tmp/codeclimate-rspec_tests-0
-          tmp/codeclimate-rspec_tests-1
-          tmp/codeclimate-rspec_tests-2
-          tmp/codeclimate-cucumber_tests-0
-          tmp/codeclimate-cucumber_tests-1
+          ${{github.workspace}}/tmp/codeclimate-*/coverage.json:simplecov
```
JamesGlover, line 213: Was looking in the wrong location. Plus, now with globbing, so we're not coupled to the steps above.

.github/workflows/ruby_test.yml
```diff
@@ -201,6 +201,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [rake_tests, rspec_tests, cucumber_tests]
     steps:
+    - uses: actions/checkout@v2
     - name: Fetch coverage results
```
JamesGlover, line 205: Need the code checked out as the report uploaded needs access to the git repo for commits/branches etc.

pjvv (approved):

PR #3010, opened 2021-01-13
GPL-278 Get rid of 'new project registered' emails

### LINKED ISSUE #2514: GPL-278	As a dev team we would like to investigate why/how we receive emails for each new project registered. This will become unsustainable if DToL creates +60K projects + project approved.

KatyTaylor:
Get rid of new project registered email
- History: Projects had to be approved post creation. Now approval is done pre generation.
- Therefore function not required and discussed this with SSRs and they would SOOO happy if it went away
- (from Neil on https://github.com/sanger/sequencescape/issues/2514)

Closes #2514

### DIFF
--- a/lib/event_factory.rb
+++ b/lib/event_factory.rb
@@ -3,7 +3,9 @@ class EventFactory # rubocop:todo Style/Documentation
   #################################
   # project related notifications #
   #################################
-  # Creates an event and sends an email when a new project is created
+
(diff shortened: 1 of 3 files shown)


JamesGlover:
Bah, coverage failures. Okay, suspected that might happen form PRs from different repos. I'll fix that one.

JamesGlover (approved):

PR #3015, opened 2021-01-14
Issue2984/gpl 793 address significant risk and debt in ancient authorization systems

JamesGlover:
This strips out and replaces the plugin. Better permissions management to follow.


JamesGlover:
This strips out the old library, but doesn't yet have the migration to the new permissions system. Mainly as that is *HUGE*

JamesGlover:
Test failures are due to a problem with the way code coverage works with pull requests from external repositories. I thought I had fixed it, but apparently not. Trying fixes for that in another branch.

app/controllers/admin/projects_controller.rb
```diff
@@ -99,7 +99,7 @@ def sort
   private
 
   def redirect_if_not_owner_or_admin(project)
-    unless current_user.owner?(project) || current_user.is_administrator?
+    unless current_user.owner?(project) || current_user.administrator?
```
stevieing, line 102: I know you only use this twice but maybe the logic could be moved into the current_user model?

app/controllers/admin/users_controller.rb
```diff
@@ -10,7 +10,7 @@ def index
   end
 
   def edit
-    @user_roles = @user.roles.where(name: %w[administrator manager internal])
+    @user_roles = @user.roles.where(name: %w[administrator manager])
```
stevieing, line 13: Again. Maybe a scope?

app/controllers/admin/users_controller.rb
```diff
@@ -34,9 +34,9 @@ def update
     @user = User.find(params[:id])
```
stevieing, line 34: And again. Although the logic is sound maybe it is better somewhere else rather than the controller?

app/controllers/projects_controller.rb
```diff
@@ -110,11 +110,11 @@ def collaborators
 
   def follow
```
stevieing, line 111: And again. It maybe feels like creating a follower sits best in your authentication / authorisation logic.

app/models/role/user_role_helper.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
```
stevieing, line 1: Nice. Well documented and using separation of concerns.

stevieing (approved):
All looks reasonable. Added a few comments about logic in controllers. Also don't know whether there is some central documentation about the different roles and user types to explain it all. You have probably though about that already but no harm in mentioning it.

app/controllers/admin/projects_controller.rb
JamesGlover: This has already been removed in the next pass.

app/controllers/admin/users_controller.rb
JamesGlover: Good catch.

app/controllers/admin/users_controller.rb
JamesGlover: Make sense

app/controllers/projects_controller.rb
JamesGlover: What did you have in mind? I'm struggling to see how I could extract much more from this, as the flash stuff belongs in the controller. The `current_user.follower_of?(@project)` bit is not governing permissions, but just setting up the endpoint as a toggle. (Which admittedly I hate, but I'm not changing existing behaviour for this pass)

PR #3027, opened 2021-01-27
Add knapsack pro

JamesGlover:
Adds knapsack pro


emrojo (approved):
I don't know too much about what is needed for configuring knapsack Pro but I haven't found anything obviously wrong :)

PR #3028, opened 2021-01-27
Issue2984/gpl 793 address significant risk and debt in ancient authorization systems

### LINKED ISSUE #2984: GPL-793 Address significant risk and debt in ancient authorization systems
**User story**
Sequencescape currently uses an ancient rails-plugin (ie. it predates Gems) to manage authorization of users, and a clunky system of groups. There are two major problems with this:

- rails-authorization-plugin is ancient, and its coding style is quite out of step with modern ruby. We have made a few minor tweaks to it as time has gone on, but its a small wonder it is still working.
- Privilege checking is implemented opaquely. Currently:
   - Users will have many groups
   - Each group may be associated with an authorizable, usually a study or a project
   - Permission checks are all over the codebase, and generally work by checking which groups a user is members of.
   - Some of this logic is a bit weird, such as granting permissions to administrators, unless they are also lab managers.
- This makes it very difficult to quickly tell what permissions a user has, and what groups then need to be added to to be awarded them. This means that over time, users end up getting added to more and more groups, and rarely know exactly what permissions or roles they need.
- Instead we should have a system of permissions, and then perform checks along the lines of `@user.permission_to?(:create_studies)`, or `@user.permission_to?(:edit_study, @study)` 
- Permissions can then be granted to groups or even individual users.

We should consider migrating to a more modern authorization (and possible authentication) framework. We'll probably also need to refresh how our current permissions operate.

Things to look at (that I'm aware of off the top of my head):

CanCanCan: Authorization library. We're already using its precursor (can-can) in a couple of places.
Devise: Mostly authentication, rather than authorization, but given its popularity it would be good to get an understanding of what it provides. 

e.g. John S (don't include surnames in public repos)

**Acceptance criteria**
- [ ] ...

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

JamesGlover:
Closes #2984

A few failing tests, which look to be permissions issues, so I'll need to tweak those, but the key changes are here.


JamesGlover:
Replaced by: https://github.com/sanger/sequencescape/pull/3031

PR #3049, opened 2021-02-05
Add lefthook

JamesGlover:
Closes #

*
*
* ...


pjvv (approved):
Looks cool and something might be worth adding to https://github.com/sanger/.github/blob/master/standards.md as we collate our favourite tools etc.

PR #3061, opened 2021-02-10
Fix knapsack coverage in queue mode

JamesGlover:
So the previous attempt at solving this wasn't quite working, and seems to be due to forked processes exiting and clobbering the coverage.

This change ensures we get a new coverage file per process, and also ensures that the sever processes spun up by capybara et al, get their coverage recorded.


JamesGlover:
Uhh, well that's a definite fail. Coverage through the floor.

PR #3065, opened 2021-02-12
Gpl 860 glasgow eagle foreign barcodes

### LINKED ISSUE #3064: GPL-860 Allow Glasgow Eagle foreign barcodes
**User story**
As an SSR I want to be able to upload manifests including plate barcodes of the format 'EGGnnnnnn'.
We should allow any barcodes of the format 'EGxnnnnnn' where x is any letter, so we don't have to keep adding new ones.

Rich C

KatyTaylor:
Closes #3064

### DIFF
--- a/app/models/barcode.rb
+++ b/app/models/barcode.rb
@@ -40,14 +40,15 @@ class Barcode < ApplicationRecord
     randox: 19,
     uk_biocentre_v4: 20,
     cambridge_a_z_v2: 21,
-    glasgow_v2: 22
+    glasgow_v2: 22,
+    eagle: 23
   }
 
   # Barcode formats which may be submitted via sample manifests
   FOREIGN_BARCODE_FORMATS = %i[cgap fluidx_barcode fluidigm
                                uk_biocentre_v1 uk_biocentre_v2 uk_biocentre_unid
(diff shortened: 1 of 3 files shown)


app/models/barcode.rb
```diff
@@ -39,14 +39,16 @@ class Barcode < ApplicationRecord
     heron_tailed: 18,
     randox: 19,
     uk_biocentre_v4: 20,
-    cambridge_a_z_v2: 21
+    cambridge_a_z_v2: 21,
```
andrewsparkes, line 42: All looks fine.
My question: 
why separate out the specific EGG and EGC from the general Eagle one?
e.g. Milton Keynes and the others are not done, so if they start appearing you'd need another change.

andrewsparkes (approved):

app/models/barcode.rb
KatyTaylor: As discussed - I'll put it here for record - it seemed easier to have a barcode just fall into one category. There is code that says "Here's a barcode, what category am I in?" and I think it will be easier for future dev to have just one answer to that and not have to guess which one it pulled back. Think the categories are used to enforce uniqueness within the categories.

PR #3066, opened 2021-02-12
Release GPL-860 Eagle Glasgow foreign barcodes

KatyTaylor:


app/models/barcode.rb
```diff
@@ -39,14 +39,16 @@ class Barcode < ApplicationRecord
...
+    glasgow_v2: 22,
+    eagle: 23
   }
 
   # Barcode formats which may be submitted via sample manifests
   FOREIGN_BARCODE_FORMATS = %i[cgap fluidx_barcode fluidigm
                                uk_biocentre_v1 uk_biocentre_v2 uk_biocentre_unid
                                alderly_park_v1 alderly_park_v2 uk_biocentre_v3 cgap_plate cgap_rack
-                               glasgow cambridge_a_z heron_tailed randox uk_biocentre_v4 cambridge_a_z_v2].freeze
+                               glasgow cambridge_a_z heron_tailed randox uk_biocentre_v4 cambridge_a_z_v2 glasgow_v2 eagle].freeze
```
JamesGlover, line 52: Should 'eagle' also be added here?

JamesGlover (changes requested):

app/models/barcode.rb
JamesGlover: My mistake (Or weird GH bug)

JamesGlover (approved):

PR #3094, opened 2021-03-04
Issue3091/gpl 900 cancel a single plate {heron}

### LINKED ISSUE #3016: GPL-885 BUG - GnT Picoplex pipeline lost cross-pool functionality
**Describe the bug**
The GnT Picoplex pipeline used to allow submissions containing multiple library types, to pool two different libraries from the same sample together. This functionality was removed in https://github.com/sanger/sequencescape/commit/96d6322e3d0c5700fb4b84ab3e36446b5ae8f5c2#diff-565afe05f44c168c516703612ee97a04bb1bb874a9b42d23662d33db6d7288fd to address an issue with incorrect well failure.

We should:
- [x] Fix the detection of outer requests in PlatePurpose::fail_stock_well_requests to use the request ids on the aliquots.
- [ ] Remove the restrictions

Old plate may not have a request_id on aliquots. It is unlikely that we will see attempt to fail requests on these wells. However we should fail noisily in this scenario. It probably doesn't make much sense to maintain the old behaviour in this situation.

**RT Ticket Number**
https://rt.sanger.ac.uk/Ticket/Display.html?id=709590

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**

### LINKED ISSUE #3091: GPL-900 Cancel a single plate [Heron]
This represents the sequencescape work required for https://github.com/sanger/limber/issues/648

Original issue:
**User story**
As a lab user I would like to be able to cancel a single plate, without any further repercussions (no requests cancelled etc), to signal that it should not be processed further. It should behave consistently regardless of which plate it is. This will be helpful as a visual indicator that this plate should not be processed further, but allow me to go back to an earlier plate in the pipeline and carry on from there under the same submission.

James G, Katy

**Acceptance criteria**
When I cancel a plate in Limber:
- [ ] It cancels the transfer requests into that plate
- [ ] It does not cancel any Requests or Submissions
- [ ] It does not block processing on any plates other than the one I am cancelling
- [ ] There is some signposting in the app telling me what will happen when I cancel the plate

JamesGlover:
Closes #3091
Closed #3016

* Cancelling tubes and plates does not propagate to associated requests
* Extract state change behaviour from plate purpose
* Remove legacy plate purposes
* Simplify request lookup for plate state change to resolve #3016
* First pass simplification of tube state changes

I *think* we can essentially consolidate the tube state changes, as I'm not aware of a need to maintain the current scope of behaviour. However there are already enough changes going on here that I'm exercising caution.

Note: There *may* be further changes required here.


.rubocop_todo.yml
```diff
@@ -1,12 +1,12 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config --exclude-limit 10000`
-# on 2021-01-29 08:53:40 UTC using RuboCop version 1.8.1.
+# on 2021-03-03 09:15:46 UTC using RuboCop version 1.10.0.
 # The point is for the user to remove these configuration records
```
JamesGlover, line 4: Lots of files removed/significantly changed so regenerated to remove unneccessary ignores

app/models/illumina_b/mx_tube_purpose.rb
```diff
@@ -1,31 +1,11 @@
...
-  #
-  # @param tube [Tube] The tube for which to find the stock_plate
-  #
-  # @return [Plate, nil] The stock plate if found
-  #
-  def stock_plate(tube)
-    tube.requests_as_target.where_is_a(IlluminaB::Requests::StdLibraryRequest).first.asset.plate
-  end
-  deprecate stock_plate: 'Stock plate is nebulous and can easily lead to unexpected behaviour'
+#  - Update any purposes using this class by running `bundle exec rake remove:deprecated_purposes`
```
JamesGlover, line 4: Lots of these going on. We'll be able to lose the actual classes themselves once the listed rake task has been run.

app/models/illumina_htp/initial_stock_tube_purpose.rb
```diff
@@ -1,36 +1,12 @@
+# frozen_string_literal: true
+
 # Class initially used to represent the first tube in a pipeline following pooling
 # from a plate. However appears to be a little more widespread in Limber pipelines.
 # eg. LB Custom Pool Norm
+# In fact, in the limber pipeline it appears to be responsible for any intermediate
+# tubes.
 class IlluminaHtp::InitialStockTubePurpose < IlluminaHtp::StockTubePurpose
-  def valid_transition?(outer_request, target_state)
```
JamesGlover, line 5: I've extracted state change logic into the state changers. I should probably do the same for the state calculators.

app/models/illumina_htp/mx_tube_purpose.rb
```diff
@@ -4,38 +4,8 @@
 # pipeline. These tubes represent the cleaned-up normalized libraries at the end
 # of the process that can pass directly into a {SequencingPipeline}.
 # State changes on these tubes will automatically update the requests into the tubes
-# @note Most current activity is on subclasses of this purpose, especially IlluminaHtp::MxTubeNoQcPurpose
-#       As of 2019-10-01 only used directly by 'Lib Pool Norm' and 'Lib Pool SS-XP-Norm' which haven't been
-#       used since 2017-04-28 14:16:03 +0100
 class IlluminaHtp::MxTubePurpose < Tube::Purpose
-  # Called via Tube#transition_to
-  # Updates the state of tube to state
```
JamesGlover, line 12: More state change extraction. I had hoped to be able to eliminate a few of the tube state changers, and still think it should be possible, but they weren't quite as similar as I'd have liked. So keeping risk low for the time being.

app/models/plate.rb
```diff
@@ -61,6 +61,14 @@ def construct!
...
     # Returns the wells with their pool identifier included
+    # @note This is a method defined on the well association, and can be considered a scope.
+    # Ie. You can do plate.wells.with_pool_id.located_at(['A1'])
+    # It is a little special in that its behaviour is dependent on the {PlatePurpose#pool_wells plate purpose}.
+    # In practice this lets wells on PlatePurpose::Initial to pull their information from the requests, whereas
+    # other purposes jump back to the stock wells.
+    # Given this is usually called by the transfers themselves, prior to the transfer of the aliquots
+    # it can't be replaced purely by using request on aliquot (as it is partly responsible for setting that)
+    # however we can almost certainly simplify this whole process somewhat.
     def with_pool_id
```
JamesGlover, line 72: This was where my brain melted. I've tried to put a few pointers in the comments, and also added an issue  for a more long term fix.

app/models/plate_purpose/broadcast_library_complete.rb
```diff
@@ -1,25 +0,0 @@
-module PlatePurpose::BroadcastLibraryComplete # rubocop:todo Style/Documentation
```
JamesGlover, line 1: We have a number of these modules which were confined to a single purpose class, in some cases one which has been deprecated.

I've inlined all modules which only belonged to a single class.

app/models/plate_purpose/input.rb
```diff
@@ -10,10 +12,46 @@
 #       flexibility, such as by allowing library-manifest plates to pass straight in to the
 #       pipeline.
 class PlatePurpose::Input < PlatePurpose
-  include PlatePurpose::Stock
+  self.state_changer = StateChanger::InputPlate
 
+  UNREADY_STATE  = 'pending'
+  READY_STATE    = 'passed'
   WELL_STATE_PRIORITY = %w[pending started passed failed cancelled].freeze
```
JamesGlover, line 20: All the additions here are inlined from PlatePurpose::Stock as this was the only place it was used.

app/models/pulldown/plate_purposes.rb
```diff
@@ -62,45 +62,16 @@ module Pulldown::PlatePurposes
     ], ISCH_PURPOSE_FLOWS.first
   ].freeze
 
-  PLATE_PURPOSE_TYPE = {
```
JamesGlover, line 65: Unused constants

app/models/purpose.rb
```diff
@@ -40,7 +41,9 @@ class Purpose < ApplicationRecord
 
   before_validation :set_default_barcode_prefix
 
+  # rubocop:todo Rails/UniqueValidationWithoutIndex
   validates :name, format: { with: /\A\w[\s\w.\-]+\w\z/i }, presence: true, uniqueness: { case_sensitive: false }
+  # rubocop:enable Rails/UniqueValidationWithoutIndex
```
JamesGlover, line 47: Weirdly it was the removal of the brackets around 'plate_purposes' above that let rubocop detect this. There are a load of them in the rubocop_ignore file as well. SHould be easy enough to fix.

app/models/state_change.rb
```diff
@@ -1,4 +1,6 @@
-# Performs a change of state on an asset.
+# frozen_string_literal: true
+
```
JamesGlover, line 2: A bit of reorganization here to group together validations/callbacks etc. I'll try and highlight the actual changes as well.

app/models/state_change.rb
```diff
@@ -11,47 +13,55 @@ class StateChange < ApplicationRecord
   include Uuid::Uuidable
   include Asset::Ownership::ChangesOwner
 
-  belongs_to :user
-  validates :user, presence: true
+  attr_accessor :customer_accepts_responsibility
```
JamesGlover, line 17: belongs_to associations are required by default in Rails 5, but this functionality isn't enabled in SS. Slowly updating our belongs to as I come across them.

app/models/state_change.rb
```diff
@@ -11,47 +13,55 @@ class StateChange < ApplicationRecord
...
-  private :record_current_state_of_target
+
+  def state_changer
+    target.state_changer
+  end
 
   # After creation update the state of the target asset, leaving it to do the right thing.
   # After state change, update the owner
-  after_create :update_state_of_target
   def update_state_of_target
```
JamesGlover, line 58: Main change. Use our state changers. Allows us to remove transition_to from a whole host of places, and centralise state transitions.

features/api/illumina_b/top_of_the_pipeline.feature
```diff
@@ -121,107 +121,27 @@ Feature: The top of the Illumina-B pipeline
           "transfers": {
             "A1":  "A1",
             "A2":  "A2",
```
JamesGlover, line 123: Surprisingly enough, this was because I accidentally fixed a bug. Previously the transition to built up a list of 'bad_wells' and subtracted them from the transition hash. However this list didn't include wells that didn't actually exist.

spec/api/state_change_spec.rb
```diff
@@ -0,0 +1,198 @@
+# frozen_string_literal: true
```
JamesGlover, line 1: Migrated from the cucumber tests.

.rubocop.yml
```diff
@@ -32,6 +32,11 @@ Rails/SkipsModelValidations:
     - 'db/migrate_asset_refactor/*.rb'
     - 'features/support/step_definitions/*.rb'
```
JamesGlover, line 34: I was having some issues with the .rubocop.yml files in sub-directories, so I've pulled the configuration into here.

app/heron/factories/tube_rack.rb
```diff
@@ -97,7 +97,7 @@ def create_contents!
 
       def containers_for_locations
         @tube_rack.racked_tubes.each_with_object({}) do |racked_tube, memo|
-          memo[unpad_coordinate(racked_tube.coordinate)] = racked_tube.tube
+          memo[unpad_coordinate(racked_tube.coordinate)] = racked_tube.tube.receptacle
         end
```
JamesGlover, line 101: Tubes and other single receptacle labware were delegating aliquots and aliquots= to the receptacle. This was only ever intended to be a temporary situation when the labware-receptacle split happened. It was disrupting eager-loading of aliquots and their tags in the API, resulting it taking ages to get a tube.

Fixing this had a handful of knock on effects where we needed to modify that aliquot association. This was one example.

app/models/bulk_submission.rb
```diff
@@ -149,13 +149,6 @@ def process
 
             begin
               orders_processed = orders.map(&method(:prepare_order)).compact
-              library_types = orders_processed.map do |order|
-                order.request_options['library_type'] unless order.request_options.nil?
-              end.uniq
-              if library_types.size > 1
-                errors.add :spreadsheet, "Submission #{submission_name} has multiple library types: #{library_types.join(', ')}."
-                next
-              end
```
JamesGlover, line 152: We've dramatically simplified the way we find the request to fail, making this restriction uneccessary.

app/models/concerns/aasm/extensions.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
```
JamesGlover, line 1: This functionality was repeated in multiple places, so I extracted it.,

app/models/state_changer/base.rb
```diff
@@ -0,0 +1,39 @@
...
+  # - Potentially failing the associated {Request requests}
+  class Base
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    # The labware to update the state of
+    # @return [Labware] The labware to update the state of
+    attr_accessor :labware
+    # The user performing the action that led to the state change
+    # @return [Labware] The user performing the action that led to the state change
```
KatyTaylor, line 17:
```suggestion
    # @return [User] The user performing the action that led to the state change
```
?

app/models/state_changer/base.rb
```diff
@@ -0,0 +1,39 @@
...
+    #  @!attribute rw contents
+    #    @return [nil, Array<String>] Array of well locations to update, leave nil or empty for ALL wells
+    attribute :contents, default: nil
+    #  @!attribute rw customer_accepts_responsibility
+    #    @return  [Boolean] The customer proceeded against advice and will still
+    #                       be charged in the the event of a failure
+    attribute :customer_accepts_responsibility, :boolean, default: false
+
+    # Updates the state of the labware to the target state.  The basic implementation does this by updating
+    # all of the TransferRequest instances to the state specified.  If contents is blank then the change is assumed to
```
KatyTaylor, line 32:
```suggestion
    # all of the TransferRequest instances to the state specified.  If 'contents' is blank then the change is assumed to
```
Just because I read this three times before I realised 'contents' was an attribute, not just part of the sentence

app/models/state_changer/standard_mx_tube.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module StateChanger
+  # Handles the basic transitions of the {Tube::StandardMx} purpose tubes
+  # @note As of 201-10-01 only used for 'Standard MX' and 'Tag MX' tubes (Gatekeeper)
```
KatyTaylor, line 5: what's '201-10-01' meant to be? :)

KatyTaylor: Ah, I see that's not a new comment, it's copied

app/models/state_changer/standard_plate.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+module StateChanger
+  # Handles the basic transitions of a standard plate
+  class StandardPlate < StateChanger::Base
+    # Updates the state of the labware to the target state.  The basic implementation does this by updating
+    # all of the TransferRequest instances to the state specified.  If contents is blank then the change is assumed to
```
KatyTaylor, line 7:
```suggestion
    # all of the TransferRequest instances to the state specified.  If 'contents' is blank then the change is assumed to
```

As suggested elsewhere

app/models/state_changer/tube_base.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module StateChanger
+  # Shared behaviour for tubes
+  class TubeBase < StateChanger::Base
+    # Updates the state of the labware to the target state.  The basic implementation does this by updating
+    # all of the TransferRequest instances to the state specified.
+    # @return [Void]
+    def update_labware_state
+      update_associated_requests
```
KatyTaylor, line 10: this method isn't implemented, so this class shouldn't be used on its own?

app/models/state_changer/stock_tube.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    private
+
+    def transfer_requests
+      @transfer_requests ||= labware.transfer_requests_as_target
+                                    .where.not(state: TERMINATED_STATES)
+    end
+
+    def update_associated_requests
+      return unless target_state == 'failed'
```
KatyTaylor, line 16: on the plate equivalent I think you've got `if target_state == cancelled` ? is there a reason for the reversal of the logic?

app/models/transfer_request.rb
```diff
@@ -2,9 +2,17 @@
 
 # Every request "moving" an asset from somewhere to somewhere else without really transforming it
 # (chemically) as, cherrypicking, pooling, spreading on the floor etc
+#
+# @note The setting of submission_id and outer_request is quite complicated, and depends
+# on the exact route by which the transfer request has been created. The preferred route
+# is by setting outer_request explicitly. However much of the historic code handles it
+# via submission_id, either set explicitly, (eg Transfer::BetweenPlates#calculate_location_submissions)
+# or extracted from the pool_id attribute on well, which itself if populated as part of an
```
KatyTaylor, line 10:
```suggestion
# or extracted from the pool_id attribute on well, which itself is populated as part of an
```

config/default_records/tag_layout_templates/limber.dev.yml
```diff
@@ -136,3 +136,13 @@ IDT for Illumina v1 - 384 Quadrant:
...
+TSsc384-PCR1-nCoV-2019/V3/B:
+  :tag_group_name: TS_sc384_i7
+  :tag2_group_name: TS_sc384_i5
+  :direction_algorithm: TagLayout::InColumns
+  :walking_algorithm: TagLayout::WalkWellsOfPlate
+TSsc384-PCR2-nCoV-2019/V3/B:
+  :tag_group_name: TS_sc384_i7
+  :tag2_group_name: TS_sc384_i5
+  :direction_algorithm: TagLayout::InColumns
+  :walking_algorithm: TagLayout::WalkWellsOfPlate
```
KatyTaylor, line 148: Nice, I think I have worked around this not being there in the past 😆

sequencescape.code-workspace
```diff
@@ -4,5 +4,10 @@
 			"path": "."
 		}
 	],
-	"settings": {}
-}
\ No newline at end of file
+	"settings": {
+		"cSpell.words": [
+			"barcode",
```
KatyTaylor, line 9: it didn't know the word 'barcode'?

KatyTaylor (approved):

app/models/state_changer/base.rb
JamesGlover: Using {#contents} instead as then yard will link the documentation. Also needed to adjust the documentation on the attributes above to ensure it got picked up properly

app/models/state_changer/standard_mx_tube.rb
JamesGlover: I think it was supposed to be a date. I'll see if I can work it out.

JamesGlover: 2019

app/models/state_changer/tube_base.rb
JamesGlover: Yep, added documentation and explicit 'raise NotImplementedError'

app/models/state_changer/stock_tube.rb
JamesGlover: Plate seems to be `if target_state == 'failed'` but gates calling of `update_associated_requests` rather than acting as a guard clause. But `InitialStockTube` does use  return if target_state == 'cancelled'.

I need to have a quick step through the logic (as they may behave differently for states like 'started') but IIRC I should be able to standardize.

JamesGlover: Yep, its partly to handle the transition to started in those cases. (This is one of those cases where I *think* we can simplify, as the transfer requests should handle that themselves)

sequencescape.code-workspace
JamesGlover: Huh, hadn't realised I'd added this to the workspace. Oddly was experimenting with it in a different branch.

app/models/state_changer/stock_tube.rb
JamesGlover: Hmm, okay, realising I need to drastically extend the test coverage here. I was mostly relying on the existing tests as I refactored, but could do with some explicit ones to ensure coverage

JamesGlover: Auto-generated some tests to cover every permutation of the existing behaviour. Will use these to refactor, and then look at introducing something a little more sane.

JamesGlover:
So extended the test-suite using a number of automatically generated tests, and in doing so noticed the behaviour of `StateChanger::InitialStockTube` is incorrect. However the issue appears to be two-fold:

1) It's supposed to trigger request start where appropriate, but this was never updated to include the new process_1 and process_2 states. In practice we don't use these for tubes, so no biggie.
2) The request finding looks like it may be broken. Going to investigate this with real-world data.

app/models/state_changer/stock_tube.rb
JamesGlover: Auto tests there, and refactor done. But in doing so I noticed that `StateChanger::InitialStockTube` had an untriggered bug (It would do the wrong thing in the event of a state change we don't actually use on tubes).

app/models/state_changer/base.rb
```diff
@@ -6,34 +6,64 @@ module StateChanger
...
+  #       we actually depend on a lot of the differences in behaviour.
+  #
+  # Usage as of 2021-03-21
+  # Class                            Used   Last used
+  # StateChanger::MxTube             43758  2021-03-23
+  # StateChanger::StandardPlate      121793 2021-03-23
+  # StateChanger::InputPlate         14     2021-03-02
+  # StateChanger::InitialStockTube   30823  2021-03-23
+  # StateChanger::StockTube          5650   2021-03-11
+  # StateChanger::QcableLibraryPlate 213    2018-08-16
```
JamesGlover, line 21: I don't remove QcableLibraryPlate as its behaviour is quite distinct, and the pipeline its associated with is unused, rather than removed. We can remove it when we look at https://github.com/sanger/gatekeeper/issues/100

KatyTaylor (approved):

PR #3116, opened 2021-03-12
Aggregate comments show if request has no comments

### LINKED ISSUE #3115: GPL-918 Requests comments page not showing aggregated comments if no direct comments on request
The request comments index page if checking the request comments relationship when determining if there are comments to show, not the @comments variable.

JamesGlover:
Fixes #3115

- Use the @comments variable, not the association
- Convert the commentable to the base class in the shared view
- Add the commentable class name to the comment to better distinguish
  the source

### DIFF
--- a/app/views/requests/comments/_list.html.erb
+++ b/app/views/requests/comments/_list.html.erb
@@ -1,10 +1,10 @@
 
 <ul id="request_annotations_<%= @request.id %>" class="list-group">
-  <% if @request.comments.size > 0 %>
+  <% if @comments.present? %>
     <% @comments.each do |comment| %>
       <li class="list-group-item">
         <%= render partial: "shared/comment" ,
-                   locals: { comment: comment, commentable: @request.becomes(Request), update_target: "request_annotations_#{@request.id}" }%>
+                   locals: { comment: comment, update_target: "request_annotations_#{@request.id}" }%>
       </li>
     <% end %>
   <% else %>
--- a/app/views/shared/_comment.html.erb
+++ b/app/views/shared/_comment.html.erb
@@ -1,7 +1,7 @@
-<% commentable ||= comment.commentable %>
+<% commentable ||= comment.commentable.becomes(comment.commentable.class.base_class) %>
 <div class="comment-box">
   <div class="comment">
-    <h5 class="comment-title"><%= comment.title %></h5>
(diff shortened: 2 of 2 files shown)


app/views/shared/_comment.html.erb
```diff
@@ -1,7 +1,7 @@
-<% commentable ||= comment.commentable %>
+<% commentable ||= comment.commentable.becomes(comment.commentable.class.base_class) %>
 <div class="comment-box">
```
JamesGlover, line 2: This ensures that rails generates a link to, for example, /requests/3/comments rather than /request_subclass/3/comments which fails. I've got no idea why Rails isn't smart enough to perform this conversion itself in cases where the route doesn't exist for the subclass.

app/views/shared/_comment.html.erb
```diff
@@ -1,7 +1,7 @@
...
   <div class="comment">
-    <h5 class="comment-title"><%= comment.title %></h5>
+    <h5 class="comment-title"><%= comment.title %> <%= badge(commentable.class.name, style: 'info') %></h5>
     <div class="comment-description"><%= comment.description %></div>
```
JamesGlover, line 5: Because requests can contain comments for the labware/receptacles I just added a little badge to help determine where the comment came from.

app/views/requests/comments/_list.html.erb
```diff
@@ -1,10 +1,10 @@
 
 <ul id="request_annotations_<%= @request.id %>" class="list-group">
-  <% if @request.comments.size > 0 %>
+  <% if @comments.present? %>
     <% @comments.each do |comment| %>
```
JamesGlover, line 4: The actual bug was looking at the comments on request specifically, not the aggregate in @comments.

app/views/requests/comments/_list.html.erb
```diff
@@ -1,10 +1,10 @@
...
       <li class="list-group-item">
         <%= render partial: "shared/comment" ,
-                   locals: { comment: comment, commentable: @request.becomes(Request), update_target: "request_annotations_#{@request.id}" }%>
+                   locals: { comment: comment, update_target: "request_annotations_#{@request.id}" }%>
```
JamesGlover, line 7: We're no longer just looking at request comments, so don't override the commentable. I've moves the 'becomes' bit into the partial itself.

harrietc52 (approved):
Took me a while to understand whats going on (with Eduardo's help) but all good!

PR #3119, opened 2021-03-12
Add update Submission permission to all users

JamesGlover:


spec/models/ability_spec.rb
```diff
@@ -75,7 +75,7 @@
       Robot => %i[read],
       Sample => %i[edit release accession],
       Study => %i[create read print_asset_group_labels],
-      Submission => %i[create read new order_fields study_assets],
+      Submission => %i[create read update new order_fields study_assets edit],
```
harrietc52, line 78: Guess `edit` can be removed from [here](https://github.com/sanger/sequencescape/pull/3119/files#diff-a29fd9111cd60bf700042f0bf64897f4996c7f6b5281a9e4023ff00659db58d5R163) too now it is a basic permission

harrietc52 (approved):

PR #3122, opened 2021-03-16
Rubocop enable pending

JamesGlover:
- Enables pending cops
- Uses safe fixers
- Adds todo's for anything that can't be auto-corrected
- Manually validate the unsafe autocorrects
- Fixes a safe autocorrect that wasn't as safe as it claimed


app/models/bulk_submission.rb
```diff
@@ -212,6 +212,7 @@ def validate_entry(header, pos, row, index)
     return [translate(header), row[pos].try(:strip)] unless header.nil? && row[pos].present?
 
     errors.add(:spreadsheet, "Row #{index}, column #{pos + 1} contains data but no heading.")
+    nil
```
KatyTaylor, line 215: What's the reason for this? I feel like there are probably many methods like this, that end in an `errors.add` statement

app/models/bulk_submission.rb
```diff
@@ -224,24 +225,24 @@ def submission_structure
       csv_data_rows.each_with_index do |row, index|
         next if row.all?(&:nil?)
 
-        details = Hash[headers.each_with_index.map do |header, pos|
-                         validate_entry(header, pos, row, index + start_row)
-                       end ].merge('row' => index + start_row)
+        details = headers.each_with_index.filter_map do |header, pos|
+          validate_entry(header, pos, row, index + start_row)
+        end.to_h.merge('row' => index + start_row)
```
KatyTaylor, line 230: Related to above comment - not quite sure if the new version is equivalent to the old?

KatyTaylor (approved):

app/models/bulk_submission.rb
JamesGlover: This is a result of `Hash[array]` and `array.to_h` not being absolutely identical. Rubocop actually tries to auto-correct it, but we ended up with a failing test:

```ruby
[1] pry(#<BulkSubmission>)> array
=> [["user login", "user"],
 ["template name", "Illumina-A - Cherrypick for pulldown - Pulldown WGS - Paired end sequencing"],
 ["study name", "abc123_study"],
 ["Row 2, column 4 contains data but no heading."],
 ["submission name", "sub"],
 ["asset names", "testing123"],
 ["asset group name", "assetgroup123"],
 ["read length", "100"],
 ["fragment size from", nil],
 ["fragment size to", nil],
 ["library type", "Standard"],
 ["comments", "hello there"]]
[2] pry(#<BulkSubmission>)> Hash[array]
=> {"user login"=>"user",
 "template name"=>"Illumina-A - Cherrypick for pulldown - Pulldown WGS - Paired end sequencing",
 "study name"=>"abc123_study",
 "Row 2, column 4 contains data but no heading."=>nil,
 "submission name"=>"sub",
 "asset names"=>"testing123",
 "asset group name"=>"assetgroup123",
 "read length"=>"100",
 "fragment size from"=>nil,
 "fragment size to"=>nil,
 "library type"=>"Standard",
 "comments"=>"hello there"}
[3] pry(#<BulkSubmission>)> array.to_h
ArgumentError: wrong array length at 3 (expected 2, was 1)
from (pry):3:in `to_h'
```

In this case it's the error that's causing the problem, as its returning a single element array. When `Hash[]` finds this it sets it as a key with no value, whereas `to_h` instead throws an exception.

Both changes here fix this issue:
1) This map a translated column title to its value. As a result, the original behaviour, while harmless, isn't correct, as the error isn't the column title.
2) So we return nil instead, and then call filter_map to compact down to the valid columns only. (Filter map is cool)

```ruby
pry(#<BulkSubmission>)> [1,2,3,4,5,6].filter_map {|v| v*3 if v.even? }
=> [6, 12, 18]
```

So there is a slight behaviour change, but its one to more 'correct' behaviour.

That said, bulk submission could definitely do with a refactor pass. `validate_entry` is somehow also handling column translation as well.

KatyTaylor: Cool thanks for explanation :)

PR #3127, opened 2021-03-24
Update yanked Mimemagic

JamesGlover:
*All* versions of mimemagic below 0.3.6 were janked due to
a GPL license violation.

https://github.com/minad/mimemagic/issues/97


JamesGlover:
Marked as urgent as breaks all builds

KatyTaylor (approved):

Issue #3109, opened 2021-03-10
GPL-916-1 Add Sequencescape sample_metadata fields to MLWH sample table


neilsycamore:
See sibling issue here - https://github.com/sanger/unified_warehouse/issues/212

DTOL (Shane M) raised via https://rt.sanger.ac.uk/Ticket/Display.html?id=713449 that the "organism part" field in sample_metadata table 
was missing in the mlwh sample table

It is considered that rather than adding requested columns ad hoc that all missing columns should be added.
Attached is a comparison of the fields in each table with TO ADD marked against those that are absent.
[missing_mlwh_sample_fields.xlsx](https://github.com/sanger/crawler/files/6116727/missing_mlwh_sample_fields.xlsx)

Naming new fields:

Note some field names compare across tables (ss => mlwh)
eg cohort => cohort
whilst some differ
sample_public_name => public_name
some not so obvious
sample_strain_att => strain
and some just outright left field
sample_sra_hold => sample_visibility

The 'sample_' prefix naming convention in the sample_metadata table is a legacy that does not persist in the mlwh (apart it would seem for sample_visibility!). It would be good to not replicate sample_ 

NB
Look at the sample_reference_genome_old field.
I added reference_genome_id way back (when I still had all my teeth) to reference the new table (before Ref genome names were hard coded in) and sample_reference_genome_old was supposed to be removed (never happened)

neilsycamore:
The request owner has just come back with "No need to get this added to mlwh. I'll find a work around."

JamesGlover:
Going to suggest that `concentration` goes in as `customer_supplied_concentration` to avoid confusion with measured concentrations.

JamesGlover:
We should also avoid any _id fields if the corresponding record is not in the WH. So user_id_of_consent_withdrawn for example should be consent_withdrawn_by and then the user login

JamesGlover:
Note that donor_id is distinct from the above in that it doesn't reference a donors table.

KatyTaylor:
What's the story of supplier_plate_id?
If I do the following on the training db, I get 0 rows:
```
SELECT supplier_plate_id
FROM sample_metadata
WHERE supplier_plate_id IS NOT NULL
  AND supplier_plate_id != ''
LIMIT 100
```

KatyTaylor:
Went down a bit of a rabbit hole with the `rnai` field, to find a representative value for a test. Ended up with this where clause:
```
WHERE rnai IS NOT NULL
[...]

PR #3137, opened 2021-03-29
Gpl 916 add sample metadata fields mlwh

KatyTaylor:


Gemfile.lock
```diff
@@ -35,44 +35,44 @@ GEM
   specs:
     aasm (5.1.1)
       concurrent-ruby (~> 1.0)
-    actioncable (5.2.4.5)
-      actionpack (= 5.2.4.5)
+    actioncable (5.2.5)
+      actionpack (= 5.2.5)
```
KatyTaylor, line 39: I did a bundler update as I was having mimemagic issues. It seemed to get rid of the dependency altogether.

app/models/api/sample_io.rb
```diff
@@ -68,6 +68,36 @@ def json_root
...
+    map_attribute_to_json_attribute(:organism_part)
+    map_attribute_to_json_attribute(:time_point)
+    map_attribute_to_json_attribute(:disease)
+    map_attribute_to_json_attribute(:subject)
+    map_attribute_to_json_attribute(:treatment)
+    map_attribute_to_json_attribute(:genome_size)
+    map_attribute_to_json_attribute(:saphyr)
+    map_attribute_to_json_attribute(:pacbio)
+    map_attribute_to_json_attribute(:date_of_consent_withdrawn)
+    map_attribute_to_json_attribute(:user_id_of_consent_withdrawn)
```
KatyTaylor, line 100: The following fields are all blank or not applicable, so wondering whether to include them:

supplier_plate_id
rnai
genome_size
saphyr
pacbio
sample_reference_genome_old (not quite all blank)

More detail in comments on issue - https://github.com/sanger/sequencescape/issues/3109

KatyTaylor, line 100: For the ones where we want to change the name, it looks like the place to do that is here? https://github.com/sanger/unified_warehouse/blob/develop/app/models/sample.rb

Gemfile.lock
JamesGlover: Yep, they updated rails/marcel to remove it entirely. Which should make things simpler for us in future.

app/models/api/sample_io.rb
JamesGlover: I'd vote for performing the translation in Sequencescape, rather than the warehouse, isolates the weirdness to one code-base.

JamesGlover: So looking at existing translations:
```
id: :id_sample_lims,
uuid: :uuid_sample_lims,
```
Feel like they belong in the WH, as they deal with specific WH implementation.

In an ideal world I guess:
```
      sample_common_name: :common_name,
      sample_description: :description,
      sample_ebi_accession_number: :accession_number,
      sample_taxon_id: :taxon_id,
      sample_public_name: :public_name,
      sample_sra_hold: :sample_visibility,
      sample_strain_att: :strain
 ```

Would all belong in Sequencescape world, as it feels like they are dealing with SS implementation decisions. (It mostly comes from time when the WH was actually the v0.5 API. But that is no longer exposed elsewhere, so we're relatively safe making changes)

One thing I will note with translations, is that you can still supply a message to the WH with the translated term, and it will work. So its actually not too tricky to fix these issues.

File: lib/label_printer/label_printer/label/batch_tube.rb [hunk 3]
```diff
@@ -32,7 +32,7 @@ def round_label_top_line(tube)
       end
 
       def round_label_bottom_line(tube)
-        tube.is_a?(PacBioLibraryTube) ? source_plate_barcode(tube).split('').last(4).join : super
+        tube.is_a?(PacBioLibraryTube) ? source_plate_barcode(tube)[-4..] : super
       end
```
JamesGlover, line 36: I made a similar fix when updating rails in another branch, this actually doesn't quite work:
```
irb(main):022:0> 'abcdefghijk'[-4..]
=> "hijk"
irb(main):023:0> 'ijk'[-4..]
=> nil
```
I don't know what the risk of us getting strings of three characters here, but I decided to play it safe and went with:
```
      def round_label_bottom_line(tube)
        if tube.is_a?(PacBioLibraryTube)
          source_plate = source_plate_barcode(tube)
          # Return the last 4 characters. The fallback handles scenarios where
          # we have fewer than 4 characters in the string, as:
          #     'abc'[-4..] # => nil
          source_plate[-4..] || source_plate
        else
          super
        end
      end
      ```

File: app/models/api/sample_io.rb [hunk 4]
```diff
@@ -68,12 +68,47 @@ def json_root
...
```
JamesGlover, line 100: Regardless of where we translate the rest, user_id_of_consent_withdrawn should definitely be something else, as the message contains the login, not the id.

app/models/api/sample_io.rb
```diff
@@ -68,12 +68,47 @@ def json_root
...
 
   extra_json_attributes do |_object, json_attributes|
     if json_attributes['reference_genome'].blank?
       json_attributes['reference_genome'] = nil
     end
+
+    user_id = json_attributes['user_id_of_consent_withdrawn']
+    if user_id.present? && User.find(user_id)
+      json_attributes['user_id_of_consent_withdrawn'] = User.find(user_id)&.login
+    end
```
JamesGlover, line 111: User.find(x) raises ActiveRecord::RecordNotFound if the user is missing.

Instead this should probably be:
```ruby
   if user_id.present?
      json_attributes['user_id_of_consent_withdrawn'] = User.find_by(id: user_id)&.login
    end
```
find_by(id:x) returns nil if not found. This approach also gets rid of the double lookup of the user.

app/models/api/sample_io.rb
JamesGlover: I'll pop the comments about this on the story

app/models/api/sample_io.rb
KatyTaylor: Good spot thanks

JamesGlover (approved):

KatyTaylor:
Deployed this to UAT, without yet deploying the warehouse changes. Messages fail at the warehouse consumer end, with `unknown attribute 'customer_measured_volume' for Sample.` (see [kibana](http://uat.kibana.psd.sanger.ac.uk/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:'2021-04-08T12:50:33.609Z',mode:absolute,to:'2021-04-08T12:51:33.609Z'))&_a=(columns:!(beat.hostname,log_message),index:a475edd0-a16d-11e9-a8f6-dde4b738d716,interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc))) log)

PR #3147, opened 2021-04-06
Remove ... to allow flog to work

JamesGlover:


JamesGlover:
So looks like code climate doesn't include the specs for the flog run, eliminating a lot of false positives. Lots of issues being picked up here, and a quick oggle of the top ones looks like its grabbing some pretty crunch methods.

PR #3154, opened 2021-04-08
Move ancestor of purpose onto labware

### LINKED ISSUE #3153: Method missing errors on receptacles page
On visitng the receptacles page for an MX library tube:

    A NoMethodError occurred in receptacles#show:

      undefined method `ancestors_of_purpose' for #<MultiplexedLibraryTube:0x00007f33868771c8>
    Did you mean?  ancestor_of_purpose
      app/models/purpose.rb:70:in `block in source_plates'

This is caused by ancestors_of_purpose being defined on plate, not labware
and yet getting referenced by from plate purpose.

Should just be able to move the methods on to labware.

JamesGlover:
Ancestor of purpose was getting called on tubes, but was failing.
The logic holds for all labware, so should be defined on the base class.

Fixes #3153

### DIFF
--- a/app/models/tube.rb
+++ b/app/models/tube.rb
@@ -40,12 +40,6 @@ class Tube < Labware
     submissions.ids.first
   end
 
-  def ancestor_of_purpose(ancestor_purpose_id)
-    return self if plate_purpose_id == ancestor_purpose_id
-
-    ancestors.order(created_at: :desc).find_by(plate_purpose_id: ancestor_purpose_id)
-  end
-
   def name_for_label
     primary_sample&.shorten_sanger_sample_id.presence || name
   end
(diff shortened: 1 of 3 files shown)


app/models/tube.rb
```diff
@@ -40,12 +40,6 @@ def pool_id
     submissions.ids.first
   end
 
-  def ancestor_of_purpose(ancestor_purpose_id)
-    return self if plate_purpose_id == ancestor_purpose_id
-
-    ancestors.order(created_at: :desc).find_by(plate_purpose_id: ancestor_purpose_id)
```
KatyTaylor, line 46: I expect you noticed but just pointing out that this one uses 'created_at' rather than 'id' to order. But we have made a lot of changes recently in this direction, so changing it is probably a good thing?

JamesGlover: Oh, nice catch. I hadn't actually noticed. I did actually eyeball them to make sure they were the same, but obviously missed that.

You're right though, this should be a good thing.

PR #3165, opened 2021-04-20
GPL-985-2 Added accepted plate format HSLnnnnnnnnn

### LINKED ISSUE #3164: GPL-985-2 Support new barcode format HSLnnnnnn
**User story**
As PAM, I want to support HSL barcodes in Sequencescape for automatic creation through lighthouse workflows (Sentinel and Beckman)

The new barcode format to support is: HSLnnnnnn (e.g. HSL123456)

Original story [GPL-983](https://github.com/sanger/baracoda/issues/111)

Alan K.

**Acceptance criteria**
- [x] Support creating a plate with barcode HSLnnnnnn in Sequencescape

**Dependencies**
This story is related with story: https://github.com/sanger/baracoda/issues/111
This story is coupled with Crawler https://github.com/sanger/crawler/issues/316

**Additional context**

Alan K… I can confirm that the HSL barcode that we should build into the system is HSLnnnnnn (e.g. HSL123456).
They will use an old eagle barcode (EGG123456) as a temporary solution.

harrietc52:
Closes #3164 
Coupled with Crawler [PR](https://github.com/sanger/crawler/pull/323)

### DIFF
--- a/app/models/barcode.rb
+++ b/app/models/barcode.rb
@@ -49,7 +49,8 @@ class Barcode < ApplicationRecord
     randox_eagle: 28,
     randox_v2: 29,
     glasgow_v3: 30,
-    uk_biocentre_v5: 31
+    uk_biocentre_v5: 31,
+    health_services_laboratories: 32
   }
 
   # Barcode formats which may be submitted via sample manifests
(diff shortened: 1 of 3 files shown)


emrojo (approved):

KatyTaylor (approved):

pjvv (commented):
It feels like we keep adding regex here which is becoming superfluous since there are probably quite a few which are matching the same patterns. Did @JamesGlover have an idea about tidying this up?

JamesGlover:
Yep, on holiday this week. Thought I put an issue in but can't find it. Might have been a comment one one of the previous barcode format issues.

Essentially:
1. There is already a catch all barcode format. (Something like external barcode formats)
2. The array of acceptable formats was intended to describe only those that could come in via manifests, hence why the catch all was missing.
3. The heron plate import currently shares the lookup method with the one used by manifests,
4. Instead it could:
 - always use the catch all format
 - Use the catch all as fallback

Only issue is the Eagle route, as that does come. in via manifest. (Although check was mainly to provide validation that a customer hadn't just arbitrarily changed the field)

PR #3169, opened 2021-05-04
Issue3056/gpl 951 remove oracle adapters + ubw intergration

### LINKED ISSUE #3056: GPL-951 Remove oracle adapters + UBW intergration
**Description**
The oracle adapters were initially added for SNP integration, which since been removed. However by the point at which this was finally turned off, they also supported agresso integration. However I do not believe this feature has ever been used in anger.

Oracle driver requirements add complexity to:
1) Deployment process - especially as download of latest versions cannot be automated
2) Development - The warehouse gem group needs to be disabled. bin/setup tries to automate this.
3) CI - Again we need to constantly disable the warehouse gem group

If we can't remove it, we should look at flagging the group as 'optional' and only installing it where explicitly needed.

James Glover

Rich R will probably know what usage is like

Any other useful context or information that may be useful.

JamesGlover:
Closes #3056

* Remove oracle dependencies
* Remove ubw billing reports/items and integrations
* Archive billing tables


db/migrate/20210504153704_archive_billing_tables.rb
```diff
@@ -0,0 +1,13 @@
...
+# Archive the billing tables to env_sequencescape_archive
+class ArchiveBillingTables < ActiveRecord::Migration[5.2]
+  include MigrationExtensions::DbTableArchiver
+
+  def change
+    check_archive!
+    archive!('billing_items')
+    archive!('billing_products')
+    archive!('billing_product_catalogues')
+  end
```
emrojo, line 12: We have to check if we have permissions to create new databases in production with current user.

JamesGlover: We've archive tables before, and the database did exist then, but someone seems to have dropped it?!

JamesGlover: We have `GRANT ALL PRIVILEGES ON `sequenc%`.* TO 'psdp_rw'@'%'` though, so should be fine.

JamesGlover: Table was archived to NFS.

README.md
```diff
@@ -115,14 +115,6 @@ gem install bundler
...
 ```
 
-If you experience `An error occurred while installing ruby-oci8` then disable the oracle gem installation:
-
-```shell
-bundle config without warehouse
-```
-
-The oracle adapters are not needed for development.
-
```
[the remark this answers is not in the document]
JamesGlover: Removed.

db/migrate/20210504153704_archive_billing_tables.rb
JamesGlover: We do have the correct permissions, I double checked. Deleted the comment with the actual grants as it included the username. Not really a major concern, but I'll avoid leaking it if I can. (Edit kept the comment history)

emrojo: Ok

emrojo (approved):

PR #3177, opened 2021-05-06
Fix buffer addition to zero concentration

JamesGlover:
- Apply buffer to zero concentration wells.
- Reduce used of nested min/max arrays in favour of
 clamps and named limits.

### DIFF
--- a/app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
+++ b/app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
@@ -1,4 +1,24 @@
-module Cherrypick::VolumeByNanoGramsPerMicroLitre # rubocop:todo Style/Documentation
+# Included in {Well} to provide calculations for cherrypicking
+module Cherrypick::VolumeByNanoGramsPerMicroLitre
+  #
+  # Used in the cherrypicking process to calculate the relative volumes of source
+  # and buffer required to reach the target volume and concentration.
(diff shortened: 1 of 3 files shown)


app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
```diff
@@ -1,36 +1,68 @@
...
                                                                         source_concentration)
 
+    # @note Here we appear to set the concentration based on the required concentration, regardless of whether we hit it
+    # it or not. Checking if this behaviour is desired RT#719205
     well_attribute.concentration    = concentration_required
     well_attribute.requested_volume = volume_required
+    # Similarly we set current volume based on required. This is only untrue in rare edge cases though
+    # (When you have almost all your required volume from your source, then add more buffer than intended
+    #  due to minimum robot picks)
     well_attribute.current_volume   = volume_required
```
emrojo, line 35: This line is not part of the changes, but I dont see a match: volume_required  is what is intended to pick, but not necessarily will happen? Then current_volume is going to be that but there will be cases where the real current_volume will be: well_attribute.buffer_volume+ well_attribute.picked_volume. So picked_volume_buffer_volume is what really defines the current volume?

emrojo (approved):
All good, keeps all the logic but fixing the 0.0 buffer assignation when source concentration 0 that was causing the problem.

PR #3179, opened 2021-05-07
Fix tag duplication

JamesGlover:
Tags were being duplicated on each deploy!

### DIFF
--- a/lib/record_loader/tag_group_loader.rb
+++ b/lib/record_loader/tag_group_loader.rb
@@ -15,7 +15,7 @@ module RecordLoader
 
       TagGroup.create_with(options).find_or_create_by!(name: name).tap do |tag_group|
         tag_attributes = tags.map { |map_id, oligo| { map_id: map_id, oligo: oligo } }
-        tag_group.tags.import(tag_attributes)
+        tag_group.tags.import(tag_attributes) if tag_group.tags.empty?
       end
     end
   end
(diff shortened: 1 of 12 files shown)


lib/record_loader/tag_group_loader.rb
```diff
@@ -15,7 +15,7 @@ def create_or_update!(name, options)
 
       TagGroup.create_with(options).find_or_create_by!(name: name).tap do |tag_group|
         tag_attributes = tags.map { |map_id, oligo| { map_id: map_id, oligo: oligo } }
-        tag_group.tags.import(tag_attributes)
+        tag_group.tags.import(tag_attributes) if tag_group.tags.empty?
       end
```
JamesGlover, line 19: Fix the main issue

spec/lib/record_loader/barcode_printer_type_loader_spec.rb
```diff
@@ -5,10 +5,12 @@
 
 # This file was initially generated via `rails g record_loader`
 RSpec.describe RecordLoader::BarcodePrinterTypeLoader, type: :model, loader: true do
-  subject(:record_loader) do
+  def a_new_record_loader
     described_class.new(directory: test_directory, files: selected_files)
   end
```
JamesGlover, line 11: When attempting to reproduce the issue in tests I couldn't, as it was using the same instance of record-loader.
I've ported the updated behaviour to all other record loaders, just in case we were missing other similar bugs.
I'll also update the template in record loader itself.

stevieing (approved):

emrojo (approved):

PR #3186, opened 2021-05-13
Experiment with prettier

JamesGlover:
Using guides here:
https://prettier.io/docs/en/install.html
and here:
https://github.com/prettier/plugin-ruby

Decided to opt straight for installing via yarn, as mentions editor
integration better:

`yarn add --dev --exact prettier`
`echo {}> .prettierrc.json`

Then built up the `.prettierignore` file from `.gitignore`
and added in `.rubocop_todo.yml` as its autogenerated
and `.history/*` to handle one of my plugins.

Added
```
inherit_from:
  - node_modules/@prettier/plugin-ruby/rubocop.yml
```

To `rubocop.yml` and ran prettier.

Needed to fix duplicate keys in a couple of yaml files.

Then ran rubocop, some conflicts, but regenerated
my todo and down to 9 all Lint/RedundantCopDisableDirective, these seem
to be associated with metrics cops.

Given this is a massive commit anyway, decided to
inline all the cop todos for these.

Actually, may inline *all* metrics cops (bar line length)

Done this, had a bit of a battle with rubocop infinite
loops, but managed to fix those.

Now an eslint battle
`yarn add --dev eslint-config-prettier`
Then add 'prettier' to the esLintrc files
under extends (in last place)
Removed some custom formatting configuration

Added `config/cucumber.yml` and `config/database.yml`
to prettierignore as it
seems to be confused by the erb

Also, some battle with rubocop metrics cops still.
Mostly doesn't like inline disables, issue in to fix this:
https://github.com/prettier/plugin-ruby/issues/819
Manually fixed these by switching to the block
rathet than inline style

Closes #

*
*
* ...


JamesGlover:
Ahh, rubocop now dependent on prettier being installed for configuration. Thats a bit annoying.
This is mostly so see what people think of the formatting decsions though, do won't bother fixing until we know this is something we want to progress.

JamesGlover:
Vote on adoption here with thumbs up thumbs down reactions.

PR #3206, opened 2021-06-01
DPL-027 bug - Sample edit page adding consent withdrawn metadata

### LINKED ISSUE #3203: DPL-027 [BUG] Sample edit page adding consent withdrawn metadata
When editing a sample date_of_consent_withdrawn and user_id_of_consent_withdrawn is being saved in the sample_metadata.
Note: The consent_withdrawn flag is NOT being set.

This combined with 
https://github.com/sanger/sequencescape/issues/2937
https://github.com/sanger/sequencescape/issues/3104
makes sample editing a conduit for bad data

example id: 4495321, name: “CV001_KM8853544”

When fixed the affected data will require fixing
select count(*) from sample_metadata where date_of_consent_withdrawn is not null and consent_withdrawn = 0
800 records

KatyTaylor:
Closes #3203


KatyTaylor:
Would appreciate opinions on whether you think this is correct behaviour now. For instance, it's not obvious whether editing consent withdrawn from true to false should blank out the timestamp and user id fields - what if it's a mistake?

andrewsparkes (approved):

spec/features/pipelines/cherrypick/cherrypick_pipeline_spec.rb
```diff
@@ -190,8 +188,7 @@ def initialize_plates(plates)
         [plates[0], plates[2], plates[1]]
       end
 
-      let(:expected_tecan_file) do
-        <<~TECAN
+      let(:expected_tecan_file) { <<~TECAN }
```
andrewsparkes, line 191: has prettier done the right thing there? should the close curly brace not be after the second TECAN? or left as a do end

KatyTaylor: You're right that does look dodgy, thanks

PR #3209, opened 2021-06-02
Gpl 694 sample edit not saving data 2

KatyTaylor:
Closes #3208 

* Get rid of checktext_field and replace with text_field
* This fixes the bug
* Gets rid of technical debt - a lot of complicated code around this
* Not convinced of any UX benefit to the different design of the checktext fields
* Not sure why these fields use one style of UI and the other fields don't
* Even though the checktext field default appeared to be 'N/A' (when viewing in the UI), when the page was saved, a null value was saved in the database


KatyTaylor:
With checktext fields:
Screenshot 2021-06-02 at 11 42 31

Without checktext fields:
Screenshot 2021-06-02 at 13 56 19

andrewsparkes (approved):

PR #3217, opened 2021-06-10
Supplier name validation fix

KatyTaylor:
don't apply the validation to supplier name if you're blanking it

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -230,7 +230,7 @@ class Sample < ApplicationRecord # rubocop:todo Metrics/ClassLength
                 with: /\A[[:ascii:]]+\z/,
                 message: 'only allows ASCII'
               },
-              if: :supplier_name_changed?
+              if: :supplier_name_changed? && :supplier_name?
 
     # here we are aliasing ArrayExpress attribute from normal one
     # This is easier that way so the name is exactly the name of the array-express field
--- a/spec/models/sample_spec.rb
+++ b/spec/models/sample_spec.rb
@@ -149,4 +149,22 @@ RSpec.describe Sample, type: :model, accession: true, aker: true do
       end
     end
   end
+
+  context 'updating supplier name' do
+    let(:sample) { create :sample }
+
+    it 'validates that supplier name allows only ASCII characteres' do
+      expect(sample.sample_metadata.supplier_name).to be_nil
+      sample.sample_metadata.supplier_name = 'भारत'
+      expect(sample.sample_metadata.save).to eq false
+    end
+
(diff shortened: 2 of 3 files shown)


spec/models/sample_spec.rb
```diff
@@ -149,4 +149,22 @@
...
+    it 'validates that supplier name allows only ASCII characteres' do
+      expect(sample.sample_metadata.supplier_name).to be_nil
+      sample.sample_metadata.supplier_name = 'भारत'
+      expect(sample.sample_metadata.save).to eq false
+    end
+
+    it 'can have the supplier name blanked' do
+      expect(sample.sample_metadata.supplier_name).to be_nil
+      sample.sample_metadata.update!(supplier_name: 'something')
+      expect(sample.sample_metadata.supplier_name).not_to be_nil
```
harrietc52, line 165: Would save eq true be more useful than `update!`?

spec/models/sample_spec.rb
```diff
@@ -149,4 +149,22 @@
...
+      sample.sample_metadata.update!(supplier_name: nil)
+      expect(sample.sample_metadata.supplier_name).to be_nil
```
harrietc52, line 167: And save eq true here too?
Is `null` acting the same as `nil`?

app/models/sample.rb
```diff
@@ -230,7 +230,7 @@ class Metadata
                 with: /\A[[:ascii:]]+\z/,
                 message: 'only allows ASCII'
               },
-              if: :supplier_name_changed?
+              if: :supplier_name_changed? && :supplier_name?
```
harrietc52, line 233: Does this automatiicaly trigger the `touch` which resends the sample to the MLWH?

harrietc52 (approved):
Looks good 👍 few minor questions

app/models/sample.rb
KatyTaylor: Hopefully it should be sent to the warehouse when we update to nil. Didn't test it locally though so I will def check this in UAT.

spec/models/sample_spec.rb
KatyTaylor: `nil` is just Ruby for `null`

spec/models/sample_spec.rb
KatyTaylor: Not sure, what else would it tell us?

app/models/sample.rb
KatyTaylor: Tested this in UAT - updates MLWH fine when filling in and blanking supplier_name

PR #3231, opened 2021-06-21
DPL-024 bug incorrect uuids due to stock resource problem

### LINKED ISSUE #3230: DPL-024 BUG - incorrect uuids due to stock_resource problem [Heron]
**Describe the bug**
Some records in the lighthouse_sample table have uuids that do not match their corresponding records in the sample table.
This is because they were missed during [the migration](https://github.com/sanger/unified_warehouse/issues/204), due to the labware_human_barcode in the stock_resource table referring to the DN... sanger barcode rather than the (lighthouse) foreign barcode - meaning the join didn't find them.

It looks like there is a bug with the sample manifest behaviour where, if there is a foreign barcode, it registers the 'Sanger' stock plate barcode rather than the foreign barcode. Maybe `register_stock!` is called before the second barcode is registered?

See RT ticket for an example record.

**RT Ticket Number**
https://rt.sanger.ac.uk/Ticket/Display.html?id=720844

**Acceptance Criteria**
- Fix the data so that the join on uuid between lighthouse_sample and sample gets these records
- Make sure it doesn't happen again - probably involves changing the Sequencescape manifest code

KatyTaylor:
Closes #3230 

Several changes to the manifest upload process.

Before:
* Samples were created as a side effect of the `sample` method on Row, so were created whenever this happened to be called first, which was when calling valid?
* register_stock! was being called on the wells when the samples were created, which was before the specialised fields were processed - so if the manifest contained foreign barcodes, the stock resource table referred to the old sanger barcode rather than the foreign barcode.

After:
* Samples are created explicitly during the 'process' phase
* The sample metadata and specialised field validation that relied on samples being present has therefore also been moved to the 'process' phase.
* The 2 specialised fields relating to foreign barcodes have been moved to before sample creation, so that register_stock! is not called until after the foreign barcodes are inserted


KatyTaylor:
Update:

I've made some changes based on @JamesGlover & @stevieing 's feedback (thank you!).

- Removed some of the many places validation was being done (only the unnecessary ones).
- Refactored the top level controller method.
- Added tests to make sure the validation doesn't cause sample creation as a side effect.
- Moved the Rails transaction one level up, so that it includes the validation (in case record creation creeps back in there in the future)
- Checked out the logs to confirm that the transaction causes the records to be committed to the database, and THEN the Rabbit MQ messages get fired off.

Key understanding from chat with James - it didn't matter so much that the method register_stock! was getting called before the foreign barcode was created - this should work if they are both inside a transaction. The problem was that the validation was outside the transaction, and it had this side effect of creating the samples.

I was thinking about the 'initialize -> validate -> save' workflow that both of you mentioned, and would like to implement that here as it would be simpler than what I've got. But, I got stuck because I think it would involve 'building' / initializing several linked objects - samples, aliquots, maybe study_samples - and I'm not sure how to do that (and might be time consuming). So I've left it like this for now.

app/sample_manifest_excel/sample_manifest_excel/upload/row.rb
```diff
@@ -19,8 +19,8 @@ class Row # rubocop:todo Metrics/ClassLength
       validates :number, presence: true, numericality: true
       validate :sanger_sample_id_exists?, if: :sanger_sample_id
       validates_presence_of :data, :columns
-      validate :check_sample_present
-      validate :sample_can_be_updated
+
+      validate :check_early_specialised_fields
```
JamesGlover, line 23: So there's a fair chunk of complexity coming out of the separation of early and late specialised fields. So just want to double check my understanding.

1) With the transaction moved outwards, this change is no longer actually required in order to fix the bug.
2) However it still results in the mini stock message being added to the transaction before the stock barcode is updated, which just feels a bit wrong
3) I think you also mentioned a worry that if we moved the generate stock event later and separate it from the register stock event, then we may end up with samples getting created, but not registered.

I'm not sure that the advantages of point 3 outweighs the complexity, especially when the counterpoint it we'd registering stocks that hadn't been created correctly, and so would be definition be faulty. (Such as what was happening before if validation failed.) 

Definitely a point of discussion though, but inclined to suggest moving `receptacle.register_stock!` to a callback after manifest processing.

app/sample_manifest_excel/sample_manifest_excel/upload/base.rb
```diff
@@ -66,40 +66,36 @@ def derive_sample_manifest
...
 
       def data_at(column_name)
         required_column = columns.find_by(:name, column_name)
         rows.data_at(required_column.number) if required_column.present?
       end
 
+      def complete
+        sample_manifest.finished!
+      end
+
```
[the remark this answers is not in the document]
JamesGlover: Oh, and I also hate it when a code review pulls up a method which all I did was move somewhere else in the file.

app/sample_manifest_excel/sample_manifest_excel/upload/row.rb
```diff
@@ -87,8 +87,9 @@ def update_sample(tag_group, override) # rubocop:todo Metrics/AbcSize
         if sample.updated_by_manifest && !override
           @sample_skipped = true
         else
-          update_specialised_fields(tag_group)
+          update_late_specialised_fields(tag_group)
           asset.save!
+          update_metadata_fields
```
JamesGlover, line 92: Trying to unpick this one, as it also appears below, as part of sample validation. (Which makes sense, as you can't validate the data before its set) Why is it also called here?

app/sequencescape_excel/sequencescape_excel/specialised_field/sanger_plate_id.rb
```diff
@@ -18,32 +18,34 @@ class SangerPlateId
       # rubocop:todo Metrics/PerceivedComplexity
       # rubocop:todo Metrics/MethodLength
       # rubocop:todo Metrics/AbcSize
-      def update(attributes = {}) # rubocop:todo Metrics/CyclomaticComplexity
-        return unless valid? && attributes[:aliquot].present? && foreign_barcode_format.present?
+      def update(_attributes = {}) # rubocop:todo Metrics/CyclomaticComplexity
+        return unless valid? && foreign_barcode_format.present?
+
+        well = sample_manifest_asset.asset
```
JamesGlover, line 24: You can actually just do asset here as SpecialisedField::Base has:
`delegate :asset, :sample, to: :sample_manifest_asset`

JamesGlover: Might make sense to add sample manifest to that delegate as well

app/sequencescape_excel/sequencescape_excel/specialised_field/sanger_tube_id.rb
```diff
@@ -18,32 +18,33 @@ class SangerTubeId
       # rubocop:todo Metrics/PerceivedComplexity
```
JamesGlover, line 18: Not sure we still need two separate classes here now tubes/plates are both labware with receptacles. But not worth touching at the moment.

features/sample_manifest/sample_manifest.feature
```diff
@@ -388,7 +388,7 @@ Feature: Sample manifest
     Examples:
       | initial                           | update                            | state     | message                                                           |
       | sample_manifest_a_accessions.csv  | sample_manifest_no_accessions.csv | Completed | Sample manifest successfully uploaded                             |
-      | sample_manifest_a_accessions.csv  | sample_manifest_b_accessions.csv  | Completed | The accession number does not match the existing accession number |
+      | sample_manifest_a_accessions.csv  | sample_manifest_b_accessions.csv  | Failed    | The accession number does not match the existing accession number |
```
JamesGlover, line 391: Wuh? Well that was almost certainly my fault. As I both wrote the original code, and ported that feature across.

spec/features/sample_manifests/uploader_for_manifests_with_tag_sequences_spec.rb
```diff
@@ -97,7 +97,7 @@
           visit('sample_manifest_upload_with_tag_sequences/new')
           attach_file('File to upload', test_file)
           click_button('Upload manifest')
-          expect(page).to have_content('Your sample manifest couldn\'t be uploaded. See errors below.')
+          expect(page).to have_content('Your sample manifest couldn\'t be uploaded.')
```
JamesGlover, line 100: Was this just because the errors are now displayed differently, or do you no longer see a breakdown?

spec/features/sample_manifests/uploader_for_manifests_with_tag_sequences_spec.rb
```diff
@@ -225,7 +225,7 @@
...
           expect(page).to have_content('Same tags AA, TT are used on rows 10, 15.')
```
JamesGlover, line 229: Ahh, from this I assume the errors are still shown.

JamesGlover (changes requested):
A few minor suggestions, and then the more significant discussion about the early/late fields. Generally though a nice refactor.

app/sample_manifest_excel/sample_manifest_excel/upload/processor/base.rb
```diff
@@ -26,13 +26,44 @@ def initialize(upload)
...
+
+          upload.rows.each do |row|
+            unless row.validate_sample
+              upload.errors.add(:base, row.errors.full_messages.join(', ').to_s)
+              all_valid = false
+            end
+          end
+
+          all_valid
+        end
```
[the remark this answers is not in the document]
stevieing: Personally i would put the sample validation in the rows but I can understand why it is here.

app/sample_manifest_excel/sample_manifest_excel/upload/processor/base.rb
```diff
@@ -26,13 +26,57 @@ def initialize(upload)
...
+
+        def create_samples_if_not_present
+          upload.rows.each(&:sample)
+        end
+
+        def samples_valid?
+          all_valid = true
+
+          upload.rows.each do |row|
+            unless row.validate_sample
```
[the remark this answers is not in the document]
stevieing: This still feels a bit weird where you are validating some parts and then processing and then validating other bits. Maybe worth coming back to at a future date?

stevieing (approved):
All good. Nice work to get it all sorted. Still think there is some more stuff that could be refactored further but maybe in the future.

spec/features/sample_manifests/uploader_for_manifests_with_tag_sequences_spec.rb
KatyTaylor: The former. Because of the refactoring in sample_manifest_upload_with_tag_sequences_controller

app/sample_manifest_excel/sample_manifest_excel/upload/processor/base.rb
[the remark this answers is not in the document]
KatyTaylor: Yes agreed. I think the way forward is to make the rest of the specialised field validation not dependent on the sample being inserted - and then the `return unless valid?` on line 27 could deal with all the validation.

app/sample_manifest_excel/sample_manifest_excel/upload/processor/base.rb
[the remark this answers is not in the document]
KatyTaylor: Thanks for the suggestion, James, yes also not sure - think it's more concise but less readable unless you are used to `reduce`.

Steve - agreed, it looks like it should go there - I did have a go but it caused me more troubles due to the errors getting added onto the wrong object, etc, so decided to leave it here.

app/sequencescape_excel/sequencescape_excel/specialised_field/sanger_plate_id.rb
KatyTaylor: Cool thanks

app/sample_manifest_excel/sample_manifest_excel/upload/row.rb
KatyTaylor: This was the commit - https://github.com/sanger/sequencescape/pull/3231/commits/e7ba7cb30bab000720bc93e7d27a8f362c1ec00e - think the message explains it OK.

I think it's only actually necessary for the test, as it gets set as part of the sample validation as you say. Do you think I should rewrite the test so it's not needed, and take it out here?

app/sample_manifest_excel/sample_manifest_excel/upload/row.rb
KatyTaylor: Yep I think that all makes sense. I think in addition to point 2, it's a feeling that future people looking at the code might not realise the importance of the transaction with delaying the registering of the stock resource, so it's better to make that explicit somehow. Agreed though, that it's actually a pain if the stocks get registered if the sample creation was faulty - so I'll have a quick look at moving that `register_stock` call to at the end.

KatyTaylor: I've altered it substantially based on this - still waiting to see if tests pass, and doing some manual testing - but would appreciate another look at the last couple of commits to see if this matches what you were thinking. Thank you!

app/sample_manifest_excel/sample_manifest_excel/upload/row.rb
JamesGlover: Probably, but let's leave it for now. As you said, the commit message explains it pretty well. (Sorry for missing it)

JamesGlover (approved):

PR #3234, opened 2021-06-22
Adjust cop configuration

JamesGlover:
- Disable class length cops in tests. Tests are good.
- Disable aggregate examples cops
  The are a bit ardent to have enabled globally, and while they improve
  test performance, it is at the expense of readability. Plus they get
  in a fight with other rspec cops.

Closes #

*
*
* ...


KatyTaylor:
Allowing longer classes in tests makes sense.
I don't actually know what aggregate examples are - is it the same as this? https://relishapp.com/rspec/rspec-expectations/docs/aggregating-failures

KatyTaylor (approved):

JamesGlover:
> Allowing longer classes in tests makes sense.
> I don't actually know what aggregate examples are - is it the same as this? https://relishapp.com/rspec/rspec-expectations/docs/aggregating-failures

Yep. Although the cop itself argues otherwise (and got renamed to aggregate examples as a result). Essentially, the cop is detecting separate examples, that it reckons you should lump together and use 'aggregate_failures' for performance reasons. But I think the differences boil down to pedantry and semantics.

PR #3240, opened 2021-07-08
Bug - Unhandled error when tube rack scan file missing

### LINKED ISSUE #3239: DPL-nnn [BUG] Unhandled error when tube rack scan file missing
**Describe the bug**
If you try to upload a tube rack manifest, and the scan file is missing, it throws an unhandled exception.
Error:
```
A TypeError occurred in sample_manifest_upload_with_tag_sequences#create:

  no implicit conversion of nil into String

  lib/csv_parser_client.rb:13:in `get_tube_rack_scan_results'
```

**To Reproduce**
1. Go to the page of a study
2. Click on right hand link 'Sample Manifests'
3. Click on right hand link 'Create manifest for tube racks'
4. Select template 'Default Tube Rack'
5. Select any Supplier
6. Click on button 'Create manifest'
7. Click on button 'Download Blank Manifest'
8. Edit and save excel file - add some supplier names and tube barcodes, and the tube rack barcode near the top.
9. Re-upload file to the page, and click 'Upload manifest'
10. See unhandled exception page

**Expected behaviour**
You should get a friendly error telling you there's no scan file available for that tube rack

KatyTaylor:
Tube rack manifest upload was not handling the situation where Wrangler responded with no body

- this happens when there is no csv found for the tube rack scan
- changed to handle this error more gracefully
- found during testing of DPL-024

Closes #3239

### DIFF
--- a/lib/csv_parser_client.rb
+++ b/lib/csv_parser_client.rb
@@ -9,6 +9,14 @@ module CsvParserClient
     endpoint = configatron.tube_rack_scans_microservice_url
     response = Net::HTTP.get_response(URI(endpoint + tube_rack_barcode))
 
+    if response.body.nil?
+      error_message =
+        "Scan could not be retrieved for tube rack with barcode #{tube_rack_barcode}. Service responded with status code #{response.code}"
+      error_message += " and message #{response.message}."
+      object_to_add_errors_to.errors.add(:base, error_message)
+      return nil
+    end
+
     begin
       scan_results = JSON.parse(response.body)
     rescue JSON::JSONError => e


stevieing (approved):
Not sure exactly what is going on but it all seems ok.

lib/csv_parser_client.rb
```diff
@@ -9,6 +9,14 @@ def self.get_tube_rack_scan_results(tube_rack_barcode, object_to_add_errors_to)
     endpoint = configatron.tube_rack_scans_microservice_url
     response = Net::HTTP.get_response(URI(endpoint + tube_rack_barcode))
 
+    if response.body.nil?
+      error_message =
+        "Scan could not be retrieved for tube rack with barcode #{tube_rack_barcode}. Service responded with status code #{response.code}"
+      error_message += " and message #{response.message}."
```
sdjmchattie, line 15: Not a big deal at all, but if three lines are needed anyway (presumably to satisfy the linters) it might look neater if you split that long first string into two parts at the break between the sentences.  Then you won't have a short line with just the variable name and a really long one following it.

sdjmchattie (approved):
Looks good to me -- just one minor suggestion.

PR #3242, opened 2021-07-09
Investigate set descriptions handler

JamesGlover:
Working on documenting some to the pipeline steps, and cleaning up unused code


JamesGlover:
In preparation of the OKR work to improve user self support I have been:

1) Looking into how the set_descriptor_task works, and ensuring this is documented.
2) Stripping out unused code
3) Simplifying a few parts as I go.

I was planning to wait until I stripped out the old Library Creation code, but that kills more code than I anticipated, so I'll do this in a few steps.

.prettierignore
```diff
@@ -6,6 +6,8 @@ tmp/*
 .DS_Store
 public/doc/*
 db/*.sql
+db/schema.rb
+db/views_schema.rb
 db/views_schema.old
```
JamesGlover, line 11: Prettier touching the db/schema is just annoying. It explains the massive schema diff here. Did try and get that in a two part commit, but it didn't work out as intended and somehow ended up as one. Sorry about that.

app/models/pipeline.rb
```diff
@@ -58,8 +58,6 @@ class Pipeline < ApplicationRecord # rubocop:todo Metrics/ClassLength
   delegate :item_limit, :batch_limit?, to: :workflow
 
   belongs_to :control_request_type, class_name: 'RequestType'
-  belongs_to :next_pipeline, class_name: 'Pipeline'
-  belongs_to :previous_pipeline, class_name: 'Pipeline'
```
JamesGlover, line 61: These are pretty ancient, and I don't think have been used the entire time I've been here. My code-archeology suggests that they used to be part of automated qc processes, with woudl advance requests through pipelines.

app/models/tasks/set_descriptors_handler.rb
```diff
@@ -4,10 +4,17 @@ module Tasks::SetDescriptorsHandler # rubocop:todo Style/Documentation
...
     @requests = @batch.ordered_requests
 
-    # if qc_state is qc_manual then update it
+    # If qc_state is qc_manual then update it
+    # @note Only seems to have been used over a 7 month period between
+    #       2009-12-09 14:53:15 and 2010-06-16 12:20:11
+    #       Not sure why its duplicated.
+    #       I believe we originally had additional tasks within SS that would be
+    #       performed after the run had been processed.
     if @batch.qc_state == 'qc_manual'
```
JamesGlover, line 18: Obviously this is a clear candidate for plain deletion.

app/models/tasks/set_descriptors_handler.rb
```diff
@@ -69,17 +100,30 @@ def do_set_descriptors_task(_task, params) # rubocop:todo Metrics/CyclomaticComp
                 end
               end
```
JamesGlover, line 101: Another candidate for deletion

app/models/tasks/set_descriptors_handler.rb
```diff
@@ -4,10 +4,17 @@ module Tasks::SetDescriptorsHandler # rubocop:todo Style/Documentation
   # rubocop:todo Metrics/AbcSize
```
JamesGlover, line 4: Plan is to pretty much re-implement, rather than refactor this. Even ignoring the complexity of the code here, the Workflows controller is a bit fat. So this documentation is to try and help with identifying the key behaviours.

JamesGlover: Not all of which appear to be used or needed.

db/schema.rb
```diff
@@ -10,2437 +10,1932 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_05_04_153704) do
```
JamesGlover, line 13: Prettier got hold of the db/schema file. I'f added it to the ignore, but this one is pretty large as a result. I had tried to get it into two commits, but the first seemingly failed and I didn't notice.

emrojo (approved):

PR #3254, opened 2021-07-15
Rails6/upgrade branch

JamesGlover:
Adds rails 6 dual booting:

- Setting the ENV BUNDLE_GEMFILE to GemfileNext will run in rails 6 mode. 
- Added CI tests for any branches beginning rails6/ or tags with rails6.
- A scattering of backwards compatible fixes

Few test failures, so this should be an easy one.

See: https://github.com/sanger/sequencescape/issues/2360


app/models/aliquot.rb
```diff
@@ -87,7 +87,7 @@ class InsertSize < Range
           ).order('tag1s.map_id ASC, tag2s.map_id ASC')
         }
   scope :untagged, -> { where(tag_id: UNASSIGNED_TAG, tag2_id: UNASSIGNED_TAG) }
-  scope :any_tags, -> { where.not(tag_id: UNASSIGNED_TAG, tag2_id: UNASSIGNED_TAG) }
+  scope :any_tags, -> { where.not(tag_id: UNASSIGNED_TAG).or(where.not(tag2_id: UNASSIGNED_TAG)) }
```
JamesGlover, line 91: This one is actually a bug-fix.
It triggered a deprecation warning, but it turns out that the future behaviour, is the one we actually want. The previous code was finding aliquots where neither tag_id or tag2_id were -1. The new code finds any aliquots with at least one tag.

app/models/labware.rb
```diff
@@ -70,11 +70,29 @@ def receptacle_proxy
...
+              # If we use joins(:ancestor) here, Rails 6 seems to try to find an ancestor
+              # relationship on Labware.
+              joins('INNER JOIN labware AS sbl ON sbl.id = asset_links.ancestor_id')
+                .where(sbl: { sti_type: 'SpikedBuffer' })
+                .direct
+            },
+            class_name: 'AssetLink',
+            foreign_key: :descendant_id,
+            inverse_of: :descendant
+  end
```
JamesGlover, line 95: This one is a little unfortunate. As far as I'm concerned, both versions should work with either version of rails. However, each falls foul of strange version specific behaviour.

app/models/labware.rb
```diff
@@ -224,7 +242,7 @@ def find_from_any_barcode(source_barcode)
         nil
       elsif /\A[0-9]{1,7}\z/.match?(source_barcode)
         # Just a number
-        joins(:barcodes).where('barcodes.barcode LIKE "__?_"', source_barcode).first # rubocop:disable Rails/FindBy
+        joins(:barcodes).order(:id).find_by('barcodes.barcode LIKE "__?_"', source_barcode)
```
JamesGlover, line 245: This was mostly to make rubocop happy. I added order by to ensure actual behaviour isn't changed. (Although in practice duplicate barcodes are exceedingly rare)

app/models/phi_x.rb
```diff
@@ -55,7 +55,7 @@ def self.tag_options
   # Returns the names of valid tag options for creation of PhiX libraries.
   # @return [Array] Valid tag options
   def self.tag_option_names
-    tag_options.keys
+    tag_options.keys.map(&:to_s)
   end
```
JamesGlover, line 59: A lot of the changes here are due to the way rails 6 handles config loading. Fortunately this is all backwards compatible.

app/models/well.rb
```diff
@@ -96,7 +96,7 @@ def attach(wells)
           )
         }
 
-  scope :without_report, ->(product_criteria) { where.not(id: Well.with_report(product_criteria)) }
+  scope :without_report, ->(product_criteria) { where.not(id: with_report(product_criteria)) }
```
JamesGlover, line 100: A bit like the change above, this is due to changes in the way in which rails handles scopes. https://github.com/rails/rails/issues/14003

app/models/labware.rb
JamesGlover: Ahh, found a universal approach that's also a lot nice than my previous attempt.

.talismanrc
```diff
@@ -0,0 +1,19 @@
+fileignoreconfig:
```
stevieing, line 1: Should this be git ignored? I know there have been issues in other repos with conflicts as the checksum will be different for different users.

stevieing (approved):

.talismanrc
JamesGlover: Agh, yes, you are right... not sure how that one slipped in there

PR #3262, opened 2021-07-23
Ensure flowcell broadcast by batch touch

### LINKED ISSUE #3248: DPL-053 [BUG] tag substitutions made by manifest or via substitution page are not being updated in MLWH
As title "tag substitutions made by manifest or via substitution page are not being updated in MLWH"

Requires manual intervention via rails console

JamesGlover:
Fixes  #3248

- Add test for flowcell message creation
- After commit callback on batch triggers immediate broadcast
- Move batch callback inside transaction

### DIFF
--- a/spec/models/tag_substitutions_spec.rb
+++ b/spec/models/tag_substitutions_spec.rb
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 
-describe TagSubstitution do
+describe TagSubstitution, warren: true do
   # Works for: Library manifests, old tube pipelines
   # We have two samples, each with unique tags, which only exist
   # in aliquots identified by the library id. We don't need to consider:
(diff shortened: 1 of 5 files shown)


spec/models/tag_substitutions_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 
-describe TagSubstitution do
+describe TagSubstitution, warren: true do
```
KatyTaylor, line 5: what does the `warren: true` do, is that new since your recent work on warren?

spec/models/tag_substitutions_spec.rb
```diff
@@ -23,27 +23,35 @@
...
 
-    it 'assigns comments to the lanes and tubes' do
-      expect(library_tube_a.receptacle.comments.map(&:description)).to eq [comment]
-      expect(library_tube_b.receptacle.comments.map(&:description)).to eq [comment]
-      expect(mx_library_tube.receptacle.comments.map(&:description)).to eq [comment]
-      expect(lane.comments.map(&:description)).to eq [comment]
+    it 'rebroadcasts the flowcell' do
+      expect { subject.save }.to change {
+        Warren.handler.messages_matching("queue_broadcast.messenger.#{flowcell_message.id}")
+      }.by(1)
```
KatyTaylor, line 54: Nice, wasn't aware of there being testing helpers internal to Warren

spec/models/tag_substitutions_spec.rb
JamesGlover: Nope, its actually been around for a little while. It enables logging of messages for testing purposes:
https://rubydoc.info/gems/sanger_warren/Warren/Handler/Test

KatyTaylor (approved):

spec/models/tag_substitutions_spec.rb
JamesGlover: I'll make an issue to fix the markup on that documentation... its not supposed to insert whole iframes for the 'useful' messages section, and they'd be better off as simple links.

PR #3265, opened 2021-07-23
Improve performance of cherrpick specs

JamesGlover:
page.all('td', text: /.+v13.0 b0.0/) waits for 10 seconds
if no cells match the regexp, and will then return an empty array.

However:
1) We have some tests for different volumes,
so we increase the permissiveness of the regexp
2) We know the page is already rendered, so don't wait.
3) If we *don't* find anything, that's probably a bug. Lets fail.

This also moves the barcode array generation out of the loop.
Its a minor saving, but one I made while tracking down the
bottle-neck.

These tests are still pretty slow, but this change knoxk about 10 seconds
off each pass through, saving a minute or so over all.

Closes #

*
*
* ...

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -161,3 +161,7 @@ Metrics/ClassLength:
   Exclude:
     - "spec/**/*"
     - "test/**/*"
+
+RSpec/MultipleExpectations:
+  Exclude:
+    - "spec/features/**/*"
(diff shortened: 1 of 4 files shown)


harrietc52 (approved):
Thanks for the comment on the PR too explaining the gains, made the context clear :)

PR #3266, opened 2021-07-23
Update sequencing inbox eager loading

JamesGlover:

* Eager load required records
* Remove non-required records


harrietc52 (approved):
`:scanned_into_lab_event` has been nested within labware, assuming that's okay?

JamesGlover:
> `:scanned_into_lab_event` has been nested within labware, assuming that's okay?

Yep, its the association that's actually used. The one on receptacle should probably be removed, its a leftover from when receptacle and labware were the same thing (Asset).

PR #3271, opened 2021-07-29
DPL-060: Fix TubeRack view

### LINKED ISSUE #3252: DPL-060 [BUG] Tube Rack 'show' page fails to load
** Time box this to half a day and put back in triage if it turns out to be harder than expected to fix **

**Describe the bug**
The Tube Rack 'show' page fails to load. For instance, try going to http://training.sequencescape.psd.sanger.ac.uk/labware/27048269

Error is:
```
A NoMethodError occurred in labware#show:

  undefined method `stock_plate' for #<TubeRack:0x000055c34ea28a70>
  app/models/purpose.rb:72:in `block in source_plates'
```

sdjmchattie:
Closes #3252 

* Do more checks on the labware object before calling `stock_plate`.  TubeRack doesn't implement this method.

### DIFF
--- a/app/models/purpose.rb
+++ b/app/models/purpose.rb
@@ -63,8 +63,6 @@ class Purpose < ApplicationRecord
         labware.ancestors_of_purpose(source_purpose_id)
       elsif labware.respond_to?(:stock_plate)
         labware.stock_plate
-      else
-        nil
       end
     end
   end
--- a/app/models/purpose.rb
+++ b/app/models/purpose.rb
@@ -59,11 +59,7 @@ class Purpose < ApplicationRecord
     ActiveSupport::Deprecation.silence do
       # Rails 6 lets us do this:
       # ActiveSupport::Deprecation.allow(:stock_plate) do
-      if source_purpose_id.present?
-        labware.ancestors_of_purpose(source_purpose_id)
-      elsif labware.respond_to?(:stock_plate)
-        labware.stock_plate
-      end
+      source_purpose_id.present? ? labware.ancestors_of_purpose(source_purpose_id) : labware.try(:stock_plate)
     end
   end
 
@@ -73,13 +69,7 @@ class Purpose < ApplicationRecord
     ActiveSupport::Deprecation.silence do
       # Rails 6 lets us do this:
       # ActiveSupport::Deprecation.allow(:stock_plate) do
-      if source_purpose_id.present?
-        labware.ancestors_of_purpose(source_purpose_id)
-      elsif labware.respond_to?(:stock_plate)
-        [labware.stock_plate].compact
-      else
-        []
-      end
+      source_purpose_id.present? ? labware.ancestors_of_purpose(source_purpose_id) : [labware.try(:stock_plate)].compact
     end
   end
 
--- a/app/models/purpose.rb
+++ b/app/models/purpose.rb
@@ -59,7 +59,7 @@ class Purpose < ApplicationRecord
     ActiveSupport::Deprecation.silence do
       # Rails 6 lets us do this:
       # ActiveSupport::Deprecation.allow(:stock_plate) do
-      source_purpose_id.present? ? labware.ancestors_of_purpose(source_purpose_id) : labware.try(:stock_plate)
+      source_purpose_id.present? ? labware.ancestor_of_purpose(source_purpose_id) : labware.try(:stock_plate)
     end
   end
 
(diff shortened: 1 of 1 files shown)


app/models/purpose.rb
```diff
@@ -59,7 +59,11 @@ def source_plate(labware)
     ActiveSupport::Deprecation.silence do
       # Rails 6 lets us do this:
       # ActiveSupport::Deprecation.allow(:stock_plate) do
-      source_purpose_id.present? ? labware.ancestor_of_purpose(source_purpose_id) : labware.stock_plate
+      if source_purpose_id.present?
+        labware.ancestors_of_purpose(source_purpose_id)
+      elsif labware.respond_to?(:stock_plate)
+        labware.stock_plate
```
JamesGlover, line 65: This can simply become labware.try(:stock_plate) (Similarly below)

```
[1] pry(main)> TubeRack.new.try(:stock_plate)
   (1.7ms)  SET NAMES utf8,  @@SESSION.sql_mode = 'TRADITIONAL',  @@SESSION.group_concat_max_len = 67108864, @@SESSION.sql_auto_is_null = 0, @@SESSION.wait_timeout = 2147483
=> nil
```

Although I'm wondering if we should be defining stock_plate on labware (Although the fact we've deprecated it makes me wary)

app/models/purpose.rb
```diff
@@ -69,7 +73,13 @@ def source_plates(labware)
     ActiveSupport::Deprecation.silence do
       # Rails 6 lets us do this:
       # ActiveSupport::Deprecation.allow(:stock_plate) do
-      source_purpose_id.present? ? labware.ancestors_of_purpose(source_purpose_id) : [labware.stock_plate].compact
+      if source_purpose_id.present?
+        labware.ancestors_of_purpose(source_purpose_id)
+      elsif labware.respond_to?(:stock_plate)
+        [labware.stock_plate].compact
```
JamesGlover, line 79: See above regarding the use of try()

app/models/purpose.rb
JamesGlover: I was also going to suggest that we might be able to work out the source plates, but from glancing at the way the features actually used, it wouldn't be much use.

I think we can actually avoid needing to use this method at all here, and instead migrate to ancestors_of_purpose. However I need to work out some of the details.

Background: This is mostly used to generate links to plate summaries, which provide an overview of all actions that happened to a plate as it was processed in a pipeline. As a result, each plate or tube tries to link back to the plate at the top of the pipeline, using the source purpose. However this is a bit hard coded, and doesn't really take into account all the variation that can happen now. Instead it would be better to grab a list of any ancestor that could theoretically be of interest, based on a list of all suitable plates.

`labware.ancestor_of_purpose(PlatePurpose.considered_stock_plate)` would get us most the way there, but there might be a few edge cases I'm not thinking of at the moment.

JamesGlover (changes requested):
Switch to using .try rather than the null check, as its essentially an active support method for exactly that behaviour.

The bigger comment regarding the switch to `ancestor_of_purpose` was mostly musings by me (Still worth making the update to the source_plates(s) methods regardless)

app/models/purpose.rb
sdjmchattie: This sounds like a lot of change to resolve an issue that hasn't been reported yet.  I think given that:
a) `:stock_plate` has been deprecated
b) The TubeRack is able to load now with just the use of `:try`
that we should go for the simplest solution for now.  If it's desirable to see the source for TubeRacks and this doesn't cover that use case, then that should be a different ticket (i.e. not this bug report ticket)

sdjmchattie:
I don't know enough about Cucumber or how to read its logs to know what I should be doing to fix the failed tests.  Would appreciate some pointers.

JamesGlover (approved):

JamesGlover:
> I don't know enough about Cucumber or how to read its logs to know what I should be doing to fix the failed tests. Would appreciate some pointers.

Looks like you managed to track it down.

Unfortunately when we switched to knapsack_pro for running cukes, the failures became a bit harder to find, although searching for 'Failing' allows you to track down the failing scenarios. Not quite sure what happened to the actual failure messages though, looks like something is eating them. I wonder if my attempt at exporting screenshots/html is inadvertently over-riding the usual failure message. Didn't notice this when I added that behaviour, but it seems to most likely culprit.

Cucumber was never the easiest thing to track down failures from though.

PR #3280, opened 2021-08-09
Dpl 033 add phix during library prep

KatyTaylor:

* Previously, PhiX was included in Lanes at sequencing stage by making the SpikedBuffer tube (which contains the PhiX) a direct parent of the Lane Labware.
* Now, the SpikedBuffer tube can also be an indirect ancestor of the Lane Labware
  * When looking for the relevant SpikedBuffer tube, direct parents will be considered first. If there is no direct parent, the most recently created ancestor will be chosen.
    * This is because if we just considered the most recently created ancestor, it would find different tubes for some historical lanes (see this comment on story - https://github.com/sanger/limber/issues/740#issuecomment-893468497)
* This modification is necessary so that the PhiX can be added at an earlier stage for the Heron pipeline (as a parent of the multiplexed library tube).
* I had to change `spiked_in_buffer` on Labware from a `has_one` association to a method, to accommodate the need to check direct parent first, then ancestors.
* The two interfaces affected here are 
  * a) the message that goes to MLWH iseq_flowcell table (flowcell_io class), and 
  * b) the batch xml interface, which is used by NPG's front end and perl API (show.xml.builder)

I'm merging into an epic branch as the changes needed to the sequencing pipeline interface and Limber library prep pipeline are quite separate and will be done in separate pull requests.

### DIFF
--- a/app/models/labware.rb
+++ b/app/models/labware.rb
@@ -71,7 +71,8 @@ class Labware < Asset # rubocop:todo Metrics/ClassLength
 
   belongs_to :purpose, foreign_key: :plate_purpose_id, optional: true, inverse_of: :labware
 
-  # Gets the SpikedBuffer tube that is a direct parent of this labware, if it exists
+  # Gets the SpikedBuffer tube that is a direct parent of this labware, if it exists.
+  # The original implementation of spiked_in_buffer just supported direct parent tubes.
   has_one :spiked_in_buffer_links,
           # If we try to use the Rails 6 version, Rails 5 doesn't seem to perform the join
           -> { includes(:ancestor).references(:ancestor).where(labware: { sti_type: 'SpikedBuffer' }).direct },
@@ -79,8 +80,8 @@ class Labware < Asset # rubocop:todo Metrics/ClassLength
           foreign_key: :descendant_id,
           inverse_of: :descendant
 
-  # Gets the most recent SpikedBuffer tube ancestor, if it exists, to use if `spiked_in_buffer_links` returns nil
-  # For when PhiX gets added during library prep rather than at sequencing time.
+  # Gets the most recent SpikedBuffer tube ancestor, if it exists, to use if there is no direct parent SpikedBuffer tube.
+  # Added to support PhiX being added during library prep rather than at sequencing time (for Heron).
   has_one :spiked_in_buffer_most_recent_links,
           # If we try to use the Rails 6 version, Rails 5 doesn't seem to perform the join
           -> {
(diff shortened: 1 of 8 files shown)


app/models/labware.rb
```diff
@@ -71,14 +71,29 @@ def receptacle_proxy
...
+  # Gets the SpikedBuffer tube that is a direct parent of this labware, if it exists.
+  # The original implementation of spiked_in_buffer just supported direct parent tubes.
   has_one :spiked_in_buffer_links,
           # If we try to use the Rails 6 version, Rails 5 doesn't seem to perform the join
           -> { includes(:ancestor).references(:ancestor).where(labware: { sti_type: 'SpikedBuffer' }).direct },
           class_name: 'AssetLink',
           foreign_key: :descendant_id,
           inverse_of: :descendant
 
-  has_one :spiked_in_buffer, through: :spiked_in_buffer_links, source: :ancestor
```
JamesGlover, line 81: Rather than removing this relationship, I'd follow a similar parrent to the links:

```ruby
has_one :direct_spiked_in_buffer, through: :spiked_in_buffer_links, source: :ancestor
has_one :most_recent_spiked_in_buffer, through: :spiked_in_buffer_most_recent_links, source: :ancestor
```
As then the `spiked_in_buffer` method below can be simpler (See below)

app/models/labware.rb
```diff
@@ -153,6 +168,15 @@ def ancestors_of_purpose(ancestor_purpose_id)
     ancestors.order(id: :desc).where(plate_purpose_id: ancestor_purpose_id)
   end
 
+  # Gets the relevant SpikedBuffer tube, if one exists, by using the two associations.
+  # A direct parent SpikedBuffer tube is used if it exists, otherwise the most recent ancestor.
+  # This was necessary to avoid affecting historical data, for which the direct parent should be used,
+  # even though there is another ancestor that was created more recently.
+  def spiked_in_buffer
```
JamesGlover, line 175: If you still define an association with the buffer itsef this can become simply
```ruby
direct_spiked_in_buffer || most_recent_spiked_in_buffer
```

app/views/batches/show.xml.builder
```diff
@@ -12,7 +12,14 @@ xml.batch do
           request: [
             {
               target_asset: {
-                spiked_in_buffer: [:index, { aliquots: %i[library tag tag2 aliquot_index sample] }],
+                labware: {
+                  spiked_in_buffer_links: {
+                    ancestor: [:index, { aliquots: %i[library tag tag2 aliquot_index sample] }]
```
JamesGlover, line 17: If you adjust the associations as suggested above, then don't forget to update this.

JamesGlover (changes requested):

app/models/labware.rb
KatyTaylor: Yes OK, that's probably nicer, thanks

PR #3284, opened 2021-08-12
Issue3228/dpl 043 manifest upload causing duplications in ega

### LINKED ISSUE #3228: DPL-043 Manifest upload causing duplications in EGA
When manifests are uploaded duplicate entries are being made in the EGA database (nearly always)

Example manifest (non sensitive data generated for testing)
[2405stdy_manifest_17293_170621.xlsx](https://github.com/sanger/sequencescape/files/6669512/2405stdy_manifest_17293_170621.xlsx)
Duplicate entries (scraped from the EGA Webin page (https://www.ebi.ac.uk/ena/submit/sra/#home)
[repeat_EGA_entries.xlsx](https://github.com/sanger/sequencescape/files/6669514/repeat_EGA_entries.xlsx)

When 2405STDY10409901 is viewed in sequencescape_warehouse.samples ('old' warehouse) there are two accession entries EGAN00003219314 and then EGAN00003219315, both with the same timestamps.

James G's comments (from Slack 17/06/21)
"Suspect we’re creating and then updating samples, and generating an accession job for each action. I’m not quite sure why both accessions are succeeding and returning different number though, I’m assuming the EBI have a similar race condition

JamesGlover:
Closes #3228 

- Prevents saves in the middle of manifest processing
- Deliberately disable accessioning while processing to avoid risking similar issues in future

### DIFF
--- a/app/controllers/admin/studies_controller.rb
+++ b/app/controllers/admin/studies_controller.rb
@@ -78,7 +78,8 @@ class Admin::StudiesController < ApplicationController # rubocop:todo Style/Docu
       redirect_to controller: 'admin/studies', action: 'update', id: @study.id
     end
   rescue ActiveRecord::RecordInvalid => e
-    logger.warn "Failed to update attributes: #{@study.errors.map(&:to_s)}}"
+    errors = @study.errors.full_messages
+    logger.warn "Failed to update attributes: #{errors}}"
     flash[:error] = 'Failed to update attributes for study!'
     render action: :show, id: @study.id and return
   end
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -20,6 +20,11 @@ require 'rexml/text'
 # - Heron: Heron samples get registered via the Api::V2::Heron::PlatesController
 # - Special samples: Samples such as {PhiX} are generated internally
 class Sample < ApplicationRecord # rubocop:todo Metrics/ClassLength
+  # See https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html
+  class Current < ActiveSupport::CurrentAttributes
+    attribute :processing_manifest
+  end
+
   GC_CONTENTS = ['Neutral', 'High AT', 'High GC'].freeze
   GENDERS = ['Male', 'Female', 'Mixed', 'Hermaphrodite', 'Unknown', 'Not Applicable'].freeze
   DNA_SOURCES = [
@@ -354,7 +359,11 @@ class Sample < ApplicationRecord # rubocop:todo Metrics/ClassLength
   validates_associated(:ena_study, allow_blank: true, on: :accession)
 
   before_destroy :safe_to_destroy
-  after_save :accession
+
+  # processing_manifest is true if we're currently processing a manifest. We
+  # disable accessioning, as we'll perform it explicitly later. This avoids
+  # accidental calls to save triggering duplicate accessions
+  after_save :accession, unless: -> { Sample::Current.processing_manifest }
 
   # NOTE: Samples don't tend to get released through Sequencescape
   # so in reality these methods are usually misleading.
--- a/app/models/sample_manifest/uploader.rb
+++ b/app/models/sample_manifest/uploader.rb
@@ -27,17 +27,11 @@ class SampleManifest::Uploader
       SampleManifestExcel::Upload::Base.new(file: file, column_list: self.configuration.columns.all, override: override)
   end
 
-  # rubocop:disable Metrics/MethodLength
   def run!
     ActiveRecord::Base.transaction do
       return false unless valid?
 
-      if upload.process(tag_group)
-        upload.finished!
-        upload.broadcast_sample_manifest_updated_event(user)
-        upload.register_stock_resources
-        return true
-      end
+      return true if process_upload_and_callbacks
 
       # One of our post processing checks failed, something went wrong, so we
       # roll everything back
@@ -49,9 +43,17 @@ class SampleManifest::Uploader
     false
   end
 
-  # rubocop:enable Metrics/MethodLength
-
   private
+
+  def process_upload_and_callbacks
+    return false unless upload.process(tag_group)
+
+    upload.finished!
+    upload.broadcast_sample_manifest_updated_event(user)
+    upload.register_stock_resources
+    upload.trigger_accessioning
+    true
+  end
 
   def create_tag_group
     TagGroup.find_or_create_by!(name: configuration.tag_group) if configuration.tag_group.present?
--- a/app/sample_manifest_excel/sample_manifest_excel/upload/base.rb
+++ b/app/sample_manifest_excel/sample_manifest_excel/upload/base.rb
@@ -67,12 +67,18 @@ module SampleManifestExcel
       # An upload can only be processed if the upload is valid.
       # Processing involves updating the sample manifest and all of its associated samples.
       def process(tag_group)
+        # Temporarily disable accessioning until we invoke it explicitly
+        # If we don't do this, then any accidental triggering of sample
+        # saves will result in duplicate accessions
+        Sample::Current.processing_manifest = true
         sample_manifest.last_errors = nil
         sample_manifest.start!
         @cache.populate!
         processor.run(tag_group)
 
         processed?
+      ensure
+        Sample::Current.processing_manifest = false
       end
 
       def data_at(column_name)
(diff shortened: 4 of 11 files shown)


app/controllers/admin/studies_controller.rb
```diff
@@ -78,7 +78,8 @@ def managed_update # rubocop:todo Metrics/AbcSize
       redirect_to controller: 'admin/studies', action: 'update', id: @study.id
     end
   rescue ActiveRecord::RecordInvalid => e
-    logger.warn "Failed to update attributes: #{@study.errors.map(&:to_s)}}"
+    errors = @study.errors.full_messages
+    logger.warn "Failed to update attributes: #{errors}}"
```
JamesGlover, line 82: Not been able to get to the bottom of this problem, but the former way throws:

>   wrong number of arguments (given 1, expected 0)
> 
>   app/controllers/admin/studies_controller.rb:81:in `map'

Seen it in other similar contexts

app/models/sample.rb
```diff
@@ -20,6 +20,11 @@
 # - Heron: Heron samples get registered via the Api::V2::Heron::PlatesController
 # - Special samples: Samples such as {PhiX} are generated internally
 class Sample < ApplicationRecord # rubocop:todo Metrics/ClassLength
+  # See https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html
+  class Current < ActiveSupport::CurrentAttributes
+    attribute :processing_manifest
+  end
+
```
JamesGlover, line 27: I considered setting a basic attribute on sample to handle this, but that would mean doing so for new samples, as well as found ones. Plus, if we ever reloaded the samples, the attribute would get blitzed.

app/models/sample_manifest/uploader.rb
```diff
@@ -27,17 +27,11 @@ def initialize(file, configuration, user, override)
...
     ActiveRecord::Base.transaction do
       return false unless valid?
 
-      if upload.process(tag_group)
-        upload.finished!
-        upload.broadcast_sample_manifest_updated_event(user)
-        upload.register_stock_resources
-        return true
-      end
+      return true if process_upload_and_callbacks
```
JamesGlover, line 34: Refactoring here mostly due to Rubocop.

app/sample_manifest_excel/sample_manifest_excel/upload/base.rb
```diff
@@ -82,14 +88,18 @@ def data_at(column_name)
 
       def broadcast_sample_manifest_updated_event(user)
         # Send to event warehouse
-        sample_manifest.updated_broadcast_event(user, samples_to_be_broadcasted.map(&:id))
+        sample_manifest.updated_broadcast_event(user, changed_samples.map(&:id))
```
JamesGlover, line 91: Just renamed these methods as I'm now using them in a different context.

app/sample_manifest_excel/sample_manifest_excel/worksheet/test_worksheet.rb
```diff
@@ -13,7 +13,6 @@ class TestWorksheet < SequencescapeExcel::Worksheet::Base # rubocop:todo Metrics
```
JamesGlover, line 13: The rules for accessioning are painfully complicated, and depend on the study. Updated this to let my inject my own studies.

app/sequencescape_excel/sequencescape_excel/specialised_field/control_type.rb
```diff
@@ -19,7 +19,6 @@ def update(_attributes = {})
           sample.control = false
           sample.control_type = nil
         end
-        sample.save
```
JamesGlover, line 22: And this was the source of all the pain.

app/sequencescape_excel/sequencescape_excel/specialised_field/priority.rb
```diff
@@ -13,7 +13,6 @@ def update(_attributes = {})
         return unless valid?
 
         sample.priority = value if value.present?
-        sample.save
```
JamesGlover, line 16: Or possibly this one. I forget which the tests actually detected. But the other got missed as it depends on which templates are being used. Hence the more robust checks to ensure we don't accidentally re-introduce the bug.

spec/models/sample_manifest_spec.rb
```diff
@@ -35,7 +35,7 @@
 
       teardown { Delayed::Worker.delay_jobs = true }
 
-      # rubocop:todo Metrics/BlockLength
+      # rubocop:disable Metrics/BlockLength
```
JamesGlover, line 38: Switched these to disables, as I don't think block length is a problem here.

spec/spec_helper.rb
```diff
@@ -151,6 +151,17 @@ def enable_chrome_headless_downloads(driver)
     Warren.handler.disable!
   end
 
+  config.around(:each, accessioning_enabled: true) do |ex|
```
JamesGlover, line 154: There are a few messy bits to set up, so I've pulled them into an rspec configuration.

KatyTaylor: Could you put a short comment, something like:
```
# For tests with the accessioning_enabled flag, do:
# setup: set the accession folder to a test folder and turn on accessioning
# teardown: turn accessioning off again if it was off previously.
```

...if that is accurate?
What is the accession folder for exactly and why don't you reset that in the teardown?

spec/sample_manifest_excel/upload/upload_spec.rb
```diff
@@ -332,21 +330,26 @@
...
       end
 
       it 'updates all of the data' do
-        upload = SampleManifestExcel::Upload::Base.new(file: test_file, column_list: plate_columns, start_row: 9)
         upload.process(nil)
         expect(upload).to be_processed
       end
+
+      context 'when accessioning is enabled', accessioning_enabled: true do
+        it 'surpresses accessioning' do
```
KatyTaylor, line 349:
```suggestion
        it 'suppresses accessioning' do
```
:)

spec/sample_manifest_excel/upload/upload_spec.rb
```diff
@@ -332,21 +330,26 @@
...
+          expect { upload.process(nil) }.not_to change(Delayed::Job, :count)
```
KatyTaylor, line 350: Maybe it would be worth a short comment here saying something like:
```
# accessioning is triggered explicitly after upload
```
if that is accurate?

spec/models/sample_manifest/uploader_spec.rb
```diff
@@ -157,12 +156,24 @@
...
+      download =
+        build(
+          :test_download_plates,
+          manifest_type: 'plate_full',
+          columns: SampleManifestExcel.configuration.columns.plate_full.dup,
+          study: create(:open_study, accession_number: 'acc')
+        )
+      download.save(test_file_name)
+      uploader = described_class.new(test_file, SampleManifestExcel.configuration, user, false)
+      expect { uploader.run! }.to change(Delayed::Job, :count).by(4)
```
KatyTaylor, line 174: 4 is the number of samples in the test file? If so maybe could make that explicit by grabbing it dynamically somehow, or assigning it to a variable with a name, or just a short comment?

app/sequencescape_excel/sequencescape_excel/specialised_field/control_type.rb
KatyTaylor: So the sample was getting saved twice as part of the import process?

KatyTaylor: Sorry read the commit message now :)

app/controllers/admin/studies_controller.rb
KatyTaylor: I've seen times when the .map(&:to_s) syntax doesn't work, but the .map{ |thing| thing.to_s } does, maybe it's that?

app/sequencescape_excel/sequencescape_excel/specialised_field/control_type.rb
KatyTaylor: So looks like they all get saved in `update_sample` in `Row.rb`

KatyTaylor (approved):
Looks fine as far as I can tell!

app/controllers/admin/studies_controller.rb
JamesGlover: Yep, its the same bug. In this case opted to pull it out as I figured it was less likely to get reverted than an over-eager Rubocop. Perhaps I should stick a comment above to stop an over-eager developer from reverting it.

JamesGlover: Although I'm just noticing that I also switched to full messages, so could probably inline it again anyway.

app/sequencescape_excel/sequencescape_excel/specialised_field/control_type.rb
JamesGlover: Yep! Which should also improve speeds a little, as we'll have a single write per sample.

spec/models/sample_manifest/uploader_spec.rb
JamesGlover: Good suggestion, had done that elsewhere but then moved the test and seemingly lost that.

spec/spec_helper.rb
JamesGlover: It sets the configuration it loads the config from. I did actually try to reset it, but it seemed to blow away a bit more state than expected, and broke a load of other tests.

PR #3287, opened 2021-08-13
Improve batch summary page for multiple events

JamesGlover:
- Use tables for more compact display
- Add timestamps to event information
- Ensure events are sorted in chronological order

### DIFF
--- a/app/models/api/messages/flowcell_io.rb
+++ b/app/models/api/messages/flowcell_io.rb
@@ -96,8 +96,9 @@ class Api::Messages::FlowcellIO < Api::Base
         end
 
         def detect_descriptor(name)
-          lab_events.each { |e| e.descriptor_value_for(name).tap { |bc| return bc if bc.present? } }
-          nil # We have no flowcell barcode
+          # Sort here goes by id. Which should
+          lab_events.sort.reverse_each { |e| e.descriptor_value_for(name).tap { |bc| return bc if bc.present? } }
+          nil # We have no matching descriptor
         end
       end
     end
--- a/app/models/api/messages/flowcell_io.rb
+++ b/app/models/api/messages/flowcell_io.rb
@@ -96,7 +96,7 @@ class Api::Messages::FlowcellIO < Api::Base
         end
 
         def detect_descriptor(name)
-          # Sort here goes by id. Which should cover our needs
+          # Sort here goes by id ascending, so we use a reverse each, in order to find the most recent descriptor with the passed in 'name'
           lab_events.sort.reverse_each { |e| e.descriptor_value_for(name).tap { |bc| return bc if bc.present? } }
           nil # We have no matching descriptor
         end
--- a/app/models/api/messages/flowcell_io.rb
+++ b/app/models/api/messages/flowcell_io.rb
@@ -96,7 +96,7 @@ class Api::Messages::FlowcellIO < Api::Base
         end
 
         def detect_descriptor(name)
-          # Sort here goes by id. Which should
+          # Sort here goes by id. Which should cover our needs
           lab_events.sort.reverse_each { |e| e.descriptor_value_for(name).tap { |bc| return bc if bc.present? } }
           nil # We have no matching descriptor
         end
--- a/app/models/lab_event.rb
+++ b/app/models/lab_event.rb
@@ -1,6 +1,12 @@
 require_dependency 'broadcast_event/lab_event'
 
-class LabEvent < ApplicationRecord # rubocop:todo Style/Documentation
+# Lab events are created as part of the traditional Sequencescape based pipelines
+# the track the information supplied in each step, mostly in the form of descriptors:
+# key value pairs.
+# They can be associated with individual requests, or the batch as a whole.
+# The information is mainly displayed on the batch summary screen, but also acts
+# as a source of information for the FlowcellIO message
+class LabEvent < ApplicationRecord
   include ActsAsDescriptable
 
   CHIP_BARCODE_STEPS = ['Cluster generation', 'Add flowcell chip barcode', 'Loading'].freeze
--- a/app/models/lab_event.rb
+++ b/app/models/lab_event.rb
@@ -30,8 +30,7 @@ class LabEvent < ApplicationRecord # rubocop:todo Style/Documentation
   end
 
   def descriptor_value_for(name)
-    descriptors.each { |desc| return desc.value if desc.name.casecmp(name.to_s).zero? }
-    nil
+    descriptors.detect { |desc| desc.name.casecmp?(name.to_s) }&.value
   end
 
   def add_new_descriptor(name, value)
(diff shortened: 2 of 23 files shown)


app/models/api/messages/flowcell_io.rb
```diff
@@ -96,8 +96,9 @@ def detect_float_descriptor(name, ignored_unit)
         end
 
         def detect_descriptor(name)
-          lab_events.each { |e| e.descriptor_value_for(name).tap { |bc| return bc if bc.present? } }
-          nil # We have no flowcell barcode
+          # Sort here goes by id. Which should cover our needs
```
KatyTaylor, line 99:
```suggestion
          # Sort here goes by id descending, in order to find the most recent descriptor with the passed in 'name'
```

What does variable `bc` stand for?

app/models/lab_event.rb
```diff
@@ -1,6 +1,12 @@
 require_dependency 'broadcast_event/lab_event'
 
-class LabEvent < ApplicationRecord # rubocop:todo Style/Documentation
+# Lab events are created as part of the traditional Sequencescape based pipelines
+# the track the information supplied in each step, mostly in the form of descriptors:
```
KatyTaylor, line 4:
```suggestion
# they track the information supplied in each step, mostly in the form of descriptors:
```

app/models/lab_event.rb
```diff
@@ -1,6 +1,12 @@
...
+# key value pairs.
+# They can be associated with individual requests, or the batch as a whole.
+# The information is mainly displayed on the batch summary screen, but also acts
+# as a source of information for the FlowcellIO message
```
KatyTaylor, line 8: Nice docs thank you

db/migrate/20210817144508_drop_unused_columns.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
```
KatyTaylor, line 1: So we have to make sure this happens after the above migration, and doesn't run if the above one fails? (To make sure descriptor_fields are all migrated out before dropping the column)

KatyTaylor: I saw why descriptor_fields is going, because the info is stored in the 'descriptors' field, but why are the other 3 fields being axed?

db/migrate/20210813103531_migrate_descriptor_fields_into_descriptors.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+# Descriptor fields is redundant now hashes are ordered. This migration
+# updates historical data to ensure:
+# - Hashes are sorted in the same order defined in order-fields
+# - Any keys present in the array, but absent from the hash are recorded
+# - Data missing a + in the hash key is repaired
+class MigrateDescriptorFieldsIntoDescriptors < ActiveRecord::Migration[5.2]
```
KatyTaylor, line 8: Think this will need careful testing in training :)

lib/acts_as_descriptable.rb
```diff
@@ -46,20 +43,13 @@ def descriptor_value(key)
...
-  private
-
-  def each_descriptor
-    descriptor_fields.each do |field|
-      next if field.blank?
-
-      yield(field, descriptor_hash[field])
-    end
+  def descriptor_fields=(fields)
+    # Ignore for now while we remove this
```
KatyTaylor, line 53: Maybe a slightly fuller explanation? Could see this persisting into the future due to confusion...

app/models/tasks/set_descriptors_handler.rb
```diff
@@ -72,7 +72,7 @@ def do_set_descriptors_task(_task, params) # rubocop:todo Metrics/CyclomaticComp
               # params[:fields] => <ActionController::Parameters {"1"=>"Operator", "2"=>"Workflow (Standard or Xp)", "3"=>"DPX1", "4"=>"DPX2", "5"=>"DPX3", "6"=>"NovaSeq Xp Mainfold", "7"=>"Pipette Carousel", "8"=>"PhiX lot #", "9"=>"PhiX %", "10"=>"Lane loading concentration (pM)", "11"=>"Comment"} permitted: true>
               # I believe that some of this complexity predates ordered hashes in ruby, and was an attempt to maintain field order.
               unless params[:descriptors].nil?
-                event.descriptors = params[:descriptors]
+                event.descriptors = params[:descriptors].to_unsafe_hash
```
KatyTaylor, line 75: what's the reason for this?

app/models/api/messages/flowcell_io.rb
JamesGlover: Sort actuall is ascending, which is why we do a reverse_each

KatyTaylor (approved):
I'm approving as I can't see anything wrong with the code, although slightly unsure if the table format is an improvement in UX - left a comment on the story under the screenshot explaining what I think might help.

app/models/api/messages/flowcell_io.rb
JamesGlover: But as to bc, that's a very good question. I ket the previous variable names in place, but that could definitely do with a change.

I did attempt a large change here, but couldn't work anything else out.

JamesGlover: Huh, although no reason to tap this, as we're not mapping.

app/models/tasks/set_descriptors_handler.rb
JamesGlover: Descriptors are key value pairs defined by the task, and just get stored as a serialised hash. While we could potentially reconstruct the range of permitted keys from the task, there doesn't seem to be much value in doing so, as the only real attack here is adding new keys. I suppose potentially there is a risk of adding a potentially misleading novel key into the mix, 'witnessed by' or something, but it didn't seem worth the added complexity.

db/migrate/20210813103531_migrate_descriptor_fields_into_descriptors.rb
JamesGlover: Already done several runs through of it, including some runs that were self validating. We *do* see some changes, mostly associated with the bug fix restoring missing data, but there are a few other places where we'll also show up data that was previously hidden. Mostly:
- Robot_id was previously only added to the hash, not the fields, so wouldn't show up on the summary page before, and will now.
- There are a few other cases where we have additional fields that will now show up on the batch summary page.  These seem to be cases where additional data has been added later, presumably by USG.

db/migrate/20210817144508_drop_unused_columns.rb
JamesGlover: Migrations will run in order of their timestamps, and won't run if the previous one fails. You'd have to go out of your way to manually run this one first.

description isn't being axed, just changed to a string column, rather than text as its only used to store the step name. (The largest copy in the database is 33 characters long)

filename and data have never been used. They were added to allow the storing of files, but the implementation has long since decayed that it wouldn't work. If we did want to store files against steps in future, we'd use the carrierwave back documents/db_files tables.

lib/acts_as_descriptable.rb
JamesGlover: Its already gone in the head of my branch.

JamesGlover:
I'll merge in, as its better than what's there already, but follow up based on the discussion on the issue/ with Tris.

PR #3296, opened 2021-08-19
Rename intergration to integration

sdjmchattie:
Noticed this while updating integration-suite.  I'm fixing the typos there too so these line up.


JamesGlover (approved):
Cool, will leave you to merge so that you can co-ordinate it with the update on the other side

harrietc52 (approved):

PR #3307, opened 2021-08-31
Fix request detail lookup on batch show

JamesGlover:
-Request information type logic moved onto
 RequestInformationType
- Use consistent sort for labe_events
- Remove detailed_worksheet

This would have needed significant changes to
handle the above adjustments. However on
investigating it turned out that the page was
non-functional, and had been for quite some time.
In addition its contents only made sense for
Illumnia sequencing pipelines, yet it was only
enabled for PacBio.


app/controllers/batches_controller.rb
```diff
@@ -290,25 +290,24 @@ def print_barcodes
     end
   end
 
+  # Handles printing of the worksheet
   def print # rubocop:todo Metrics/AbcSize
```
JamesGlover, line 294: By default, pipelines fell back to a `detailed_worksheet`, which I discovered while looking for other users of `request.value_for`. However this worksheet was throwing exceptions, and would have been for quite some time. Furthermore, its contents made no sense in the context of the only place it was used, the Pacbio pipelines.

So it was clear that it was unused. Set the default to false instead, and added logic to redirect when there's no template. (Also hide the link)

Then rubocop complained, so did a bit of refactoring. (@workflow was unused)

app/helpers/application_helper.rb
```diff
@@ -214,11 +214,6 @@ def tab_pane(name, id: nil, tab_id: nil, active: false, &block)
     )
   end
 
-  def display_request_information(request, rit, batch = nil)
-    r = request.value_for(rit.name, batch)
-    r.presence || 'NA'
-  end
-
```
JamesGlover, line 221: Unused

app/models/api/messages/flowcell_io.rb
```diff
@@ -98,12 +98,6 @@ def detect_float_descriptor(name, ignored_unit)
             nil
           end
         end
-
-        def detect_descriptor(name)
-          # Sort here goes by id ascending, so we use a reverse each, in order to find the most recent descriptor with the passed in 'name'
-          # Lazy ensures we stop searching as soon as we find a value
-          lab_events.sort.lazy.reverse_each.map { |e| e.descriptor_value_for(name) }.detect(&:present?)
-        end
```
JamesGlover, line 106: Used in a wider context, so moved into Request

app/models/pipeline.rb
```diff
@@ -5,7 +5,7 @@
 # and to from there progress a {Batch} through the associated workflow.
 # @note Generally speaking we are trying to migrate pipelines out of the Sequencescape
 #       core.
-class Pipeline < ApplicationRecord # rubocop:todo Metrics/ClassLength
```
JamesGlover, line 8: Hah, nice. This wasn't intentional, but always good to nuke a todo

app/models/pipeline.rb
```diff
@@ -18,7 +18,6 @@ class Pipeline < ApplicationRecord # rubocop:todo Metrics/ClassLength
                   :requires_position,
                   :inbox_partial,
                   :pulldown,
-                  :prints_a_worksheet_per_task,
```
JamesGlover, line 21: False on all remaining pipeline classes, so just nuked it.

app/models/presenters/batch_submenu_presenter.rb
```diff
@@ -98,18 +101,11 @@ def load_pipeline_options # rubocop:todo Metrics/CyclomaticComplexity
...
       end
 
-      if can? :print
-        if @pipeline.prints_a_worksheet_per_task? && !pacbio_sample_pipeline?
-          @tasks.each { |task| add_submenu_option "Print worksheet for #{task.name}", action: :print, task_id: task.id }
-        else
-          add_submenu_option 'Print worksheet', :print
-        end
-      end
+      add_submenu_option 'Print worksheet', :print if worksheet? && can?(:print)
```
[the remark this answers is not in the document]
JamesGlover: The other logic represents a path that was impossible to trigger, so has been removed.

app/models/request.rb
```diff
@@ -411,27 +411,11 @@ def aliquot_attributes
...
-    return rit_value if rit_value.present?
-
-    event_value_for(name, batch)
-  end
-
-  def event_value_for(name, batch = nil)
-    list = (batch.present? ? lab_events_for_batch(batch) : lab_events)
-    list.each { |event| desc = event.descriptor_value_for(name) and return desc }
-    ''
+  def detect_descriptor(name, descriptor_batch: batch)
```
JamesGlover, line 414: Moved in from the FlowcellIO. Couple of adjustments:
1) Filters to the current batch by default. Only likely to have an impact of a request is moved between batches, but that does happen, and there are stories in the backlog to make it easier.
2) Uses the lab_events_for_batch to to the initial sorting to avoid repeating ourselves.

app/models/request.rb
```diff
@@ -411,27 +411,11 @@ def aliquot_attributes
...
-    rit_value = get_value(rit) if rit.present?
-    return rit_value if rit_value.present?
-
-    event_value_for(name, batch)
-  end
-
-  def event_value_for(name, batch = nil)
-    list = (batch.present? ? lab_events_for_batch(batch) : lab_events)
-    list.each { |event| desc = event.descriptor_value_for(name) and return desc }
-    ''
```
JamesGlover, line 434: And most of the rest of this has been moved to request_informations_types.

app/models/request_information_type.rb
```diff
@@ -1,6 +1,41 @@
+# frozen_string_literal: true
+
+# A request information type associates a pipline with request information,
+# either via {Request::Metadata} or a {LabEvent}
 class RequestInformationType < ApplicationRecord
```
JamesGlover, line 5: All the logic here used to exist on Request, but I've re-written it. The formatters stuff isn't actually used, but seemed like a solid enough idea that I'v kept it for now.

app/models/robot.rb
```diff
@@ -64,6 +64,10 @@ def generator(batch:, plate_barcode:, pick_number:)
     generation_behaviour.new(batch: batch, plate_barcode: plate_barcode, picking_data: picking_data, layout: layout)
   end
 
+  def self.default_for_verification
```
JamesGlover, line 67: Mostly because rubocop was complaining about my fat controller.

app/views/batches/_request.html.erb
```diff
@@ -1,75 +0,0 @@
-
```
JamesGlover, line 1: Part of the detailed worksheet. Used `request.value_for(key)` but without request information types. I did start to update it, but then found the page hadn't worked for some time, and didn't make any sense in the contexts in which it was used.

app/views/shared/_batch.html.erb
```diff
@@ -1,5 +1,5 @@
 
-  <table class="sortable table table-stripped">
```
JamesGlover, line 2: Typo here, so the striped style wasn't applied. Decided to just remove it, rather than ad striping.

app/views/shared/_batch.html.erb
```diff
@@ -26,7 +26,7 @@
     </thead>
     <tbody id="requests_list">
       <% requests.each_with_index do |request, i| -%>
-        <tr id="request_<%= request.id %>" class="<%= cycle("rowodd", "roweven") %> row<%= i %>">
```
JamesGlover, line 29: And legacy stuff to make stuff striped in the pre-bootstrap days.

app/models/request_information_type.rb
```diff
@@ -1,6 +1,41 @@
+# frozen_string_literal: true
+
+# A request information type associates a pipline with request information,
+# either via {Request::Metadata} or a {LabEvent}
```
KatyTaylor, line 4: Could you explain this a bit more? I don't think I've come across this area of the data model before. It looks to me like Request Information Type Id is a foreign key on the 'Request Information' table... but there are no rows in that table. Was it deprecated? How does Request Information Type link to {Request::Metadata} and {LabEvent}? I see you've added methods below that look like they pull information from those tables, but what was this object's purpose before you added those methods?

app/controllers/batches_controller.rb
```diff
@@ -290,25 +290,24 @@ def print_barcodes
...
 
-    robot_id = params.fetch(:robot_id, @batch.robot_id)
-
-    @robot = robot_id ? Robot.find(robot_id) : Robot.with_verification_behaviour.first
-
-    # Fall-back
-    @robot ||= Robot.first
-
-    if @pipeline.is_a?(CherrypickingPipeline)
+    if template == 'cherrypick_worksheet'
```
KatyTaylor, line 300: What's the reason for checking the template rather than the pipeline being a 'CherrypickingPipeline' ? Presume it's OK that `@robot` is only set if it's a cherrypicking one?

KatyTaylor (approved):

app/models/request_information_type.rb
JamesGlover: I was a bit new to request information types before this as well.

`RequestInformation` is before my time, and I'm not sure about the decisions around it. I'm not sure if it got deprecated, or never really got used. I believe it essentially lost the battle to request metadata, although given some of its functionality also ended up in LabEvents, I'm not entirely certain what happened. I do know that one of the former team leads, Lars, was generally speaking in favour of metadata tables over key value pairs, but I'm not really certain why. (Partly I think as I fall in the opposite camp)

So RequestInformationTypes appear to be left over from this battle between request-metadata and the lab-events. I'm not sure if its a unification that never happened, or a split that occurred early in its development. As a result, they end up looking in one place, and then the other.

Before my changes it still did an almost identical thing, its just all the associated logic was contained on Request. I flipped it around because:
1) Request already has a lot on its plate, whereas this is almost exclusively concerned with fetching and displaying data
2) The methods on request seemed to talk to the request information type more than they did the request.

app/controllers/batches_controller.rb
JamesGlover: It was mostly because the instance variables being set here are to be used in the worksheets. @robot only gets used in  cherrypick_worksheet, so we don't need to waste time looking it up for other worksheets. I switched to checking the worksheet explicitly, because it seemed to better cover 'why' we were setting the instance (because its used in the template) whereas the fact its a cherrypicking pipeline just happens to correlate (albeit in a pretty tight way).

I think the 'correct' solution, when you have multiple templates rendered by the same endpoint, is probably to use a separate presenter for each one.

app/models/request_information_type.rb
KatyTaylor: OK thanks for the background :)

PR #3309, opened 2021-08-31
Update prettier cops

JamesGlover:
Lots of linting changes in this as I updated ruby-prettier

The biggest impact is that prettier now enables the rubocop line layout cops. (Which helps it avoid getting into a few battles, which triggered by attempt to update)

In turn thought this pretty much force my hand in putting the line length todos inline.


app/models/batch.rb
```diff
@@ -477,13 +476,17 @@ def swap(current_user, batch_info = {}) # rubocop:todo Metrics/CyclomaticComplex
       batch_request_left.batch.lab_events.create!(
         description: 'Lane swap',
         message:
+          # rubocop:todo Layout/LineLength
```
sdjmchattie, line 479: Same comment about moving this up a line (or in this case indenting the enable line if that makes more sense)

app/models/bulk_submission.rb
```diff
@@ -116,7 +116,10 @@ def valid_header?
     return true if headers.include? 'submission name'
 
     errors.add :spreadsheet,
+               # rubocop:todo Layout/LineLength
```
sdjmchattie, line 119: Such indent

app/models/bulk_submission.rb
```diff
@@ -400,7 +417,9 @@ def prepare_order(details) # rubocop:todo Metrics/CyclomaticComplexity
       attributes[:assets] = found_assets
     elsif found_assets.present? && found_assets != attributes[:asset_group].assets
       raise StandardError,
+            # rubocop:todo Layout/LineLength
```
sdjmchattie, line 420: I'm going to stop commenting on these now.  There are more than I expected.  I know they're auto-generated -- shame the auto-generator isn't a bit smarter.

app/models/fluidigm_template_task.rb
```diff
@@ -6,7 +6,8 @@ def partial
   def plate_purpose_options(batch) # rubocop:todo Metrics/AbcSize
     requests = batch.requests.flat_map(&:next_requests)
     plate_purposes = requests.filter_map(&:request_type).uniq.map(&:acceptable_plate_purposes).flatten.uniq
-    plate_purposes = batch.requests.filter_map { |r| r.request_metadata.target_purpose }.uniq if plate_purposes.empty? # Fallback situation for the moment
+    plate_purposes = # Fallback situation for the moment
+      batch.requests.filter_map { |r| r.request_metadata.target_purpose }
```
sdjmchattie, line 10: This doesn't look like it's functionally the same.  There's no guard to prevent overwriting `plate_purposes` if it is not `empty?` and the `.uniq` has been dropped.

app/models/illumina_c/requests.rb
```diff
@@ -17,7 +17,8 @@ def update_pool_information(pool_information)
   # Stage 1:
   # Refactor to allow use of IlluminaC::Requests::LibraryRequest [Done: This commit]
   # Step 2:
-  # Write migration to update existing request types / requests to use IlluminaC::Requests::LibraryRequest [Will be done soon]
+  # Write migration to update existing request types / requests to use IlluminaC::Requests::LibraryRequest
+  # [Will be done soon]
```
sdjmchattie, line 21: Sure sure ... 😄

app/models/illumina_htp/initial_stock_tube_purpose.rb
```diff
@@ -22,9 +22,13 @@ def sibling_tubes(tube) # rubocop:todo Metrics/AbcSize
     sibling_requests = tube.submission.requests.multiplexed.opened.ids
 
     sibling_tubes =
+      # rubocop:todo Layout/LineLength
       Tube.joins(:transfer_requests_as_target).includes(:transfer_requests_as_source) # Outer join, as we don't want these
+        # rubocop:enable Layout/LineLength
         .where(transfer_requests: { submission_id: tube.submission }) # find out tubes via transfer requests
+        # rubocop:todo Layout/LineLength
         .where("transfer_requests_as_sources_#{Tube.table_name}": { id: nil }) # Make sure we have no transfers out of the tube
+        # rubocop:enable Layout/LineLength
```
sdjmchattie, line 31: I know I said I wouldn't comment on these any more, but this one is particularly frustrating.

app/models/parsers/bioanalysis_csv_parser.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-class Parsers::BioanalysisCsvParser # rubocop:todo Style/Documentation
+# rubocop:todo Metrics/ClassLength
+class Parsers::BioanalysisCsvParser # rubocop:todo Style/Documentation, Metrics/ClassLength
```
sdjmchattie, line 4: There's some weird duplication of the `:todo` markers here.  I assume those that are on the same line as a class are automatically scoped to the class without needing to be enabled again explicitly?  Otherwise is `Style/Documentation` off permanently after Rubocop hits this file?

app/models/request.rb
```diff
@@ -178,8 +177,12 @@ def date_for_state(state)
             .joins(
               add_joins + [
                 'INNER JOIN maps AS pw_location ON pw.map_id = pw_location.id',
+                # rubocop:todo Layout/LineLength
                 'INNER JOIN pre_capture_pool_pooled_requests ON requests.id=pre_capture_pool_pooled_requests.request_id',
+                # rubocop:enable Layout/LineLength
+                # rubocop:todo Layout/LineLength
                 'INNER JOIN uuids ON uuids.resource_id = pre_capture_pool_pooled_requests.pre_capture_pool_id AND uuids.resource_type="PreCapturePool"'
+                # rubocop:enable Layout/LineLength
```
sdjmchattie, line 185: Maybe consolidate this block

app/models/tasks/assign_tubes_to_wells_handler.rb
```diff
@@ -50,22 +50,25 @@ def do_assign_pick_volume_task(_task, params)
   # rubocop:todo Metrics/AbcSize
   def wells_with_duplicates(params) # rubocop:todo Metrics/CyclomaticComplexity
```
sdjmchattie, line 51: If the CyclomaticComplexity todo is scoped to the method by declaring it this way, maybe the other three above should be too since they're turned back on after the method.

app/models/transfer_request.rb
```diff
@@ -22,8 +22,8 @@ class TransferRequest < ApplicationRecord # rubocop:todo Metrics/ClassLength
   # States which are still considered to be processable (ie. not failed or cancelled)
   ACTIVE_STATES = %w[pending started passed qc_complete].freeze
 
-  # The assets on a request can be treated as a particular class when being used by certain pieces of code.  For instance,
-  # QC might be performed on a source asset that is a well, in which case we'd like to load it as such.
+  # The assets on a request can be treated as a particular class when being used by certain pieces of code.
+  #  For instance, QC might be performed on a source asset that is a well, in which case we'd like to load it as such.
```
sdjmchattie, line 26:
```suggestion
  # For instance, QC might be performed on a source asset that is a well, in which case we'd like to load it as such.
```

app/sample_manifest_excel/sample_manifest_excel/upload/base.rb
```diff
@@ -48,7 +51,9 @@ def initialize(attributes = {}) # rubocop:todo Metrics/AbcSize
       end
 
       def inspect
+        # rubocop:todo Layout/LineLength
```
sdjmchattie, line 54: Could declare this as a method scope todo at the end of the method name?

db/migrate/20180323134506_update_view_requests.rb
```diff
@@ -2,6 +2,7 @@
 
 # The requests view had been updated to include request purpose, which is now an enum.
 # This adds the previously untracked request_purpose column to the view in a compatible manner.
+# rubocop:disable Layout/LineLength
```
sdjmchattie, line 5: Could be declared at the end of the `class` line?

db/migrate/20190829125352_update_view_asset_links.rb
```diff
@@ -23,7 +23,9 @@ def self.down
     ViewsSchema.update_view(
       'view_asset_links',
       # Direct dump of current behaviour, keeping risk minimal by not adjusting white-space
+      # rubocop:disable Layout/LineLength
```
sdjmchattie, line 26: Should be `todo` instead of `disable`?  Or is this a permanent "won't fix"?  If the former, there are others in other migration files to consider changing as well.

db/seeds/0001_snp_plate_purposes.rb
```diff
@@ -72,9 +72,15 @@
 
 ActiveRecord::Base.transaction do
   # A couple of legacy pulldown types
+  # rubocop:todo Layout/LineLength
   PlatePurpose.create!(name: 'SEQCAP WG', cherrypickable_target: false) # Superceded by Pulldown WGS below (here for transition period)
+
+  # rubocop:enable Layout/LineLength
+  # rubocop:todo Layout/LineLength
```
sdjmchattie, line 79: Should be consolidated

features/support/timecop.rb
```diff
@@ -6,7 +6,16 @@ class Timecop # rubocop:todo Style/Documentation
   class << self
     # Block the use of Timecop.freeze as it upsets the Capybara...
     def freeze_with_warning(_time)
-      raise "\n\n#{'*' * 90}\nTimecop.freeze() interferes with Capybara's javascript timeouts.\nCould you either use Timecop.travel instead or not use JavaScript in this scenario?  \n\n#{'*' * 90}\n\n"
+      raise <<~EXCEPTION
+
+        #{'*' * 90}
```
sdjmchattie, line 11: I think you might be missing a new line character at the start and the end of this heredoc.

spec/models/sample_manifest_spec.rb
```diff
@@ -224,22 +224,9 @@
...
-                                                                                                                                                                                          BroadcastEvent,
-                                                                                                                                                                                          :count
-                                                                                                                                                                                        )
-                                                                                                                                                                                          .by(
-                                                                                                                                                                                          1
-                                                                                                                                                                                        )
```
sdjmchattie, line 242: Oh wow!

sdjmchattie (changes requested):
By all means, choose your battles.  I just wanted to give this a thorough review and I think some of the suggestions are of value.  The indentation of the `todo` and subsequent `enable` directives are a bit annoying, but there are too many to really care enough about to modify them manually.

spec/models/sample_manifest_spec.rb
JamesGlover: This change was actually manual. I assumed when I saw it in the diff that prettier or rubocop had done something funky in this set of changes, hadn't realised it was terrible before the last sweep

db/migrate/20190829125352_update_view_asset_links.rb
JamesGlover: It was a wont fix. I thought I switched the other view migrations to Todos as well, but may have missed some.

app/controllers/admin/bait_libraries/bait_library_suppliers_controller.rb
```diff
@@ -39,7 +39,10 @@ def destroy
     if @bait_library_supplier.bait_libraries.visible.count > 0
       respond_to do |format|
         flash[:error] =
+          # rubocop:todo Layout/LineLength
```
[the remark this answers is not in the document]
JamesGlover: Good call, but in this case I've actually gone ahead and resolved the issues. There was a little more weirdness going on:
1) Two calls to @bait_library_supplier.bait_libraries.visible.count
2) The notice was conditional on the return from @bait_library_supplier.hide despite this always being truthy.
3) Bit of code duplication with the redirect. Also removed this from the html formatter, as we're not actually rendering anything.

app/controllers/admin/bait_libraries/bait_library_types_controller.rb
```diff
@@ -40,7 +40,10 @@ def destroy
     if @bait_library_type.bait_libraries.visible.count > 0
       respond_to do |format|
         flash[:error] =
+          # rubocop:todo Layout/LineLength
```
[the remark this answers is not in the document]
JamesGlover: And done the same here, as the two were nearly identical

db/migrate/20180323134506_update_view_requests.rb
JamesGlover: Actually no. It appears that the inline declarations only work for cops which specifically concern the class/method they are inline for. It isn't a way of simply scoping a cop to a class/method.

db/migrate/20190829125352_update_view_asset_links.rb
JamesGlover: I've added a couple of missing documentation comments, but otherwise converted all the migration todos to disables as in practice we're never going to go back and update historic migration.

db/seeds/0001_snp_plate_purposes.rb
JamesGlover: Done, along with any others that needed consolidating.

features/support/timecop.rb
JamesGlover: You're right. This was half intentional, as adding two seemed absurd, however when actually printed to the console this would just result in a single blank line, not the two that the heredoc shows, so I'll add the extra new line in.

JamesGlover:
So I fixed up your suggestions, and also searched for similar patterns I could fix up. So this includes:
- Consolidation of any enable->disable pairings
- Consolidate cops onto a single line where appropriate.

There were also a few cases where you suggested moving a todo line, where I did the todo instead. Biggest one of these is in:

https://github.com/sanger/sequencescape/pull/3309/commits/1ba77ed1739856fab2b106468ea5b01b8aec6b30

I go into the changes a bit in the commit message there, as it looks like a more significant change than it actually is.

sdjmchattie:
Just wanted to check whether you spotted this comment and whether what I said is valid or not? https://github.com/sanger/sequencescape/pull/3309#discussion_r700362780

app/models/fluidigm_template_task.rb
JamesGlover: What the flip happened there?! I rolled it back, then ran rubocop over it again, only for it to spit out this beauty:
```ruby
class FluidigmTemplateTask < PlateTemplateTask # rubocop:todo Style/Documentation
  def partial
    'fluidigm_template_batches'
  end

  def plate_purpose_options(batch) # rubocop:todo Metrics/AbcSize
    requests = batch.requests.flat_map(&:next_requests)
    plate_purposes = requests.filter_map(&:request_type).uniq.map(&:acceptable_plate_purposes).flatten.uniq
    plate_purposes = # Fallback situation for the moment
batch.requests.filter_map do |r|
 r.request_metadata.target_purpose end.uniq
    plate_purposes.map { |p| [p.name, p.size, p.id] }.sort
  end
end
```

Which is also not the same. Prettier then reformatted it to something more akin to what was there, but with the uniq at least.

JamesGlover:
> Just wanted to check whether you spotted this comment and whether what I said is valid or not? [#3309 (comment)](https://github.com/sanger/sequencescape/pull/3309#discussion_r700362780)

😱  No, I missed that one. Looks like a horrible bug in the rubocop safe auto-corrects. Well spotted.

(Looks like I missed a load of comments that GH hid beneath the fold)

app/models/parsers/bioanalysis_csv_parser.rb
JamesGlover: Not quite scoped the the class as mentioned before, but more limits it to problems with the class itself. You were right that this one was redundant though, not sure why Rubocop didn't detect that

sdjmchattie (approved):

PR #3313, opened 2021-09-01
Issue3312/dpl 122 import cardinal samples via manifest

### LINKED ISSUE #3312: DPL-122 Import Cardinal samples via manifest
**User story**
As a member of the Cardinal team, I want to be able to upload samples / tubes into Sequencescape, so that I can record metadata about them, and continue along the pipeline using the LIMS.

Stef L

**Acceptance criteria**
- Import uses sample manifests
- Samples are created in tubes, of purpose "LCA Blood Tube"
- Sample metadata needed:
  - Identifier of the sample is the tube barcode, which is stored in the sample metadata `sample_supplier_name`
  - Use manifest supplier to track location / Site ID:
    - Stockport: 11025
    - Reading: 11026
    - Newcastle: 11027
    - Bristol: 11028
    - East London Genes and Health: ?
  - Sample volume (volume of blood) (Use existing volume)

**Dependencies**
References
- https://github.com/sanger/limber/issues/755 

**Additional context**

It is expected that the tubes / samples might be imported in a different way in future, for instance a more automated way to save time, or once the BMS system (Bioresource Management System) is up and running. 

This might already be doable using existing tube manifests - this story is to make sure.

JG: Yep, I think almost everything is in place. Will just need a new, super-tiny manifest format and new tube barcode.

There were these original requirements - not included in acceptance criteria due to reasons below:
- Upon receipt of the blood tubes at Sanger the blood tubes will be scanned into the LIMS. The LIMS should collate a list of the scanned-in tubes and send an e-mail to the various sites to confirm receipt of the particular tubes.  (This was agreed to be handled by the BMS - assume it will be manual until this is built / purchased).
- Following the entry of the samples into the LIMS, I need to be able to assign them into a storage location. (Assume this is already covered by LabWhere).

JamesGlover:
Closes #3312

* Add minimal cardinal manifest template (We need Sanger sample ids in there as that's how samples get linked behind the scenes) 
* Add cardinal tube purpose
* Minor tweaks to manifest page to add basic table formatting
* Minor tweaks to manifest page to remove hidden span and use time tags.

### DIFF
--- a/config/default_records/tube_purposes/005_cardinal_tube_purposes.yml
+++ b/config/default_records/tube_purposes/005_cardinal_tube_purposes.yml
@@ -1,0 +1,7 @@
+# Note: Tubes in the cardinal pipeline are under control of Limber. These are
+# the initial tubes that enter via manifest
+---
+LCA Blood Tube:
+  type: Tube::Purpose
+  target_type: SampleTube
+  stock_plate: false
--- a/config/default_records/tube_purposes/005_cardinal_tube_purposes.yml
+++ b/config/default_records/tube_purposes/005_cardinal_tube_purposes.yml
@@ -1,7 +1,7 @@
 # Note: Tubes in the cardinal pipeline are under control of Limber. These are
 # the initial tubes that enter via manifest
 ---
-LCA Blood Tube:
+LCA Blood Vac:
   type: Tube::Purpose
   target_type: SampleTube
   stock_plate: false
--- a/config/sample_manifest_excel/manifest_types.yml
+++ b/config/sample_manifest_excel/manifest_types.yml
@@ -678,3 +678,11 @@ heron:
     - :donor_id2
     - :control_type
     - :priority
+cardinal:
+  heading: "Cardinal"
+  asset_type: "1dtube"
+  columns:
+    - :sanger_tube_id
+    - :sanger_sample_id
+    - :supplier_name
+    - :volume
--- a/app/views/shared/_manifests.html.erb
+++ b/app/views/shared/_manifests.html.erb
@@ -1,5 +1,5 @@
 
-<table id="study_list" class="sortable">
+<table id="study_list" class="sortable table">
   <thead>
   <tr>
     <th>Name</th>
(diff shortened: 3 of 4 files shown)


config/sample_manifest_excel/manifest_types.yml
```diff
@@ -678,3 +678,11 @@ heron:
...
     - :control_type
     - :priority
+cardinal:
+  heading: "Cardinal"
+  asset_type: "1dtube"
+  columns:
+    - :sanger_tube_id
+    - :sanger_sample_id
+    - :supplier_name
+    - :volume
```
KatyTaylor, line 688: OK so:
- `:sanger_tube_id` -> this is auto-populated with the NT barcode, and can be pasted over for the UKB tubes, which have their own barcodes.
- `:sanger_sample_id` -> this is auto-populated
- `:supplier_name` -> this will be manually populated with the NT barcode or UKB barcode
- `:volume` -> this will be manually filled out - actually not sure if this is something that is supplied or will be measured by the bioresource team - will check

And then supplier will be filled in when generating the manifest, as per your comment on the story.

Is that correct?

KatyTaylor: Volume - Stef reckons this is measured on-site by the bioresource team - is this field the one you said is 'customer measured volume'? If so, might be misleading.

JamesGlover: Yep, saw the slack conversation, it might be that this gets dropped. But essentially that means we have no customer supplied fields, so it might be easier it the manifest registers all samples up-front. Plus sets the supplier name. That way there would be no need to upload manifests at any stage. A bit more work for us, but far less work for SSRs or anyone else. It does depend on how the volume conversation resolves though.

config/default_records/tube_purposes/005_cardinal_tube_purposes.yml
```diff
@@ -0,0 +1,7 @@
+# Note: Tubes in the cardinal pipeline are under control of Limber. These are
+# the initial tubes that enter via manifest
+---
+LCA Blood Tube:
+  type: Tube::Purpose
+  target_type: SampleTube
+  stock_plate: false
```
KatyTaylor, line 7: Remind me how these are run?

JamesGlover: There is a task:
```bash
bundle exec rake record_loader:all
```
That gets triggered automatically on deployment. (After migrations have run)

KatyTaylor (approved):

PR #3324, opened 2021-09-07
DPL-128 add message about deprecation of the run status updates

### LINKED ISSUE #3320: DPL-128 Add deprecation message for run status updates
**User story**
As developers we would like to deprecate the batch.xml API. NPG have a number of dependencies on this that they have been working through (see email chain "Sequencescape 'Batch xml'" Katy, James, David J, Marina, Rich, Jaime).

One step towards doing this is to deprecate the run status updates that NPG send, which are displayed on the request event history page (e.g. http://sequencescape.psd.sanger.ac.uk/requests/21095252/history).

We want to put a message up saying these will no longer be sent, to warn our users for about a month before the feed stops.

Katy, Marina

**Acceptance criteria**
- [ ] Add a message to the request history page to say that run status updates will no longer be sent from October 2021 (don't say a specific date), and to get in contact with PSD if you need that information.
- [ ] Set a reminder to check with NPG towards the start of October when they are stopping them, as we'll need to take the message down again. 

**Additional context**
See email chain "Sequencescape 'Batch xml'" Katy, James, David J, Marina, Rich, Jaime for initial discussion.
and "PSD issue to follow up" (Katy, Marina) for follow up.

KatyTaylor:
Closes #3320

### DIFF
--- a/app/views/requests/history.html.erb
+++ b/app/views/requests/history.html.erb
@@ -4,6 +4,12 @@
 
 <%= page_title 'Event History' %>
 
+<div>
+  <%= alert(:warning) do %>
+    From October 2021, the Event History page will no longer receive and display new updates on the status of the run. If this will be a problem for you, please let team PSD know by emailing psd-help@sanger.ac.uk
+  <% end %>
+</div>
+
 <div class="help">
   Complete event history for request <%= @request.id %>, in chronological order.
 </div>
--- a/app/views/requests/history.html.erb
+++ b/app/views/requests/history.html.erb
@@ -6,7 +6,7 @@
 
 <div>
   <%= alert(:warning) do %>
-    From October 2021, the Event History page will no longer receive and display new updates on the status of the run. If this will be a problem for you, please let team PSD know by emailing psd-help@sanger.ac.uk
+    From October 2021, the Event History page will no longer receive and display new updates on the status of the run. If this will be a problem for you, please let team PSD know by emailing <%= help_email_link %>
   <% end %>
 </div>
 


KatyTaylor:
See screenshots on issue #3320

JamesGlover (approved):

JamesGlover (commented):
Whoops, approved slightly too eagerly, we probably should adjust the email to pull from the one in the config.

JamesGlover:
There's actually a helper that generates a mailto link to the right address:
```
<%= help_email_link %>
```
https://github.com/sanger/sequencescape/blob/8cdeb0c9101ffc069639983c4773bebf40a1c5e4/app/helpers/application_helper.rb#L250

KatyTaylor:
> There's actually a helper that generates a mailto link to the right address:
> 
> ```
> <%= help_email_link %>
> ```
> 
> https://github.com/sanger/sequencescape/blob/8cdeb0c9101ffc069639983c4773bebf40a1c5e4/app/helpers/application_helper.rb#L250

Looks like that would go to seq-help, that OK?

JamesGlover:
> > There's actually a helper that generates a mailto link to the right address:
> > ```
> > <%= help_email_link %>
> > ```
> >     
> >   
> > https://github.com/sanger/sequencescape/blob/8cdeb0c9101ffc069639983c4773bebf40a1c5e4/app/helpers/application_helper.rb#L250
> 
> Looks like that would go to seq-help, that OK?

Oh that's odd. It all ends up in the same place, so it is okay, but I'm surprised the configuration hasn't been updated. We use the same parameter for the help form, and the feedback link at the bottom of the page. (And both definitely come to us) @neilsycamore might know if there was a reason the link never got updated.

JamesGlover:
@neilsycamore might know if there was a reason the link never got updated.

Asking on slack, just in case there's a better reason than 'no one remembered to update the config'.

JamesGlover (approved):

JamesGlover:
Was pretty much just a case of the email not getting updated. Apparently psd-help was introduced the week after I started as Sanger, and the two legacy emails (including seq-help) were set to forward to it.

PR #3327, opened 2021-09-08
updated sample model to have supplier method which returns its sample…

emrojo:
… manifest supplier name

Closes #

*
*
* ...

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -472,6 +472,10 @@ class Sample < ApplicationRecord # rubocop:todo Metrics/ClassLength
     sample_supplier_name_empty?(supplier_sample_name)
   end
 
+  def supplier
+    sample_manifest ? sample_manifest.supplier.name : nil
+  end
+
   def sample_supplier_name_empty?(supplier_sample_name)
     supplier_sample_name.blank? ||
       ['empty', 'blank', 'water', 'no supplier name available', 'none'].include?(supplier_sample_name.downcase)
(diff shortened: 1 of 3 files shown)


app/models/sample.rb
```diff
@@ -472,6 +472,10 @@ def sample_empty?(supplier_sample_name = name)
     sample_supplier_name_empty?(supplier_sample_name)
   end
 
+  def supplier
+    sample_manifest ? sample_manifest.supplier.name : nil
```
JamesGlover, line 476: Use the safe navigation operator:
```
sample_manifest&.supplier&.name
```

JamesGlover: Or we could actually add:
```
has_one :supplier, through: :sample_manifest
```
To pop a direct link to the supplier object. Normally you'd then expose the name as `supplier_name` although that would cause confusion here due to the confusingly named attribute in sample metadata.

I'm not sure quite what to suggest in terms of naming.

app/resources/api/v2/sample_resource.rb
```diff
@@ -14,6 +14,7 @@ class SampleResource < BaseResource
       attribute :uuid
       attribute :control
       attribute :control_type
+      attribute :supplier
```
JamesGlover, line 17: This attribute is dependent on both sample manifest and supplier. This means that for every sample, we'll generate two additional queries, which results in significant slowdown if retrieving large numbers of samples.

If you're going to add methods to the API that depend on other records, then you *need* to ensure that they are added to the default includes for that resource, otherwise you'll bog down all requests for sample.

Either that, or expose the actual supplier object, that way we could only load it when we actually needed it.

JamesGlover (commented):
I know this is a WIP, so just some comments on what's there so far.

PR #3333, opened 2021-09-13
Dpl 064 qc cell count

### LINKED ISSUE #3332: DPL-066 Cardinal - Upload cell count data
**User story**
As a user of the Cardinal pipeline, I want there to be a record of the cell count and other measurements of each sample in the LCA PBMC plate, and to have some help in deciding which wells to take forward and which to fail, based on quality measures.

Stef L

**Acceptance criteria**
- I can upload a new quant type in quanthub: Cardinal_PBMC_Count
- I can upload an excel (CSV) file to quanthub. See example below. Well name will display the name of each sample as it was in the LIMS-generated Cellaca upload file (rather than A1, A2 etc). For most plates, multiple files will need to be uploaded as only 24 samples can be read at a time on the Cellaca. Quanthub will need to assemble these multiple uploads to be processed together as a single plate. 
image
_Figure 2. Cellaca output file to be uploaded to quanthub. Up to 4 files per 96 well plate will be generated._ 

Example files (choice of 2 file formats):
[Cellaca Excel.xlsx](https://github.com/sanger/limber/files/6841852/Cellaca.Excel.xlsx)
[Cellaca Export.csv](https://github.com/sanger/limber/files/6841854/Cellaca.Export.csv)
 

- If the % viability is below 50% the well needs to be flagged, to inform the user of lower than optimal viability.  
- If live cell count is less than 400,000 cells/ml the well needs to be flagged, to inform the user of lower than optimal cell count.
- Investigate how these flags are normally created - assume Quanthub will both visually highlight them, and export the info to MLWH, and Limber will read this info and also visually highlight them. (new Limber plate presenter, importing required infor from SS, then specifically show this data in Limber)
- I can fail certain wells of choice on the LCA PBMC plate. (should be covered by standard Limber functionality) (_Q: assume this will be a manual step, choose reason why failed, which trigger event to MLWH, add more 'reasons', or could be automated with 'threshold' reason_)
- I can download the collated quality information so that it can be uploaded to an ELN - if this is not trivial, it's not necessary as they can upload the raw output from Cellaca into ELN.

**Dependencies**
- pipeline setup - https://github.com/sanger/limber/issues/772
- Decide for sure what to put as the sample identifier in the generated input file - https://github.com/sanger/limber/issues/757 (probably sample name = tube barcode)
- LCA PBMC plate created, in https://github.com/sanger/limber/issues/757

stevieing:
Closes #3332

*
*
* ...


app/models/parsers/cardinal_pbmc_count_parser.rb
```diff
@@ -0,0 +1,56 @@
...
+      'Total Mean Size',
+      'Note:',
+      'Errors:'
+    ].freeze
+
+    self.assay_type = 'Cardinal_PBMC_Count'
+    self.assay_version = 'v1.0'
+
+    def self.parses?(content)
+      content.split('\r\n')[0][0] == HEADERS
```
sdjmchattie, line 27: Worth commenting about the nested arrays that are incoming?  I assume that's why there are two `[0]` here?

stevieing: This is more about how the data is formatted rather than nested arrays for Cardinal. The original file is parsed as a csv which creates a 2 dimensional array.

sdjmchattie (approved):

PR #3339, opened 2021-09-15
DPL-067-2

harrietc52:


app/models/transfer_request.rb
```diff
@@ -224,7 +224,7 @@ def perform_transfer_of_contents
     target_asset.aliquots << aliquots_for_transfer
   rescue ActiveRecord::RecordNotUnique => e
     # We'll specifically handle tag clashes here so that we can produce more informative messages
-    raise e unless /aliquot_tags_and_tag2s_are_unique_within_receptacle/.match?(e.message)
+    raise e unless /aliquot_tag_tag2_and_tag_depth_are_unique_within_receptacle/.match?(e.message)
```
JamesGlover, line 227: There's similar code elsewhere in SS that will need to be updated:
Such as:
https://github.com/sanger/sequencescape/blob/a7c43e591668745dd1283ce3fecf2343781a39e4/app/models/tag_substitution.rb#L109

JamesGlover: Can't actually find other references at the moment, could have sworn we had a custom exception matcher on `Aliquot::TagClash` but know we also had a few bugs with handling that exception, so maybe it got ripped out / never got merged.

app/models/transfer_request.rb
```diff
@@ -17,7 +17,7 @@ class TransferRequest < ApplicationRecord # rubocop:todo Metrics/ClassLength
 
   # Determines if we attempt to filter out {Aliquot#equivalent? equivalent} aliquots
   # before performing transfers.
-  attr_accessor :merge_equivalent_aliquots
+  attr_accessor :merge_equivalent_aliquots, :tag_depth
```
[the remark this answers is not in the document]
JamesGlover: I'm not sure we want to expose this as a separate attribute on transfer request, when https://github.com/sanger/sequencescape/pull/3338 is about to expose 'aliquot_attributes'. The advantage of the latter is we can potentially set any of the attributes on aliquot easily, without needing to modify transfer request. This approach would also let us set, say, the suboptimal flag, on an aliquot automatically when creating a descendant. (Something that may be needed for cardinal)

So instead of 
```
{
    "transfer_request_collection":{
      "user": "#{user.uuid}",
      "transfer_requests": [
        {
          "source_asset":"#{asset.uuid}",
          "target_asset": "#{target_asset.uuid}",
          "submission": "#{submission.uuid}",
          "volume": 10,
          "tag_depth": 2
        }
      ]
    }
  }
```

You'd post:

```
{
    "transfer_request_collection":{
      "user": "#{user.uuid}",
      "transfer_requests": [
        {
          "source_asset":"#{asset.uuid}",
          "target_asset": "#{target_asset.uuid}",
          "submission": "#{submission.uuid}",
          "volume": 10,
          "aliquot_attributes": {
            "tag_depth": 2
          }
        }
      ]
    }
  }
```

JamesGlover:
Realise this is still a wip, so just a couple of comments on what's there at the moment.

db/migrate/20210915093828_add_tag_depth_colunn_to_aliquots_table.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Added to support the Cardinal Pipeline, which requires untagged samples to be pooled,
+# and not through a tag clash error
+# tag_depth here is unique within a pool
+class AddTagDepthColunnToAliquotsTable < ActiveRecord::Migration[6.0]
```
sdjmchattie, line 6: `Colunn` should be `Column`

spec/models/transfer_request_spec.rb
```diff
@@ -184,6 +184,56 @@
...
+      #     end
+      #   end
+
+      #   context 'when tag depth not defined' do
+      #     let(:tag_depth) {nil}
+      #     it 'does not return tag_depth' do
+      #       expect(transfer_request.aliquot_attributes).not_to have_key(:tag_depth)
+      #     end
+      #   end
+      # end
```
sdjmchattie, line 234: Perhaps this commented out code needs to be removed?

sdjmchattie (changes requested):

harrietc52 (commented):
All good just agree with the existing comments

harrietc52:
Closing, as https://github.com/sanger/limber/issues/830 explains, we are reverting the tag_depth column

PR #3340, opened 2021-09-16
Rewrite the submission template creation for Cardinal

KatyTaylor:
- driver is that I needed it to specify the project_id, so it could be used in an 'automated' way from Limber
- refactored the project querying code that is also used by heron into the helper class
- renamed a task to 'create_purposes' - makes more sense
- got rid of all the extra submission templates as the plan is for the sequencing submissions to be manual, after library prep (due to custom / multiple pooling steps)

- added in things for cardinal cell banking pipeline

### DIFF
--- a/lib/tasks/limber.rake
+++ b/lib/tasks/limber.rake
@@ -19,8 +19,7 @@ namespace :limber do
       { name: 'GBS Stock', size: 384 },
       { name: 'GnT Stock', size: 96 },
       { name: 'LTHR RT', size: 96 },
-      { name: 'LTHR-384 RT', size: 384 },
-      { name: 'LCA Blood Array', size: 96 }
+      { name: 'LTHR-384 RT', size: 384 }
     ]
 
     purposes.each do |purpose|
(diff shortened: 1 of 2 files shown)


lib/tasks/limber.rake
```diff
@@ -762,22 +754,18 @@ namespace :limber do
...
-        product_line: 'Cardinal',
-        catalogue: cardinal_catalogue
-      ).build!
+      unless SubmissionTemplate.find_by(name: 'Limber - Cardinal')
+        SubmissionTemplate.create!(
+          name: 'Limber - Cardinal',
+          submission_class_name: 'LinearSubmission',
+          submission_parameters: {
+            request_type_ids_list: [RequestType.where(key: 'limber_cardinal').pluck(:id)],
+            project_id: Limber::Helper.find_project('Project Cardinal').id
```
JamesGlover, line 763:
```suggestion
            request_type_ids_list: [RequestType.where(key: 'limber_cardinal').ids],
            project_id: Limber::Helper.find_project('Project Cardinal').id
```
Just a minor adjustment, purely personal preference, so feel free to ignore.

JamesGlover (approved):

JamesGlover (approved):

PR #3341, opened 2021-09-16
Add qc result uat action

JamesGlover:
I wanted to test with the actual qc-attributes we'd be expecting for this step, but the existing uat action only allowed for molarity and concentrations, something that was also implied by its name.

We probably want to get rid of the more specific one soon, but I left it there until we can update the UAT actions.

### DIFF
--- a/app/uat_actions/uat_actions/generate_qc_results.rb
+++ b/app/uat_actions/uat_actions/generate_qc_results.rb
@@ -1,0 +1,122 @@
+# frozen_string_literal: true
+
+# Will generate qc results for a given plate
+class UatActions::GenerateQcResults < UatActions
+  self.title = 'Generate qc results for a plate or tube'
+  self.description = 'Generate a set of randomised qc result values for a plate or tube.'
+
+  ATTRIBUTE_UNITS = {
+    'molarity' => 'nM',
+    'volume' => 'ul',
+    'concentration' => 'ng/ul',
+    'gender_markers' => 'bases',
+    'loci_passed' => 'bases',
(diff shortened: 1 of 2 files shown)


app/uat_actions/uat_actions/generate_qc_results.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
+
+# Will generate qc results for a given plate
+class UatActions::GenerateQcResults < UatActions
+  self.title = 'Generate qc results for a plate or tube'
+  self.description = 'Generate a set of randomised qc result values for a plate or tube.'
+
+  ATTRIBUTE_UNITS = {
```
andrewsparkes, line 8: I think my only comment here is that the original one for concentrations produces values with decimal places.
If the units are e.g. cells/ml you will only want integer values to be realistic (the cell counter will not give a value that is a part of a cell). Maybe it doesn't matter for testing.

andrewsparkes (approved):

app/uat_actions/uat_actions/generate_qc_results.rb
JamesGlover: Good point. There actually was a minor problem with whole numbers, because ruby decided to represent them as a Rational, which the javascript then got confused about. So everything actually does get converted to floats for processing thresholds anyway. (Picked up in the unit tests) Hadn't considered that we'd expect integers for all value here.

But are we sure that's the case here? Presumably the reading is intended to be 'average cells/ml' so could theoretically include partial cells if it were based on measurements of >1ml.

PR #3353, opened 2021-09-29
Bug dpl 064 qc cell count

stevieing:
Closes #

*
*
* ...


app/models/parsers/cardinal_pbmc_count_parser.rb
```diff
@@ -44,11 +44,18 @@ def qc_data
...
+            qc_data[get_well_location(row[0])] = {
+              viability: Unit.new(row[4]),
+              live_cell_count: Unit.new(row[2], 'cells')
+            }
           end
         end
     end
 
+    def get_well_location(cell)
+      cell.try(:split, ':')[1]
```
emrojo, line 56: Why is .try() needed? If the string does not have a ':' separator, don't we want it to fail?

stevieing: good point

sdjmchattie: This seems to be tightly coupled to the implementation here: https://github.com/sanger/sequencescape/blob/develop/app/models/well.rb#L329

Would it be worth either putting both the implementations next to each other or putting a comment in both locations to make the association obvious to someone who decides to update one or the other in the future?

PR #3355, opened 2021-10-04
Dpl 067 compound sample

### LINKED ISSUE #3348: DPL-067-5 SS records for compound sample (S)
**User story**
As a developer, we want to create a new table in SS, linking the source sample to the pseudo sample 

Harry
Eduardo

**Acceptance criteria**
- [ ]  Creation of a new empty table
- [ ]  two columns, both integers, foreign keys to the samples table

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

sdjmchattie:
Closes #3348 

* Add new compound sample ability between samples.
* Ensure the samples cannot be both components and acting as a compound at the same time.


spec/models/sample_spec.rb
```diff
@@ -167,4 +167,79 @@
       expect(sample.sample_metadata.supplier_name).to be_nil
     end
   end
+
```
KatyTaylor, line 170: Can I request that we remove all mention of 'parent' and 'child' from this? I think it's inherently confusing. 
- I would say that it's intuitive for the multiple samples to be referred to as the 'parents' (because they existed first), and the single, compound sample to be the 'child' (because they arose out of the multiple samples).
- However, you are using the terms the opposite way round here.

Therefore I suggest that we refer to them consistently as 'compound sample' and 'component samples'.

spec/models/sample_spec.rb
```diff
@@ -167,4 +167,79 @@
...
+    it 'fails validation when a compound sample is assigned a compound sample' do
+      other_sample = create(:sample)
+      parent_sample.compound_samples = [other_sample]
+
+      expect(parent_sample.save).to be_falsey
+      expect(parent_sample.errors[:compound_samples]).to include('cannot exist when component samples also exist')
+      expect(parent_sample.errors[:component_samples]).to include('cannot exist when compound samples also exist')
+
+      expect(child_sample1.save).to be_falsey
+      expect(child_sample1.errors[:compound_samples]).to include('cannot themselves have further compound samples')
```
KatyTaylor, line 230: I find these two lines a little bit strange. I would expect that adding an illegal compound sample to `parent_sample` would break the `parent_sample` record, but wouldn't affect the existing sample `child_sample1`.

Happy to be convinced though...?

spec/models/sample_spec.rb
KatyTaylor: Nice set of test cases 👍

spec/requests/api/v2/sample_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+describe 'Samples API', with: :api_v2, cardinaldo: true do
+  context 'when creating a compound sample' do
+    let(:composed_samples) { create_list(:sample, 5) }
```
KatyTaylor, line 6: More consistent to name this `component_samples` ?

db/schema.rb
```diff
@@ -70,7 +70,7 @@
     t.string "name", null: false
     t.string "key", null: false
     t.string "contact", null: false
-    t.text "description", limit: 16777215
+    t.text "description", size: :medium
```
KatyTaylor, line 73: Have you checked whether all these `limit: 16777215` -> `size: :medium` changes will cause a problem (presume it's OK...)

db/migrate/20210930172204_add_support_for_compound_samples.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Add a join table to allow samples to be parents and children of each other.
+class AddSupportForCompoundSamples < ActiveRecord::Migration[6.0]
+  def change
+    create_table 'compound_samples' do |t|
+      t.integer 'parent_id', null: false
+      t.integer 'child_id', null: false
```
KatyTaylor, line 8: See comment in the test file, regarding naming. Thanks!

db/schema.rb
sdjmchattie: This change is automatic by Rails.  As far as I can see from this Stackoverflow: https://stackoverflow.com/a/19389681/772095 `size: medium` means the same thing as `limit: 16777215`

KatyTaylor (changes requested):
Apologies, it seems quite embedded, but I'm quite against the parent-child terminology in this case, as explained in the comments.
Also not quite sure about the nested validation, but up for debate on that one!

db/schema.rb
KatyTaylor: Great :)

db/migrate/20210930172204_add_support_for_compound_samples.rb
sdjmchattie: I'm going to rename the join type from `ComponentSample` to `SampleRelationship` as I don't think it's clear what's meant by having a column in the `compound_samples` table called `compound_sample_id` when there's also an `id` that's the ID of the object with the same name.

spec/models/sample_spec.rb
sdjmchattie: Yes I think perhaps there are too many validations on the Sample in the relationships.  I think (maybe) the only constraint needs to be that a sample doesn't have both component samples and compound samples.  It shouldn't need to validate two layers away from itself.  What do you think?

KatyTaylor: I haven't thought about it hard, but yes my concern was about validating two layers away, that it would be confusing that you could break records other than the one you're touching. I know I've had a pain in Sequencescape before when some existing bad data meant I couldn't do a thing with a record that seemed unrelated to the bad data. I know it's not unrelated in this case, but I still think it might be overkill. I think checking a sample doesn't have both component and compound samples sounds like a good approach.

db/migrate/20210930172204_add_support_for_compound_samples.rb
KatyTaylor: I was imagining the join table would be called the generic name, like SampleLink or SampleRelationship. 
Then the rails relationships could be called compound_samples, and component_samples, and the id fields on the table called `compound_sample_id` and `component_sample_id`.
Is that the same as what you're suggesting?

sdjmchattie: Yes I just did a terrible job of explaining it, but what you described seems to match with what I was trying to 😁

KatyTaylor: haha OK great, glad we agree!

spec/models/sample_compound_component_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    # let variables are lazy loaded and we always want the relationship to exist
+    # even if we don't access the compound sample in the test.
+    before do
+      compound_sample.update(component_samples: [component_sample])
+      component_sample.reload
+    end
+
+    context 'when another sample becomes a compound sample of our compound sample' do
+      it 'fails to validate when the new sample adopts our compound sample as a component' do
```
KatyTaylor, line 17: "adopts" 😆

KatyTaylor (approved):

KatyTaylor:
Looks good to me, thanks for changing the terminology

PR #3359, opened 2021-10-05
Issue3358/dpl 139 add chromium visium library to be limber bespoke chromium – xxxxx templates

### LINKED ISSUE #3358: DPL-139 Add Chromium Visium library to be Limber-Bespoke - Chromium – xxxxx  templates
This represents the sequencescape work required for https://github.com/sanger/limber/issues/823

Original issue:
Chromium Visium library exists but needs to be added to Limber-Bespoke - Chromium – xxxxx  templates

At the moment Bespoke have to manually proceed through Limber and when released the library type is missing from the batch data in MLWH

From srl
"Chromium Visium is a valid library type, we (i.e. NPG) rely on the library_type being correct to determine the appropriate downstream analysis"

Lynda D - ld20 (CellGen SSR)

Referenced in [RT730498](https://rt.sanger.ac.uk/Ticket/Display.html?id=730498)

JamesGlover:
Based of master, so can go into there is we have blocking changes in develop

Fixes #3358 

Adds chromium library type to chromium pipelines (removes it from the non-chromium pipleine)


JamesGlover:
Not sure this is actually correct :( All the talk on RT implies that they want to use the standard bespoke pipelines, in which case this is the incorrect approach.

PR #3377, opened 2021-10-18
Dpl 148 bugfix belongs to models not forwarded to mlwh with rails 6

emrojo:
Closes DPL148

* Changing metadata triggers touch the sample
* Changing well_attribute triggers touch well
* Changing roles triggers touch study


emrojo:
I think there is a problem with the cucumber test but dont happen when I run it in local; so it must be one of those timing problems that we normally have in this project.

PR #3404, opened 2021-11-04
DPL-157 Cellaca file upload tweaks

KatyTaylor:
Tweak Cellaca output file parser to cope with 'NaN' values in the Viability column

### DIFF
--- a/spec/models/parsers/cardinal_pbmc_count_parser_spec.rb
+++ b/spec/models/parsers/cardinal_pbmc_count_parser_spec.rb
@@ -107,6 +107,28 @@ RSpec.describe Parsers::CardinalPbmcCountParser, type: :model do
     end
   end
 
+  context 'when a row has no cells' do
+    # this file has 1 normal row, and a couple of rows with 0 live cells, and NaN viability
+    let(:filename) { Rails.root.join('spec/data/parsers/cardinal_pbmc_count_no_cells.csv') }
+    let(:content) { read_file(filename) }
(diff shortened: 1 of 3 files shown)


spec/models/parsers/cardinal_pbmc_count_parser_spec.rb
```diff
@@ -107,6 +107,28 @@ def read_file(filename)
...
+    it 'will have three qc data entries - one for each row in the file' do
+      expect(parser.qc_data.values.length).to eq(3)
+    end
+
+    it 'will have cell count and viability metrics for the normal row' do
+      expect(parser.qc_data['A4'].keys).to eq(%i[live_cell_count viability])
+    end
+
+    it 'will have just cell count for the rows with 0 cells' do
+      expect(parser.qc_data['A5'].keys).to eq([:live_cell_count])
```
harrietc52, line 126: Could also add, `.not_to include viability` for A5 and E5

harrietc52 (approved):

spec/models/parsers/cardinal_pbmc_count_parser_spec.rb
KatyTaylor: good idea

KatyTaylor: Actually, I think this covers it already - it's checking that the array of keys is [:live_cell_count] - as in, only live_cell_count - so there can't be viabiliity in there

PR #3407, opened 2021-11-05
Issue3395/dpl 111 as a user i would like the ability to display tube racks {heron} {tor 67}

### LINKED ISSUE #3395: DPL-111 - As a user I would like the ability to display tube racks [Heron] [TOR-67]
This represents the sequencescape work required for https://github.com/sanger/limber/issues/794

Original issue:
**User story**
As a user I would like the ability to display tube racks [Heron]

Require the addition of a few attributes to the API.

Iraad / Steve / James

**Acceptance criteria**
- [ ] I would like to be able to view tube racks
- [ ] ...

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
Users want to generate CSV containing the concentration of the source tubes. (I suggested that we could just generate the robot driver files, rather then them needing to copy-paste the information into an intermediate spreadsheet). However, both approaches require a means of grouping together multiple tubes for processing at the same time.

While this doesn’t need to be a tube rack, we:

- Already have the back-end support for this
- Need to add tube racks shortly for other purposes

A meeting about Tapestation for Stef's work led to a conversation about this work for Naomi as well.  See https://github.com/sanger/quanthub/issues/102#issuecomment-906259713

JamesGlover:
Closes #3395 

This adds some behind-the scenes work to enable tube rack viewing in limber.

- Tube racks have have and display comments
- Add a few more attributes and associations to the API
- Refactor some of the validation code
- Used stubbed factories for record testing to improve performance


spec/resources/api/v2/tube_resource_spec.rb
```diff
@@ -6,7 +6,7 @@
 RSpec.describe Api::V2::TubeResource, type: :resource do
   subject(:resource) { described_class.new(resource_model, {}) }
 
-  let(:resource_model) { create :tube, barcode_number: 1 }
+  let(:resource_model) { build_stubbed :tube, barcode_number: 1 }
```
sdjmchattie, line 9: Heh .. well today I learned a new trick!

sdjmchattie (approved):

sdjmchattie:
Worth noting, it turns out it's impossible to satisfy both the push linter and the pull request linter because the version of Prettier on `develop` is different from in this branch and they don't agree with each other.  I've currently got it set up to be happy in the destination `develop` branch.

JamesGlover:
Ahhh, that's interesting. Suppose the trick for solving that kind of thing would be to rebase on HEAD.

sdjmchattie:
> Ahhh, that's interesting. Suppose the trick for solving that kind of thing would be to rebase on HEAD.

Would have been good, except the code on develop didn't appear to satisfy its own prettier version so you'd have all that to deal with as well, perhaps.

PR #3415, opened 2021-11-11
DPL-153 cherrypick minimum volumes

KatyTaylor:
Enforcing robot minimum cherrypicking volumes.

### DIFF
--- a/app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
+++ b/app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
@@ -5,76 +5,83 @@ module Cherrypick::VolumeByNanoGramsPerMicroLitre
   # Used in the cherrypicking process to calculate the relative volumes of source
   # and buffer required to reach the target volume and concentration.
   #
-  # - Ideally will aim for a target well containing volume_required at concentration_required
+  # - Ideally will aim for a target well containing final_volume_desired at final_conc_desired
   # - This will be made by combining material from the source well (picked_volume) and
   #   buffer (buffer_volume)
-  # - Maximum picked_volume is limited by source_volume and volume_required
-  # - If source_volume < volume_required then buffer will be added to make up volume_required
-  #   even if it reduces the target concentration below concentration_required
+  # - Maximum picked_volume is limited by source_volume and final_volume_desired
+  # - If source_volume < final_volume_desired then buffer will be added to make up final_volume_desired
+  #   even if it reduces the target concentration below final_conc_desired
   #
-  # @param volume_required [Float] The volume to aim for in the target well (self)
-  # @param concentration_required [Float] The concentration to aim for in the target well (self)
+  # @param final_volume_desired [Float] The volume to aim for in the target well (self)
+  # @param final_conc_desired [Float] The concentration to aim for in the target well (self)
   # @param source_concentration [Float] The concentration (ng/ul) in the source well
   # @param source_volume [Float] The volume (ul) in the source well (ie. the maximum pick)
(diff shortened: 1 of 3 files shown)


app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
```diff
@@ -5,70 +5,82 @@ module Cherrypick::VolumeByNanoGramsPerMicroLitre
...
-    concentration_required,
+    final_volume_desired,
+    final_conc_desired,
     source_concentration,
     source_volume,
-    robot_minimum_picking_volume = 0.0
+    robot_minimum_pick_vol = 0.0
   )
-    robot_minimum_picking_volume ||= 0
+    robot_minimum_pick_vol ||= 0
```
sdjmchattie, line 34: Should this not be set to a float rather than an integer?  I'm sure it doesn't make much difference, but it's weird to see this setting to `0` when the default for the argument above is `0.0`.

sdjmchattie (approved):
Looks good.  This is very similar in functionality to the calculations I'm adding to Limber for tubes.  I chose to use `target_volume`, `target_molarity`, `minimum_pick` and `source_molarity` for mine.  I don't have to worry about source volume like this one does.  My implementation seems to be a lot less complicated than this though.  https://github.com/sanger/limber/blob/DPL-112-show-tube-qc-data/app/helpers/transfers_helper.rb

app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
KatyTaylor: I didn't notice this. I'm sure Ruby does all sorts of magical conversions. Is this line actually redundant because of the default in the parameter list?

sdjmchattie: Ruby does distinguish between integers and floats.  If you do a division with two integers, you get integer results only.  `3/2 = 1`

I'm not sure about the redundancy.  If someone forcefully passes `nil` as the value for `robot_minimum_pick_vol` does it use the default value or is that only when you don't pass any value at all?

sdjmchattie:
> > Looks good. This is very similar in functionality to the calculations I'm adding to Limber for tubes. I chose to use `target_volume`, `target_molarity`, `minimum_pick` and `source_molarity` for mine. I don't have to worry about source volume like this one does. My implementation seems to be a lot less complicated than this though. https://github.com/sanger/limber/blob/DPL-112-show-tube-qc-data/app/helpers/transfers_helper.rb
> 
> Thanks for reviewing. `target` is more succinct than my version. I like having `robot` in there as it explains why it's necessary. Do you have infinite source volume?! This is quite complicated, it seems over complicated maybe but I checked it in detail with Danni and she approves.

We've been told there will always be enough volume for this.  I expect that might change in future, but in our case it's not worth supporting an edge case you're told won't happen right now.

app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
KatyTaylor: You're right, it's not redundant - there are a few tests that specifically pass in `nil`.
I'm going to go with "if it ain't broke, don't fix it" :)

sdjmchattie: I know you've merged, but as a rule of thumb I would say that if you see something that you're fairly certain isn't right and when you fix it the tests still pass, change it.  It's subtle things like this that maybe aren't causing problems today, but down the line someone makes a change where there is a difference between having a float and having and integer and then the bugs can be quite hard to identify.

Issue #3292, opened 2021-08-19
DPL-120 Print plate and tube labels to Squix from SS


rl15:
**User story**
As representative for SSR (Rich C) we would like to be able to pint plate and tube labels on the new Squix printers from SS as the Toshiba TEC are increasing unreliable.

**Who are the primary contacts for this story**
Rich C (SSR)
Katy T
James G

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] Test what happens today when try to pint from SS to one of teh Squi printers (In UAT)
- [ ] List all the templates that are required [Apps and printing templates](https://docs.google.com/spreadsheets/d/14DU8zbAMrk3EEUceUY5pt7m6hLp7343MHsWQuDyq-aU/edit?usp=sharing)
- [ ] Deprecate all the backlog items that mention Squix or barcodes

**Additional context**

rl15:
List of backlog items that mention Squix or barcodes

https://github.com/search?p=2&q=org%3Asanger+squix&type=Issues

Of note 
GPL-906
GPL-831
GPL-832

harrietc52:
LabWhere squix [diagram](https://lucid.app/lucidchart/004c0b67-4273-41dd-bbf6-a545edd53f20/edit?page=0_0#)

harrietc52:
Priority: Tube, Plate 96 (384 not needed yet)

http://sqtest.internal.sanger.ac.uk  (Jamie’s printer) lives in the SSR Pod, Level 2, Ogilvie.
Currently set up for 96 plate labels (can swap out to tube)
But hopefully soon there will be both tube and plate printers available

harrietc52:
Add printers to SS
`   <% printer_list = BarcodePrinter.alphabetical.includes(:barcode_printer_type).all.group_by {|printer| "#{printer.barcode_printer_type.name} (#{printer.active? ? 'Active' : 'Inactive'})" } %>
`

harrietc52:
[Sequencescape printing rationalisation](https://ssg-confluence.internal.sanger.ac.uk/display/PSD/Sequencescape+printing+rationalisation)

stevieing:
Tasks:

- [ ] get sqtest re-added to the ACL
- [ ] test the sqtest printer
- [ ]

harrietc52:
On Deployment

Script for running in SS
```
# Create new BarcodePrinterTypes, with the correct label template names or update existing? 
# sqsc_1dtube_label_template.yml.erb for tube
[...]

PR #3418, opened 2021-11-15
printed to sqtest on pmb uat

harrietc52:
Closes https://github.com/sanger/sequencescape/issues/3292

* Upgrade SS to use PMB v2

Coupled with https://github.com/sanger/deployment/pull/94


spec/factories/barcode_printers.rb
```diff
@@ -4,5 +4,6 @@
   factory(:barcode_printer) do
     sequence(:name) { |i| "a#{i}bc" }
     association(:barcode_printer_type, factory: :plate_barcode_printer_type)
+    # print_service { "SPrint" }
```
stevieing, line 7: Can this comment be removed?

stevieing (changes requested):
All good. Probably best to get James' view on it before we merge. Looks like there are quite a few failing cucumber tests.

lib/label_printer/label_printer/label.rb
```diff
@@ -24,7 +21,7 @@ def create_labels
       end
 
       def label(asset)
-        { main_label: create_label(asset) }
```
JamesGlover, line 27: This is probably an unnecessary level of abstraction now. I think we'd probably be better of removing it, and either calling `create_label` directly, or alternatively renaming all the `create_label` implementations to `label`. The later is more work, but I think is a better method name. (As we're not really creating anything)

lib/label_printer/label_printer/label.rb
```diff
@@ -24,7 +21,7 @@ def create_labels
       end
```
JamesGlover, line 21: I know this wasn't anything you touched, but the big fan of Enumerable in me:

```ruby
def create_labels
  assets.flat_map do |asset|
    Array.new(count, label(asset))
  end
end
```

Its just personal preference though. (Its probably also a bit faster, but not to a sufficient enough degree that it would be worth considering on that basis)

Actually interested to know if you think it's an improvement or not.

lib/label_printer/label_printer/multiple_double_labels.rb
```diff
@@ -13,8 +13,9 @@ def double_label(asset)
         [label(asset), extra_label(asset)]
       end
 
+      # TODO: remove
       def extra_label(asset)
```
JamesGlover, line 17: Yep, similar to the label/create label stuff I mentioned earlier.

I was wondering if you could actually keep these around, and use them to DRY out the `label_name: 'main_label'` and `label_name: 'extra_label'` stuff. There's obviously a bit of repetition there, but I'm not sure if this would be an improvement or not.

lib/label_printer/label_printer/print_job.rb
```diff
@@ -32,27 +31,26 @@ def execute # rubocop:todo Metrics/MethodLength
...
+      @build_attributes ||=
+        {
+          printer_name: printer_name,
+          label_template_name: label_template_name,
+          labels: labels_attribute
+        }
     end
 
+    # TODO
+    # change name of to_h to to_a
```
JamesGlover, line 43: Or see the suggestion of just using 'labels' instead.

app/models/barcode_printer_type1_d_tube.rb
```diff
@@ -4,6 +4,7 @@
 #
 class BarcodePrinterType1DTube < BarcodePrinterType
   def self.first
+    # Update this?
```
JamesGlover, line 7: Are we using new template names? If so then yep, update them here, otherwise remove this comment.

JamesGlover (changes requested):
Hey,

Mostly looking good, but you've got a few TODOs kicking around which should probably either been done, or cleared up.

lib/label_printer/label_printer/label.rb
harrietc52: I've removed the `label` method, and renamed `create_label` to `build_label`

lib/label_printer/label_printer/label.rb
```diff
@@ -4,14 +4,11 @@ module Label
     module MultipleLabels # rubocop:todo Style/Documentation
       attr_accessor :count
 
+      # TODO: rename
       def to_h
```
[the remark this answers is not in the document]
harrietc52: Have tried to do think, a few `to_h` methods also are living in other classes in `LabelPrinter` module

PR #3419, opened 2021-11-15
Added new plate barcode for the Lemington Spa Lighthouse

Sangeetha-Bheeman:
Closes #

*
*
* ...

### DIFF
--- a/app/models/barcode.rb
+++ b/app/models/barcode.rb
@@ -57,7 +57,8 @@ class Barcode < ApplicationRecord
          newcastle: 36,
          brants_bridge_v2: 37,
          uk_biocentre_v7: 38,
-         east_london_genes_and_health: 39
+         east_london_genes_and_health: 39,
+         lemington_spa_v2: 40,
        }
 
   # Barcode formats which may be submitted via sample manifests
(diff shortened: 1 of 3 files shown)


app/models/barcode.rb
```diff
@@ -97,6 +98,7 @@ class Barcode < ApplicationRecord
     brants_bridge_v2
     uk_biocentre_v7
     east_london_genes_and_health
+    lemington_spa_v2
```
sdjmchattie, line 101: I think it's spelled `leamington` rather than `lemington`.  https://www.warwickdc.gov.uk/info/20114/health_and_medical_care/1646/rosalind_franklin_laboratory_in_royal_leamington_spa

There are a few mentions in this pull request.

sdjmchattie (changes requested):

app/models/barcode.rb
Sangeetha-Bheeman: I took the spelling from the issue. I will change it.

sdjmchattie: Sure thing.  I did see that's what had happened after I posted my review 😄 .  I only know it's spelled differently because I used to live in the next town across.

sdjmchattie (approved):

KatyTaylor (approved):

PR #3420, opened 2021-11-16
prettier

harrietc52:


stevieing (approved):
You have got some cop failures. I have no idea what this Unchanged files with check annotations is all about??

JamesGlover:
This looks like tis is the older version of prettier. We updated recently to prettier 2.

JamesGlover:
@sdjmchattie Pointed out something about a similar linting failure I ran into, which might be relevant here. In this case the linting tests on the push and pull were running with two different versions of prettier, and it was impossible to satisfy both. In that case he fixed up for the newer version (ie. the pull tests) and we merged that in. However you could also rebase on top of develop, and then everything would be nice and consistent.

PR #3425, opened 2021-11-19
Move yard group

JamesGlover:
Yard should not be required for linting. It should be in the development group only.


JamesGlover:
I can't read. deployment != development

PR #3426, opened 2021-11-22
Dpl 074 1 consolidate samples

andrewsparkes:
For story https://github.com/sanger/limber/issues/773


app/models/specific_tube_creation.rb
```diff
@@ -64,13 +64,44 @@ def connect_parent_and_children
...
+    raise "Foreign Barcode: #{foreign_barcode} is already in use!"
+  end
+
+  def add_foreign_barcode_to_tube(tube, foreign_barcode)
+    # determine the format of the foreign barcode
+    foreign_barcode_format = determine_foreign_barcode_format(foreign_barcode)
+
+    raise "Cannot determine format for foreign barcode #{foreign_barcode}" if foreign_barcode_format.blank?
+
+    # check for uniqueness
```
KatyTaylor, line 83: I think you could remove most of the comments you've added here - the methods are v well named so they are just duplicating information.

config/default_records/plate_purposes/007_cardinal_purposes.wip.yml
```diff
@@ -6,7 +6,3 @@ LCA Blood Array:
   type: PlatePurpose::Input
   stock_plate: true
   cherrypickable_target: false
-LCA Blood Bank:
-  type: PlatePurpose::Input
-  stock_plate: true
-  cherrypickable_target: false
```
KatyTaylor, line 12: I notice that here we're getting rid of the 'LCA Blood Bank' as the input plate to the cell banking pipeline. Do you remember we talked about making it symmetrical - so that the submission worked the same way, starting on the LCA Blood Vac tube, for the library prep pipeline? Have you tested that path?

config/default_records/request_types/006_limber_cardinal.wip.yml
```diff
@@ -4,25 +4,23 @@
...
+  request_class_name: CustomerRequest
   for_multiplexing: false
   billable: true
-  product_line_name: Illumina-Htp
-  library_types:
-    - Cardinal Banking
+  product_line_name: Cardinal
   acceptable_plate_purposes:
-    - LCA Blood Bank
+    - LCA Blood Vac
```
KatyTaylor, line 26: When we deploy this to UAT, we'll have to remember to delete the request types beforehand, so these actually get created.

config/default_records/submission_templates/001_cardinal_submission_templates.wip.yml
```diff
@@ -3,13 +3,15 @@
 # the real release.
 ---
 Limber - Cardinal:
+  name: "limber_cardinal"
```
KatyTaylor, line 6: I think maybe these 'names' shouldn't be here - perhaps a hangover from the issue where the submission template records were being created with 'limber_cardinal' as the name rather than 'Limber - Cardinal' as the name? I think you changed your local database rather than doing the fix like this - https://github.com/sanger/sequencescape/pull/3411/files

KatyTaylor (changes requested):
Looks good, just a couple of comments - the most important I think being in 001_cardinal_submission_templates.wip.yml

app/models/request_type.rb
```diff
@@ -60,8 +60,8 @@ class RequestTypePlatePurpose < ApplicationRecord # rubocop:todo Style/Documenta
 
   # Defines the acceptable plate purposes or the request type.  Essentially this is used to limit the
   # cherrypick plate types when going into pulldown to the correct list.
-  has_many :plate_purposes, class_name: 'RequestType::RequestTypePlatePurpose'
-  has_many :acceptable_plate_purposes, through: :plate_purposes, source: :plate_purpose
+  has_many :purposes, class_name: 'RequestType::RequestTypePlatePurpose'
+  has_many :acceptable_plate_purposes, through: :purposes, source: :purpose
```
KatyTaylor, line 64: These changes are needed in order to add the Cardinal request types, because they come from a tube rather than a plate. Without these changes, the request type record loader errors on `request_type.acceptable_plate_purposes << Purpose.find_by!(name: name)`

app/models/specific_tube_creation.rb
```diff
@@ -64,13 +64,41 @@ def connect_parent_and_children
   end
   private :connect_parent_and_children
 
+  def determine_foreign_barcode_format(foreign_barcode)
+    Barcode.matching_barcode_format(foreign_barcode)
```
JamesGlover, line 68: Would it make more sense to explicitly specify the barcode format? Rather than trying to auto-detect it. Presumably we know what kind of tubes we're registering.

app/models/specific_tube_creation.rb
```diff
@@ -64,13 +64,41 @@ def connect_parent_and_children
...
+  end
+
+  def check_foreign_barcode_unique(foreign_barcode_format, foreign_barcode)
+    return unless Barcode.exists_for_format?(foreign_barcode_format, foreign_barcode)
+
+    raise "Foreign Barcode: #{foreign_barcode} is already in use!"
```
JamesGlover, line 74: This should probably be validation, rather than an exception. Also, it may be better to just use .valid? on barcode, and extract the errors from that, rather than duplicating logic all over the place.

app/models/specific_tube_creation.rb
```diff
@@ -64,13 +64,41 @@ def connect_parent_and_children
...
+  end
+
+  def add_foreign_barcode_to_tube(tube, foreign_barcode)
+    foreign_barcode_format = determine_foreign_barcode_format(foreign_barcode)
+
+    raise "Cannot determine format for foreign barcode #{foreign_barcode}" if foreign_barcode_format.blank?
```
JamesGlover, line 80: More inclined to suggest validating everything, before attempting to do anything, rather than throwing an exception mid flow. I'm assuming transactions wrap all of this, so it would be rolled back, but...

app/models/specific_tube_creation.rb
```diff
@@ -64,13 +64,44 @@ def connect_parent_and_children
   end
   private :connect_parent_and_children
 
+  def determine_foreign_barcode_format(foreign_barcode)
```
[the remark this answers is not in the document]
JamesGlover: I kind of agree, not sure that wrapping this method provides much value.

app/models/specific_tube_creation.rb
```diff
@@ -64,13 +64,41 @@ def connect_parent_and_children
...
     self.children =
       child_purposes.each_with_index.map do |child_purpose, index|
         # For each tube purpose listed in the child_purposes array
         # create a tube via the tube purpose factory, passing in our
         # custom attributes.
-        child_purpose.create!(tube_attributes[index])
+        tube_detail = tube_attributes[index].symbolize_keys
+
+        # extract foreign barcode for use after the create if one has been set
+        foreign_barcode = tube_detail.delete(:foreign_barcode)
```
JamesGlover, line 97: This may make more sense as a simple setter on tube, rather than handling the barcode logic here.

JamesGlover (changes requested):
I'm not quite so sure about the `specific_tube_creation.rb` stuff:

1) Exceptions are being used for what feels more like validations.
2) I'm not really convinced the barcode logic belongs in specific_tube_creations at all, it feels like its more a property of the tube

I also wonder if you should be explicit about what barcode format you're expecting. Heron has results in such an explosion of barcode formats that I could easily see the auto-detection being error prone. (I think we need to revist that as a whole, we talked about it at one point, but never really made any decisions)

KatyTaylor:
The diff seems a bit dodgy - showing things as changes that shouldn't be - so I'm going to try closing and reopening the PR

PR #3428, opened 2021-11-23
Rollback changes

emrojo:
Closes #

*
*
* ...


db/schema.rb
```diff
@@ -56,10 +56,9 @@
     t.boolean "suboptimal", default: false, null: false
     t.bigint "primer_panel_id"
     t.integer "request_id"
-    t.integer "tag_depth", default: 1
     t.index ["library_id"], name: "index_aliquots_on_library_id"
     t.index ["primer_panel_id"], name: "index_aliquots_on_primer_panel_id"
-    t.index ["receptacle_id", "tag_id", "tag2_id", "tag_depth"], name: "aliquot_tag_tag2_and_tag_depth_are_unique_within_receptacle", unique: true
+    t.index ["receptacle_id", "tag_id", "tag2_id"], name: "aliquot_tags_and_tag2s_are_unique_within_receptacle", unique: true
```
stevieing, line 61: Do you need to remove tag depth? Does leaving it in have any side effects?

stevieing (approved):
All looks good to me. 1 question about the migration. What is the plan? Do you want to merge this in as is and then uncomment it when ready? This is pretty much a feature flag!

sdjmchattie (approved):
Does what it says on the tin.  For future reference (and because it will be useful practice for when you have feature flags) it would probably have been cleaner to create this branch just hiding the feature and to also create a second branch that adds it back in, instead of having to put TODOs throughout the code.  It's also a good idea to label your TODOs if you include them so they're easier to search for to change this back again.

db/schema.rb
emrojo: The index name was hardcoded inside SS code originally, and that's why it has been changed in some files.
Tag_depth is removed in the schema but it has never been applied in production; the migration that added this column has been removed. We have to add again that migration when rolling back.

Issue #3383, opened 2021-10-20
DPL-067-x Create Compound sample on Batch creation (M)


harrietc52:
**User story**
We have moved out the creation of the compound sample on batch creation to a separate story so we can test the earlier Cardinal pipeline, without being dependent on this. 

See the original story [DPL-067](https://github.com/sanger/limber/issues/759), particularly [this message](https://github.com/sanger/limber/issues/759#issuecomment-947738586) for more information

**Who are the primary contacts for this story**
Harry
Eduardo

**Acceptance criteria**
To be considered successful the solution must allow:

- [ ] Update MLWH indexes to allow tag clashes though the use of `tag_depth`

On Submission Batch creation, then create the compound sample:
- [x] A new compound sample is created, from the component samples in the well, forming a "pool". 
- [x] Add the compound sample to the lane on Batch creation? Update the samples aliquot with study, library, aliquot and request ?? 
- [x] This compound sample should be added to the SS sample table  _- KT - remember to link it to the appropriate Study_ 
- [ ] Update "Sample Links" table in SS, with two columns, both foreign keys to the sample table.  A new record should be added to the SS Sample Links table,  linking the compound sample to the source samples
- [ ] This compound sample should be added to the MLWH sample table
- [ ] Update "Sample Links" table in MLWH , with two columns, both foreign keys to the sample table. A new record should be added to the MLWH Sample Links table,  linking the compound sample to the source samples

Reporting:
- [ ] Check the compound sample gets inserted into iseq_flowcell on batch release (note: if the sample isn’t in the mlwh.sample table the batch data will not be entered into the iseq_flowcell table, Neil)
- [ ] Reporting: In MLWH, from a iseq_flowcell record, the compound sample can be accessed, along with the source samples contained in the compound sample, and for each of those source samples, which study they belong to (via stock resource?)

PR #3435, opened 2021-11-26
Compound sample creation fix

KatyTaylor:
Issue discovered in UAT - compound sample creation, on sequencing batch creation (#3383) was expecting just one compound sample. We need to handle the creation of multiple compound samples.

### DIFF
--- a/app/models/request/sample_compound_aliquot_transfer.rb
+++ b/app/models/request/sample_compound_aliquot_transfer.rb
@@ -41,6 +41,8 @@ module Request::SampleCompoundAliquotTransfer
   def transfer_into_compound_sample_aliquot(source_aliquots)
     samples = source_aliquots.map(&:sample)
 
+    # Check that the component samples in the compound sample will be able to be distinguished -
+    # this is represented by them all having a unique 'tag_depth'
     if source_aliquots.pluck(:tag_depth).uniq.count != source_aliquots.size
       raise "Cannot create compound sample from following samples due to duplicate 'tag depth': #{samples.map(&:name)}"
     end
--- a/app/models/request/sample_compound_aliquot_transfer.rb
+++ b/app/models/request/sample_compound_aliquot_transfer.rb
@@ -1,21 +1,23 @@
 # frozen_string_literal: true
 
-# Module to provide support to handle creation of a compound sample
+# Module to provide support to handle creation of compound samples
 # from the list of samples at the source.
 #
 # Assumptions:
 #  - This module will be included in a Request class
 module Request::SampleCompoundAliquotTransfer
-  # Indicates if a compound sample creation is needed, if any of the aliquots
-  # at the source has a different tag_depth defined
+  # Indicates if a compound sample creation is needed:
+  # If any of the source aliquots share the same tag1 and tag2, but have distinct tag_depths
   def compound_samples_needed?
     return false if asset.aliquots.count == 1
 
     _tag_clash?
   end
 
-  # Creates a sample in a single aliquot at destination as a compound sample
-  # where the component samples are all samples at source
+  # Groups the source aliquots by their tag1 and tag2 combination
+  # For each of these groups, create a compound sample,
+  # to hide the fact from Illumina deplexing NPG software that there are
+  # actually multiple samples with this combination.
   def transfer_aliquots_into_compound_sample_aliquots
     _aliquots_by_tags_combination.each do |_tags_combo, aliquot_list|
       transfer_into_compound_sample_aliquot(aliquot_list)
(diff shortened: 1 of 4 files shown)


app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -10,25 +10,53 @@ module Request::SampleCompoundAliquotTransfer
   # at the source has a different tag_depth defined
   def compound_samples_needed?
     return false if asset.aliquots.count == 1
-    (asset.aliquots.pluck(:tag_depth).uniq.count == asset.aliquots.count)
+
+    _tag_clash?
   end
 
   # Creates a sample in a single aliquot at destination as a compound sample
```
harrietc52, line 17: Update comment here

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -10,25 +10,53 @@ module Request::SampleCompoundAliquotTransfer
...
+      transfer_into_compound_sample_aliquot(aliquot_list)
+    end
+  end
+
+  private
+
+  def transfer_into_compound_sample_aliquot(source_aliquots)
+    samples = source_aliquots.map(&:sample)
+
+    if source_aliquots.pluck(:tag_depth).uniq.count != source_aliquots.size
```
harrietc52, line 30: This could be renamed the tag_clash

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -10,25 +10,53 @@ module Request::SampleCompoundAliquotTransfer
...
+        aliquot.tag2_id = tag2_id
         aliquot.library_type = _default_library_type
         aliquot.study_id = _default_compound_study.id
         aliquot.project_id = _default_compound_project_id
         aliquot.save
       end
   end
 
-  private
+  def _tag_clash?
```
harrietc52, line 53: Maybe change this name? The method isn't really checking if there is a tag_clash I don't think? But the size of the aliquots for one tag combo?

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -10,25 +10,53 @@ module Request::SampleCompoundAliquotTransfer
...
+    if source_aliquots.pluck(:tag_depth).uniq.count != source_aliquots.size
+      raise "Cannot create compound sample from following samples due to duplicate 'tag depth': #{samples.map(&:name)}"
+    end
+
+    compound_sample = _create_compound_sample(_default_compound_study, samples)
+
+    add_aliquot(compound_sample, source_aliquots.first.tag_id, source_aliquots.first.tag2_id)
+  end
+
+  def add_aliquot(sample, tag_id, tag2_id)
```
harrietc52, line 39: I guess method definition could accept the source_aliquots, or the source_aliquots.first, instead of passing in the tags separately, but I think passing them in is explicit and probably better

spec/models/request/sample_compound_aliquot_transfer_spec.rb
```diff
@@ -51,7 +51,7 @@
     end
   end
 
-  describe '#transfer_aliquots_into_compound_sample_aliquot' do
+  describe '#transfer_aliquots_into_compound_sample_aliquots' do
```
harrietc52, line 54: Does this test need anything adding to represent the creation of multiple compound sample aliquots

spec/models/sequencing_request_spec.rb
```diff
@@ -139,13 +139,19 @@
...
+    let(:tags) { create_list :tag, 4 }
 
-    context 'when no tag_depth is defined in the source asset aliquots' do
-      let(:aliquot1) { create :aliquot, sample: samples[0], tag_id: 1, study: study }
-      let(:aliquot2) { create :aliquot, sample: samples[1], tag_id: 2, study: study }
+    before do
+      expect(tags.map(&:oligo).uniq.size == tags.size)
+    end
+
+    context 'when compound samples are not necessary because each aliquot has a unique tag combination' do
```
harrietc52, line 152: nice 👍

harrietc52 (commented):
Looks good, some minor thoughts
I havent quite finished going through the tests yet - will come back to it. (got my jab..)

app/models/request/sample_compound_aliquot_transfer.rb
KatyTaylor: Yeah it kind of is, because if there are 2 aliquots with the same tag combination, then the tags clash. 
But Eduardo found a similar bit of code confusing, so I will have a go at renaming the methods 👍

spec/models/request/sample_compound_aliquot_transfer_spec.rb
KatyTaylor: I think that's probably covered in `sequencing_request_spec` by 'when there are two tag combinations'? Don't want to create too much duplication

app/models/request/sample_compound_aliquot_transfer.rb
KatyTaylor: added a comment to explain what I'm checking for

PR #3436, opened 2021-11-29
Dpl 120 create printers in sequencescape

stevieing:
Closes #

*
*
* ...


harrietc52 (commented):
I think the `printer_type` that is being added here is a duplication of `print_service`. Where print_service PMB is the equivalent of printer_type toshiba, and print_service SPrint, is the equivalent of printer_type squix. 
With that I think printer_type isn't needed but print_service could be updated/renamed/values renamed?

stevieing:
> I think the `printer_type` that is being added here is a duplication of `print_service`. Where `print_service: PMB` is the equivalent of `printer_type: toshiba`, and `print_service: SPrint`, is the equivalent of `printer_type: squix`. Unless I'm missing somthing, I think that I think printer_type isn't needed but print_service could be updated/renamed/values renamed?

The printer type is required to create the printer in pmb. I would prefer not to use it but there was no other choice. The print service, I believe, is used by limber to identify which print service to use.

The ideal would be to not have either but unfortunately we need both.

The alternative would be to either use printer type in limber or move limber over to v2. It would be even better to not have printer in Sequencescape at all but that would be far more work.

PR #3440, opened 2021-12-01
Dpl 177 cancel submission bug

Sangeetha-Bheeman:
Closes #

*
*
* ...


stevieing (approved):

sdjmchattie (approved):

app/models/ability/administrator.rb
```diff
@@ -85,6 +85,7 @@ def grant_privileges # rubocop:todo Metrics/AbcSize
 
     # Edit and delete submissions
     can :manage, Submission
+    can :cancel, Submission
   end
```
emrojo, line 89: From reading the ticket content:
```
We should probably add can :cancel, Submission to both the lab_manager and sample_management_manager files...
```
I think to add this we would need to add the same change to app/models/ability/lab_manager and sample_management_manager

emrojo (changes requested):

sdjmchattie (approved):

emrojo (approved):

PR #3442, opened 2021-12-02
Setup building of Docker image

emrojo:
Sets up the building of a docker image for development environment


.github/workflows/automated_docker_release.yml
```diff
@@ -0,0 +1,51 @@
...
+    branches:
+      - master
+      - develop
+
+env:
+  IMAGE_NAME: ${{ github.repository }}/${{ github.event.repository.name }}
+
+jobs:
+  build:
+    name: Test
```
sdjmchattie, line 14: This doesn't seem like the right name for this job

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Install required software:
+#  - net-tools: to run ping and other networking tools
+#  - build-essential: to have a compiling environment for building gems
+#  - curl: for healthcheck
+#  - netcat: for wait for connection to database
+#  - nodejs, yarn, git, default-libmysqlclient-dev and graphviz are rails gems dependencies
+RUN apt-get update && apt-get install -y \
+  net-tools build-essential curl netcat wget vim \
+  nodejs yarn git default-libmysqlclient-dev npm graphviz
```
sdjmchattie, line 11: I know it's not what we've always done, but would it be worth splitting these onto separate lines?  I've seen on more than one occasion where duplicates end up in these lists which aren't a problem until you're removing it again in the future and can't understand why removing the one duplicate you spotted doesn't do what you thought it was going to do.  It also gives us the opportunity to put them in A-Z order to help people assess whether the one they want to be there is in fact there, also reducing the chances of duplication.

README.md
```diff
@@ -81,7 +81,38 @@ The following tools are required for development:
   Sequencescape (currently 5.7) use [this](https://gist.github.com/operatino/392614486ce4421063b9dece4dfe6c21)
   helpful link.
 
-## Getting started
+## Getting started (using Docker)
+
+To set up a local development environment in Docker, you have to build a new Docker image for
+Sequencescapek. start a stack of services that include a mysql database, and reset
```
sdjmchattie, line 87:
```suggestion
Sequencescape. start a stack of services that include a mysql database, and reset
```

README.md
```diff
@@ -81,7 +81,38 @@ The following tools are required for development:
...
+this database contents. You can do all together by running the command:
+
+```
+RESET_DATABASE="true" docker-compose up
```
sdjmchattie, line 91: I don't think the quotes are necessary in this environment variable.  The command line will interpret everything up to the first whitespace as a string to assign to the variable.

```suggestion
RESET_DATABASE=true docker-compose up
```

.github/workflows/check_docker_build.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build image
+        run: >-
+          docker-compose build
+
+      - name: Start stack and wait for all healthy
+        run: |
+          RESET_DATABASE="true" docker-compose -f docker-compose.yml up &
```
sdjmchattie, line 22:
```suggestion
          RESET_DATABASE=true docker-compose -f docker-compose.yml up &
```

README.md
```diff
@@ -81,7 +81,38 @@ The following tools are required for development:
...
+```
+
+Optionally, if this is not ther first time you start the app, you may not want to reset the
```
sdjmchattie, line 94:
```suggestion
Optionally, if this is not the first time you start the app, you may not want to reset the
```

README.md
```diff
@@ -81,7 +81,38 @@ The following tools are required for development:
...
   helpful link.
 
-## Getting started
+## Getting started (using Docker)
+
+To set up a local development environment in Docker, you have to build a new Docker image for
+Sequencescapek. start a stack of services that include a mysql database, and reset
+this database contents. You can do all together by running the command:
+
+```
```
sdjmchattie, line 90:
```suggestion
```shell
```

README.md
```diff
@@ -81,7 +81,38 @@ The following tools are required for development:
...
+RESET_DATABASE="true" docker-compose up
+```
+
+Optionally, if this is not ther first time you start the app, you may not want to reset the
+database, and you can run this command instead:
+
+```
```
sdjmchattie, line 97:
```suggestion
```shell
```

README.md
```diff
@@ -81,7 +81,38 @@ The following tools are required for development:
...
+
+With this we should have started Sequencescape server and all required services. You should be
+able to access Sequencescape by going to <http://localhost:3000> and log in with
+username and password admin/admin.
+
+**NOTE** If you ever need to recreate the image built on first start (because you made modifications
+to the Dockerfile file) you can run a building process with:
+
+```
```
sdjmchattie, line 108:
```suggestion
```shell
```

config/database.yml
```diff
@@ -19,7 +19,7 @@ mysql: &MYSQL
   pool: 5
   timeout: 5000
   reaping_frequency: 600
-  host: 127.0.0.1
+  host: <%= ENV.fetch('DBHOST', '127.0.0.1') %>
```
sdjmchattie, line 22: Nice 👍

docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+      retries: 5
+
+  mysql_server:
+    image: mysql:5.7
+    environment:
+      - MYSQL_ALLOW_EMPTY_PASSWORD="yes"
+    healthcheck:
+      test: "/etc/init.d/mysql status"
+      interval: 30s
+      retries: 120
```
sdjmchattie, line 42: This MySQL server might conflict with the one you might be running for Lighthouse and Crawler.  We might have to think about this more carefully, or at least not start this server on a local dev machine unless a flag is set, for example.

sdjmchattie (changes requested):

stevieing (changes requested):
All good. Just a question about the Aker config.

.github/workflows/automated_docker_release.yml
emrojo: Yes, I'll fix it

Dockerfile
emrojo: Yes, good idea, I'll do it alphabetical multiline

docker-compose.yml
emrojo: This would be like an option that can be used along with native;  I was thinking on leaving the decision of what would be needed to later to see what it would be the best hybrid approach for us; about doing this this way I thought this had the advantage that it will work independently to what anybody has in local; you should be able to run a full complete Sequencescape even if your local version of mysql is broken (of course as long as Docker works), so to have a full safe option to work around the problems with native installations. But this is an initial config, I think it can be modified in other stories/changes to adapt to what is better for us.

emrojo: Yes, I dont have a global solution for it. Is the same problem if you want to connect the Rabbitmq from Sequencescape with the unified_warehouse consumers, for example.  We don't have a common configuration to have all projects to run together, but this is very early stage.
The only temporal option I can think of is to have different profiles of docker-compose and so you can use the one is better for your situation, but keeping a default one that should work with the easier setup. This solution for the moment has 2 profiles: 
1) docker-compose up : runs docker-compose.yml + docker-compose.override.yml so it will start everything in local
2) docker-compose -f docker-compose.yml up : runs only docker-compose.yml so it will work in github

So it could be added another like docker-compose-local.yml that could be run like: ```docker-compose -f docker-compose-localdev.yml up```, would it work for you?

sdjmchattie: According to Docker docs: https://docs.docker.com/compose/extends/

They say that you could have a production file `docker-compose.prod.yml` which you could use in the build config for the GitHub workflow.  Ideally that would have the MySQL service in it while the un-prefixed one does not have MySQL.  Then the developer can choose to add the production file to their `docker-compose up` if they want to have the MySQL, but most of the time it is expected you have a MySQL instance already available.

emrojo: I agree but I would do it the other way. I would keep the setup as easier as possible to have all running in one command, so anybody can run and test Sequencescape in the easier way possible (if anybody wants to) and then have a separate local configuration for custom development for working with any custom flags there if they are needed.

emrojo: docker-compose up -> with mysql
docker-compose -f <localtemplate> up -> without mysql

sdjmchattie: If you can do that, that also seems fine.  The problem I was having with docker-compose create your MySQL was that it kept destroying the container and the data associated with it, so you can unexpectedly lose all your database contents that you've been setting up to test an issue because of one quick slip.  That's why I made it a completely separate docker-compose for just the database dependencies in Lighthouse and Crawler.

emrojo: In these files ```docker-compose.override.yml``` includes a volume for mysql that refers to local folder so you can destroy the container and create it again but the data remains there.
Ok, I'll add the localdev template.

sdjmchattie: Having discussed offline, there has been some misunderstanding on my part.  The port for MySQL is not being exposed and therefore won't cause an issue when trying to run alongside a local installation or another Docker container for MySQL.  I would revert the change made for the additional docker-compose file.  If it turns out that it's hard to access MySQL as a developer when the port isn't exposed by default, we should reconsider, but for now I think this is fine.

PR #3443, opened 2021-12-02
Remove mini racer

JamesGlover:
- Remove mini-racer dependency
- Add node actions to other GH actions

Miniracer is a requirement of execjs, which is used in a few places in the rails js pipeline (mostly the sprockets stuff)
It is able to use a few js engines, including node.

When originally we adopted mini-racer, we weren't using node, so it seemed the simpler dependency. However now we have a node-install available for the other bits of the asset compilation pipeline, so it makes sense to use that.

This change was driven by breaking changes in mini-race 0.5

### DIFF
--- a/spec/features/perform_a_tag_substitution_spec.rb
+++ b/spec/features/perform_a_tag_substitution_spec.rb
@@ -78,7 +78,10 @@ describe 'Perform a tag substitution', js: true do
         "#{sample_a_orig_tag.map_id} - #{sample_a_orig_tag.oligo}",
         from: 'tag_substitution[substitutions][][substitute_tag_id]'
       )
+
+    scroll_to(find_button('Substitute Tags'))
     click_button 'Substitute Tags'
+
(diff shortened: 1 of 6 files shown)


spec/features/perform_a_tag_substitution_spec.rb
```diff
@@ -115,13 +118,14 @@
...
     click_button 'Substitute Tags'
+
     expect(page).to have_content(lane.name)
     expect(page).to have_content 'Your tag substitution could not be performed.'
-    expect(
-      page
-      # rubocop:todo Layout/LineLength
-    ).to have_content "Tag pair #{sample_b_orig_tag.oligo}-#{sample_b_orig_tag2.oligo} features multiple times in the pool."
-    # rubocop:enable Layout/LineLength
+    expect(page).to have_content(
```
JamesGlover, line 127: Just fixed a linelength cop while I was here.

KatyTaylor (approved):

PR #3446, opened 2021-12-02
DPL-074-1 consolidate samples

KatyTaylor:


app/uat_actions/uat_actions/generate_tubes.rb
```diff
@@ -46,6 +51,20 @@ def perform
...
+        # using a set prefix and a subset of the machine barcode
+        prefix = 'SA'
+        suffix = tube.machine_barcode[-8..]
+
+        foreign_barcode = prefix + suffix
+
+        tube.barcodes << Barcode.new(format: foreign_barcode_format, barcode: foreign_barcode)
+      end
+
+      # set the tube primary barcode on the report
```
JamesGlover, line 67: Pulling this out into a separate method would probably appease rubocop

app/models/tube.rb
```diff
@@ -84,19 +84,32 @@ def after_comment_addition(comment)
...
+    tube.foreign_barcode = foreign_barcode if foreign_barcode
+    tube.reload
   end
 end
 
+def extract_barcode(args, attributes)
+  barcode = args.first || attributes.delete(:barcode)
+  prefix = attributes.delete(:barcode_prefix)&.prefix || default_prefix
+  [barcode, prefix]
+end
```
JamesGlover, line 106: This should probably be private, as it mutates the arguments passed in, which isn't immediately obvious.

app/models/tube.rb
```diff
@@ -84,19 +84,32 @@ def after_comment_addition(comment)
...
-    end
+    attributes = args.extract_options!.symbolize_keys
+
+    barcode, prefix = extract_barcode(args, attributes)
+    validate_barcode(barcode, prefix) if barcode.present?
     barcode ||= AssetBarcode.new_barcode
-    primary_barcode = { prefix: prefix, number: barcode }
-    create!(attributes.merge(sanger_barcode: primary_barcode), &block)
+
+    foreign_barcode = attributes.delete(:foreign_barcode)
```
JamesGlover, line 93: Why are you deleting the attribute, only to immediately set it?
Create will quite happily use the setter you've created if it sees the right attribute.
If its a case of << not working until after the tube it persisted, I believe you can use barcodes.build(...) instead and it will defer attempting to save the barcode until the parent record is persisted.

JamesGlover (approved):
Made a couple of suggestions, rubocop is complaining as well though, which will need to be fixed.

app/models/tube.rb
KatyTaylor: Hi James, I had some trouble getting the barcodes to be added in the correct order - so, the sanger one first, and then the foreign one, to ensure that the foreign one becomes the primary. This way I can control the order explicitly.

JamesGlover: Ahh, that makes some sense. I'm wondering if we *need* to generate a Sanger barcode for everything. The fact we do was more an accident of design, followed by a "sure, why not".

KatyTaylor: Ah fair enough, I didn't know the history! But sure, I can't think of a good reason why we need a Sanger barcode.

PR #3448, opened 2021-12-03
Added new libraries

emrojo:
Closes #

*
*
* ...


sdjmchattie:
This isn't the way we should do this data update.  Instead there it a new PR with the correct implementation.  https://github.com/sanger/sequencescape/pull/3452

PR #3452, opened 2021-12-07
DPL 173: Add emSEQ library type

### LINKED ISSUE #3416: DPL-173 - As Matt I would like to have a new library type for emSEQ so CASM and NPG can do effective data analysis (C=S, V=4)
**User story**
As Matt I would like to have a new library type for emSEQ so CASM and NPG can do effective data analysis

Matt / Steve

Is there any more information required for this story?

**Acceptance criteria**
- [ ] I should be able to select the emSEQ library type
- [ ] The emSEQ library type should be associated with all limber bespoke PCR templates.

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

sdjmchattie:
Closes #3416 

* Add emSEQ library type to `limber_pcr_bespoke` request type.
* Update the UAT action for creating tag groups so that no adapter type needs to be provided.

### DIFF
--- a/app/models/tube/purpose.rb
+++ b/app/models/tube/purpose.rb
@@ -67,5 +67,6 @@ end
 require_dependency 'qcable_tube_purpose'
 require_dependency 'illumina_htp/mx_tube_purpose'
 require_dependency 'illumina_htp/stock_tube_purpose'
+require_dependency 'illumina_htp/initial_stock_tube_purpose'
 require_dependency 'tube/standard_mx'
 require_dependency 'tube/stock_mx'
--- a/app/uat_actions/uat_actions/generate_tag_group.rb
+++ b/app/uat_actions/uat_actions/generate_tag_group.rb
@@ -25,7 +25,11 @@ class UatActions::GenerateTagGroup < UatActions
              :select,
              label: 'Adapter Type Name',
              help: 'The name of the adapter type for the tag group.',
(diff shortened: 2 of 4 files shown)


app/models/tube/purpose.rb
```diff
@@ -67,5 +67,6 @@ def set_default_printer_type
 require_dependency 'qcable_tube_purpose'
 require_dependency 'illumina_htp/mx_tube_purpose'
 require_dependency 'illumina_htp/stock_tube_purpose'
+require_dependency 'illumina_htp/initial_stock_tube_purpose'
```
harrietc52, line 70: Just wondering what this is for?

sdjmchattie: `bundle exec rails db:seed` and `bundle exec rails record_loader:all` both fail without it.  The classes are lazy loaded during those processes and without declaring this dependency, they fail when they try to load the tube purpose with that type:

```
Lib Pool:
  type: IlluminaHtp::InitialStockTubePurpose
  target_type: StockMultiplexedLibraryTube
  stock_plate: false
```

harrietc52: Ah sounds like a good bug fix 👍

PR #3458, opened 2021-12-09
Remove aker

stevieing:
Closes #

*
*
* ...


sdjmchattie (approved):
I'm assuming losing all these is fine?  I don't really know anything about Aker but I would assume it's a test framework and we're losing tests.

stevieing:
> I'm assuming losing all these is fine? I don't really know anything about Aker but I would assume it's a test framework and we're losing tests.

Aker was a piece of software that PSD built which was going to track samples across the institute. It was deprecated. This code was to support it in Sequencescape but never got used and never will be used.

Issue #3457, opened 2021-12-09
DPL-190 [BUG] - As a lab user when I try to print off a swipe card id it fails silently


stevieing:
**Describe the bug**
As a lab user when I try to print off a swipe card id it fails silently and no barcode is printed

**To Reproduce**
Steps to reproduce the behaviour:
1. Profile > Edit
2. Enter a swipe card id
3. Select a printer
4. Press print barcode
5. Nothing happens

**Expected behaviour**
A barcode should be printed.

**Screenshots**
![Screen Shot 2021-12-09 at 9 53 07 AM](https://user-images.githubusercontent.com/519327/145374129-8391d7ee-56e6-4b02-b976-94073be91a49.png)

```
 NoMethodError occurred in print_jobs#create:

 undefined method `each' for nil:NilClass
 app/controllers/v2/print_jobs_controller.rb:22:in `block in print_job_params'

URL        
 * HTTP Method: POST
 * IP address : 
 * Parameters : {"data"=>{"attributes"=>{"printer_name"=>"sqtest", "label_template_id"=>199, "labels"=>{"body"=>[{"main"=>{"left_text"=>"si5", "barcode"=>"229745"}}]}}}, "controller"=>"v2/print_jobs", "action"=>"create", "print_job"=>{"data"=>{"attributes"=>{"printer_name"=>"sqtest", "label_template_id"=>199, "labels"=>{"body"=>[{"main"=>{"left_text"=>"si5", "barcode"=>"229745"}}]}}}}}
 * Timestamp  : 2021-12-09 08:47:53 UTC
 * Server : 
   * Rails root : 
 * Process: 1

```

**Additional context**
Fix is currently deployed to Sequencescape Training and UAT was in use

PR #3459, opened 2021-12-09
DPL-190 Swipecard printing bug

harrietc52:
Closes https://github.com/sanger/sequencescape/issues/3457

On Deployment
- [x] Run deploy_label_templates to add swipecard_label_template to sprint (dependent on https://github.com/sanger/deployment/pull/111)
- [ ] Deploy SS


stevieing (approved):

app/views/users/show.html.erb
```diff
@@ -58,7 +58,12 @@ end %>
         </td>
       </tr>
     </table>
-    <div id="pmb-data" data-pmb-api="<%= configatron.pmb_api %>" data-pmb-template="<%= @label_template_id || 0 %>"></div>
+    <div
+      id="pmb-data"
+      data-pmb-api="<%= configatron.pmb_api %>"
+      data-swipecard_pmb_template="<%= configatron.swipecard_pmb_template %>"
```
JamesGlover, line 64: Could you remove the setting of @label_template_id from the controller. It should remove the need for an API call to PMB.

JamesGlover: Also, we're typically using kebab case for data-attributes (I think the vue-style-guide enforces this, although this isn't a vue component). You might find you'd need to switch to camel case for data retrieval (you do when going via dataset) but I'm not sure how jquery wraps this interface.

JamesGlover (changes requested):
Could you clean up the label template stuff in the users controller please. Not only does it simplify things, but removes an unnecessary call to the PMB API.

PR #3463, opened 2021-12-13
copy library_id over from source aliquots

### LINKED ISSUE #3454: DPL-189 set library id on compound samples (C=S?, V=4)
**User story**
As NPG I would like the id_library_lims field to be set on the Cardinal iseq_flowcell (MLWH) rows, so that I can use it to identify when a library is repeat sequenced, and aggregate the sequence data.

Katy, James G, Marina

**Acceptance criteria**
- [ ] When the compound sample is created, at sequencing batch creation time, the compound sample aliquot should have the library_id set
- [ ] The library_id should be set to the same as the library_id on the aliquots that are being transferred into it, if they are all the same
- [ ] If the source aliquots have differing library_ids, leave the library_id on the target aliquot as null - we don't expect this scenario to happen, but if it does happen this will hopefully cause NPG alarm bells to ring and we can improve the behaviour as required

**Additional context**
- The library_id will get propagated automatically to the id_library_lims field in the MLWH.
- This could be implemented by copying across ALL the attributes that are the same across all the source aliquots, rather than just the library id - this could mitigate against any similar future requirements.

KatyTaylor:
Closes #3454 

copy library_id over from source aliquots

### DIFF
--- a/spec/models/request/sample_compound_aliquot_transfer_spec.rb
+++ b/spec/models/request/sample_compound_aliquot_transfer_spec.rb
@@ -56,7 +56,8 @@ RSpec.describe 'Request::SampleCompoundAliquotTransfer' do
              tag_depth: 1,
              study: study,
              project: project,
-             library_type: 'Standard'
+             library_type: 'Standard',
+             library_id: 54
     end
     let(:aliquot2) do
       create :aliquot,
@@ -66,7 +67,8 @@ RSpec.describe 'Request::SampleCompoundAliquotTransfer' do
              tag_depth: 2,
              study: study,
              project: project,
-             library_type: 'Standard'
+             library_type: 'Standard',
+             library_id: 54
     end
 
     it 'creates a compound sample and transfers an aliquot of it' do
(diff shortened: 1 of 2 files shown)


spec/models/request/sample_compound_aliquot_transfer_spec.rb
```diff
@@ -76,7 +78,20 @@
       expect(sequencing_request.target_asset.aliquots.first.library_type).to eq('Standard')
       expect(sequencing_request.target_asset.aliquots.first.study).to eq(study)
       expect(sequencing_request.target_asset.aliquots.first.project).to eq(project)
+      expect(sequencing_request.target_asset.aliquots.first.library_id).to eq(54)
       expect(sequencing_request.target_asset.samples.first.component_samples.order(:id)).to eq(samples.sort)
     end
+
+    # How the library_id should be set if the source aliquots have different library_ids is not defined
+    # Therefore, set it to null for now, until we have a real requirement
```
harrietc52, line 86: I had a question about exactly this when reading the method, which this comment answers 👍 in that case maybe add this comment to the `_copy_library_id` method too? And could update the comment to `nil` not `null`

spec/models/request/sample_compound_aliquot_transfer_spec.rb
```diff
@@ -76,7 +78,20 @@
...
+    context 'with conflicting library_ids' do
+      before { aliquot1.update!(library_id: 82) }
+
+      it 'creates a compound sample with a blank library id' do
+        sequencing_request.transfer_aliquots_into_compound_sample_aliquots
```
harrietc52, line 91: Could add `expect(sequencing_request.target_asset.aliquots.count).to eq(0)` above here too like the other test, it makes it quite clear what the method is changing.
I guess you could also still check that all the other attributes are being set too: library_type, study, project etc

harrietc52 (approved):
Few minor comments

spec/models/request/sample_compound_aliquot_transfer_spec.rb
KatyTaylor: I wanted to keep this one simple so it was testing the specific thing I wanted to test (blank library id), rather than duplicating stuff from the other test and muddying the waters?

PR #3465, opened 2021-12-14
DPL183 Targeted nanoseq

andrewsparkes:


config/default_records/submission_templates/002_targeted_nanoseq_submission_templates.yml
```diff
@@ -0,0 +1,9 @@
+# This submission template is associated with the Limber Targeted NanoSeq pipeline.
+---
+Limber-Htp - Targeted NanoSeq:
+  submission_class_name: "LinearSubmission"
+  related_records:
+    request_type_keys: ["limber_targeted_nanoseq"]
+    project_name: "UAT Project"
+    product_line_name: Targeted-NanoSeq
+    product_catalogue_name: Targeted-NanoSeq
```
KatyTaylor, line 9: The `project_name`, `product_line_name` and `product_catalogue_name` were all Cardinal-related (copy-paste error?), so I've updated them to reference 'Targeted-NanoSeq' instead. `project_name` is still a placeholder though - not sure what this should be.

PR #3466, opened 2021-12-16
DPL-168-1 Moved print swipecard call from AJAX to use PMB rails client

### LINKED ISSUE #3464: DPL-168-1 Move PMB call in SS from JS to rails
**User story**
As a developer I would like SS to not send print swipecard requests to PMB via JavaScript, as this caused a bug because of lack of DRY, and it will trigger CORS issues when SS is on HTTPS and PMB is on HTTP. 

@harrietc52 

**Acceptance criteria**
- [x] In SS go to user profile, enter a swipecard and click `Print Swipecard` called from  `app/views/users/show.html.erb`
- [x] This request doesn't use a AJAX request to call PMB in `print_swipecard.js` (Can this JS file be removed?)
- [x] Instead PmbClient class could be used, and the request to PMB is via SS rails backed

**Additional context**
This is required for https://github.com/sanger/deployment/issues/115, until PMB is on HTTPS. But even then its better to not have PMB being called form two places

BenTopping:
Closes #3464 

* Moved PMB swipecard print request to rails
* added users controller request tests


app/controllers/users_controller.rb
```diff
@@ -7,7 +7,7 @@ class UsersController < ApplicationController # rubocop:todo Style/Documentation
   authorize_resource
 
   def show
-    @printer_list = BarcodePrinter.alphabetical.where(barcode_printer_type: BarcodePrinterType96Plate.all)
+    @printer_list = BarcodePrinter.alphabetical.where(barcode_printer_type: BarcodePrinterType96Plate.all).map(&:name)
```
JamesGlover, line 10: Use `.pluck(:name)` This will:

- Optimise the select query to only bring back the name field
- Avoid instantiating a `BarcodePrinter` for each record, and just create strings.

Not only can this be substantially faster, but it also reduces memory footprint.

app/controllers/users_controller.rb
```diff
@@ -32,11 +32,38 @@ def study_reports
...
+  end
+
   private
 
   def find_user
     @user = User.find(params[:id])
   end
+  
+  def print_swipecard_with_pmb(swipecard, printer)
+    LabelPrinter::PmbClient.print({
```
JamesGlover, line 53: Most other controllers interact with `PrintJob#execute` which gives you your error handling, and centralises all our label classes in one place. This makes it much easier to review how printing it being used, and keeps the controllers nice and light.

app/views/users/show.html.erb
```diff
@@ -35,37 +35,21 @@ end %>
   <% end %>
 <% end %>
 </div>
+
 <div class="card-default">
 <%= panel(:default, title: 'Print Swipecard Barcode') do %>
-  <form onkeypress="return event.keyCode != 13;">
+  <%= form_tag(@user.id.to_s + '/print_swipecard') do %>
```
JamesGlover, line 41: Use rails path helpers here, rather than composing strings yourself.

app/controllers/users_controller.rb
JamesGlover: Part of me also wonders if this should be an entirely separate controller, but perhaps that's something we can put off until we rationalise the rest of our printing. I've got a branch in progress at the moment that is stripping out a load of the legacy batch stuff, which will put us in a much stronger position.

JamesGlover (changes requested):

JamesGlover (approved):

JamesGlover:
Although you probably should add a bit of documentation for the new Label::Swipecard class, obviously doesn't have to be anythign substantial, but possibly just explain that it is used to print of swipecards in barcode format for users who don't have access to a proximity reader.

PR #3472, opened 2021-12-20
Sample Manifest UAT action tweak

KatyTaylor:
Don't specify the library_id on the aliquot
- this is not appropriate for the cardinal pipeline - it should be set later on when the tags are added
- it is probably appropriate for other types of manifest, however the only use I know of so far is the cardinal pipeline, so keep it simple

### DIFF
--- a/app/uat_actions/uat_actions/generate_sample_manifest.rb
+++ b/app/uat_actions/uat_actions/generate_sample_manifest.rb
@@ -78,7 +78,7 @@ class UatActions::GenerateSampleManifest < UatActions
       raise 'Manifest for plates is not supported yet' unless asset_type == '1dtube'
 
       create_sample("Sample_#{asset.human_barcode}_1", sample_manifest).tap do |sample|
-        asset.aliquots.create!(sample: sample, study: study, library: asset)
+        asset.aliquots.create!(sample: sample, study: study)
         study.samples << sample
       end
     end


app/uat_actions/uat_actions/generate_sample_manifest.rb
```diff
@@ -78,7 +78,7 @@ def create_samples(sample_manifest)
       raise 'Manifest for plates is not supported yet' unless asset_type == '1dtube'
 
       create_sample("Sample_#{asset.human_barcode}_1", sample_manifest).tap do |sample|
-        asset.aliquots.create!(sample: sample, study: study, library: asset)
+        asset.aliquots.create!(sample: sample, study: study)
```
harrietc52, line 81: I thought this PR https://github.com/sanger/sequencescape/pull/3463/files was adding in library_id, maybe it could be `nil`?

harrietc52 (approved):
Not sure if it's related but thought ..

app/uat_actions/uat_actions/generate_sample_manifest.rb
harrietc52: Just read your comment in the description field 👍 can ignore the above in that case that it is added later

KatyTaylor: Hey, yes so it should be nil at this sample manifest upload stage - then it should be added in when the tags get added, later in the pipeline (I think) - and then that PR you've linked to makes sure it gets transferred to the compound sample when it's created.

I will do a test though, before merging this in, to make sure that it does get added at tagging stage.

KatyTaylor: OK tested locally, and if you don't assign a library id up front in the manifest, then when you get to the PCR XP plate, all aliquots in a well have the same library id, which is what we want - because then when you make the sequencing batch, the compound sample aliquot should also have this library id

PR #3474, opened 2021-12-22
DPL-188 simplified heron request graph

### LINKED ISSUE #3451: DPL-188 - Simplified Heron Request Graph (C=M, V=4)
**User story**
As a developer I would like to simplify the heron request graph in order to reduce the number of edge-cases in the Heron pipeline. (Note: This story also has limber requirements)

- Heron submissions should build *just* the limber_heron request type (or possibly a V2 flavour?) [Edit: Yep, a V2 request type would make a seamless transition far easier]
- This request type should now be flagged as `for_multiplexing`
- This request should cover the *entire* library creation pipeline in limber
- No longer create the final tube up-front (Will require a new tube purpose)
- Remove the charge and pass step form the pool plate
- Instead should charge and pass at end of pipeline and build sequencing requests then.

This change:
- Avoids the need to explicitly know which RT plates will be processed together when building submissions
- Removed the pain points where you can't easily jump back to a previous step
- Avoids the need to 'charge and pass' before you get the QC data on the XP tube

Edit:

Note: Complexity high due to uncertainty. The actual core changes are fairly simple, and we may get lucky. But I'm anticipating a number of unknowns.

James

**Acceptance criteria**
- [x] Update LucidChart doc
- [ ] Update Confluence "Limber mistakes" page
- [ ] Communicate changes to Heron leads

KatyTaylor:
Closes #3451 

~TODO: write code for updating old submission templates to fill in 'superceded_by_id' and 'superceded_at' fields.~

See Limber and Integration Suite PRs of the same name

### DIFF
--- a/config/default_records/request_types/004_limber_high_throughput.yml
+++ b/config/default_records/request_types/004_limber_high_throughput.yml
@@ -156,6 +156,20 @@ limber_heron_lthr:
     - Sanger_tailed_artic_v1_384
     - PCR amplicon tailed adapters 96
     - PCR amplicon tailed adapters 384
+limber_heron_lthr_v2:
+  <<: *limber_htp_library
+  name: Limber Heron LTHR V2
+  request_class_name: IlluminaHtp::Requests::HeronTailedRequest
+  for_multiplexing: true
+  acceptable_purposes:
+    - LTHR RT
+    - LTHR-384 RT
+    - LTHR Cherrypick
+  library_types:
+    - Sanger_tailed_artic_v1_96
+    - Sanger_tailed_artic_v1_384
+    - PCR amplicon tailed adapters 96
+    - PCR amplicon tailed adapters 384
 limber_pwgs-384:
   <<: *limber_htp_library
   name: Limber pWGS-384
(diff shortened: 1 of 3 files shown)


config/default_records/request_types/004_limber_high_throughput.yml
```diff
@@ -156,6 +156,21 @@ limber_heron_lthr:
     - Sanger_tailed_artic_v1_384
     - PCR amplicon tailed adapters 96
     - PCR amplicon tailed adapters 384
+limber_heron_lthr_v2:
+  <<: *limber_htp_library
+  name: Limber Heron LTHR V2
+  request_class_name: IlluminaHtp::Requests::HeronTailedRequest
+  for_multiplexing: true
+  target_purpose_name: LB Lib Pool Norm
```
JamesGlover, line 164: I think you may have mentioned this in stand-up, but that will still tie us to fixed-submissions, as we'll still be generating the Pool tube up-front. But I think you said you'd not got to that bit yet.

JamesGlover (approved):

PR #3479, opened 2022-01-05
Merge develop into master

harrietc52:
Ready for deployment to Production tomorrow 6pm
@stevieing @KatyTaylor @JamesGlover Can you check everything you expect to be here, is?


JamesGlover:
All my changes in there are safe to go out. I highlighted the bit I think is causing your permissions issues with swipecard printing.

stevieing (changes requested):
All fine with me but needs release version updating. I am not sure what is going on but is there an issue that only admins can print barcodes? Is that going to be changed before release?

KatyTaylor (approved):

PR #3481, opened 2022-01-06
DPL-188 deployment fix

KatyTaylor:
move new Heron submission templates to record loader
- because deployment was failing
- they are dependent on the new request type which was in record loader, and limber:setup is run before record loader, on deploy
- had to also move some existing request types over otherwise it didn't work locally when running db:setup


KatyTaylor:
Not ready to go yet as needs:
- things removing from seeds
- things removing from sequencing.rake
- and extra things adding like RequestType::Validator, and other records

KatyTaylor:
This approach is too complex (too many dependencies). Going to try a different approach.

PR #3483, opened 2022-01-07
Issue3253/dpl 088 {bug} editing of batches to swap lane position {c m v 5}

### LINKED ISSUE #3253: DPL-088 [BUG] Editing of batches to swap lane position [C=M, V=5]
If a pool has been assigned to a lane in the wrong order edit batch has a facility where the pool can be swapped with another to position it correctly. 
However this is inconsistent in it's success and fails to update the MLWH. 
It also fails to work if the pools are not next to each other (2 moves).

JamesGlover:
Closes #3253

* Use datatables inbuilt sorting
* Explicitly require a save to update batch
* Ensure broadcast messages have updated timestamp with touch
* Update dependencies


.github/workflows/append_release.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v2
         with:
-          node-version: "12"
+          node-version: "16"
```
JamesGlover, line 20: We were on node 12, which was pretty old. The update process was nice and smooth once I updated webpacker.

app/controllers/batches_controller.rb
```diff
@@ -206,7 +206,9 @@ def fail_items # rubocop:todo Metrics/AbcSize
 
   def sort
     @batch.assign_positions_to_requests!(params['requests_list'].map(&:to_i))
-    @batch.rebroadcast
+
+    # Touch the batch to update its timestamp and trigger re-broadcast
+    @batch.touch # rubocop:disable Rails/SkipsModelValidations
```
JamesGlover, line 211: We were sending the message, but it didn't have an updated timestamp (not to mention was at risk of being generated before the transaction closed). Rather than faffing around pulling in the batch_request timestamps, we just update the batch itself with a touch. Touch in turn triggers rebroadcast at the close of the transaction.

app/views/shared/_batch.html.erb
```diff
@@ -1,99 +1,78 @@
```
JamesGlover, line 1: The whole table was indented, so lots of purely whitespace changes here.

app/views/shared/_batch.html.erb
```diff
@@ -1,99 +1,78 @@
...
-  <table class="sortable table">
-    <thead>
-      <tr>
-        <% if edit -%>
-          <th width="3%">Drag</th>
-        <% elsif stage %>
-          <th class="nosort" width="7%"><input type='checkbox' class='select_all'></th>
+<table  id="batch-<%= edit ? 'edit' : 'show' %>"
+        class="table"
+        data-batch-id="<%= @batch&.id %>">
```
JamesGlover, line 4: Removed the sortable class, and we'll mount Datatables by id instead. This allows us to mount the edit page with different options.

app/views/shared/_batch.html.erb
```diff
@@ -1,99 +1,78 @@
 
-  <table class="sortable table">
-    <thead>
-      <tr>
-        <% if edit -%>
-          <th width="3%">Drag</th>
```
JamesGlover, line 6: We can use the whole row as a drag handle, so can remove this. In turn this also means we don't need to specify a custom index column for the sort functionality.

app/views/shared/_batch.html.erb
```diff
@@ -1,99 +1,78 @@
...
+      <th>Target Type</th>
+      <th>Target barcode</th>
+      <% if @batch.display_tags? %>
+        <th>Tag #</th>
+      <% end %>
+      <% @rits.each do |rit| %>
+        <th><%= rit.label %></th>
+      <% end %>
+      <th>State</th>
+      <th>Comments</th>
```
JamesGlover, line 24: Just whitespace

app/views/shared/_batch.html.erb
```diff
@@ -1,99 +1,78 @@
...
-      $( "#requests_list" ).sortable({
-        update: function(event, elem) {
-          $.post('/batches/sort?batch_id=' + <%= @batch.id %>, batchListPostParametersFromSortable(event.target));
-        }
-      });
-      $( "#requests_list" ).disableSelection();
-    }());
-
-<% end %>
-<% end %>
```
JamesGlover, line 99: All handled in a proper js file now.

app/views/shared/_batch.html.erb
```diff
@@ -1,99 +1,78 @@
...
-        <% @rits.each do |rit| %>
-          <th><%= rit.label %></th>
+
+        <% if request.target_asset %>
+          <td><%= link_to request.target_asset.sti_type, receptacle_path(request.target_asset) %></td>
+          <td><%= request.target_asset.try(:human_barcode) %></td>
+        <% else %>
+          <td></td>
+          <td></td>
         <% end %>
```
JamesGlover, line 53: Get rid of the nested if here and simplify. It does slightly change the behaviour (before we'd hide a target asset if we didn't have an asset, now we show it) but this feels more useful.

package.json
```diff
@@ -1,12 +1,13 @@
 {
   "dependencies": {
-    "@rails/webpacker": "^5.1.1",
-    "datatables.net-bs4": "^1.10.19",
-    "datatables.net-buttons-bs4": "^1.5.6",
-    "datatables.net-fixedcolumns-bs4": "^3.2.6",
-    "datatables.net-fixedheader-bs4": "^3.1.5",
-    "datatables.net-responsive-bs4": "^2.2.3",
-    "datatables.net-rowgroup-bs4": "^1.1.0",
```
JamesGlover, line 9: Lots of updates. Webpacker in particular is nice, as it drops the node-sass dependency which regularly gives me pain.

spec/features/batches/sort_requests_spec.rb
```diff
@@ -3,10 +3,11 @@
 require 'rails_helper'
```
JamesGlover, line 3: Not much actually changes in the overall workflow. Just add tests for the message generation and now we click a link, not a button.

stevieing (approved):

PR #3510, opened 2022-01-31
Improvements to compound sample creation

KatyTaylor:
- to do with setting the study id on the target aliquots (the study that will end up in iseq_flowcell)
- if the user specifies a study on the sequencing submission/ request, use that (it was previously being ignored)
- otherwise, same logic as before - use the one from the source aliquots if it's consistent, error if not

### DIFF
--- a/app/models/compound_aliquot.rb
+++ b/app/models/compound_aliquot.rb
@@ -9,7 +9,7 @@ class CompoundAliquot # rubocop:todo Style/Documentation
 
   attr_accessor :request, :source_aliquots
 
-  attr_reader :component_samples
+  attr_reader :component_samples, :compound_sample
 
   validate :tag_depth_is_unique
 
@@ -33,10 +33,23 @@ class CompoundAliquot # rubocop:todo Style/Documentation
 
   # Generates the compound sample, under the default study, using the component samples
   def create_compound_sample
-    default_compound_study.samples.create!(
-      name: SangerSampleId.generate_sanger_sample_id!(default_compound_study.abbreviation),
-      component_samples: component_samples
-    )
+    @compound_sample =
+      default_compound_study.samples.create!(
+        name: SangerSampleId.generate_sanger_sample_id!(default_compound_study.abbreviation),
+        component_samples: component_samples
+      )
+  end
+
+  def aliquot_attributes
+    {
+      tag_id: tag_id,
+      tag2_id: tag2_id,
+      library_type: default_library_type,
+      study_id: default_compound_study.id,
+      project_id: default_compound_project_id,
+      library_id: copy_library_id,
+      sample: compound_sample
+    }
   end
 
   # Default study that the new compound sample will use
--- a/app/models/compound_aliquot.rb
+++ b/app/models/compound_aliquot.rb
@@ -33,7 +33,7 @@ class CompoundAliquot
   # Check that the component samples in the compound sample will be able to be distinguished -
   # this is represented by them all having a unique 'tag_depth'
   def tag_depth_is_unique
-    return unless source_aliquots.pluck(:tag_depth).uniq.count != source_aliquots.size
+    return unless source_aliquots.pluck(:tag_depth).uniq!
 
     errors.add("#{DUPLICATE_TAG_DEPTH_ERROR_MSG}: #{component_samples.map(&:name)}")
   end
--- a/app/models/compound_aliquot.rb
+++ b/app/models/compound_aliquot.rb
@@ -31,6 +31,14 @@ class CompoundAliquot # rubocop:todo Style/Documentation
     source_aliquots.first.library_type
   end
 
+  # Generates the compound sample, under the default study, using the component samples
+  def create_compound_sample
+    default_compound_study.samples.create!(
+      name: SangerSampleId.generate_sanger_sample_id!(default_compound_study.abbreviation),
+      component_samples: component_samples
+    )
+  end
+
   # Default study that the new compound sample will use
   # Uses the one from the request if it's present,
   # otherwise, the one from the source aliquots if it's consistent.
--- a/app/models/compound_aliquot.rb
+++ b/app/models/compound_aliquot.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+class CompoundAliquot # rubocop:todo Style/Documentation
+  include ActiveModel::Model
+
+  DUPLICATE_TAG_DEPTH_ERROR_MSG = "Cannot create compound sample from following samples due to duplicate 'tag depth'"
+  MULTIPLE_STUDIES_ERROR_MSG =
+    'Cannot create compound sample due to the component samples being under different studies.'
+
+  attr_accessor :request, :source_aliquots
+
+  attr_reader :component_samples
+
+  validate :tag_depth_is_unique
+
+  def initialize(attributes)
+    super
+
+    @component_samples = source_aliquots.map(&:sample)
+  end
+
+  # Check that the component samples in the compound sample will be able to be distinguished -
+  # this is represented by them all having a unique 'tag_depth'
+  def tag_depth_is_unique
+    return unless source_aliquots.pluck(:tag_depth).uniq.count != source_aliquots.size
+
+    errors.add("#{DUPLICATE_TAG_DEPTH_ERROR_MSG}: #{component_samples.map(&:name)}")
+  end
+
+  def default_library_type
+    source_aliquots.first.library_type
+  end
+
+  # Default study that the new compound sample will use
+  # Uses the one from the request if it's present,
+  # otherwise, the one from the source aliquots if it's consistent.
+  def default_compound_study
+    request.initial_study ||
+      begin
+        raise MULTIPLE_STUDIES_ERROR_MSG if studies.count > 1
+
+        studies.first
+      end
+  end
+
+  def studies
+    source_aliquots.map(&:study).uniq
+  end
+
+  # Default project that the new compound sample will use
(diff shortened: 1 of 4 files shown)


app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -108,13 +106,28 @@ def _create_compound_sample(study, component_samples)
   end
 
   # Default study that the new compound sample will use
+  # Uses the one from the request if it's present,
+  # otherwise, the one from the source aliquots if it's consistent.
   def _default_compound_study
```
JamesGlover, line 111: Do you know the reason for the underscores in front of these method names?

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -108,13 +106,28 @@ def _create_compound_sample(study, component_samples)
...
-    _studies.first
+    _initial_study ||
+      begin
+        raise MULTIPLE_STUDIES_ERROR_MSG if _studies.count > 1
+
+        _studies.first
+      end
+  end
+
+  def _initial_study
```
JamesGlover, line 120: Why not just used `initial_study`? This will prevent us from eager loading studies, and force us to look it up again each time.

app/models/request/sample_compound_aliquot_transfer.rb
JamesGlover: Ahh I see, its because its a module...

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -108,13 +106,28 @@ def _create_compound_sample(study, component_samples)
   end
 
   # Default study that the new compound sample will use
+  # Uses the one from the request if it's present,
+  # otherwise, the one from the source aliquots if it's consistent.
   def _default_compound_study
-    _studies.first
+    _initial_study ||
+      begin
+        raise MULTIPLE_STUDIES_ERROR_MSG if _studies.count > 1
```
JamesGlover, line 114: It might make sense to raise a custom exception class which we can then rescue and return a useful error message to the end user. Could even point them to request additional sequencing as a potential solution then.

JamesGlover: Actually, isn't this validation too strict? It's looking at the studies on *All* aliquots in the receptacle, not just those that will compounded together.

JamesGlover (changes requested):
I think the validation is being too strict, as it requires a single study for *all* aliquots in the receptacle, not just those which will be compounded together. Also, not sure on why the initial_study method has been re-implemented.

app/models/compound_aliquot.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+class CompoundAliquot # rubocop:todo Style/Documentation
```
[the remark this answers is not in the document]
JamesGlover: Probably makes sense to add this, as compound aliquots are a bit confusing, and this is probably where people will end up. Also good to make clear that this is just a factory.

app/models/compound_aliquot.rb
```diff
@@ -0,0 +1,72 @@
...
+  def initialize(attributes)
+    super
+
+    @component_samples = source_aliquots.map(&:sample)
+  end
+
+  # Check that the component samples in the compound sample will be able to be distinguished -
+  # this is represented by them all having a unique 'tag_depth'
+  def tag_depth_is_unique
+    return unless source_aliquots.pluck(:tag_depth).uniq.count != source_aliquots.size
```
JamesGlover, line 25: Should probably use size or count on both sides here. Given both are arrays I thing size, count and length are all essentially synonyms.

Also, you could do:

```
return unless source_aliquots.pluck(:tag_depth).uniq!
```
As uniq! returns any duplicate elements, or nil otherwise.

I'm a bit surprised that Array doesn't just provide `duplicates?` as it feels like a very common requirement.

app/models/compound_aliquot.rb
```diff
@@ -0,0 +1,72 @@
...
+    source_aliquots.first.library_type
+  end
+
+  # Default study that the new compound sample will use
+  # Uses the one from the request if it's present,
+  # otherwise, the one from the source aliquots if it's consistent.
+  def default_compound_study
+    request.initial_study ||
+      begin
+        raise MULTIPLE_STUDIES_ERROR_MSG if studies.count > 1
```
JamesGlover, line 40: Possibly move this to validation instead, so the check can be performed up-front before you do anything.

app/models/compound_aliquot.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def studies
+    source_aliquots.map(&:study).uniq
+  end
+
+  # Default project that the new compound sample will use
+  # Uses the one from the request if it's present,
+  # otherwise, one grabbed from a source aliquot.
+  def default_compound_project_id
+    request.initial_project_id || source_aliquots.first.project_id
```
JamesGlover, line 54: Perhaps we want the same check for studies as for projects.

app/models/compound_aliquot.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  # If the library_id is the same on all source aliquots, we can confidently transfer it to the target aliquot
+  # How the library_id should be set if the source aliquots have different library_ids is not defined
+  # Therefore, set it to nil for now, until we have a real requirement
+  def copy_library_id
```
JamesGlover, line 60: Sorry, pedantic naming complaint. copy_library_id implies that this is a verby method, and its actually copying the library_id to the aliquots, instead its a nouny method, and is returning the expected library id. I'd go along with the same pattern used for study and project, and call it default_library_id

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -94,31 +51,43 @@ def _aliquots_by_tags_combination
...
   end
 
-  def _studies
-    @studies ||= asset.aliquots.map(&:study).uniq
+  def _transfer_into_compound_sample_aliquot(source_aliquots)
+    compound_aliquot = CompoundAliquot.new(request: self, source_aliquots: source_aliquots)
+
+    raise compound_aliquot.errors unless compound_aliquot.valid?
+
+    compound_sample = _create_compound_sample(compound_aliquot)
```
JamesGlover, line 59: I'd extract this into compound aliquot

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -94,31 +51,43 @@ def _aliquots_by_tags_combination
...
-    _studies.first
-  end
-
-  # Default project that the new compound sample will use
-  def _default_compound_project_id
-    asset.aliquots.first.project_id
+  def _add_aliquot(sample, compound_aliquot)
+    target_asset
+      .aliquots
+      .create(sample: sample)
```
JamesGlover, line 78: Rather than setting the aliquot attributes individually here, you could do:

```
target_asset.aliquots.create(compound_aliquot.aliquot_attributes)
```

That would let you move a whole chunk of behaviour into compound aliquot and reduce the hits to the database. (This currently hits twice, once on the create, and then on the following save.

JamesGlover (changes requested):
Much better, there are a few other methods that would probably better belong in CompoundSample, and then some minor niggles with naming of method consistency.

app/models/compound_aliquot.rb
KatyTaylor: Yeah I wondered whether to, but the reason it's here is because you only really care about this check if there's no 'initial_study'. I suppose we could also do that check in the validation though.

app/models/compound_aliquot.rb
KatyTaylor: I don't know. I don't know much about projects and where they're used, other than the fact that they're used in billing. Will it matter which project id ends up on the compound sample aliquot?

app/models/compound_aliquot.rb
JamesGlover: You can make validation conditional: `validate :aliquot_studies_unique, unless: :initial_study` rather than including it in one method

app/models/compound_aliquot.rb
JamesGlover: Oh dear... we might need to check this.

So usually projects are billed based on the fraction of a lane their aliquots take up, with the assumption that all aliquots are loaded at approximately equal concentrations. So if a single project was assigned to a compound aliquot, they'd be billed, but the other projects would get it for free.

But the additional gotcha, is that I'm not sure its true to assume that a compound aliquot it the same 'size' as other aliquots. Billing reports will need to re-inflate them.

app/models/compound_aliquot.rb
```diff
@@ -0,0 +1,104 @@
...
+  # Check that the component samples in the compound sample will be able to be distinguished -
+  # this is represented by them all having a unique 'tag_depth'
+  def tag_depth_is_unique
+    return unless source_aliquots.pluck(:tag_depth).uniq!
+
+    errors.add("#{DUPLICATE_TAG_DEPTH_ERROR_MSG}: #{component_samples.map(&:name)}")
+  end
+
+  def default_library_type
+    source_aliquots.first.library_type
```
JamesGlover, line 42: I thought we were only setting library type if it was the same for all aliquots. (Sorry about not catching this one sooner, I think I got it mixed up with library_id)

JamesGlover (commented):
Yep, looks good, but just want to chat about the project stuff.

JamesGlover (approved):

Issue #3511, opened 2022-02-02
Seemingly non-deterministic failures on download testing in RSpec tests


JamesGlover:
When running RSpec tests we see failures of the form:

```
  1) Cherrypicking pipeline where picking by ng for a tecan robot behaves like a cherrypicking procedure running the cherrypicking pipeline
     Failure/Error: raise StandardError, "Could not open #{file} in #{PATH} after #{timeout} seconds"
     
     StandardError:
       Could not open 238_batch_DN4V_1.gwl in /Users/jg16/projects/sanger/sequencescape/tmp/capybara after 10 seconds
     Shared Example Group: "a cherrypicking procedure" called from ./spec/features/pipelines/cherrypick/cherrypick_pipeline_spec.rb:328
     # ./spec/support/download_helper.rb:26:in `rescue in wait_for_download'
     # ./spec/support/download_helper.rb:23:in `wait_for_download'
     # ./spec/support/download_helper.rb:13:in `downloaded_file'
     # ./spec/features/shared_examples/cherrypicking.rb:341:in `block (6 levels) in <top (required)>'
     # ./spec/features/shared_examples/cherrypicking.rb:240:in `block (5 levels) in <top (required)>'
     # ./spec/features/shared_examples/cherrypicking.rb:167:in `block (4 levels) in <top (required)>'
     # ./spec/features/shared_examples/cherrypicking.rb:165:in `each'
     # ./spec/features/shared_examples/cherrypicking.rb:165:in `block (3 levels) in <top (required)>'
     # ./spec/features/shared_examples/cherrypicking.rb:163:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # Timeout::Error:
     #   execution expired
     #   ./spec/support/download_helper.rb:24:in `sleep'
```

This happens because the downloaded files are saved to the project root, rather than the `Capybara.save_path`
(./tmp/capybara). However, when runn in isolation, this failure is not reproduced.

Furthermore, even with a fixed test order, failure does not appear to be deterministic. When running:

```
[...]

PR #3512, opened 2022-02-02
Issue3314/dpl 123 as a long read user i would like a sample manifest with library type added {c s v 4}

### LINKED ISSUE #3314: DPL-123 - As a long read user I would like a sample manifest with library type added (C=S, V=4)
**User story**
As a long read user I would like a sample manifest with library type added so that we can identify IsoSeq runs

e.g. John S (don't include surnames in public repos)

**Acceptance criteria**
- [x] When I go to the sample manifest page there will be an option to generate a long read manifest with library type
- [x] I should be able to select the library type from a drop-down list.

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

JamesGlover:
Closes #3314

* Add library type column to Long Read manifests
* Add library type record loader for pacbio library types
* Manifest loader supports parameters for scopes
* Library type has scope for loading from record loader files
*  Remove capybara-selenium - This one was a red herring, but is entirely redundant now. Hasn't been touched for years, isn't needed, and is just a security risk.
* Improve performance of sample manifest pages - This was when I thought the story was simple, and I'd be hitting this page a lot.
* Library type must now have a unique name - This was pretty much always a requirement, and violations were considered bugs. Now enforced.
* Fixed tests resulting from the above.
* Added documentation re: https://github.com/sanger/sequencescape/issues/3511 - I made some slight adjustments here while investigating to remove a few legacy configuration options which supported chrome on old CI setups.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -929,7 +929,6 @@ Rails/FilePath:
     - 'lib/accession/accession.rb'
     - 'lib/tasks/cucumber.rake'
     - 'lib/tasks/sample_manifest_excel_tests.rake'
-    - 'lib/tasks/test.rake'
     - 'test/test_helper.rb'
     - 'test/unit/fluidigm_file_test.rb'
     - 'test/unit/import_fluidigm_data_test.rb'
--- a/app/controllers/sdb/sample_manifests_controller.rb
+++ b/app/controllers/sdb/sample_manifests_controller.rb
@@ -55,8 +55,10 @@ class Sdb::SampleManifestsController < Sdb::BaseController # rubocop:todo Style/
   end
 
   def index
-    pending_sample_manifests = SampleManifest.pending_manifests.paginate(page: params[:page])
-    completed_sample_manifests = SampleManifest.completed_manifests.paginate(page: params[:page])
+    pending_sample_manifests =
+      SampleManifest.pending_manifests.includes(:study, :supplier, :user).paginate(page: params[:page])
+    completed_sample_manifests =
+      SampleManifest.completed_manifests.includes(:study, :supplier, :user).paginate(page: params[:page])
     @display_manifests = pending_sample_manifests | completed_sample_manifests
     @sample_manifests = SampleManifest.paginate(page: params[:page])
   end
--- a/app/models/library_type.rb
+++ b/app/models/library_type.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require './lib/record_loader/library_type_loader'
+
 # A library type is an identifier which describes how a library has been produced
 # Whereas {RequestType} is largely internal to sequencescape, a library type directly
 # reflects the science, and thus has a many to many relationship with request type.
--- a/app/models/library_type.rb
+++ b/app/models/library_type.rb
@@ -16,7 +16,7 @@
 class LibraryType < ApplicationRecord
   include SharedBehaviour::Named
 
-  scope :long_read, -> { alphabetical.where(name: records_in('001_long_read')) }
+  scope :from_record_loaders, ->(loader) { alphabetical.where(name: records_in(loader)) }
 
   has_many :library_types_request_types, inverse_of: :library_type, dependent: :destroy
   has_many :request_types, through: :library_types_request_types
--- a/app/models/library_type.rb
+++ b/app/models/library_type.rb
@@ -1,9 +1,29 @@
 # frozen_string_literal: true
-class LibraryType < ApplicationRecord
-  validates :name, presence: true
 
-  scope :alphabetical, -> { order(:name) }
+# A library type is an identifier which describes how a library has been produced
+# Whereas {RequestType} is largely internal to sequencescape, a library type directly
+# reflects the science, and thus has a many to many relationship with request type.
+# LibraryType gets exposed in the multi-lims warehouse, and can affect downstream
+# processing.
+#
+# Some library types have no request type associated. These may be used by
+# external pipelines, such as Traction, or may be of use exclusively for library
+# manifests.
+#
+# @note For historical reasons, library type is *not* a relationship on request_metadata
+# or on aliquot. This is because library_types used to by hardcoded in the {Request}
+# class.
+class LibraryType < ApplicationRecord
+  include SharedBehaviour::Named
+
+  scope :long_read, -> { alphabetical.where(name: records_in('001_long_read')) }
 
   has_many :library_types_request_types, inverse_of: :library_type, dependent: :destroy
   has_many :request_types, through: :library_types_request_types
+
+  validates :name, presence: true, uniqueness: { case_sensitive: false }
+
+  def self.records_in(file_name)
+    RecordLoader::LibraryTypeLoader.new(files: [file_name]).names
+  end
 end
--- a/app/models/request_type/validator.rb
+++ b/app/models/request_type/validator.rb
@@ -70,10 +70,14 @@ class RequestType::Validator < ApplicationRecord
     def default
       nil
     end
+
+    def valid_options
+      []
+    end
   end
 
-  belongs_to :request_type
-  validates :request_type, :request_option, :valid_options, presence: true
+  belongs_to :request_type, optional: false
+  validates :request_option, :valid_options, presence: true
   serialize :valid_options
 
   delegate :include?, to: :valid_options
--- a/app/models/sample_manifest.rb
+++ b/app/models/sample_manifest.rb
@@ -15,7 +15,7 @@
 # All {Sample samples} in a given manifest will initially belong to a single
 # {Study}, although it is possible for them to become associated with additional
 # studies over time.
-class SampleManifest < ApplicationRecord # rubocop:todo Metrics/ClassLength
+class SampleManifest < ApplicationRecord
   include Uuid::Uuidable
   include ModelExtensions::SampleManifest
   include SampleManifest::BarcodePrinterBehaviour
@@ -60,8 +60,8 @@ class SampleManifest < ApplicationRecord # rubocop:todo Metrics/ClassLength
 
   def template; end
 
-  belongs_to :supplier
-  belongs_to :study
+  belongs_to :supplier, optional: false
+  belongs_to :study, optional: false
   belongs_to :project
   belongs_to :user
   belongs_to :purpose
@@ -75,8 +75,6 @@ class SampleManifest < ApplicationRecord # rubocop:todo Metrics/ClassLength
   serialize :last_errors
   serialize :barcodes
 
-  validates :supplier, presence: true
-  validates :study, presence: true
   validates :count, numericality: { only_integer: true, greater_than: 0, allow_blank: false }
   validates :asset_type, presence: true, inclusion: { in: SampleManifest::CoreBehaviour::BEHAVIOURS }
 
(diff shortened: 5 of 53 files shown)


.rubocop_todo.yml
```diff
@@ -929,7 +929,6 @@ Rails/FilePath:
     - 'lib/accession/accession.rb'
     - 'lib/tasks/cucumber.rake'
     - 'lib/tasks/sample_manifest_excel_tests.rake'
-    - 'lib/tasks/test.rake'
```
JamesGlover, line 932: ToDone!

app/controllers/sdb/sample_manifests_controller.rb
```diff
@@ -55,8 +55,10 @@ def show
   end
 
   def index
-    pending_sample_manifests = SampleManifest.pending_manifests.paginate(page: params[:page])
-    completed_sample_manifests = SampleManifest.completed_manifests.paginate(page: params[:page])
+    pending_sample_manifests =
+      SampleManifest.pending_manifests.includes(:study, :supplier, :user).paginate(page: params[:page])
+    completed_sample_manifests =
+      SampleManifest.completed_manifests.includes(:study, :supplier, :user).paginate(page: params[:page])
```
JamesGlover, line 61: Performance optimisation. Not entirely convinced by the split, feel like it may make morse sense to allow filtering of manifests by state.

app/models/library_type.rb
```diff
@@ -1,9 +1,31 @@
...
+# LibraryType gets exposed in the multi-lims warehouse, and can affect downstream
+# processing.
+#
+# Some library types have no request type associated. These may be used by
+# external pipelines, such as Traction, or may be of use exclusively for library
+# manifests.
+#
+# @note For historical reasons, library type is *not* a relationship on request_metadata
+# or on aliquot. This is because library_types used to by hardcoded in the {Request}
+# class.
```
JamesGlover, line 17: Document all the things. I'm not actually sure why rubocop wasn't shouting about this one.

app/models/library_type.rb
```diff
@@ -1,9 +1,31 @@
...
 class LibraryType < ApplicationRecord
-  validates :name, presence: true
```
JamesGlover, line 3: Moved down below the associations

app/models/library_type.rb
```diff
@@ -1,9 +1,31 @@
...
+  include SharedBehaviour::Named
 
-  scope :alphabetical, -> { order(:name) }
```
JamesGlover, line 5: Used the shared module which already provides this

app/models/library_type.rb
```diff
@@ -1,9 +1,31 @@
...
+  scope :from_record_loaders, ->(loader) { alphabetical.where(name: records_in(loader)) }
 
   has_many :library_types_request_types, inverse_of: :library_type, dependent: :destroy
   has_many :request_types, through: :library_types_request_types
+
+  validates :name, presence: true, uniqueness: { case_sensitive: false }
+
+  def self.records_in(file_name)
+    RecordLoader::LibraryTypeLoader.new(files: [file_name]).names
+  end
```
JamesGlover, line 30: This is a little bit circular, as we immediately pluck the names from the record we just loaded. But I figured this is rare enough that we can spend the round trip to the database for the sake of consistency.

app/models/request_type/validator.rb
```diff
@@ -70,10 +70,14 @@ def validate?(_value)
     def default
       nil
     end
+
+    def valid_options
+      []
+    end
   end
```
JamesGlover, line 77: Mostly to help cases in the test suite where we didn't have library types associated with a request type. It would fall back to the permissive 'NullValidator' but in some cases would try and get a list of options for drop-downs etc. We might need to revisit this is its something that starts happening in production (where it might make more sense to find *all* library types) but currently this is more of a theoretical scenario, so we just don't suggest any options.

app/models/request_type/validator.rb
```diff
@@ -70,10 +70,14 @@ def validate?(_value)
...
 
-  belongs_to :request_type
-  validates :request_type, :request_option, :valid_options, presence: true
```
JamesGlover, line 76: This is related to some of the cop issues @stevieing ran into. Migrating the validators to the new approach while I'min this class. Essentially no behaviour changes, but future proofs us for future Rails upgrades.

app/models/sample_manifest.rb
```diff
@@ -60,8 +60,8 @@ def barcode_printer; end
 
   def template; end
 
-  belongs_to :supplier
-  belongs_to :study
```
JamesGlover, line 64: More validation updates

app/models/sample_manifest.rb
```diff
@@ -130,27 +128,17 @@ def default_filename
...
   scope :pending_manifests,
         -> {
-          order('sample_manifests.id DESC')
-            .joins(
-              # rubocop:todo Layout/LineLength
-              'LEFT OUTER JOIN documents ON documentable_type="SampleManifest" AND documentable_id=sample_manifests.id AND documentable_extended="uploaded"'
-              # rubocop:enable Layout/LineLength
-            )
-            .where('documents.id IS NULL')
+          order(id: :desc).includes(:uploaded_document).references(:uploaded_document).where(documents: { id: nil })
```
JamesGlover, line 133: Replace SQL with basic Arel.

app/models/sample_manifest.rb
```diff
@@ -130,27 +128,17 @@ def default_filename
...
-            .joins(
-              # rubocop:todo Layout/LineLength
-              'LEFT OUTER JOIN documents ON documentable_type="SampleManifest" AND documentable_id=sample_manifests.id AND documentable_extended="uploaded"'
-              # rubocop:enable Layout/LineLength
-            )
-            .where('documents.id IS NOT NULL')
+          order(updated_at: :desc)
+            .includes(:uploaded_document)
+            .references(:uploaded_document)
+            .where.not(documents: { id: nil })
```
JamesGlover, line 141: Replace SQL with basic Arel

app/models/sample_manifest/library_tube_behaviour.rb
```diff
@@ -17,6 +17,10 @@ def initialize(manifest)
       @tubes = []
     end
 
+    def generate
+      @tubes = generate_tubes(Tube::Purpose.standard_library_tube)
+    end
+
```
JamesGlover, line 23: Moved this up to maintain the same organization with the other behaviour classes.

app/models/tag_substitution/substitution.rb
```diff
@@ -19,7 +19,7 @@ class TagSubstitution::Substitution
   delegate :disable_match_expectation, to: :tag_substituter, allow_nil: true
 
   delegate :friendly_name, to: :sample, prefix: true
-  validates :sample_id, :library_id, presence: true
+  validates :sample_id, presence: true
```
JamesGlover, line 22: We don't generate a library id for the long read manifests, as they aren't libraries. But we *do* update the aliquot attributes, like library-type (to indicate what library will be made). This triggers the tag substitution mechanism on re-upload.

This removes the need for library id to be provided, but will still match against it. (So we'll only be able to update the attributes for non-library aliquots)

app/sequencescape_excel/sequencescape_excel/column.rb
```diff
@@ -91,6 +91,7 @@ def attribute_value(detail)
     end
 
     def specialised_field?
+      # We can't use const_defined? here as we want to make sure we trigger rails class loading
```
JamesGlover, line 94: I broke things by trying to be clever. Exception handling isn't the fastest in Ruby, and const_defined? is ten times faster than rescuing the exception. However it fails to trigger the Rails class loading behaviour, so will never trigger loading of the specialised field classes in test and development. I added this as a warning to either my future self, or other moments of hubris.

db/migrate/20220128112324_add_uniqueness_index_to_library_type_name.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Library types should be unique, otherwise things get a little confusing
+class AddUniquenessIndexToLibraryTypeName < ActiveRecord::Migration[6.0]
+  def change
+    add_index :library_types, :name, unique: true
+  end
+end
```
JamesGlover, line 8: Database backing for the added constraint. Good practice, and Rubocop actually insists on it now.

features/api/library_creation_requests.feature
```diff
@@ -19,7 +19,7 @@ And I have a "full" authorised user with the key "cucumber"
```
JamesGlover, line 19: So what follows is a lot of fixes for tests. Unfortunately the added uniqueness validation broke things, as the tests kept re-creating the 'Standard' library type.
The fixes entailed:
- Don't automatically associate library_type and library type validators with requests
- Explicitly do so when we need to test them

features/api/library_creation_requests.feature
```diff
@@ -19,7 +19,7 @@ And I have a "full" authorised user with the key "cucumber"
 
     Given I have an active study called "Testing the sequencing requests API"
     And the UUID for the study "Testing the sequencing requests API" is "11111111-2222-3333-4444-000000000000"
-    And I have a library tube of stuff called "tube_1"
```
JamesGlover, line 22: This step actually creates an empty library tube. There was a todo above it, written ten years ago, saying perhaps it should be a full tube. I've updated the step text to reflect what it actually does, and removed the todo.

features/support/step_definitions/5070230_requesting_additional_hiseq_sequencing_steps.rb
```diff
@@ -1,24 +1,21 @@
 # frozen_string_literal: true
```
JamesGlover, line 1: I had to add the library type association with the request type in here, so just tidied up a bit. Modern Cucumber doesn't rely quite so heavily on regexs.

lib/record_loader/library_type_loader.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
```
JamesGlover, line 1: Library type record loaders are nice and simple, and are pretty much just the template from record-loader itself.

lib/tasks/test.rake
```diff
@@ -25,10 +25,8 @@ namespace :test do
     desc 'Verify that all FactoryBot factories are valid'
     task lint: :environment do
       require 'factory_bot'
-      Dir
-        .glob(File.expand_path(File.join(Rails.root, %w[spec factories ** *.rb])))
-        .sort
-        .each { |factory_filename| require factory_filename }
+
```
JamesGlover, line 28: Rubocop

spec/factories/request_type_factories.rb
```diff
@@ -8,6 +8,16 @@
     end
   end
 
+  trait :with_library_types do
```
JamesGlover, line 11: Extract into a trait, rather than always doing it.

spec/models/parsers/cardinal_pbmc_count_parser_spec.rb
```diff
@@ -16,9 +16,7 @@
 # DN871908M:H1	1029	1940000	8.37	74.00%	357	675000	2.35	1386	2610000	8.57
 
 def read_file(filename)
-  content = nil
-  File.open(filename, 'r') { |fd| content = fd.read }
```
JamesGlover, line 20: I'm not quite sure why I ended up in this file.

JamesGlover: Ahh yes, the well factory had briefly broken. And I was tidying up while I investigated. I then realised I'd be better off fixing factory_bot:lint as a first port of call.

spec/spec_helper.rb
```diff
@@ -46,14 +46,15 @@
...
 Capybara.register_driver :headless_chrome do |app|
-  enable_chrome_headless_downloads(Capybara.drivers[:selenium_chrome_headless].call(app))
+  Capybara.drivers[:selenium_chrome_headless].call(app).tap { |driver| enable_chrome_headless_downloads(driver) }
 end
 
 def enable_chrome_headless_downloads(driver)
   driver.options[:options].add_preference(:download, default_directory: Capybara.save_path)
-  driver.browser.download_path = Capybara.save_path
-  driver
 end
```
JamesGlover, line 58: I lost a day to this problem, and didn't manage to get to the bottom of it. However I:
1) Created an issue so at least other people can track their findings
2) Removed some legacy stuff that was acting as a red-herring

stevieing (approved):

PR #3513, opened 2022-02-04
Dpl098-2 Config for submission template for LCM pipeline

emrojo:
Closes #

*
*
* ...


app/resources/api/v2/request_resource.rb
```diff
@@ -66,6 +67,10 @@ def pre_capture_pool_id
         _model.pooled_request&.pre_capture_pool_id
       end
 
+      def library_type
```
stevieing, line 70: You can also use try which returns nil if it does not exist e.g. `_model.try(:library_type)`

db/seeds/9998_submission_templates.rb
```diff
@@ -128,6 +128,15 @@
 ].freeze
 
 DEV_ONLY = [
```
stevieing, line 130: Can these comments be removed?

stevieing (approved):
All good. Just a couple of comments.

PR #3521, opened 2022-02-17
library plate manifest fix

KatyTaylor:
don't think this 'name' is meant to be here
- means that the 'library_type' column on the library plate manifest is filled with only long read library types
- 'name' doesn't appear on the config for any of the other columns


KatyTaylor:
closing as I did a hotfix, I'll pull in master instead

PR #3533, opened 2022-02-23
Fix long read manifest upload

JamesGlover:

* Removes the ability to customize a column 'name' as this results in the later column definitions over-writing earlier ones.
* Add updates: attribute to allow over-riding behaviour instead.

### DIFF
--- a/app/sequencescape_excel/sequencescape_excel/column.rb
+++ b/app/sequencescape_excel/sequencescape_excel/column.rb
@@ -10,6 +10,7 @@ module SequencescapeExcel
     include Helpers::Attributes
 
     setup_attributes :name,
+                     :updates,
                      :heading,
                      :number,
                      :type,
@@ -39,6 +40,7 @@ module SequencescapeExcel
 
     def initialize(attributes = {})
       super(default_attributes.merge(attributes))
+      self.updates ||= name
     end
 
     ##
@@ -46,7 +48,6 @@ module SequencescapeExcel
     # create a new validation object
     def validation=(validation)
       return if validation.nil?
-
       @validation = validation.is_a?(Hash) ? Validation.new(validation) : validation.dup
     end
 
@@ -79,11 +80,11 @@ module SequencescapeExcel
     end
 
     def metadata_field?
-      @metadata_field ||= Column.sample_metadata_model.respond_to?(name) unless specialised_field?
+      @metadata_field ||= Column.sample_metadata_model.respond_to?(updates) unless specialised_field?
     end
 
     def update_metadata(metadata, value)
-      metadata.send("#{name}=", value) if metadata_field?
+      metadata.send("#{updates}=", value) if metadata_field?
     end
 
     def attribute_value(detail)
@@ -173,7 +174,7 @@ module SequencescapeExcel
     attr_reader :attribute
 
     def classify_name
-      "SpecialisedField::#{name.to_s.delete('?').classify}"
+      "SpecialisedField::#{updates.to_s.delete('?').classify}"
     end
   end
 end
--- a/app/sequencescape_excel/sequencescape_excel/column.rb
+++ b/app/sequencescape_excel/sequencescape_excel/column.rb
@@ -97,7 +97,7 @@ module SequencescapeExcel
     end
 
     def specialised_field
-      @specialised_field ||= SequencescapeExcel.const_get(classify_name, false)
+      @specialised_field ||= SequencescapeExcel::SpecialisedField.const_get(classify_name, false)
     rescue NameError
       nil
     end
@@ -144,7 +144,7 @@ module SequencescapeExcel
       attr_reader :arguments
 
       def initialize(args, key, default_conditional_formattings)
-        @arguments = args.reverse_merge(name: key)
+        @arguments = args.merge(name: key)
         combine_conditional_formattings(default_conditional_formattings)
       end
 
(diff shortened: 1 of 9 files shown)


spec/data/sample_manifest_excel/columns.yml
```diff
@@ -741,8 +741,8 @@ disease:
   unlocked: true
   conditional_formattings:
     empty_cell:
-donor_id2:
-  name: :donor_id
+legacy_donor_id:
```
JamesGlover, line 744: I renamed this to make it clearer what it is. Plus I have a horrible feeling that we've probably been inserting this heading into our manifests, rather than the correct one.

config/sample_manifest_excel/columns.yml
```diff
@@ -783,8 +784,8 @@ disease:
   unlocked: true
   conditional_formattings:
     empty_cell:
-donor_id2:
-  name: :donor_id
+legacy_donor_id:
```
JamesGlover, line 787: I renamed this to make it clearer what it is. Plus I have a horrible feeling that we've probably been inserting this heading into our manifests, rather than the correct one.

config/sample_manifest_excel/columns.yml
```diff
@@ -181,6 +181,7 @@ library_type:
     is_error:
 library_type_long_read:
   heading: LIBRARY TYPE
+  updates: library_type
```
JamesGlover, line 184: Restore the upload behaviour with the new route

config/sample_manifest_excel/columns.yml
KatyTaylor: Yes, I tested before and after your change. With the Default Plate manifest, which has the `:donor_id` column:
- Before your change, heading was DONOR ID (required for cancer samples)
- After your change, heading was DONOR ID (required for EGA)
- Both uploaded into the donor_id field on sample_metadata

So the column in several manifests will have been displaying the wrong column heading, but at least the data is in the right place.

Is it worth actually removing this `:legacy_donor_id` column, as I don't think it's used anywhere now you've removed it from the Heron manifest?

Also, I searched for `donor_id2` in the repo and there are still quite a few references hanging around in the spec folder.

config/sample_manifest_excel/manifest_types.yml
```diff
@@ -679,7 +679,6 @@ heron:
     - :subject
     - :disease
     - :sample_ebi_accession_number
-    - :donor_id2
```
KatyTaylor, line 682: Why was this not appearing in the Heron manifest?

app/sequencescape_excel/sequencescape_excel/column.rb
```diff
@@ -143,7 +144,7 @@ class ArgumentBuilder
       attr_reader :arguments
 
       def initialize(args, key, default_conditional_formattings)
-        @arguments = args.reverse_merge(name: key)
+        @arguments = args.merge(name: key)
```
KatyTaylor, line 147: Wonder if this mattered? It was changed in the other direction here - https://github.com/sanger/sequencescape/commit/af0da03dd966f64463878d83c8f128aae737a591#diff-0622c2fd372210896155392045c637aa66eb184af2a6c9768e42e3f79e33df5d

KatyTaylor (changes requested):

config/sample_manifest_excel/columns.yml
JamesGlover: Legacy_donor_id will allow the upload of manifests using the old column headers, and ensure donor_id still gets correctly updated. I did consider adding a 'legacy_headers' section to the config instead, but this would require a few more changes, including to code that is currently very general purpose.

config/sample_manifest_excel/manifest_types.yml
JamesGlover: Because the column configuration contained a name: attribute, which meant it was over-writing the donor_id column definition instead. It appears that if a column definition is missing, we just skip the column. (Didn't look into exactly what happens here too closely. )

app/sequencescape_excel/sequencescape_excel/column.rb
JamesGlover: Hah, well that was the cause of the bug, or at least part of it. As we then index the columns by name, and use that for the lookup. Hence the later definitions over-writing the first.

Looks like I introduced it when attempting to ensure that donor_id2 updated donor_id, something now handled by 'updates'.

config/sample_manifest_excel/columns.yml
KatyTaylor: I see, thanks. Could you add a comment in the yml to explain why `legacy_donor_id` is still present? I can see someone removing it in the future.

JamesGlover: Just occurred to me that I need to do something similar for the date columns in the compliance branch.

config/sample_manifest_excel/columns.yml
```diff
@@ -856,3 +850,14 @@ priority:
   conditional_formattings:
     empty_cell:
     is_number:
+# LEGACY COLUMNS
+# If a column heading changes, we still need to map historical manifests back
+# to the correct entry, as manifests may be uploaded, or re-uploaded a long time
+# after originally generated.
```
KatyTaylor, line 856: nice :)

KatyTaylor (approved):

JamesGlover:
For future reference:
This bug actually had wider repercussions than first assumed, as it affects *All* library manifests. This is as all columns map back to the long_read_library_type column on upload. Not a problem when it works, as the behaviour is identical, but is a problem otherwise.

Might be worth a revisit to add some validation to make sure that all columns with the same header have the same 'update' behaviour.

PR #3536, opened 2022-02-24
Experiment with vite

JamesGlover:
This drops webpacker in favour of vite.

I'm also in the process of migrating our sprockets dependencies to vite, which should modernise the legacy js.

Note: This drops babel from the build pipeline in favour of @vitejs/plugin-legacy however I'm not sure what the browser support is for this, so we may run into issues with our XP users. (I really wish we could ignore XP)


config/initializers/content_security_policy.rb
```diff
@@ -12,28 +12,36 @@
...
-  #   # policy.report_uri "/csp-violation-report-endpoint"
 
   # Snippet provided after running
   # `bundle exec rails webpacker:install:vue`
   # > You need to enable unsafe-eval rule.
   # > This can be done in Rails 5.2+ for development environment in the CSP initializer
   # > config/initializers/content_security_policy.rb with a snippet like this:
   if Rails.env.development?
-    policy.script_src :self, :https, :unsafe_eval
+    # Also allow @vite/client to hot reload javascript changes in development
```
stevieing, line 22: Hot reloads. Whoop!

package.json
```diff
@@ -10,15 +9,20 @@
     "datatables.net-rowreorder-bs4": "^1.2.8",
     "jquery": "^3.5.0",
```
stevieing, line 10: What do we still use jquery for? Is there legacy javascript code that is using it?

stevieing, line 10: What do we still use jquery for? Is there legacy javascript code that is using it?

stevieing (approved):

package.json
JamesGlover: Yep, some legacy stuff, most of which is just the selectors, so fairly easy to update. 
However datatables also has a dependency on jquery, as does the version of bootstrap we are using.When it comes to updating bootstrap I know the newer versions don't support the latest versions of FF and Chrome to support XP, but I don't know if that is 'catastrophically bad rendering' or 'just doesn't look quite as nice'.

PR #3542, opened 2022-03-01
Dpl 048 1 fix root sample ids

Jonnie-Bevan:
Script to run in the Rails console. Tested locally. Basically just takes the original ID from the CSV file, searches for it in the DB and if it exists, replaces it with the new ID from the CSV file.


sdjmchattie:
Looks good to me.  Don't worry about linting etc.  This is a one off script and not getting merged into the code base.

data/data_fixes/fix_sample_ids.rb
```diff
@@ -0,0 +1,34 @@
...
+success_ids = Array.new()
+
+data.each_with_index do |row, index|
+    if index % 1000 == 0
+        puts "."
+    end
+    samples = Sample::Metadata.where("sample_description" => row["original_root_sample_id"])
+    if samples != []
+        samples.each do |sample|
+            puts "updating: #{sample.sample_description}"
```
emrojo, line 17: I think we could also log the sampleMetadata id so we know the specific row we are changing

data/data_fixes/fix_sample_ids.rb
```diff
@@ -0,0 +1,34 @@
...
+            sample.sample_description = row["root_sample_id"]
+            saved = sample.save
```
emrojo, line 19: When we test in uat we have to check if this save will trigger updating the warehouse too.

data/data_fixes/fix_sample_ids.rb
```diff
@@ -0,0 +1,34 @@
...
+            if saved
+                success_ids << row["original_root_sample_id"]
+            else
+                failed_ids << row["original_root_sample_id"]
+            end
```
emrojo, line 24: I think it would make more sense to store the id of the sample_metadata where the update failed; if not we would have a list that will contain several times the same root_sample_id without indication of which specific sample failed to update

data/data_fixes/fix_sample_ids.rb
emrojo: why we don't write these lines in in the log file too? They seem they could be useful

emrojo (commented):
Some coments

data/data_fixes/fix_sample_ids.rb
```diff
@@ -0,0 +1,34 @@
...
+logger = Logger.new("sample_ids_log.log")
+data = CSV.parse(File.read("data/data_fixes/root_sample_ids.csv"), headers: true)
+failed_ids = Array.new()
+success_ids = Array.new()
+
+data.each_with_index do |row, index|
+    if index % 1000 == 0
+        puts "."
+    end
+    samples = Sample::Metadata.where("sample_description" => row["original_root_sample_id"])
```
KatyTaylor, line 14: It will be interesting to test this on training - I think it will be v slow, because sample_description isn't indexed (and can't be I think because it's a long text field), but I can't think of a better option 😆

KatyTaylor: I just tested this approach with a single record on UAT, and it updated the 'updated_at' timestamp, and updated the MLWH 👍

KatyTaylor (approved):

PR #3547, opened 2022-03-04
Remove sprockets

JamesGlover:
This builds on the removal of webpacker https://github.com/sanger/sequencescape/pull/3536 in attempts to unify our entire asset pipeline.

Things get a bit more complicated here, as there is a lot more to move, and some of it is pretty ancient.

I actually found a fair bit of broken JS while doing this, and either fixed it up (where it appeared useful) or dropped it where it was superfluous.

Still some things to fix up here.


JamesGlover:
I've made this ready for review. It still uses sprockets for Images and CSS, but gets rid of its use for js.

JamesGlover:
And back to draft, as I think I can do some useful cherrypicking to break this one down. (Ironically to remove flexible cherrypicking)

app/frontend/entrypoints/organism_validation.js
```diff
@@ -0,0 +1,120 @@
+/*====================================================================
```
JamesGlover, line 1: Actually migrated from:
app/assets/javascripts/organism_validation.js

app/frontend/entrypoints/organism_validation.js
```diff
@@ -0,0 +1,120 @@
...
+ *  Author: Tony Cox (avc@sanger.ac.uk)
```
stevieing, line 2: Tony Cox!!!

config/application.rb
```diff
@@ -1,6 +1,21 @@
...
+require 'active_storage/engine'
+require 'action_controller/railtie'
+require 'action_view/railtie'
+require 'action_mailer/railtie'
+require 'active_job/railtie'
+require 'action_cable/engine'
+require 'action_mailbox/engine'
+require 'action_text/engine'
+require 'rails/test_unit/railtie'
+# require 'sprockets/railtie'
```
stevieing, line 18: I guess this comment can be removed?

stevieing (approved):

config/application.rb
JamesGlover: I left it in there for reference ; it is part of the standard 'rails/all' inclusions and I wanted the two to be comparable to assist with upgrades. Sprockets is dropped in the default Rails 7 install though, so will definitely be able to remove it then.

We can probably also kill a few others of these, but that is probably for another PR.

JamesGlover:
Cool, I'll see if I can get a test build and deploy into training or UAT on Tuesday, before I go ahead and merge this, just in case I break the build unexpectedly.

PR #3551, opened 2022-03-08
Hotfix - new barcode format for Cardinal

KatyTaylor:
The customer has said that the barcode middle section, previously said to be '046', could be a different number. This is because it denotes the study id (their end), and we could receive some samples that are under a different primary study.

They have said that the best option would be to make the "S2-" prefix fixed and the rest variable.

It looks like the Cardinal team have a workaround for the sample that came in today with an unexpected barcode format (relabelling as a different barcode), but I'm putting this in as a hotfix anyway in case that doesn't work or we get another one tomorrow.

### DIFF
--- a/app/models/barcode/format_handlers.rb
+++ b/app/models/barcode/format_handlers.rb
@@ -503,4 +503,15 @@ module Barcode::FormatHandlers
   class EastLondonGenesAndHealth < BaseRegExBarcode
     self.format = /\A(?<prefix>S2)-046-(?<number>\d+)\z/
   end
+
+  # Support for East London Genes and Health
+  # As part of the Cardinal pipeline, as above
+  # Expected formats:
+  # S2-nnn-nnnnnn
+  # where n is a digit
+  # This is needed in addition to the above because the middle section represents the study
+  # in ELGH's system, and some participants come in under a different primary study.
+  class EastLondonGenesAndHealthV2 < BaseRegExBarcode
+    self.format = /\A(?<prefix>S2)-\d+-(?<number>\d+)\z/
+  end
 end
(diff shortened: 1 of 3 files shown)


app/models/barcode/format_handlers.rb
```diff
@@ -503,4 +503,15 @@ class UkBiocentreV7 < BaseRegExBarcode
...
+
+  # Support for East London Genes and Health
+  # As part of the Cardinal pipeline, as above
+  # Expected formats:
+  # S2-nnn-nnnnnn
+  # where n is a digit
+  # This is needed in addition to the above because the middle section represents the study
+  # in ELGH's system, and some participants come in under a different primary study.
+  class EastLondonGenesAndHealthV2 < BaseRegExBarcode
+    self.format = /\A(?<prefix>S2)-\d+-(?<number>\d+)\z/
```
andrewsparkes, line 515: that doesn't restrict it to 3 digits in the middle, is that ok?

andrewsparkes (approved):

app/models/barcode/format_handlers.rb
sdjmchattie: If you want it to, just put `{3}` after `\d` instead of `+`

KatyTaylor: I thought I'd leave it more flexible, as they didn't actually say it would always be 3 digits

KatyTaylor: Actually they kind of did. I don't know. I'll leave it open.

emrojo: There is a part of the barcode that is not classified as <prefix> or <number>, why it couldnt go as prefix?

KatyTaylor: Hi Eduardo, sorry I think I merged this just before you commented, or before I saw your comment. I didn't quite know where to put the middle section of the barcode. The other barcode formats seemed to have a static prefix, and this middle bit is dynamic. I don't know if that matters!

PR #3553, opened 2022-03-08
Ensure styles apply to column types

JamesGlover:
Fixes UAT issue:

Validation on numeric column types was failing as excel wasn't coercing the new text type.

I've extended the styles option to allow styles to be composed, avoiding needing to add every possible combination.


app/sequencescape_excel/sequencescape_excel/download.rb
```diff
@@ -1,36 +0,0 @@
-# frozen_string_literal: true
```
JamesGlover, line 1: This file wasn't needed, as we don't inherit from it. So I've removed it to avoid confusion.

app/sequencescape_excel/sequencescape_excel/helpers/worksheet.rb
```diff
@@ -11,8 +11,7 @@ module Worksheet
           border: {
             style: :thin,
             color: '00'
-          },
-          format_code: '@'
```
JamesGlover, line 15: I've remove this as the default type, and now it only applies to 'string' columns.

app/sequencescape_excel/sequencescape_excel/helpers/worksheet.rb
```diff
@@ -28,6 +27,12 @@ module Worksheet
             color: '00',
             edges: %i[left right top bottom]
           }
+        },
+        string: {
+          format_code: '@'
+        },
+        integer: {
```
JamesGlover, line 34: Integer rendered as such. I haven't applied custom styles to floats, as we'll probably have a range of approaches wrt decimal places and the like.

app/sequencescape_excel/sequencescape_excel/helpers/worksheet.rb
```diff
@@ -44,10 +49,21 @@ def styles
...
       def create_styles
         styles.tap { |s| STYLES.each { |k, style| s[k] = Style.new(workbook, style) } }
       end
 
+      private
+
+      def generate_style(components)
+        style = components.reduce({}) { |store, component| store.merge!(STYLES.fetch(component, {})) }
+        Style.new(workbook, style)
+      end
```
JamesGlover, line 65: Build a style by merging all those specified.

We ignore unrecognised styles here so as to let new types (and even locked) fall back to the default styles.

app/sequencescape_excel/sequencescape_excel/helpers/worksheet.rb
KatyTaylor: What does the @ do?

app/sequencescape_excel/sequencescape_excel/worksheet/data_worksheet.rb
```diff
@@ -1,118 +0,0 @@
-# frozen_string_literal: true
-
-module SequencescapeExcel
```
KatyTaylor, line 3: Was this not needed as the one in the sample_manifest_excel folder is used instead?

KatyTaylor (approved):

app/sequencescape_excel/sequencescape_excel/column.rb
```diff
@@ -79,6 +79,10 @@ def unlocked?
       unlocked
     end
 
+    def style
+      [unlocked? ? :unlocked : :locked, type]
```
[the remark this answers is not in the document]
JamesGlover: It doesn't have any specified styles. At one point I was returning nil, but then just ended up needing to add support for missing styles anyway, so it felt more useful to pass through a token that we could potentially hang behaviour off.

In particular, there is an issue with pasted content being automatically locked, which I think is related to properties of the 'default' style. So we might be better explicitly locking cells, rather than unlocking them.

app/sequencescape_excel/sequencescape_excel/download.rb
JamesGlover: When I added spreadsheet generation for bulk submission, I copied a lot of the behaviour here, and originally intended to inherit from this. But this never really happened for downloads, so this file just sat there orphaned and I didn't clean it up at the time.

app/sequencescape_excel/sequencescape_excel/helpers/worksheet.rb
JamesGlover: Marks it as a text field. I should probably add a comment, as excel format_codes are somewhat impeneterable

PR #3555, opened 2022-03-09
Dpl269 duplicate pacbio runs

emrojo:
Closes DPL269

* Re-forward redundant Pacbio messages


JamesGlover:
This will immediately revert the problem if the runs ever get touched. If this wasn't intentional this is okay, as the message would get rejected once the constraints are in place, but it will be impossible to update the batches intentionally.

Is there a reason you didn't update the actual message template?

lib/tasks/rebroadcast_pac_bio_multiple_requests_run.rake
```diff
@@ -0,0 +1,74 @@
...
+                  with_association(:tag_group) { map_attribute_to_json_attribute(:name, 'tag_set_name') }
+                end
+              end
+            end
+          end
+        end
+      end
+    end
+    
+    list_of_runs = [70489, 99999]
```
JamesGlover, line 67: Wow, really just these two. I think I'd only looked at the number of rows back from the original query, and assumed it mapped to more.

lib/tasks/rebroadcast_pac_bio_multiple_requests_run.rake
```diff
@@ -0,0 +1,74 @@
...
+namespace :pac_bio_run do
+  desc 'Update all PacBio runs in the warehouse where there is a duplicate for the index ' \
+  '[id_pac_bio_run_lims, well_label, comparable_tag_identifier, comparable_tag2_identifier]' \
+  ' because there is more than one sequencing request for the run'
+
+  task rebroadcast_duplicates: :environment do
+    # This task redefines Receptacle to be able to only use the last active request
+    Receptacle.class_eval do
+      def most_recent_active_requests_as_target_group_by_same_source
+        requests_as_target.where(state: ['started']).group(:asset_id).having('created_at=max(created_at)')
```
JamesGlover, line 14: Might want to have a quick zoom about this, as it seems to be trying to do something I've struggled with for a while,  but in my experience this approach shouldn't work (although will still be fine for this purpose).

emrojo: I changed this to get the most recent request no matter the state (also the condition 'having to' I found out was wrong).

PR #3577, opened 2022-03-30
Dpl331 fail rtq plate with failed wells

andrewsparkes:
See https://github.com/sanger/limber/issues/1038

Migration to increase size of contents field on state_changes table


db/schema.rb
```diff
@@ -259,7 +259,7 @@
     t.index ["updated_at"], name: "index_batches_on_updated_at"
   end
 
-  create_table "bkp_lab_events", id: false, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
```
JamesGlover, line 262: You've accidentally included some charset changes on other tables. These *should* be utf8mb4, and are in production. 

Not 100% sure what's happened here.

Both the mysql8 and mysql5.6 documentation states:

> utf8: An alias for utf8mb3.

So your dump is matching that, but being more explicit (which is good), just explicitly wrong. However if I try and red-dump the schema on my system, its still spitting out the vague 'utf8'. 

I'm assuming it's probably something to do with MySQL versions.

Any chance you could correct these to all be utf8mb4 to match production.

We'll also need to see if we can get utf8mb4 to be the default character set (can do that separately) for new tables in development, otherwise I think we'll keep chasing this issue. Pretty sure I made that change in prod, which is why the tables are okay there.

JamesGlover (changes requested):
There's some accidental schema 'changes' included in the commit. I think its a case of your system being nice and explicit in the dump, but unfortunately its being explicitly 'wrong'. (Although unless we end up with a country called 😀 I can't see much scope for problems)

Still, probably should fix them up and follow up with a proper fix a little later. (I'll see if I can do it tomorrow morning)

db/schema.rb
andrewsparkes: Ok removed changes as discussed

PR #3580, opened 2022-03-31
Update development schema from production

JamesGlover:
We've had some minor issues with different default column types
on production vs development. Production is using the more robust
utf8mb4, which supports emoji. While we don't expect many 4 byte
characters, we should be consistent.

Update default database encoding

Include migration to fix up schema

Closes #

*
*
* ...

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -13,9 +13,10 @@ mysql: &MYSQL
   adapter: mysql2
   username: <%= ENV.fetch('DBUSERNAME','root') %>
   password: <%= ENV['DBPASSWORD'] %>
-  encoding: utf8
+  encoding: utf8mb4
+  collation: utf8mb4_unicode_ci
   properties:
-    characterSetResults: utf8
+    characterSetResults: utf8mb4
   pool: 5
   timeout: 5000
   reaping_frequency: 600
--- a/db/migrate/20220331093754_fix_development_encoding_issues.rb
+++ b/db/migrate/20220331093754_fix_development_encoding_issues.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+# The default database encoding for the development database was wrong
+# We've fixed it up for newly created databases, but rather than forcing
+# developers to rebuild their database we'll set up a special development
+# only migration to keep things clean. This will:
+# - Update the default encoding
+# - Fix the incorrect encoding on any tables
+# Production and UAT are both correct, so we don't try and do anything there to
+# be safe.
+class FixDevelopmentEncodingIssues < ActiveRecord::Migration[6.0]
+  include MigrationExtensions::EncodingChanges
+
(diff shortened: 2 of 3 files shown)


config/database.yml
```diff
@@ -13,9 +13,10 @@ mysql: &MYSQL
   adapter: mysql2
   username: <%= ENV.fetch('DBUSERNAME','root') %>
   password: <%= ENV['DBPASSWORD'] %>
-  encoding: utf8
+  encoding: utf8mb4
+  collation: utf8mb4_unicode_ci
   properties:
-    characterSetResults: utf8
+    characterSetResults: utf8mb4
```
JamesGlover, line 19: This is what sets the defaults for the database, but it only happens on db:create, so I've also included a migration to update any dev databases.

db/migrate/20220331093754_fix_development_encoding_issues.rb
```diff
@@ -0,0 +1,52 @@
...
+  include MigrationExtensions::EncodingChanges
+
+  def up
+    say 'Checking Environment'
+    if migrate?
+      say "#{Rails.env} found, proceeding"
+      update_encoding
+      update_tables
+    else
+      say "#{Rails.env} found, skipping"
```
JamesGlover, line 21: The migration will only run in development/test. This is as production et al are setup correctly and we want to avoid accidentally introducing changes.

db/migrate/20220331093754_fix_development_encoding_issues.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  def update_encoding
+    execute("SET  @@character_set_database = 'utf8mb4', @@collation_database = 'utf8mb4_unicode_ci'")
```
JamesGlover, line 26: This ensures that any new tables created will have the correct encoding.

db/migrate/20220331093754_fix_development_encoding_issues.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def update_tables
+    change_encoding('bkp_lab_events', from: 'utf8', to: 'utf8mb4')
+    change_encoding('isndc_countries', from: 'utf8', to: 'utf8mb4')
+    change_encoding('pick_lists', from: 'utf8', to: 'utf8mb4')
+    change_encoding('racked_tubes', from: 'utf8', to: 'utf8mb4')
+    change_encoding('sample_compounds_components', from: 'utf8', to: 'utf8mb4')
+    change_encoding('tube_rack_statuses', from: 'utf8', to: 'utf8mb4')
+  end
```
JamesGlover, line 36: Whereas these are the tables which were out of sync.

andrewsparkes (approved):
Nice

PR #3593, opened 2022-04-13
DPL-309 Add lifespan field to the V2 API Purpose resource

KatyTaylor:
- so it can be accessed through the V2 API by Asset Audits


harrietc52 (approved):

KatyTaylor:
> Could any test be updated too? :)

Good point, I think it was such a tiny change I forgot about tests! I will check.

PR #3606, opened 2022-05-18
Clean up gemfile

JamesGlover:
- Remove transient dependencies from Gemfile
We had a few transient dependencies which we had to declare explicitly for historic reasons.

- Move yard-activerecord out of lint gemset
- Remove transactional_capybara as rails now handles this natively

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-next_rails = ENV['BUNDLE_GEMFILE']&.end_with?('GemfileNext')
+next_rails = ENV.fetch('BUNDLE_GEMFILE', 'Gemfile')&.end_with?('GemfileNext')
 
 source 'https://rubygems.org'
 
@@ -112,12 +112,13 @@ group :development do
 
   # Automatically generate documentation
   gem 'yard', require: false
+  gem 'yard-activerecord', '~> 0.0.16'
 
   # MiniProfiler allows you to see the speed of a request conveniently on the page.
   # It also shows the SQL queries performed and allows you to profile a specific block of code.
   gem 'rack-mini-profiler'
 
-  # find unused routes and controller actions by runnung `rake traceroute` from CL
+  # find unused routes and controller actions by running `rake traceroute` from CL
   gem 'traceroute'
 
   # Pat of the JS assets pipleine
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-next_rails = ENV['BUNDLE_GEMFILE']&.end_with?('GemfileNext')
+next_rails = ENV.fetch('BUNDLE_GEMFILE', 'Gemfile')&.end_with?('GemfileNext')
```
JamesGlover, line 2: Rubocop got involved.

Gemfile
```diff
@@ -112,12 +112,13 @@ group :development do
 
   # Automatically generate documentation
   gem 'yard', require: false
+  gem 'yard-activerecord', '~> 0.0.16'
```
JamesGlover, line 115: For some reason this was under the linting group.

Gemfile
```diff
@@ -192,24 +192,13 @@ group :test, :cucumber do
...
-
   # Keep webdriver in sync with chrome to prevent frustrating CI failures
   gem 'webdrivers'
 end
 
 group :cucumber do
   gem 'cucumber_github_formatter'
   gem 'cucumber-rails', require: false
-  gem 'mime-types'
-  gem 'rubyzip'
```
JamesGlover, line 212: These are both transient dependencies of other gems that we don't use directly. They were added historically when either we needed to pin the version, or the dependencies were incorrectly declared. Letting bundler resolve this now.

features/support/env.rb
```diff
@@ -7,10 +7,10 @@
...
 
-ENV['RAILS_ENV'] ||= 'cucumber'
+selected_env = ENV['RAILS_ENV'] ||= 'cucumber'
 
-if ENV['RAILS_ENV'] != 'cucumber'
-  puts "You are running the cucumber specs with the #{ENV['RAILS_ENV']} environment."
+if selected_env != 'cucumber'
+  puts "You are running the cucumber specs with the #{selected_env} environment."
   puts "This can cause problems with gem loading. Please use 'cucumber' instead."
 end
```
JamesGlover, line 15: More rubocop

stevieing (approved):

Issue #3589, opened 2022-04-12
DPL-340 Cardinal - re-use compound samples [C=M, V=4]


KatyTaylor:
**User story**
As NPG, in cases where the same multiplexed library tube is loaded onto all 4 lanes, I would like the compound sample names to be consistent across the lanes, so that the standard pipeline knows to merge the CRAM files, and the Cell Ranger software treats them as a single sample.

Similarly, if the same combination of samples is re-sequenced, we would like the compound sample to be re-used, rather than a new one created. 

This behaviour is consistent with how normal samples work - if a sample is re-sequenced, it will appear in the MLWH as the same sample (unless it is re-uploaded in a new manifest). However, depending on what stage it is re-worked from, it may have a different library or pool id in the warehouse.

This is important because it's currently causing an issue with NPG's processing in the production Cardinal pipeline. NPG have agreed to work around the issue for now, and process the existing data more manually, but it's not sustainable once the pipeline is going at full capacity.

**Who are the primary contacts for this story**
Katy, James, Marina, Steven L

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] When compound samples are made, a check is done to see if that combination of samples already has an existing compound sample. If so, the existing one is re-used.
- [ ] The relevant code runs in the context of a sequencing request. Make sure there are no issues caused by race conditions due to this running four times in parallel due to the four lanes in a flowcell.

**Additional context**
This came out of this research story - https://github.com/sanger/sequencescape/issues/3586

Once this is implemented, tested and deployed, NPG (Marina & Steven L) should be informed, so that they can stop doing their workaround.

mgcam:
The same multiplexed library tube can be loaded onto 2 or 3 lanes out of 4 or onto 2 out of 2. We have not seen these cases yet, but this might happen.

harrietc52:
Q:  "I would like the compound sample names to be consistent across the lanes". Does consistent mean the exact same name for all compound samples in each lane? How are they currently named? 

Q: Whats the relationship with tube? Does a multiplexed library tube, contain 1 compound sample, which contains many samples?

Q: When are compound samples made? At which stage?

Q: What does "runs in the context of a sequencing request" mean?

Q: Unsure what this means: "Make sure there are no issues caused by race conditions due to this running four times in parallel due to the four lanes in a flowcell."

PR #3608, opened 2022-05-18
Find a compound sample if it exist on transfer

harrietc52:
Closes https://github.com/sanger/sequencescape/issues/3589

UAT deployment
`ansible-playbook -vD deploy_sequencescape.yml -i environments/uat -e "github_branch=v14.11.0-uat-DPL-320" -e "full_deploy=true reload_nginx=false skip_migrations=true with_quant_files=false"`

### DIFF
--- a/app/models/compound_aliquot.rb
+++ b/app/models/compound_aliquot.rb
@@ -7,7 +7,7 @@
 #   request           A Request for transferring source_aliquots into a target receptacle.
 #                     At time of writing, only used for a SequencingRequest, from a multiplex tube into a lane.
 #   source_aliquots   A list of Aliquots that should be transferred into a single Aliquot in a target Receptacle,
-#                     that containes  contain a single compound sample.
+#                     that contains a single compound sample.
 #                     The list of component samples is derived from source_aliquots.
 #                     Some attributes are transferred from the source aliquots onto the compound aliquot.
 #
--- a/app/models/compound_aliquot.rb
+++ b/app/models/compound_aliquot.rb
@@ -54,13 +54,34 @@ class CompoundAliquot
     errors.add(:base, "#{MULTIPLE_PROJECTS_ERROR_MSG}: #{component_samples.map(&:name)}")
   end
 
+  def find_or_create_compound_sample(source_aliquots)
+    #  Check if a compound sample already exists with the source_aliquot samples
+    existing_compound_sample = find_compound_sample_for_component_samples(source_aliquots)
+
+    @compound_sample = existing_compound_sample || create_compound_sample
+  end
+
+  # Due to previous implementation, there may be multiple compound samples with the provided component samples.
+  # NPG have confirmed we do not need to fix the data where there are multiple compound samples with
+  # the same component samples
+  def find_compound_sample_for_component_samples(source_aliquots)
+    # Get all the component samples
+    component_samples = source_aliquots.map(&:sample)
+
+    # Get all the compound samples of the first component sample
+    # .includes eager loads the component_samples upfront, otherwise you risk lots of hits to the database.
+    compound_samples = component_samples[0].compound_samples.includes(:component_samples).order(id: :desc)
+
+    # Find the latest compound sample with contains the given component samples
+    compound_samples.find { |compound_sample| compound_sample.component_samples == component_samples }
+  end
+
   # Generates the compound sample, under the default study, using the component samples
   def create_compound_sample
-    @compound_sample =
-      default_compound_study.samples.create!(
-        name: SangerSampleId.generate_sanger_sample_id!(default_compound_study.abbreviation),
-        component_samples: component_samples
-      )
+    default_compound_study.samples.create!(
+      name: SangerSampleId.generate_sanger_sample_id!(default_compound_study.abbreviation),
+      component_samples: component_samples
+    )
   end
 
   def aliquot_attributes
--- a/app/models/compound_aliquot.rb
+++ b/app/models/compound_aliquot.rb
@@ -20,9 +20,9 @@ class CompoundAliquot
   MULTIPLE_PROJECTS_ERROR_MSG =
     'Cannot create compound sample due to the component samples being under different projects.'
 
-  attr_accessor :request, :source_aliquots
+  attr_accessor :request, :source_aliquots, :compound_sample
 
-  attr_reader :component_samples, :compound_sample
+  attr_reader :component_samples
 
   validate :tag_depth_is_unique
   validate :source_aliquots_have_same_study
--- a/app/models/compound_aliquot.rb
+++ b/app/models/compound_aliquot.rb
@@ -65,9 +65,6 @@ class CompoundAliquot
   # NPG have confirmed we do not need to fix the data where there are multiple compound samples with
   # the same component samples
   def find_compound_sample_for_component_samples
-    # Get all the component samples
-    component_samples = source_aliquots.map(&:sample)
-
     # Get all the compound samples of the first component sample
     # .includes eager loads the component_samples upfront, otherwise you risk lots of hits to the database.
     compound_samples = component_samples[0].compound_samples.includes(:component_samples).order(id: :desc)
--- a/app/models/compound_aliquot.rb
+++ b/app/models/compound_aliquot.rb
@@ -20,9 +20,9 @@ class CompoundAliquot
   MULTIPLE_PROJECTS_ERROR_MSG =
     'Cannot create compound sample due to the component samples being under different projects.'
 
-  attr_accessor :request, :source_aliquots, :compound_sample
+  attr_accessor :request, :source_aliquots
 
-  attr_reader :component_samples
+  attr_reader :component_samples, :compound_sample
 
   validate :tag_depth_is_unique
   validate :source_aliquots_have_same_study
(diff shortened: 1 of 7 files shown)


app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -50,14 +50,40 @@ def aliquots_by_tags_combination
...
+      compound_aliquot.create_compound_sample
+    end
     target_asset.aliquots.create(compound_aliquot.aliquot_attributes)
   end
+
+  def find_compound_sample_for_components_samples(source_aliquots)
+    # Get all the component samples
+    component_samples = source_aliquots.map(&:sample)
+
+    # Get all the compound samples the first component sample
```
KatyTaylor, line 76:
```suggestion
    # Get all the compound samples of the first component sample
```

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -50,14 +50,40 @@ def aliquots_by_tags_combination
...
+    compound_samples = component_samples[0].compound_samples
+
+    # If there is only 1 compound sample, return it. Otherwise, return the last created compound sample
+    # NPG have confirmed we do not need to fix the data where there are multiple compound samples with
+    # the same component samples
```
KatyTaylor, line 81: I would move this comment from `# NPG have...` onwards to under the `# Due to previous impl..` one below, because I think it makes more sense after reading that.

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -50,14 +50,40 @@ def aliquots_by_tags_combination
...
+    compound_samples.count == 1 ? compound_samples[0] : get_last_compound_sample(compound_samples, component_samples)
```
KatyTaylor, line 82: I think this is a bit wrong, because if there is only one compound sample, you are not checking whether all of its component samples match the ones we care about - you are only doing that check in the `get_last_compound_sample` method?

spec/features/shared_examples/sequencing.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    select('XP', from: 'Workflow (Standard or Xp)')
+    fill_in('Lane loading concentration (pM)', with: 23)
+    fill_in('+4 field of weirdness', with: 'Check stored')
+
+    click_on 'Next step'
+
+    find('#sample-2-checkbox').uncheck
+
+    # Pending question on issue#3225 may be populated with previous v
```
KatyTaylor, line 43: is this meant to say 'with previous value'?

spec/features/shared_examples/sequencing.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
```
KatyTaylor, line 1: Has this code been copy pasted from another spec? If so can we reuse it between multiple specs?

app/models/request/sample_compound_aliquot_transfer.rb
harrietc52: Ah yes, I had just moved this method about thinking it was a refactor, but actually removed that crucial functionalty. will revert 👍

spec/features/shared_examples/sequencing.rb
harrietc52: assume so! will update

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -50,14 +50,36 @@ def aliquots_by_tags_combination
...
+
+  # Due to previous implementation, there may be multiple compound samples with the provided component samples.
+  # NPG have confirmed we do not need to fix the data where there are multiple compound samples with
+  # the same component samples
+  def find_compound_sample_for_component_samples(source_aliquots)
+    # Get all the component samples
+    component_samples = source_aliquots.map(&:sample)
+
+    # Get all the compound samples of the first component sample
+    compound_samples = component_samples[0].compound_samples.order(:id)
```
JamesGlover, line 80: Rather than ordering by id, then calling reverse, you can simply `order(id: :desc)`
I'd also suggest doing `compound_samples.include(:samples).order(id: :desc)` as it will eager load the samples upfront, otherwise you risk lots of hits to the database.

spec/models/request/sample_compound_aliquot_transfer_spec.rb
```diff
@@ -75,15 +75,98 @@
...
+        expect(sequencing_request.target_asset.samples.first.component_samples[0]).to eq samples[0]
+        expect(sequencing_request.target_asset.samples.first.component_samples[1]).to eq samples[1]
+      end
+    end
+
+    context 'when a compound sample exists with different component samples' do
+      let(:compound_sample) { create(:sample, name: 'compound_sample_1') }
+
+      before do
+        compound_sample.update(component_samples: [samples[0], samples[2]])
```
JamesGlover, line 99: Its generally better to pass the samples in to the factory, than mutate it later.

1) It makes it easier to see at a glance the state of the compound_sample, as you only need to look at the let.
2) You won't be caught out if the factory is also creating samples
3) In many cases it is faster, as it reduces the trips to the database - not sure if that will be true in this context though.

In this case you don't actually refer to compound sample outside of the before block, so you can just make the call to the factory there, rather than having a named let block.

spec/models/request/sample_compound_aliquot_transfer_spec.rb
```diff
@@ -75,15 +75,98 @@
...
+        expect { sequencing_request.transfer_aliquots_into_compound_sample_aliquots }.to change(Sample, :count).by(1)
+
+        expect(sequencing_request.target_asset.aliquots.count).to eq(1)
+        expect(sequencing_request.target_asset.aliquots.first.library_type).to eq('Standard')
+        expect(sequencing_request.target_asset.aliquots.first.study).to eq(study1)
+        expect(sequencing_request.target_asset.aliquots.first.project).to eq(project1)
+        expect(sequencing_request.target_asset.aliquots.first.library_id).to eq(54)
+        expect(sequencing_request.target_asset.samples.first.component_samples.count).to eq 2
+        expect(sequencing_request.target_asset.samples.first.component_samples[0]).to eq samples[0]
+        expect(sequencing_request.target_asset.samples.first.component_samples[1]).to eq samples[1]
```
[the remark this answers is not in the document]
harrietc52: `aggregate_failures` is good to know! thanks

app/models/request/sample_compound_aliquot_transfer.rb
```diff
@@ -50,13 +50,15 @@ def aliquots_by_tags_combination
...
+  # For a group of source aliquots, find or create a compound sample containing the component samples
+  # Assign the compound sample to the target asset
   def transfer_into_compound_sample_aliquot(source_aliquots)
     compound_aliquot = CompoundAliquot.new(request: self, source_aliquots: source_aliquots)
     unless compound_aliquot.valid?
       raise Request::SampleCompoundAliquotTransfer::Error, compound_aliquot.errors.full_messages
     end
 
-    compound_aliquot.create_compound_sample
+    compound_aliquot.find_or_create_compound_sample(source_aliquots)
```
KatyTaylor, line 61: I don't think you have to pass in `source_aliquots` here, because you already passed it in on line 56?

app/models/compound_aliquot.rb
```diff
@@ -20,9 +20,9 @@ class CompoundAliquot
   MULTIPLE_PROJECTS_ERROR_MSG =
     'Cannot create compound sample due to the component samples being under different projects.'
 
-  attr_accessor :request, :source_aliquots
+  attr_accessor :request, :source_aliquots, :compound_sample
```
KatyTaylor, line 23: I saw that the above conversation was resolved, so sorry if I missed anything, but I think this `compound_sample` could go back to the `attr_reader` as it doesn't need to be set from outside any more?

app/models/compound_aliquot.rb
```diff
@@ -54,13 +54,34 @@ def source_aliquots_have_same_project
...
+  # the same component samples
+  def find_compound_sample_for_component_samples(source_aliquots)
+    # Get all the component samples
+    component_samples = source_aliquots.map(&:sample)
+
+    # Get all the compound samples of the first component sample
+    # .includes eager loads the component_samples upfront, otherwise you risk lots of hits to the database.
+    compound_samples = component_samples[0].compound_samples.includes(:component_samples).order(id: :desc)
+
+    # Find the latest compound sample with contains the given component samples
```
KatyTaylor, line 75:
```suggestion
    # Find the latest compound sample which contains the given component samples
```

app/models/compound_aliquot.rb
```diff
@@ -54,13 +54,34 @@ def source_aliquots_have_same_project
...
+
+    @compound_sample = existing_compound_sample || create_compound_sample
+  end
+
+  # Due to previous implementation, there may be multiple compound samples with the provided component samples.
+  # NPG have confirmed we do not need to fix the data where there are multiple compound samples with
+  # the same component samples
+  def find_compound_sample_for_component_samples(source_aliquots)
+    # Get all the component samples
+    component_samples = source_aliquots.map(&:sample)
```
KatyTaylor, line 69: Think you don't need to pass `source_aliquots` into this method, and can just use `component_samples` straight away, as you set it in the `initialize` method?

harrietc52: Yes of course. Thanks. Wow, my rails knowledge has really dropped!

PR #3618, opened 2022-05-30
DPL-377 [DRAFT] 'library_complete' and 'pool_released' events

### LINKED ISSUE #3607: DPL-377 'library_complete' and 'pool_released' events [C=M, V=4]
**User story**
As a DNAP team lead I would like the 'library_complete' event to fire when libraries are 'charged and passed' in Limber, as this is more representative of when the library is complete than the current trigger (when the multiplexed library tube is passed), which can be delayed for long periods of time. 

I would also like to add a new event type called 'pool_released', fired when the multiplexed library tube is passed, to allow more detail in the reports.

Liz E, Katy

**Acceptance criteria**
- [x] The library_complete event (Event WH) is triggered by the 'Charge and pass' button being clicked in Limber, instead of the current trigger (which is when the multiplexed library tube is passed)
- [x] The current trigger (when the multiplexed library tube is passed) is retained as an event point - but the event fired here is called 'pool_released' instead.
- [x] Event documentation table at the bottom of this page is updated - https://ssg-confluence.internal.sanger.ac.uk/pages/viewpage.action?pageId=137076202

**References**
This came out of this story - https://github.com/sanger/General-Backlog-Items/issues/169

KatyTaylor:
Duplicate LibraryComplete event to create new event type, PoolReleased
- Change StateChanger::MxTube to fire PoolReleased event rather than LibraryComplete
- Add 'todos' in LibraryComplete and PlateLibraryComplete event classes

Closes #3607

### DIFF
--- a/app/models/broadcast_event/pool_released.rb
+++ b/app/models/broadcast_event/pool_released.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
-# TODO: change the contents of this class - to be fired at 'charge and pass' stage instead.
-class BroadcastEvent::LibraryComplete < BroadcastEvent # rubocop:todo Style/Documentation
-  set_event_type 'library_complete'
+# At time of creation, this event type is fired when multiplexed library tubes are passed.
+# BroadcastEvent::LibraryComplete events used to be fired at this time point, however they were
+# redefined to be fired at the 'charge and pass' stage, because that made more sense to the users.
+class BroadcastEvent::PoolReleased < BroadcastEvent
+  set_event_type 'pool_released'
 
   # Properties takes :order_id
 
--- a/app/models/broadcast_event/plate_library_complete.rb
+++ b/app/models/broadcast_event/plate_library_complete.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
-class BroadcastEvent::PlateLibraryComplete < BroadcastEvent # rubocop:todo Style/Documentation
+# TODO: is this event type actually fired anywhere?
+# BroadcastEvent::PlateLibraryComplete records are present 2015-2017 only
+class BroadcastEvent::PlateLibraryComplete < BroadcastEvent
   set_event_type 'library_complete'
 
   # Properties takes :order_id
--- a/app/models/state_changer/mx_tube.rb
+++ b/app/models/state_changer/mx_tube.rb
@@ -14,7 +14,7 @@ module StateChanger
 
     def generate_events_for(orders)
       orders.each do |order_id|
-        BroadcastEvent::LibraryComplete.create!(seed: labware, user: user, properties: { order_id: order_id })
+        BroadcastEvent::PoolReleased.create!(seed: labware, user: user, properties: { order_id: order_id })
       end
     end
 


app/models/work_completion/plate_completion.rb
```diff
@@ -8,21 +8,28 @@
 # @author Genome Research Ltd.
 #
 class WorkCompletion::PlateCompletion
-  attr_reader :target_plate, :submission_ids
+  attr_reader :target_plate, :submission_ids, :order_ids
```
sdjmchattie, line 11: I know this is an existing issue, but if these two classes (PlateCompletion and TubeCompletion) share attributes and the logic is quite similar, should they have a shared super class?  Or will that not work because of Rails?

sdjmchattie (commented):
Quick review as requested 😁

app/models/work_completion/plate_completion.rb
KatyTaylor: Thanks, I think it would make sense. To be honest I kept reading them as if they inherited from WorkCompletion - so had similar thoughts.

app/models/work_completion/labware_completion.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+# Shared behaviour of WorkCompletion::PlateCompletion and WorkCompletion::TubeCompletion
+#
+# @author Genome Research Ltd.
+#
+class WorkCompletion::LabwareCompletion
```
sdjmchattie, line 7: 👍 Very pleased to see this was possible.  Good work.

KatyTaylor: Yes think it worked out well! Thanks for the suggestion.

app/models/broadcast_event/library_complete.rb
```diff
@@ -1,23 +1,27 @@
 # frozen_string_literal: true
-class BroadcastEvent::LibraryComplete < BroadcastEvent # rubocop:todo Style/Documentation
+
+# Event that indicates library prep is complete.
+# It used to be fired (with slightly different data attached) when the multiplexed library tube was passed.
```
emrojo, line 4: Because there could be more changes in how it works in future, maybe we could only describe how it currently works? Probably how it worked before is not going to be needed

app/models/broadcast_event/library_complete.rb
```diff
@@ -37,5 +41,5 @@ def order
   has_metadata(:order_type) { |_, e| e.order.order_role.try(:role) || 'UNKNOWN' }
   has_metadata(:submission_template) { |_, e| e.order.template_name }
 
-  has_metadata(:team) { |tube, _e| tube.team || 'UNKNOWN' }
+  has_metadata(:team) { |_, e| RequestType.find(e.order.request_types.try(:first))&.product_line&.name || 'UNKNOWN' }
```
emrojo, line 44: Why it needs RequestType.find()? and not e.order.request_types.try(:first) directly? Maybe it needs reload?

app/models/broadcast_event/pool_released.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# At time of writing, this event type is fired when multiplexed library tubes are passed.
+# BroadcastEvent::LibraryComplete events used to be fired at this time point, however they were
+# redefined to be fired at the 'charge and pass' stage, because that made more sense to the users.
```
emrojo, line 5: Same thing, I would only describe the current behaviour in this code, so if there is any change is easier to maintain the coments.

app/models/illumina_htp/mx_tube_purpose.rb
```diff
@@ -48,7 +48,7 @@ def source_plate_scope(tube)
       .joins(wells: :requests)
       .where(
         requests: {
-          target_asset_id: tube.id,
+          target_asset_id: tube.receptacle.id,
```
emrojo, line 51: why the change here? (or why it worked before)?

app/models/plate.rb
```diff
@@ -455,6 +455,8 @@ def update_qc_values_with_parser(parser)
     true
   end
 
+  # Finds the product line (= team) of the requests coming out of this plate's 'stock plate'.
+  # Written at a time when requests weren't recorded on the aliquot, so could be re-written in a less convoluted way.
```
emrojo, line 459: Again I think is historical description which is interesting but seems difficult to maintain over new changes in future.

app/models/work_completion/labware_completion.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    # We don't try and pass failed requests.
+    # I'm not 100% convinced this decision belongs here, and instead
+    # we may want to let the client specify wells to pass, and perform
+    # validation to ensure this is correct. However this increases
+    # the complexity of both the code and the interface, with only
+    # marginal system simplification.
+    upstream_request.pass if upstream_request.may_pass?
+    upstream_request.save!
+  end
```
emrojo, line 46: Very good comments in this file! I think is very useful info

spec/factories/submission_factories.rb
```diff
@@ -67,7 +71,16 @@
     factory :library_order do
       assets { create_list :untagged_well, 1 }
       request_types { [create(:library_request_type).id] }
-      request_options { { fragment_size_required_from: 100, fragment_size_required_to: 200, library_type: 'Standard' } }
+      request_options do
+        {
+          fragment_size_required_from: 100,
+          fragment_size_required_to: 200,
+          library_type: 'Standard',
+          bait_library_name: 'Bait'
```
emrojo, line 79: Why is bait_library_name needed now?

spec/models/broadcast_event/pool_released_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  let(:event) { described_class.create!(seed: tube, user: create(:user), properties: { order_id: order.id }) }
+  let(:subject_hash) { event.as_json['event'][:subjects].group_by(&:role_type) }
+  let(:metadata) { event.as_json['event'][:metadata] }
+
+  before do
+    request1
+    request2
+    library_request
+  end
```
emrojo, line 41: If you defined this variables with let! it creates them immediately code in that part of code, however I've seen other places where this before{} is used because it doesn work exactly that way. Did you try let!() ?

spec/models/state_changer/mx_tube_spec.rb
```diff
@@ -15,11 +15,21 @@
 
   let(:user) { build_stubbed :user }
   let(:customer_accepts_responsibility) { false }
-  let(:labware) { create :tube }
-  let!(:transfer_request) { create :transfer_request, target_asset: labware.receptacle, state: transfer_request_state }
-  let!(:request) { create :request, target_asset: labware.receptacle, state: request_state }
```
emrojo, line 20: Yes like this, probably it was creating problems using let!() :) sorry the previous comment then

emrojo (approved):
All code very good and seems very well tested. I wrote some questions because of my lack of knowledge of how events work and some opinionated suggestions which I dont think are needed to change.

app/models/broadcast_event/library_complete.rb
KatyTaylor: The data up will reflect the old version of the code, up until the time when this goes live - i.e. there will be library_complete events that were fired when the multiplexed library tube was passed. I'm not planning to do any kind of data migration to change these. So I thought it was worth recording somewhere in case it helped with confusion over reports etc?

app/models/broadcast_event/library_complete.rb
KatyTaylor: Ah because the request_types field on Order is just an array of ids - it's not a proper Rails relationship (unless I'm misunderstanding it)

app/models/illumina_htp/mx_tube_purpose.rb
KatyTaylor: I'm pretty sure it's a bug and it didn't work before. I mean, it wouldn't have thrown an exception, but it would have returned the wrong data. The target_asset field is meant to be a receptacle id, not a labware id.

app/models/work_completion/labware_completion.rb
```diff
@@ -0,0 +1,59 @@
...
+    @target_labware = labware
+    @submission_ids = submission_ids
+    @order_ids = []
+    @work_completion = work_completion
+  end
+
+  def process
+    connect_requests
+    fire_events
+  end
```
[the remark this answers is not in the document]
KatyTaylor: Yes good idea, will do

spec/factories/submission_factories.rb
KatyTaylor: Because this is where it gets the data from for `it 'has a bait_library'` test in library_complete_spec and pool_released_spec

spec/models/broadcast_event/pool_released_spec.rb
KatyTaylor: I didn't try let!, I considered it but I found this more explicit 🤷‍♀️ 
let! would probably work too

spec/models/state_changer/mx_tube_spec.rb
KatyTaylor: OK yes I forgot about this, I think the reason I changed it here was because I had some problem with the order of execution when nesting contexts within contexts.

PR #3619, opened 2022-05-31
initial adding of support scripts

### LINKED ISSUE #3601: DPL-363 - As User support (Neil) I would like to create a directory in Sequencescape and move all the current scripts there so that they are kept centrally and available to use (C=S, V=3)
**User story**
As User support (Neil) I would like to create a directory in Sequencescape and move all the current scripts there so that they are kept centrally and available to use

Neil / James

**Acceptance criteria**
- [ ] There will be a newe directory in Sequencescape to keep support scripts
- [ ] All of the current scripts will be available in the directory
- [ ] All developers will have access to those scripts and be able to use them

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

### LINKED ISSUE #3617: DPL-401 - As SSRs / Sequencing teams we would like to have improved NovaSeq pipelines inboxes to include the NovaSeq flowcell options as these are currently added in the comments inbox to provide clearer information to the Sequencing team and to reduce the probability of batching errors
**User story**
STORY: Request for new Improved NovaSeq Pipelines Inboxes within Sequencescape
 
This story is regarding the current page in sequencescape:  /pipelines/51
 
Why do we need improved NovaSeq Pipelines Inboxes within Sequencescape?

- Most of faculty sequencing requests are now moving over to NovaSeq (away from HiSeq V4, 4000 and X10)
- There are different NovaSeq flowcell options (SP, S4 and S2) now available to choose from as oppose to just one when the current NovaSeq inbox was designed
- As there is only one inbox available, the only current way to indicate to the sequencing team which flowcell to use, whether the XP is required, and if a custom read length is required is in the COMMENTS
- These comments are easy to miss, and in a busy inbox, the probability of batching errors is now increasing
- In summary having one NovasSeq inbox is now no longer appropriate

 
How could we improve this?

- We (as the SSR team and Sequencing team) would like 3 distinct NovaSeq inboxes to be available – one per flowcell as follows:

  - NovaSeq SP 6000 PE
  - NovaSeq S4 6000 PE
  - NovaSeq S2 6000 PE

- Within these inboxes we would like the following read lengths to be visible if possible:

  - 150bp
  - 100bp
  - OTHER
  - The idea of the OTHER option is to indicate clearly to the sequencing team that the SSR has selected a Custom Read length and therefore they NEED to look at the Comments

- It would also be nice (although not essential) to have a COMMENTS column visible in the inbox too so that the sequencing team can toggle between seeing all the comments and not seeing all the comments maybe (rather than clicking on each one as they have to now)

 
How could we achieve this?

- We would ask for new NovaSeq submission templates to be made as (we believe) the current inboxes are driven by the submission templates
- We would need one NovaSeq submission template for each flowcell type to be available, rather than just NovaSeq 6000 as it is now – so each of the current 6000 submission templates would need to be replaced by:
  - A NovaSeq SP 6000 PE option
  - A NovaSeq S4 6000 PE option
  - A NovaSeq S2 6000 PE option
- Within these templates we would need the accepted read lengths to be:
  - 150bp (as it is now)
  - 100bp (as it is now)
  - OTHER (not yet available)
- We would then still write the custom read length in the COMMENT column in the template

 
**Acceptance Criteria**

- [ ] That the current NovaSeq 6000 PE pipeline is replaced with 3 New Pipelines NovaSeq Sp, NovaSeq S4 and NovaSeq S2 making batching much clearer and easier for the sequencing team and therefore reducing batching errors.
- [ ] That it is very clear (within the inbox) to the sequencing team that a custom read length has been requested.

Siobhan

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

neilsycamore:
Closes #3601 3601

* added sample_move script
*
* ...


script/support/sample_move.rb
```diff
@@ -0,0 +1,296 @@
...
+  asset_group_sample_hash.each do |ag_name, samples|
+    # puts "ag => #{ag_name}"
+    asset_group = AssetGroup.find_by(name: ag_name)
+    order = Order.find_by(asset_group_id: asset_group.id)
+    aliquot_count = asset_group.assets.map(&:aliquots).flatten.size
+    if aliquot_count == samples.size
+      puts "#{order.study_id} #{order.submission_id} #{order.submission.orders.size} #{ag_name} - #{aliquot_count}/#{samples.size} :: All assets to be moved"
+
+      whole_asset_groups << asset_group
+      # if assetgroup asset == MultiplexedLibraryTube then only the aliquots can be updated
```
sdjmchattie, line 60: Does this comment relate to the `elsif` below?  It's either in the wrong position or incorrectly indented if so.

script/support/sample_move.rb
```diff
@@ -0,0 +1,296 @@
...
+    end
+  end
+  hash
+end
+
+# rubocop:disable Metrics/PerceivedComplexity
+# rubocop:disable Metrics/MethodLength
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/CyclomaticComplexity
+# rubocop:disable Layout/LineLength
```
sdjmchattie, line 46: If we aren't concerned about these things in scripts, would it be better to exclude files that we don't care about from Rubocop rather than disable the checkers on a file by file basis?

script/support/sample_move.rb
```diff
@@ -0,0 +1,296 @@
...
+  end
+end
+# rubocop:enable Rails/SkipsModelValidations
+# rubocop:enable Metrics/PerceivedComplexity
+# rubocop:enable Metrics/MethodLength
+# rubocop:enable Metrics/AbcSize
+# rubocop:enable Metrics/CyclomaticComplexity
+# rubocop:enable Layout/LineLength
+# rubocop:enable Metrics/ParameterLists
+# rubocop:enable Metrics/BlockLength
```
sdjmchattie, line 296: If these must stay in this file (see other comment above) I think I'd prefer to see some symmetry in when they're disabled and when they're re-enabled.  At the moment they're turned off sporadically in the code and then all turned on at the end.  It would be better, in my mind, to disable them as the first thing in this file if they will be turned on all together at the end of it.  It will make it easier to keep track of what's off and back on in the future.

sdjmchattie (changes requested):
I know this is still marked as draft, but it's in the Code Review column.

script/support/sample_move.rb
```diff
@@ -0,0 +1,296 @@
+# frozen_string_literal: true
+
+class StudySample < ApplicationRecord
+end
+
+# added sample class as comments are not commentable at present 21/3/2019
+class Sample
+  include Commentable
+end
```
JamesGlover, line 9: We should probably just make samples commentable

script/support/sample_move.rb
```diff
@@ -0,0 +1,296 @@
...
+
+class Asset
+  has_many :submitted_assets
+  has_many :orders, through: :submitted_assets, as: :asset
+end
```
JamesGlover, line 14: Asset shouldn't be used any more, it's only still around because I haven't had a change to kill it completely.

script/support/sample_move.rb
```diff
@@ -0,0 +1,296 @@
+# frozen_string_literal: true
+
+class StudySample < ApplicationRecord
+end
```
JamesGlover, line 4: We shouldn't need to reopen StudySample here

script/support/sample_move.rb
JamesGlover: I think  these should be todo's, not disables. Disable should be reserved only for false positives. I think we mostly want to fix these over time.

Also, the disabling should be scoped only to the bits of the script that is necessary, not the whole file.

script/support/sample_move.rb
```diff
@@ -0,0 +1,296 @@
...
+
+def update_fluidigm_plates(fluidigm_plates)
+  fluidigm_plates.each do |plate_id|
+    puts "Updating Fluidigm plate: #{plate_id}"
+    plate = Plate.find_by id: plate_id
+
+    # rubocop:disable Rails/SkipsModelValidations
+    # correct behaviour
+    plate.touch
+    plate.save!
```
JamesGlover, line 24: touch already triggers a save, so save! is not required

JamesGlover (commented):
I can see quite a lot of scope for improvements, but its probably risky doing them until we've got some test coverage. I guess we'll catch up soon and go over our next steps.

stevieing (approved):

PR #3656, opened 2022-08-02
Add caching

JamesGlover:
Going through some un-merged branches in my folders, most of this is housekeeping/refactoring that I realised was necessary when doing other work, but wanted to keep separate to keep PRs focused.

This was an attempt to speed up manifest processing by reducing N+1 query problems. I think I never submitted a PR at the time as I was hoping to make further improvements, as the last commit message attests to.

I did try and track down this cache object in another piece of work recently, so glad to see it make an appearance here.


stevieing:
Technical debt. Difficult to fix. No reason to do this.

PR #3657, opened 2022-08-02
Move legacy pulldown purposes to record loader

JamesGlover:
Found this WIP on my laptop as I cleared up before moving on. Mostly
about moving some old stuff off into record loader.


stevieing:
Not going to get done.

PR #3660, opened 2022-08-09
Dpl378 duplex tnano seq change sample progression

andrewsparkes:
Sequencescape changes for Story https://github.com/sanger/limber/issues/1077


app/controllers/api/v2/bait_libraries_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+module Api
+  module V2
+    # Provides a JSON API controller for BaitLibraryResource
+    # See: http://jsonapi-resources.com/ for JSONAPI::Resource documentation
+    class BaitLibrariesController < JSONAPI::ResourceController
+      # By default JSONAPI::ResourceController provides most the standard
+      # behaviour, and in many cases this file may be left empty.
+    end
+  end
+end
```
KatyTaylor, line 12: Is this needed, adding bait library to the API? Bait library is set as part of the request isn't it (the thing we were working on together)? But maybe you are also separately reading / creating bait libraries, from Limber, independently of a request?

app/resources/api/v2/bait_library_resource.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # Provides a JSON API representation of BaitLibrary
+    # See: http://jsonapi-resources.com/ for JSONAPI::Resource documentation
+    class BaitLibraryResource < BaseResource
+      # Constants...
+
+      # immutable # uncomment to make the resource immutable
```
KatyTaylor, line 10: Does it need to be mutable?

app/resources/api/v2/shared_behaviour/receptacle.rb
```diff
@@ -38,6 +38,8 @@ module Api::V2::SharedBehaviour::Receptacle
     # Attributes
     attribute :uuid, readonly: true
     attribute :name, delegate: :display_name, readonly: true
+
+    # TODO: these can be removed later, moved onto request_metadata
```
KatyTaylor, line 42: What needs to be done here? Data migration? Worth creating a story for it?

app/controllers/api/v2/bait_libraries_controller.rb
KatyTaylor: OK I see you're accessing it in the Limber PR,  `app/models/labware_creators/pcr_cycles_binned_plate/csv_file/row.rb`

Could be set to immutable though potentially?

app/controllers/api/v2/bait_libraries_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
```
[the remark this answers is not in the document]
andrewsparkes: What's this comment on about bot??

app/resources/api/v2/bait_library_resource.rb
andrewsparkes: What does mutable mean in terms of API v2? Nothing can be updated via v2 right?
I am just doing a lookup, so I guess I can set it.

app/controllers/api/v2/bait_libraries_controller.rb
andrewsparkes: Yeah I have to check the bait library name they put in the customer file matches a real bait library.
In terms of the controller it's the same as all the other v2 controllers.

app/models/attributable/attribute.rb
```diff
@@ -22,7 +22,7 @@ def initialize(owner, name, options = {})
     end
 
     def from(record)
-      record[name]
+      record[name] || record.send(name)
```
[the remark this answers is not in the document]
andrewsparkes: ok, added

app/resources/api/v2/shared_behaviour/receptacle.rb
andrewsparkes: There are 4 receptacle columns that could be removed.
The data I guess could be moved into the library requests, but the store and request metadata attributes are specific to the new DilutionAndCleanup request which will not exist for those old runs.
Will make a backlog story.

app/resources/api/v2/bait_library_resource.rb
KatyTaylor: Yes, some things can be updated via API v2

app/resources/api/v2/shared_behaviour/receptacle.rb
KatyTaylor: Hmm I'm not sure we should remove columns if there's nowhere to put the existing data?

KatyTaylor (approved):

andrewsparkes:
Closed as work deprecated.

Issue #3646, opened 2022-07-22
DPL-440 Cardinal sample import - single manifest


KatyTaylor:
**User story**
As a member of the Bioresource team, who receives samples from multiple collection centres per day, I would like to be able to import them all at once into Sequencescape, rather than having to create a manifest per collection centre. This will reduce the manual work I have to do.

**Who are the primary contacts for this story**
@KatyTaylor
Laura L (will do UAT)

**Acceptance criteria**
Decided by team consensus to go for the following solution, where we move how the collection centre information is stored, from the 'Supplier' to a new field against the sample.
- [x] There is a new (free text) field on the sample_metadata table, called 'collected_by' (to match the naming of the ENA field https://www.ebi.ac.uk/ena/browser/view/ERC000011) - this will be used to store the collection centre. (Q: can't see link, assume it is NULLable)
- [x] The new field should show on the sample pages in Sequencescape, like other sample metadata fields.
- [x] Above field is available in the Cardinal tube manifest, as a dropdown with limited values to select. See the table under 'Creating Suppliers' in the [documentation](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/Cardinal+Pipeline+Documentation) for the list of values that should go in the dropdown. (`Resulting 'name' field`)
- [x] In Limber, the pooling step in the Cardinal pipeline that uses the collection centre information should be updated to get that information from the new field instead (https://github.com/sanger/limber/blob/9c4a9d7fbf5740fe969db930084e7580f6b49620/app/models/labware_creators/cardinal_pools_plate.rb#L147)
- [x] Inform Laura and a SSR (?) that this is being worked on, and who best to contact for testing
- [x] Sequencescape tests
- [x] Create script to back-populated `collected_by` field on existing Cardinal samples, using the manifest -> supplier link. 
- [x] Add script to data-patch-archive GitLab repo
- [x] [Documentation](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/Cardinal+Pipeline+Documentation) 'Importing samples into Sequencescape' and 'Creating Suppliers' sections are updated to reflect new process. The 'supplier' associated with the manifest should now be filled out as 'Sanger' or similar.
- [x] Check Integration Suite passes
- [x] Deploy to UAT
- [ ] Ask Laura and Cardinal Team to test
- [ ] Deploy to Training
- [ ] Deploy to Prod
- [ ] Inform Laura, Stephen W, Vivek & Matiss once this has gone live.

[...]

PR #3678, opened 2022-10-06
DPL-440 Add `collected_by` field to cardinal tube manifest

harrietc52:
Closes https://github.com/sanger/sequencescape/issues/3646

There are 4 parts to review for DPL-440
1. This PR for Sequencescape
2. Limber PR: https://github.com/sanger/limber/pull/1175
3. `data-patch-archive` Rails console script [PR here](https://gitlab.internal.sanger.ac.uk/psd/data-patch-archive/-/merge_requests/8)
4. Documentation update [here](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/Cardinal+Pipeline+Documentation)

* Add collected_by field to sample metadata table
* Show collected by data in Sample summary view
* Add collected by field to Cardinal tube manifest
* Provide drop down list for collected by in manifest

Todo:
* Create script to back populate `collected_by` sample metatdata field, from associated sample manifest


config/sample_manifest_excel/ranges.yml
```diff
@@ -74,3 +74,11 @@ insdc_country:
...
   scope: :sorted_for_select
   scope_on: Insdc::Country
+collected_by:
+  options:
+    - UKB_Reading_11026
+    - UKB_Newcastle_11027
+    - UKB_Stockport_11025
+    - UKB_Bristol_11028
+    - ELGH_London_21032
+    - ELGH_Bradford_21034
```
KatyTaylor, line 84: Ah, I was hoping that it would be possible to provide this list specifically for the Cardinal manifest (because the values `UKB_Reading_11026`, etc. are only relevant for the Cardinal project, and other projects might want different values). Do you know if that is possible? I wonder whether the `scope` thing I can see on a few of the fields is relevant?

db/migrate/20221005133322_add_collected_by_to_sample_metadata.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Adds the column: collected_by, to the sample_metadata table
+# To allow for manifests uploading samples from multiple collection sites
+class AddCollectedByToSampleMetadata < ActiveRecord::Migration[6.0]
+  def change
+    add_column :sample_metadata, :collected_by, :string
```
KatyTaylor, line 7: I know most of our columns don't have descriptions, but it's something I think we could improve on. I wonder if we could add a description? The description of the equivalent field in EBI is 'name of persons or institute who collected the specimen', which seems alright?

db/schema.rb
```diff
@@ -1,4 +1,4 @@
-# This file is auto-generated from the current state of the database. Instead
+2# This file is auto-generated from the current state of the database. Instead
```
KatyTaylor, line 1:
```suggestion
# This file is auto-generated from the current state of the database. Instead
```

KatyTaylor (changes requested):
Just a couple of small comments. Think the thing in the schema is a typo and might break it?

harrietc52:
Also see Katy's comment here: 
https://github.com/sanger/sequencescape/issues/3646#issuecomment-1273105065

harrietc52:
When discussing multiple collected_by values, to provide a way of using a different range in a different manifest.

Could you have two new 'columns' in the columns.yml, called for instance collected_by_cardinal and collected_by_default , and make them refer to the same db field, but different ranges? (Katy)

KatyTaylor (approved):

PR #3680, opened 2022-10-10
Not remove aliquots if any batch released already

emrojo:
Closes #

*
*
* ...

### DIFF
--- a/app/models/receptacle.rb
+++ b/app/models/receptacle.rb
@@ -367,4 +367,11 @@ class Receptacle < Asset
     save!
     events.create_external_release!(!external_release_nil_before) unless external_release.nil?
   end
+
+  def allow_to_remove_downstream_aliquots?
+    # DPL-451:
+    # If the labware has descendants that are involved in a sequencing batch that has
+    # been released, do not remove any data from downstream aliquots, or from the MLWH.
+    downstream_assets.map(&:creation_batches).flatten.all? { |batch| !batch.released? }
+  end
 end
--- a/app/models/transfer_request.rb
+++ b/app/models/transfer_request.rb
@@ -284,7 +284,8 @@ class TransferRequest < ApplicationRecord # rubocop:todo Metrics/ClassLength
   end
 
   def on_failed
-    target_asset&.remove_downstream_aliquots
+    target_asset&.remove_downstream_aliquots if target_asset&.allow_to_remove_downstream_aliquots?
   end
   alias on_cancelled on_failed
+
 end


app/models/receptacle/downstream_aliquots_removal.rb
```diff
@@ -0,0 +1,44 @@
...
+    # 1) Get the submission from the outer requests that wrap this current node
+    # 2) From the first submission, get the multiplexed labware, which is the labware
+    #    that is the target of a multiplexed request (outer request before sequencing)
+    # 3) From this multiplexed labware, it goes to the next labware created from it (sequencing tube).
+    # 4) For this labware, it access to the creation batches (groups of sequencing requests that will go into
+    #    a flowcell)
+    #
+    # @return [Array[Batch]] List of batches
+    def self.creation_batches_for_requests(instance)
+      PrivateMethods.submissions_for_requests(instance)&.first&.multiplexed_labware&.children&.map(&:creation_batches)
```
KatyTaylor, line 41: `PrivateMethods.submissions_for_requests(instance)&.first&.multiplexed_labware`
... will this work if the submission we're looking at is a library-only submission, and the sequencing request is made separately, later (e.g. using the 'request additional sequencing' button)?

spec/models/receptacle/downstream_aliquots_removal_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+RSpec.describe 'Receptacle::DownstreamAliquotsRemovalSpec::Mixin' do
+  describe '#allow_to_remove_downstream_aliquots?' do
+    # Labware
+    let(:original_plate) { create(:plate_with_untagged_wells, well_count: 1) }
+    let(:original_well) { original_plate.wells.first }
+    let(:plates) { create_list(:plate, 3, well_count: 1) }
+    let(:wells) { plates.map(&:wells).flatten }
+    let(:assets) { [original_well, wells].flatten }
```
KatyTaylor, line 9: Can we try not to add more use of the word 'asset'? In this case, call it 'wells' or 'receptacles' instead? Because 'asset' is an old term, which was replaced by both labware and receptacle, so wherever it's still used I find it ambiguous.

app/models/receptacle/downstream_aliquots_removal.rb
emrojo: Yes, I suppose not. I'll modify it so it check all submissions.

spec/models/receptacle/downstream_aliquots_removal_spec.rb
emrojo: Yes, it is not a good name. Also just realised is not in use anywhere :( I'll remove it

Sangeetha-Bheeman (approved):
Reviewed on a higher level tried to understand from the comments. Approving since it was already reviewed.

PR #3685, opened 2022-10-14
Dpl 364 rake task changes

Sangeetha-Bheeman:
Closes #

*
*
* ...


.release-version
```diff
@@ -1 +1 @@
-14.15.0
\ No newline at end of file
+14.15.2
```
sdjmchattie, line 1: Can we add the blank line to the end of this file?  It is the POSIX standard that a line without a newline character at the end isn't a line of the file.

It probably doesn't matter in this file, but it's bad practice to leave them out which is why GitHub identifies them in pull requests.

sdjmchattie (approved):
A minor niggle, but otherwise all good

.release-version
Sangeetha-Bheeman: Sure, Thank you. I will add it.

PR #3698, opened 2022-11-02
Dpl497 Adding retention instructions to plate manifests

### LINKED ISSUE #3673: DPL-497 Record retention instructions against stock plates
**User story**
As a Sample Management user, I would like the sample retention instructions to be recorded against stock plates when they are imported into Sequencescape, so that I know what to do with them later on.

Danni W, Siobhan A-G, Katy

Danni W, Siobhan A-G

**Acceptance criteria**
- [x] All plate manifests have a new mandatory field, 'Retention instructions', with the following options:
    - Destroy after 2 years
    - Return to customer after 2 years
    - Long term storage
- [x] Above field is mandatory - data must be provided
- [x] This information is saved on manifest upload, at the labware level (custom metadata?)

**Additional context**
This does not need to include a way to view this data once uploaded - that will be in another story (https://github.com/sanger/sequencescape/issues/3676)

N.B. If the 'Retention instructions' are stored in the custom metadata table, the functionality for https://github.com/sanger/sequencescape/issues/3676 could be achieved using Limber, as functionality has been recently developed to allow viewing and editing of custom metadata. You would need to add configuration to Limber to show the new field against the relevant plate purposes. Custom metadata can also already be viewed (but not edited) on the labware page in Sequencescape.

andrewsparkes:
Closes #3673


config/sample_manifest_excel/ranges.yml
```diff
@@ -85,3 +85,8 @@ collected_by_for_cardinal:
 collected_by_for_controls:
   options:
     - Sanger Institute
+retention_instruction_options:
+  options:
+    - Destroy after 2 years
+    - Return to customer after 2 years
+    - Long term storage
```
KatyTaylor, line 92: Doesn't really matter, but it looks like all the other fields (apart from Harry's recent one, because that's got 2 ranges for 1 field), follow a naming convention where the yml key is just the field name? i.e. `retention_instruction:`

config/sample_manifest_excel/columns.yml
```diff
@@ -895,6 +895,27 @@ collected_by_for_controls:
...
+retention_instruction:
+  heading: RETENTION INSTRUCTION
+  updates: retention_instruction
+  unlocked: true
+  validation:
+    options:
+      type: :list
+      formula1: "$A$1:$A$2"
+      allowBlank: false
+      showDropDown: false
```
KatyTaylor, line 907: Did you figure out what `showDropDown` does? Just curious as it seems intuitively like it should be `true` for this field.

app/sequencescape_excel/sequencescape_excel/specialised_field/retention_instruction.rb
```diff
@@ -0,0 +1,76 @@
...
+
+        current_metadata = current_collection.metadata.symbolize_keys
+
+        # ok if we have other metadata with different keys
+        return unless current_metadata.key?(:retention_instruction)
+
+        # ok if the already stored metadata value matches (set by another well already)
+        return if current_metadata[:retention_instruction] == value
+
+        errors.add(:base, "the retention instruction value #{value} must match for all wells on the same plate.")
```
KatyTaylor, line 43: Wondering if there's a scenario where this error message won't make sense. What if a manifest has been uploaded previously, and they are uploading it again in order to change the values in the fields (pretty sure this happens sometimes). Then the new manifest could have all wells matching on the plate, but different to the previously saved value.

app/sequencescape_excel/sequencescape_excel/specialised_field/retention_instruction.rb
```diff
@@ -0,0 +1,76 @@
...
+              retention_instruction: value
+            }
+          )
+        current_plate.custom_metadatum_collection = cm
+        cm.save!
+      end
+
+      def check_and_update_existing_custom_metadatum_collection
+        current_collection = asset.plate.custom_metadatum_collection
+        current_metadata = current_collection.metadata.symbolize_keys
```
KatyTaylor, line 64: Code is nice and clear, could be a little less verbose? You're accessing the `asset.plate`, and the `asset.plate.custom_metadatum_collection`, and the `asset.plate.custom_metadatum_collection.keys` in multiple places in the code - perhaps you could make it a bit more concise using separate methods for some or all of them, using the `||=` thing I've forgotten the name of. Up to you.

KatyTaylor (changes requested):
Left a couple of style suggestions. Main thing I'm wondering though is does it behave correctly in the scenario where they want to re-upload a manifest with a different value?

config/sample_manifest_excel/columns.yml
yoldas: Can we also remove "for the plate " part of the prompt so the column definition can be reused for tubes (e.g. new tube_extraction manifest type) without modification?

prompt: "Please select a retention instruction for the plate from the provided list."

app/sequencescape_excel/sequencescape_excel/specialised_field/retention_instruction.rb
andrewsparkes: Refactored to validate at plate level before row updates

app/sequencescape_excel/sequencescape_excel/specialised_field/retention_instruction.rb
andrewsparkes: Refactored to validate at plate level before row updates

config/sample_manifest_excel/columns.yml
andrewsparkes: showDropDown - Axlsx gem - "A boolean value indicating whether to display a dropdown combo box for a list type data validation." Seems to work fine set to false, I click and can see a dropdown and can select an option.

plate - the whole set of code I've added is plate-specific. The requirement is it only applies only to plate manifests. Maybe we can change that later if they wish to make it useable for tubes. (The vast majority of samples come in on plates, but not all).

KatyTaylor (approved):
Looks good to me!

PR #3725, opened 2022-12-08
Develop into master

yoldas:
Closes #

* Bump release version to 14.16.1
* DPL-555 ss label and display update for hmdmc number field on the study creation screen
    * Changes HMDMC approval number text to HuMFre approval number
    * Displays HuMFre approval number in new study, edit, manage, and properties views, for any data release strategy.


config/locales/metadata/en.yml
```diff
@@ -496,7 +496,7 @@ en:
           ebi_db: ENA-STUDY
 
         hmdmc_approval_number:
```
andrewsparkes, line 498: shouldn't this be renamed if it's now labelled HuMFre? hmdmc_approval_number

app/views/shared/_ethical_approval_upload.html.erb
```diff
@@ -6,11 +6,13 @@
     <td width="60%"><%= form.check_box :ethically_approved, disabled: !can?(:change_ethically_approved, @study) %></td>
   </tr>
 </table>
+<!--
```
andrewsparkes, line 9: If this section is commented out why not delete it?

config/locales/metadata/en.yml
yoldas: No, the same field name is used for database columns and form fields. It is the same approval number with some change in the meaning. We just need to change the label.

.release-version
```diff
@@ -1 +1 @@
-14.16.0
+14.16.1
```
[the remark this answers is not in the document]
yoldas: I have updated the MINOR part instead of PATCH part of the version string as suggested.

app/views/shared/_ethical_approval_upload.html.erb
yoldas: I have deleted the commented code now.

PR #3731, opened 2023-01-18
Triggers broadcast to warehouse on well attribute update

Sangeetha-Bheeman:
Closes #

*
*
* ...


app/models/well_attribute.rb
```diff
@@ -74,4 +76,13 @@ def current_volume=(current_volume)
...
     super
   end
+
+  def broadcast_warehouse_message
+    message = Messenger.find_by(target_id: well_id)
+    if message.present?
+      message.resend
+    else
+      Messenger.new(target: self, template: 'WellStockResourceIO', root: 'stock_resource').broadcast
+    end
```
emrojo, line 86: This is changing current behaviour. I think is not right because is treating every well as if it was a stock well, but we may be updating other wells that are not stock and end up sending messages from them as stock wells.

emrojo (changes requested):
I think I would remove the lines where we create a new messenger and try how is the performance; it may work but we might find also that this enters in a recursive loop of well and well attribute firing the update event to each other.

emrojo (approved):

Issue #3603, opened 2022-05-11
DPL-369 update data text, contact and information in 'Reason for delaying release' section of the manage study page


neilsycamore:
**User story**
As Head of Research Governance (Carol Smee) I  would like the 'Reason for delaying release' section of the Study Create, Edit, and Manage pages changed.
The bullet points should be removed and the only text should be
**"To apply for a delay, please contact datasharing@sanger.ac.uk"**
example is  https://sequencescape.psd.sanger.ac.uk/admin/studies/6868

sequencescape/config/locales/metadata/en.yml
the help text for the data_release_delay_reason requires amending.
The data sharing contact also needs updating.

~~Also where HMDMC is mentioned, please change to 'HMDMC/HuMFre' in all instances.~~ (This requirement was managed as part of https://github.com/sanger/sequencescape/issues/3697)

For ref. RT is [749904](https://rt.sanger.ac.uk/Ticket/Display.html?id=749904)

Customer comment: "... the information is very out-of-date by several years!"

neilsycamore:
DPL-481 closed, superseded by DPL-555
Acceptance criteria "Also where HMDMC is mentioned, please change to 'HMDMC/HuMFre' in all instances." superceeded by DPL-555.
Bullet point amendments still required

SujitDey2022:
Need to update the story with the requirements completed.

neilsycamore:
I have today contacted Carol to confirm the remaining work is required

neilsycamore:
Can this go ahead as planned, please?

I confirm that the wording in the bullet points is very out-of-date and no longer applicable. Please replace with "To apply for a delay, please contact [datasharing@sanger.ac.uk](mailto:datasharing@sanger.ac.uk)'.

Thank you.

Kind regards,

Carol
Head of Research Governance

SujitDey2022:
Hello @neilsycamore, I could not understand how to add my query to the RT Ticket and so raising it here. Could we confirm if we need to reflect this change across Study Create, Edit, and Manage pages as they are related? 

![image](https://user-images.githubusercontent.com/115623404/222761762-13a0d566-3d4c-4230-87bd-53e92d3aa8fc.png)

yoldas:
We have also changed the data sharing contact name and email for sequencescape production ( dpl-369-data-sharing-config in deployment repo).

yoldas:
Attached screenshot.

[...]

PR #3765, opened 2023-03-14
3603-dpl-369-update-data-text-contact-and-information-in-reason-for-delaying-release-section-of-the-manage-study-page

yoldas:
Closes #

* Replace the help text for reason for delaying data release
* Show help text for reason for delaying data release only when "other" option is selected

See the UI at https://github.com/sanger/sequencescape/issues/3603#issuecomment-1487792973


emrojo (approved):

config/locales/metadata/en.yml
```diff
@@ -436,7 +436,7 @@ en:
 
         data_release_delay_reason:
           label: Reason for delaying release
-          help: "To apply for a delay, please send the following information to <%= configatron.data_sharing_contact.name %> (<%= configatron.data_sharing_contact.email %>):<ul><li>SAC sponsor</li><li>Study title and, where available, data set(s) ID</li><li>Study description (should describe the data types that will be produced)</li><li>Data sharing plan (to include the following):<ul><li>- Which repository/repositories will be used</li><li>- If data will be made available under a managed access mechanism</li><li>- When data will be shared</li></ul></li><li>Reason(s) for delaying data release</li></ul>"
+          help: "To apply for a delay, please contact <%= configatron.data_sharing_contact.email %>"
```
sdjmchattie, line 439: Might it be worth wrapping the email address in an `<a href="mailto:address@here.com>` so that it becomes clickable to start the email?

yoldas: Thanks. Attached screenshot:  https://github.com/sanger/sequencescape/issues/3603#issuecomment-1468418670

config/config.rb
```diff
@@ -124,8 +124,8 @@
   configatron.beckman_precision = 2
   configatron.hamilton_precision = 2
 
-  configatron.data_sharing_contact.name = 'John Doe'
-  configatron.data_sharing_contact.email = 'foo'
+  configatron.data_sharing_contact.name = 'Datasharing'
```
sdjmchattie, line 127: It looks like maybe this name isn't used any more.  Otherwise all looks good.

sdjmchattie: I meant that I don't think the name is used in the code.  I haven't checked all of the code, but you removed a reference to the name in that `data_release_delay_reason:help:` text and that might have been the only reference.

sdjmchattie: I take it that this is still referenced in the code somewhere?  I don't think it is, but I haven't gone through the code to find out.

config/locales/metadata/en.yml
```diff
@@ -436,7 +436,7 @@ en:
 
         data_release_delay_reason:
           label: Reason for delaying release
-          help: "To apply for a delay, please send the following information to <%= configatron.data_sharing_contact.name %> (<%= configatron.data_sharing_contact.email %>):<ul><li>SAC sponsor</li><li>Study title and, where available, data set(s) ID</li><li>Study description (should describe the data types that will be produced)</li><li>Data sharing plan (to include the following):<ul><li>- Which repository/repositories will be used</li><li>- If data will be made available under a managed access mechanism</li><li>- When data will be shared</li></ul></li><li>Reason(s) for delaying data release</li></ul>"
+          help: 'To apply for a delay which is not associated with a PhD study, please contact <a href="mailto:<%= configatron.data_sharing_contact.email %>"><%= configatron.data_sharing_contact.email %></a>'
```
sdjmchattie, line 439: This is where i'm seeing you dropped the reference to `configatron.data_sharing_contact.name` entirely and replaced it with two instances of `configatron.data_sharing_contact.email` so I don't think the name is used unless it's somewhere else.

config/config.rb
yoldas: It is still in the configuration. configatron.data_sharing_contact.name locale string is not referenced in the code. 

They did not included the name in the help text in their request for this story; they asked only for the email to be included. I would like to keep the name in the configuration:

Previously, the contact information was for an actual person. The name of the person and the email were in the help text. It was changed to a new service contact information name: Datasharing, email: datasharing@sanger.ac.uk . Only the email is included in the help text. Although we do not use configatron.data_sharing_contact.name in the code, I think, it is still part of the configuration. 

The word "Datasharing" has a mention in a few places, but they are hardcoded; not loaded from the locale string and I think fixing them is not part of this story. Searching the word in the project will reveal several instances mainly "Contact Datasharing".

Another point to mention that the name and email are configured using the deployment project. If we have to remove the name, we need to remove it from both Sequencescape and Deployment. If we need them again, or we want to improve the code, we will need to change two projects.

I will now attempt using configatron.data_sharing_contact.name whenever Datasharing appears in other places, possibly by creating new locale strings and doing interpolation not to drop the item.

PR #3777, opened 2023-04-04
DPL-717: Add new default records to support deep well stamping

sdjmchattie:


andrewsparkes (approved):
Approved pending confirmation of plate purpose name.

PR #3794, opened 2023-04-28
Changes for Novaseqx

Sangeetha-Bheeman:
Closes #

*
*
* ...


lib/record_loader/submission_template_loader.rb
```diff
@@ -26,13 +26,22 @@ def generate_derived_options(related_records)
 
     def submission_parameters(related_records)
       params = {}
-      params[:request_type_ids_list] =
-        RequestType.where(key: related_records['request_type_keys']).ids if related_records['request_type_keys']
+      params[:request_type_ids_list] = sort_request_type_ids(related_records['request_type_keys']) if related_records[
+        'request_type_keys'
+      ]
+
+      # RequestType.where(key: related_records['request_type_keys']).ids if related_records['request_type_keys']
```
sdjmchattie, line 33: Probably need to remove this commented out code

sdjmchattie (changes requested):
One question I have is whether we should keep appending new records to the bottom of the `default_records.yml` files?  Would it be better to create new indexed files like you have done with `config/default_records/request_types/011_novaseqx_sequencing.yml`? 
 That way you can put a comment at the top of the file explaining what the records are for and separate out the things being created so that it's more obvious that they're all related.

emrojo (approved):

PR #3798, opened 2023-05-03
Load transfer templates from config

yoldas:
Closes #

*
*
* ...


config/default_records/transfer_templates/bioscan_transfer_templates.yml
```diff
@@ -0,0 +1,84 @@
+# These transfer templates are associated with the Bioscan pipelines.
```
andrewsparkes, line 1: It's only the '384 plate to tube' one that is associated with Bioscan, should probably leave all the others in seeds as we don't know what they are used for?

spec/data/record_loader/transfer_templates/bioscan_transfer_templates.yml
```diff
@@ -0,0 +1,84 @@
+# These transfer templates are associated with the Bioscan pipelines.
```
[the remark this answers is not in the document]
andrewsparkes: but this test could use something generic, and not name the file with bioscan_

config/default_records/transfer_templates/bioscan_transfer_templates.yml
yoldas: I have removed the others from the config.

yoldas: We have now only "384 plate to tube" in the bioscan_transfer_templates.yml file .

andrewsparkes (approved):

PR #3805, opened 2023-05-15
Add rake task to generate tag groups, templates and mbrave config for

emrojo:
limber

Closes #

*
*
* ...


lib/tasks/create_mbrave_tags.rake
```diff
@@ -0,0 +1,215 @@
...
+
+namespace :mbrave do
+  desc 'Modifying tag groups'
+
+  #
+  # How to use it:
+  # bundle exec rake mbrave:create_tag_plates <tag_layout_template> <login> <num_plates>
+  #
+  # Example:
+  # bundle exec rake mbrave:create_tag_plates Bioscan_384_template_1_v3 admin 5
```
andrewsparkes, line 13: I think the more likely use case for this is 'As the Lab we want a set of v3 plates for the month'. So 1 instance of each tag plate created from all 24 layout templates for v3, so 24 barcodes total. It would be risky for them to have several tag plates e.g. of layout number 4 of v3 around. And they physically can only make one tag plate at at a time from the masters. Ask Scott maybe?

lib/tasks/create_mbrave_tags.rake
```diff
@@ -0,0 +1,215 @@
...
+        qcable.update!(state: 'available')
+        puts " - #{qcable.asset.machine_barcode}"
+      end
+    end
+    exit
+  end
+
+  #
+  # How to use it:
+  # bundle exec rake mbrave:create_tag_groups <forward_file> <reverse> <version>
```
andrewsparkes, line 50: forward_filename and reverse_filename

lib/tasks/create_mbrave_tags.rake
```diff
@@ -0,0 +1,215 @@
...
+  #
+  # Example:
+  # bundle exec rake mbrave:create_tag_groups ./forward.csv ./reverse.csv v3
+  task :create_tag_groups, %i[arg1 arg2] => :environment do |_t, _args|
+    # rubocop:todo Lint/ConstantDefinitionInBlock
+    # rubocop:todo Metrics/AbcSize
+    # rubocop:todo Metrics/MethodLength
+    # Class to support creation of tag groups, tag layou templates and generation of the
```
andrewsparkes, line 58: typo layou to layout

lib/tasks/create_mbrave_tags.rake
```diff
@@ -0,0 +1,215 @@
...
+            mbrave_tags = []
+          end
+        end
+      end
+
+      def create_tag_layout_templates
+        puts 'Creating tag layout templates:'
+        @reverse_groups.each_with_index do |reverse_group, index|
+          puts " - #{tag_layout_template_name(index)}"
+          TagLayoutTemplate.create(
```
andrewsparkes, line 132: What happens if the tag group or tag plate layout has already been generated before for this version number? e.g. if there's a mistake and we try to create a v3 set when a v3 set already exists.

lib/tasks/create_mbrave_tags.rake
```diff
@@ -0,0 +1,215 @@
...
+      def _add_to_yaml(yaml_filename, tag_group_name, mbrave_tags, version)
+        {}.tap do |obj|
+          record = {}
+          record['name'] = tag_group_name
+          record['version'] = version
+          record['tags'] = mbrave_tags
+
+          contents = YAML.safe_load(File.read(yaml_filename))
+          contents[tag_group_name] = record
+          #contents.push(obj)
```
andrewsparkes: unneeded comment line?

lib/tasks/create_mbrave_tags.rake
```diff
@@ -0,0 +1,215 @@
+# frozen_string_literal: true
+require 'csv'
+require 'yaml'
+
+namespace :mbrave do
+  desc 'Modifying tag groups'
```
andrewsparkes, line 6: Creating tag plates?

lib/tasks/create_mbrave_tags.rake
```diff
@@ -0,0 +1,205 @@
...
+
+  #
+  # How to use it:
+  # bundle exec rake mbrave:create_tag_plates <tag_layout_template> <login> <num_plates>
+  #
+  # Example:
+  # bundle exec rake mbrave:create_tag_plates Bioscan_384_template_1_v3 admin 5
+  task :create_tag_plates, %i[arg1 arg2] => :environment do |_t, _args|
```
sdjmchattie, line 14: I'm not sure I follow the syntax here as I've not seen it before, but I think this is declaring that you should pass in two arguments, but then they're not used in this task because you've prefixed `_t` and `_args` with underscores.  Are they used by the `:environment` task?  It would probably be good to name these arguments if they're needed, even if it's not here.

Or is this just a way to ensure the user passes in arguments?  You seem to check for `ARGV` in your task which I think is a (wrong?) way to access the arguments for the task.  Perhaps there is something that can be done here to make things clearer.  Like access the args via the `_args` parameter and naming them appropriately.

sdjmchattie (approved):

lib/tasks/create_mbrave_tags.rake
emrojo: Added better description

lib/tasks/create_mbrave_tags.rake
emrojo: Fixed

lib/tasks/create_mbrave_tags.rake
emrojo: It will break on tag group creation.

yoldas (approved):

PR #3808, opened 2023-05-18
3747-dpl-659-bioscan-add-paging-to-samples-list-in-labware-summary-screen-to-cope-with-9216-samples

yoldas:
Closes #

* Pagination of samples in labware, receptacles and studies

### DIFF
--- a/app/controllers/labware_controller.rb
+++ b/app/controllers/labware_controller.rb
@@ -31,10 +31,8 @@ class LabwareController < ApplicationController # rubocop:todo Metrics/ClassLeng
   def show
     @source_plates = @asset.source_plates
     respond_to do |format|
-      format.html do
-        @aliquots = @asset.aliquots.include_summary.paginate(page: params[:page], per_page: 384)
-      end
-      format.xml { render xml: @asset.to_xml}
+      format.html { @aliquots = @asset.aliquots.include_summary.paginate(page: params[:page], per_page: 384) }
+      format.xml { render xml: @asset.to_xml }
       format.json { render json: @asset.to_json }
     end
   end
--- a/app/controllers/labware_controller.rb
+++ b/app/controllers/labware_controller.rb
@@ -2,7 +2,7 @@
 
 # Handles viewing {Labware} information
 # @see Labware
-class LabwareController < ApplicationController # rubocop:todo Metrics/ClassLength
+class LabwareController < ApplicationController
   before_action :discover_asset, only: %i[show edit update summary print_assets print history]
 
   def index # rubocop:todo Metrics/AbcSize, Metrics/MethodLength
--- a/app/controllers/labware_controller.rb
+++ b/app/controllers/labware_controller.rb
@@ -30,11 +30,12 @@ class LabwareController < ApplicationController # rubocop:todo Metrics/ClassLeng
 
   def show
     @source_plates = @asset.source_plates
-    @aliquots = @asset.aliquots.include_summary.paginate(page: params[:page], per_page: 384)
     respond_to do |format|
-      format.html
-      format.xml
-      format.json { render json: @asset }
+      format.html do
+        @aliquots = @asset.aliquots.include_summary.paginate(page: params[:page], per_page: 384)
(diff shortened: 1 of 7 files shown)


app/controllers/labware_controller.rb
```diff
@@ -31,7 +31,7 @@ def index # rubocop:todo Metrics/AbcSize, Metrics/MethodLength
   def show
     @source_plates = @asset.source_plates
     respond_to do |format|
-      format.html
+      format.html { @aliquots = @asset.aliquots.include_summary.paginate(page: params[:page], per_page: 384) }
```
harrietc52, line 34: Maybe pull 384 out to a constant with a name suggesting why 384, or this whole block could be a in a private method possibly (same for the other controllers below)

app/views/assets/samples_partials/_asset_samples.erb
```diff
@@ -8,10 +8,11 @@
     </h3>
     </a>
   </header>
+  <div id="asset-samples-table-collapser" class="collapse in show">
```
harrietc52, line 11: I think the `card-body` div is not indented correctly?

app/controllers/receptacles_controller.rb
```diff
@@ -30,7 +30,14 @@ def index # rubocop:todo Metrics/AbcSize
   def show
     @source_plates = @asset.source_plates
     respond_to do |format|
-      format.html
+      format.html do
+        @aliquots =
+          if @asset.is_a?(AliquotIndexer::Indexable)
```
harrietc52, line 35: What does this `AliquotIndexer` do? How come that means it cannot be paginated? 

It looks like `app/views/receptacles/_samples.erb` handles the check `<% indexed = @asset.is_a?(AliquotIndexer::Indexable) %>`  within the erb. Not sure if it is worth having the same approach for both, if possible?

app/views/receptacles/_samples.erb
```diff
@@ -8,11 +8,14 @@
...
   </header>
-  <div class="card-body">
-    If the tags below are not correct, you may wish to <%= link_to 'perform tag substitution', new_receptacle_tag_substitution_path(@asset) %>.
-  </div>
   <div id="asset-samples-table-collapser" class="collapse in show">
+    <div class="card-body">
+      If the tags below are not correct, you may wish to <%= link_to 'perform tag substitution', new_receptacle_tag_substitution_path(@asset) %>.
+    </div>
     <% indexed = @asset.is_a?(AliquotIndexer::Indexable) %>
+    <% unless indexed %>
```
harrietc52, line 16: See other comment about the various ways of going this conditional 🙂

app/views/studies/samples/index.html.erb
```diff
@@ -7,7 +7,15 @@
...
 </div>
 
+
+
 <% if @samples.size > 0 %>
+  <header>
+    <h3>Samples
+      <%= counter_badge @study.samples.count %>
+    </h3>
+  </header>
```
harrietc52, line 17: Might the other paged benefit from a `count` badge too?

harrietc52 (approved):
This looks fine, just a few minor optional suggestions

app/views/studies/samples/index.html.erb
yoldas: The other ones have already the count badge. I have added here as it shows the total number of records.

app/controllers/receptacles_controller.rb
yoldas: I should have asked why it might be important to them.

app/views/assets/samples_partials/_asset_samples.erb
yoldas: I moved the collapser div up to include the comment about the tags in the collapsed section as well. Otherwise, you collapse the section but the comment related to the section stays visible.

"If the tags below are not correct, you may wish to <%= link_to 'perform tag substitution', new_receptacle_tag_substitution_path(@asset.receptacle) %>."

PR #3823, opened 2023-06-01
Bioscan epic merged develop

emrojo:
Closes #

*
*
* ...


yoldas (approved):
40 changes shown here look OK. I have made other checks for this PR:

- 17 files would be deleted if we merged develop back into SS bioscan-epic; This PR solves that issue.
- 6 merge conflicts are done correctly.
- Modifications to 16 files would be lost; This PR solves that issue.
- There will be a need to run migrations.
- Executing integration suite with number_of_lilys_plates, number_of_lysate_plates, number_of_wells_with_samples = 5, 2, 96 works.
- Double checked a few changes to see them in the branch because they caused an issue in another merge such as 

> transfers column of transfer_templates is varchar 10240 and there is a migration

PR #3842, opened 2023-06-21
Create single barcode for tag plates created in rake task

yoldas:
Closes #

*
*
* ...

### DIFF
--- a/app/models/qcable_creator.rb
+++ b/app/models/qcable_creator.rb
@@ -9,7 +9,7 @@ class QcableCreator < ApplicationRecord # rubocop:todo Style/Documentation
   validates :user, presence: true
   validates :lot, presence: true
 
-  attr_accessor :count, :barcodes
+  attr_accessor :count, :barcodes, :use_supplied_barcode
 
   after_create :make_qcables!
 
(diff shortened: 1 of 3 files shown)


andrewsparkes (approved):

app/models/qcable_creator.rb
```diff
@@ -9,22 +9,26 @@ class QcableCreator < ApplicationRecord # rubocop:todo Style/Documentation
...
 
-  attr_accessor :count, :barcodes
+  attr_accessor :count, :barcodes, :supplied_barcode
 
   after_create :make_qcables!
 
   def make_qcables!
     qcables_by_count! if count.present?
     qcables_by_barcode! if barcodes.present?
+    qcables_by_supplied_barcode! if supplied_barcode.present?
```
emrojo, line 19: should we add this in the first line as:
```suggestion
    return qcables_by_supplied_barcode! if supplied_barcode.present?
```
I'm thinking we don't want to merge all the behaviour, but to add an optional branch for it

emrojo: Not sure about the return but what I mean is to not run the other lines when supplied_barcode is present

emrojo (approved):

PR #3861, opened 2023-07-17
GPL 963 lcmb pipeline

andrewsparkes:
Changes for LCMB pipeline


lib/tasks/limber.rake
```diff
@@ -223,10 +223,6 @@ namespace :limber do
         )
       end
 
-      lcmb_catalogue =
```
harrietc52, line 226: Does this need to be replaced with anything else ?

harrietc52 (approved):

lib/tasks/limber.rake
andrewsparkes: I think it's all handled by the record loader additions above

PR #3894, opened 2023-08-31
wip: scRNA Core Cell Extraction setup

StephenHulme:
Closes #

- [x] Cardinal banking pipeline is duplicated, and given a new name "scRNA core cell extraction"
- [ ] Integration Suite test for the above is also duplicated
    - see https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/84
- [ ] New pipeline uses new plate and tube purposes (similar to Cardinal banking pipeline, but "LCA" prefix replaced with "LRC", for "Limber RNA Core")
    - see https://github.com/sanger/limber/pull/1363
- [x] Submission + associated things are duplicated, and given appropriate names
    - [x] submission template
    - [x] request type
    - [x] (library type should not be needed, as this is not a library prep pipeline)


config/default_records/product_catalogues/009_scrna_core_cell_extraction_product_catalogues.yml
```diff
@@ -0,0 +1,6 @@
+# This product catalogue is associated with the Limber Cardinal pipeline.
```
andrewsparkes, line 1: scRNA Core Cell Extraction not cardinal

config/default_records/submission_templates/009_scrn_core_cell_extraction_submission_templates.yml
```diff
@@ -0,0 +1,12 @@
+# These submission templates are associated with the Limber scRNA Core Cell Extraction pipeline.
+# They are in a separate file to allow us to soft-release them before
+# the real release.
+---
+Limber - scRNA Core Cell Extraction:
+  submission_class_name: "LinearSubmission"
+  related_records:
+    request_type_keys: ["limber_scrna_core_cell_extraction"]
+    project_name: "Project scRNA Core"
```
andrewsparkes, line 9: NB. projects and studies with these names must exist in the various environments. Usually it is the job of the SSRs to create them. Do these exist in Production (and in UAT - if not we can create those ones manually via UAT actions)?

andrewsparkes (approved):
Just change that comment from Cardinal and I think this is good to go.

config/default_records/request_types/012_limber_scrna_core_cell_extraction.yml
```diff
@@ -0,0 +1,14 @@
...
+# the real release.
+---
+limber_scrna_core_cell_extraction:
+  name: scRNA Core Cell Extraction
+  asset_type: SampleTube
+  order: 1
+  request_class_name: CustomerRequest
+  for_multiplexing: false
+  billable: true
+  product_line_name: scRNA Core Cell Extraction
```
KatyTaylor, line 12: I don't think we should be creating a specific Product Line for this pipeline. From the product_line.rb model:
```
# ProductLine represents the team doing the work. Used primarily to group together
# {SubmissionTemplate submission templates} for display, but is also used in downstream
# reporting.
#
# @note This value is becoming increasingly unreliable as the teams become more adaptable
#       with sharing workload. It is probably not a good idea to base critical behaviour off this.
```

I think to add to the note above, it's also becoming increasingly unreliable as we haven't known what it's for in the past so have just been making new ones for new pipelines 😞 

I will post in Slack asking what the correct one to use is - think from looking at the list it might be 'Illumina-Htp', but not sure. The values we have probably also don't represent the actual teams very well either.

config/default_records/submission_templates/009_scrn_core_cell_extraction_submission_templates.yml
KatyTaylor: @StephenHulme and I discussed this one. I think what we want is to *not* specify the Project or Study at all here, and use the 'autodetect_studies_projects' feature in the API to choose the appropriate ones to put the requests under. (Unlike in Cardinal)

config/default_records/request_types/012_limber_scrna_core_cell_extraction.yml
```diff
@@ -0,0 +1,14 @@
+# These purposes are associated with the Limber scRNA Core Cell Extraction pipeline.
```
KatyTaylor, line 1:
```suggestion
# These request types are associated with the Limber scRNA Core Cell Extraction pipeline.
```

KatyTaylor (approved):

KatyTaylor:
Merging this in, although it's not UAT-ed yet, as it should be harmless in develop, and we want to simplify the branching in git.

Issue #3581, opened 2022-04-01
DPL-nnn Legacy js removal


JamesGlover:
**Describe the Housekeeping**
While migrating to vite, I noted a fail chunk of legacy inline javascript. We should probably move this into proper js files, and pass in properties via data attributes.

You should be able to find the legacy stuff by searching for `legacy_javascript_tag`

**Blocking issues**
Describe any other issues or tickets that may be blocking this change.

**Additional context**

PR #3907, opened 2023-10-04
Fix Create Plate Template page

StephenHulme:

The Create Plate Template page no longer works with the error:

> [Content Security Policy “Refused to execute inline event handler” error](https://stackoverflow.com/questions/58652892/content-security-policy-refused-to-execute-inline-event-handler-error)

More info: [Chrome CSP Docs](https://developer.chrome.com/docs/apps/contentSecurityPolicy/)

This PR fixes this by replacing the inline functionality with proper event handlers.
A properly integrated fix would be more desirable, but this will do to just get it working.

#### Related to

#3581 

### DIFF
--- a/app/views/plate_templates/_layout_plate.html.erb
+++ b/app/views/plate_templates/_layout_plate.html.erb
@@ -64,9 +64,9 @@
   function toggleEmptyWell(event) {
     const well = event.target;
     const wellPosition = well.dataset.wellPosition;
-    const row = parseInt(well.dataset.row);
-    const col = parseInt(well.dataset.col);
-    const plateCols = parseInt(well.dataset.plate_cols);
+    const row = parseInt(well.dataset.row, 10);
+    const col = parseInt(well.dataset.col, 10);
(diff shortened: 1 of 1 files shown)


app/views/plate_templates/_layout_plate.html.erb
```diff
@@ -50,13 +56,24 @@
...
+  for (let well of wells) {
+    well.addEventListener('click', toggleEmptyWell);
+  }
+
+  function toggleEmptyWell(event) {
+    const well = event.target;
+    const wellPosition = well.dataset.wellPosition;
+    const row = parseInt(well.dataset.row);
+    const col = parseInt(well.dataset.col);
+    const plateCols = parseInt(well.dataset.plate_cols);
```
emrojo, line 69:
```suggestion
    const row = parseInt(well.dataset.row, 10);
    const col = parseInt(well.dataset.col, 10);
    const plateCols = parseInt(well.dataset.plate_cols, 10);
```

emrojo (approved):
All good; we probably have a lot of other places where the events are inline and will probably fail in Chrome

PR #3911, opened 2023-10-05
Add more environment indicators to title and navbar

StephenHulme:

- Change favicon colour for different environments
  Screenshot 2023-10-26 at 11 44 15

- Changes background colour of the navigation bar for different environments
  - Development (and unknown) 
     Screenshot 2023-10-05 at 14 32 15
   - Training
      Screenshot 2023-10-26 at 11 24 30
   - Staging (UAT) is unchanged
     Screenshot 2023-10-05 at 14 33 58
   - Production is unchanged
     Screenshot 2023-10-05 at 14 34 12


app/frontend/stylesheets/all/header.scss
```diff
@@ -23,6 +23,12 @@ body.production {
   }
 }
 
+body.development {
+  .main-nav {
+    @extend .bg-secondary;
+  }
+}
+
```
yoldas, line 31: Are we trying to change the main-nav colour for each environment?
Why are we doing this only for the development environment here?
If we do this, doesn't it mean that we should do it for each environment?

Reminders:

In the existing code, we have the main-nav as red (for anything), we override it only for production by extending bg-primary and turn it to blue. See `app/frontend/stylesheets/all/header.scss`

In the existing code, logo_container/app_name div, the text below logo is for recognising the environment. See `app/views/shared/_header.html.erb`

app/views/layouts/application.html.erb
```diff
@@ -17,7 +17,22 @@
...
+          test: '🧪',
+          training: '🎓',
+        }
+
+        # Get an emoji for the current environment
+        current_environment = Rails.env.to_sym
+        emoji = environments[current_environment] || '❓' 
+      %>
+      <%= "#{emoji}:" unless Rails.env.production? %> Sequencescape : <%= controller.controller_name.capitalize %> - <%= @page_name || controller.action_name %>
     </title>
```
yoldas, line 36: It is changing the title text. It is putting an emoji between the favicon and the existing title. I think a neater way of doing this would be changing the favicon (logo) colour. 

When there are too many tabs open, it looks like a clutter and emoji (proposed start of title) goes hidden.

yoldas: - These are matching the navigation bar colours. 
- They remove the clutter from browser tab buttons. 
- They have simplicity compared to the emojis.
- They use the top field above twirling lines as the environment colour.
- They fill the space between white twirling lines same as the top field colour, following the same design as Production icon, to provide contrast and put the purpose in the middle as well.
- For Dev and UAT icons only, a single colour could be considered if it helps distinguishing in browser tab buttons but I think it is not necessary and against the original icon design.

Additional: Login page's login panel image could use the same pattern.

app/frontend/stylesheets/all/header.scss
yoldas: This is way better. The colours chosen are easy to remember. Thanks.

yoldas (approved):
I suggest changing the Login page's login panel image to use the same pattern. [Update: alt text of them needs changing as well if we do this] but the current changes for navigation bar and favicons look good.

emrojo (approved):

PR #3928, opened 2023-10-23
Dpl957 fix local integration suite

emrojo:
Closes #

- 


andrewsparkes (approved):
This is great.
Do we need something that also works for tubes? e.g. we might make a lot of vac tubes for scRNA or Cardinal. or FluidX tubes in racks for scRNA.

emrojo:
The problem was only happening with plates in development, because we were using a mocked service instead of Baracoda and returning the barcodes generated in memory, with no record from a previous one. In case of tubes the barcodes are created in the database directly, which is creating the record in the database before serving it, so the race condition should not happen there.

PR #3929, opened 2023-10-23
Dpl 926 - scRNA core - updating request state check

andrewsparkes:
Closes #https://github.com/sanger/sequencescape/issues/3957

Close off requests on wells when updating state of well to failed, but not if other wells on plate have the same request


spec/models/state_changer/standard_plate_spec.rb
```diff
@@ -84,5 +84,138 @@
...
+
+        # A1 request should be failed
+        expect(reloaded_library_requests.first).to be_failed
+
+        # Other requests in B1 and C1 should be unchanged
+        expect([reloaded_library_requests[1], reloaded_library_requests.last]).to all(be_started)
+      end
+    end
+
+    context 'when a specific well is failed but the request is shared by other wells' do
```
KatyTaylor, line 105: Small refactoring suggestions, saves a few lines. Only functionality change is that when you're grabbing the parent well requests, I think you should be specifying the request type (library), as limber_shared_context seems to add multiplexing requests too - I've done that below using `library_requests`.
```
include_context 'a limber target plate with submissions'
      let(:tested_wells) { 2 }
      let(:target_state) { 'failed' }

      # override default target plate creation
      let(:target_plate) { create :plate, well_count: 4 }
      let(:target_wells) { target_plate.wells.index_by(&:map_description) }

      before do
        parent_wells = input_plate.wells.index_by(&:map_description)

        source_well_a1_request = library_requests.detect { |r| r.source_well == parent_wells['A1'] } # or just library_requests[0]
        source_well_b1_request = library_requests.detect { |r| r.source_well == parent_wells['B1'] } # or just library_requests[1]

        # set up wells in target plate
        # A1, B1 and C1 created by transferring from parent A1 (1st library request)
        target_wells['A1'].stock_well_links <<
          build(:stock_well_link, target_well: target_wells['A1'], source_well: parent_wells['A1'])
        create :transfer_request,
               asset: parent_wells['A1'],
               target_asset: target_wells['A1'],
               outer_request: source_well_a1_request

        target_wells['B1'].stock_well_links <<
          build(:stock_well_link, target_well: target_wells['B1'], source_well: parent_wells['A1'])
        create :transfer_request,
               asset: parent_wells['A1'],
               target_asset: target_wells['B1'],
               outer_request: source_well_a1_request

        target_wells['C1'].stock_well_links <<
          build(:stock_well_link, target_well: target_wells['C1'], source_well: parent_wells['A1'])
        create :transfer_request,
               asset: parent_wells['A1'],
               target_asset: target_wells['C1'],
               outer_request: source_well_a1_request

        # D1 created by transferring from parent B1 (2nd library request)
        target_wells['D1'].stock_well_links <<
          build(:stock_well_link, target_well: target_wells['D1'], source_well: parent_wells['B1'])
        create :transfer_request,
               asset: parent_wells['B1'],
               target_asset: target_wells['D1'],
               outer_request: source_well_b1_request
      end
```

spec/models/state_changer/standard_plate_spec.rb
KatyTaylor, line 1: Is it worth also considering the scenario where there are wells with the same request in wells on other plates?

app/models/state_changer/standard_plate.rb
```diff
@@ -56,8 +56,20 @@ def update_associated_requests
       raise_request_error if associated_requests.empty? && associated_submission?
 
       associated_requests.each do |request|
+        # customer_accepts_responsibility = nil
```
KatyTaylor, line 59: Is this comment meant to be here? In what scenario is it nil?

KatyTaylor (changes requested):

yoldas (commented):
Looks good.

spec/models/state_changer/standard_plate_spec.rb
```diff
@@ -84,5 +84,138 @@
...
       end
     end
+
+    context 'when a specific well is failed' do
+      include_context 'a limber target plate with submissions'
+      let(:target_state) { 'failed' }
+      let(:contents) { ['A1'] } # This is the well that will be failed
+
+      it 'allows failure of the specified requests', :aggregate_failures do
+        expect { state_changer.update_labware_state }.not_to change(BroadcastEvent::LibraryStart, :count)
```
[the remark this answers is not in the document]
andrewsparkes: Nor me, but it's in other tests in that file where it doesn't seem relevant either. Some state changes must trigger events and this is checking we've not triggered one inadvertently?

spec/models/state_changer/standard_plate_spec.rb
andrewsparkes: I also thought about that, but how to implement this? We are on this specific plate here. The request is linked to any ancestor and descendent wells which would have to be filtered out of any search (or would they? what if you create 2 child plates from the same parent, they each get the request? how does that work currently if they then fail a well on one of those? I expect they are supposed to cancel rather than fail).
I feel like this code is an improvement which solves our current situation. Maybe it would be enough to leave a TODO comment saying this might need expanding to include other plates?

spec/models/state_changer/standard_plate_spec.rb
andrewsparkes: Thanks, changed.
Had originally created a parent plate with only a library request, not multiplexing, but couldn't get it all to fully work (factory pain).

app/models/state_changer/standard_plate.rb
andrewsparkes: no, removed it
I think it's value was nil when I was running tests and was leaving myself a comment then forgot to remove it

spec/models/state_changer/standard_plate_spec.rb
KatyTaylor: True, ancestor and descendent wells make it more complicated to find relevant wells. Happy to leave it at this for now. I think adding that TODO comment is a good idea, in case it helps debugging an issue related to this in future.

KatyTaylor (approved):

PR #3930, opened 2023-10-25
Fix training integration run

emrojo:
Closes #

- 

### DIFF
--- a/app/uat_actions/uat_actions/generate_tag_group.rb
+++ b/app/uat_actions/uat_actions/generate_tag_group.rb
@@ -61,14 +61,8 @@ class UatActions::GenerateTagGroup < UatActions
     # All the form fields are accessible as simple attributes.
     # Return true if everything works
     report[:name] = name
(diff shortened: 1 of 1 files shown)


app/uat_actions/uat_actions/generate_tag_group.rb
```diff
@@ -61,7 +61,8 @@ def perform
     # All the form fields are accessible as simple attributes.
     # Return true if everything works
     report[:name] = name
-    return true if existing_tag_group
+
+    existing_tag_group&.destroy
```
emrojo, line 65: The same tag group name UAT 96 Tag Group is used in some tests with adapter type Chromium and in other tests without it. This is causing that the tag group is not displayed in the tag group interface if one of the tests was executed before the other, because then the configuration of the tag group is different and so they interface cannot find it causing a test to fail.
Instead of retrieving the tag group, as it could have changed, we destroy and recreate it instead, which it makes more sense in case of testing data.

PR #3946, opened 2023-11-03
DPL-911 correct study and project for cell extraction work

KatyTaylor:

- Implement autodetection for studies and projects independently, rather than requiring the same behaviour for both.

Feels pretty verbose for a small change, so open to refactoring suggestions.

### DIFF
--- a/app/api/io/order.rb
+++ b/app/api/io/order.rb
@@ -38,6 +38,8 @@ class Io::Order < Core::Io::Base
                                            user <=  user
 
                     autodetect_studies_projects <= autodetect_studies_projects
+                             autodetect_studies <= autodetect_studies
+                            autodetect_projects <= autodetect_projects
   '
   )
 end
--- a/app/api/io/order.rb
+++ b/app/api/io/order.rb
@@ -37,7 +37,6 @@ class Io::Order < Core::Io::Base
 
                                            user <=  user
 
-                    autodetect_studies_projects <= autodetect_studies_projects
                              autodetect_studies <= autodetect_studies
                             autodetect_projects <= autodetect_projects
   '
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -106,9 +106,11 @@ class Order < ApplicationRecord # rubocop:todo Metrics/ClassLength
   end
 
   def autodetect_studies?
-    # need this check here, as this method is called before validation
+    # Method is called before validation is run - check if we've got valid autodetection instructions.
     return autodetection_default if autodetection_clash?
 
+    # If the instructions are valid, use whichever one is not nil.
+    # Use explicit nil checks here to distinguish between nil (behaviour not specified) and false (a valid instruction).
     return autodetect_studies unless autodetect_studies.nil?
 
     return autodetect_studies_projects unless autodetect_studies_projects.nil?
@@ -117,9 +119,11 @@ class Order < ApplicationRecord # rubocop:todo Metrics/ClassLength
   end
 
   def autodetect_projects?
-    # need this check here, as this method is called before validation
+    # Method is called before validation is run - check if we've got valid autodetection instructions.
     return autodetection_default if autodetection_clash?
 
+    # If the instructions are valid, use whichever one is not nil.
+    # Use explicit nil checks here to distinguish between nil (behaviour not specified) and false (a valid instruction).
     return autodetect_projects unless autodetect_projects.nil?
 
     return autodetect_studies_projects unless autodetect_studies_projects.nil?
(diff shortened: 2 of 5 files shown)


config/default_records/submission_templates/009_scrna_core_cell_extraction_submission_templates.wip.yml
```diff
@@ -6,7 +6,6 @@ Limber - scRNA Core Cell Extraction:
   submission_class_name: "LinearSubmission"
   related_records:
     request_type_keys: ["limber_scrna_core_cell_extraction"]
-    project_name: "UAT Project"
-    study_name: "UAT Study"
+    project_name: "S0842"
```
KatyTaylor, line 9: N.B. had to delete the submission template, and call `WIP=009_scrna_core_cell_extraction_submission_templates bundle exec rake record_loader:all` to get this to update.

KatyTaylor: TODO: Integration Suite test update.

yoldas (changes requested):
I have suggested using the present checks, instead of nil checks and negations, in conditionals to make them easier to read.

config/default_records/submission_templates/009_scrna_core_cell_extraction_submission_templates.wip.yml
KatyTaylor: Update: Looks like Integration Suite update not necessary.

app/models/order.rb
```diff
@@ -97,6 +105,32 @@ def render_class
     end
   end
 
+  def autodetect_studies?
+    # need this check here, as this method is called before validation
+    return autodetection_default if autodetection_clash?
+
+    return autodetect_studies unless autodetect_studies.nil?
+
+    return autodetect_studies_projects unless autodetect_studies_projects.nil?
```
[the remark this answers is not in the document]
KatyTaylor: I thought about that, but I don't think it actually is exactly the opposite of nil?

It doesn't distinguish between `nil` and `false`, which is very important here, as I want to know if the attribute has been purposefully set, or just left blank.

Screenshot 2023-11-06 at 09 49 07

spec/models/orders/order_spec.rb
```diff
@@ -41,10 +49,110 @@
...
+        let(:autodetect_projects) { true }
+
+        it 'fails validation' do
+          expect(order).not_to be_valid
+          expect(
+            order.errors.full_messages.one? { |message| message.include? 'autodetect_studies_projects' }
+          ).to be true
+        end
+
+        it 'uses the defaults' do
```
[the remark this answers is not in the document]
KatyTaylor: This is because it actually uses the value before the validation runs - see the following line:
`before_validation :set_study_from_aliquots, unless: :cross_study_allowed, if: :autodetect_studies?`

So I've got to make a decision about the behaviour even when there's a clash, because it won't have run the validation.

Happy to change it - what would you do here? (Or can just add a comment)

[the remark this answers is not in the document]
yoldas: I would just add a comment in the test about it I think.

app/models/order.rb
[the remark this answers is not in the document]
yoldas: I see why you need to check nil explicitly. Keeping it and some comment about the logic may be helpful.

app/api/io/order.rb
```diff
@@ -38,6 +38,8 @@ def self.json_field_for(attribute)
                                            user <=  user
 
                     autodetect_studies_projects <= autodetect_studies_projects
```
andrewsparkes, line 40: can autodetect_studies_project go if you have the split options? doesn't look like it's used much elsewhere (couple places in Limber where set to true)

KatyTaylor: Good suggestion, have implemented this now.

config/default_records/submission_templates/009_scrna_core_cell_extraction_submission_templates.wip.yml
andrewsparkes: That looks like a cost code rather than a project name, is it correct to use that?

andrewsparkes (approved):
Just the cost code comment otherwise looks good

config/default_records/submission_templates/009_scrna_core_cell_extraction_submission_templates.wip.yml
KatyTaylor: That's how Rich C set it up in SS - how he'd like them all to be.

PR #3955, opened 2023-11-10
dpl-942 bug unable to print swipecard id barcode

yoldas:
Closes #

- Use swipecard_pmb_template for swipe card printing, instead of 
the sqsc_96plate_label_template_code39 plate template picked based on printer
- Change label_class to return values for swipecard_pmb_template fields instead of the plate template
- Change download directory to string for Chrome Options in order to avoid creating temporary files in the project directory instead of tmp.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -55,7 +55,8 @@ class UsersController < ApplicationController
         printer,
         LabelPrinter::Label::Swipecard,
         user_login: @user.login.truncate(10, omission: '..'),
-        swipecard: swipecard
+        swipecard: swipecard,
+        label_template_name: configatron.swipecard_pmb_template
       )
     if print_job.execute
       flash[:notice] = print_job.success
--- a/lib/label_printer/label_printer/print_job.rb
+++ b/lib/label_printer/label_printer/print_job.rb
@@ -42,7 +42,15 @@ module LabelPrinter
       @labels = label_class.new(options.merge(printer_type_class)).labels
     end
 
+    # Returns the name of the label template to use for this print job.
+    # If not specified in the options during initialisation, the label template
+    # name configured for the printer in the database is used.
+    #
+    # @return [String] the name of the label template to use for this print job
+    #
     def label_template_name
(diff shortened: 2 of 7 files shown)


lib/label_printer/label_printer/print_job.rb
```diff
@@ -42,7 +42,15 @@ def labels_attribute
...
     end
 
+    # Returns the name of the label template to use for this print job.
+    # If not specified in the options during initialisation, the label template
+    # name configured for the printer in the database is used.
+    #
+    # @return [String] the name of the label template to use for this print job
+    #
     def label_template_name
+      return options[:label_template_name] if options[:label_template_name]
```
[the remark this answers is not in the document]
yoldas: The first call is checking if options[:label_template_name] is truthy (i.e., not nil or false). If it is, the method immediately returns its value. If it's not, the method continues to the next lines to find the printer and return the label_template_name from the printer's barcode printer type.

yoldas:
Check https://www.dockerstatus.com as tests are failing because of Docker Registry.

app/controllers/users_controller.rb
```diff
@@ -55,7 +55,8 @@ def print_swipecard_with_pmb(swipecard, printer) # rubocop:todo Metrics/MethodLe
         printer,
         LabelPrinter::Label::Swipecard,
```
KatyTaylor, line 56: Do you think it would be appropriate to store the label template name inside this `LabelPrinter::Label::Swipecard` class?

KatyTaylor (approved):
Approved but left one idea.

app/controllers/users_controller.rb
yoldas: It could be done but it would need need more thinking.
- Label class is providing values for placeholders only and it provides data. Template is doing the presentation.
- Template name is passed separately to PMB API as part of  (printer name, label data and template name)
- I checked Limber and a few other places, template and label data are handled separately.
- Passing template name to Label class would give a chance to make decisions based on template, but it does not look necessary.

PR #3966, opened 2023-11-17
Group UAT Actions

StephenHulme:
Makes it much easier to locate the action you are looking for

- Groups UAT action into logical categories based on title
- Adjusts action card height for consistency
- Fixes the capitalisation of the page title

| Before | After |
| --------------- | --------------- | 
| before | after (1) | 

### DIFF
--- a/app/uat_actions/uat_actions.rb
+++ b/app/uat_actions/uat_actions.rb
@@ -51,10 +51,10 @@ class UatActions
     # Returns a hash of all registered uat_actions grouped by category and sorted
     def grouped_and_sorted_uat_actions
       # raise error if any categories are not in the list
-      uat_actions.each do |_, uat_action|
+      all.each do |uat_action|
         unless CATEGORY_LIST.include?(uat_action.category)
           raise "Category '#{uat_action.category}' from '#{uat_action}' is not in the list" \
-                " of categories #{CATEGORY_LIST}"
+                  " of categories #{CATEGORY_LIST}"
         end
       end
 
--- a/spec/uat_actions/uat_actions_spec.rb
+++ b/spec/uat_actions/uat_actions_spec.rb
@@ -2,57 +2,79 @@
 require 'rails_helper'
 
 # Test classes for UatActions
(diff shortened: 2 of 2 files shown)


KatyTudor:
@StephenHulme I think you have included me but you might have meant to include Katy Taylor

app/helpers/uat_actions_helper.rb
BenTopping, line 1: Does this need to be in a helper file? I wouldn't say its reusable. Maybe better suited as a self method on uat_actions?

BenTopping (commented):
Definitely makes it more readable 👌. Maybe worth adding a test file for uat_actions, but doesn't add loads of value so up to you.

app/uat_actions/uat_actions.rb
```diff
@@ -48,6 +48,11 @@ def category
       CATEGORY_LIST.detect { |category| name.include?(category) } || CATEGORY_LIST.last
     end
 
+    # Returns a hash of all registered uat_actions grouped by category and sorted
+    def grouped_and_sorted_uat_actions(uat_actions)
+      uat_actions.group_by(&:category).sort_by { |category, _| UatActions::CATEGORY_LIST.index(category) }
```
BenTopping, line 53: uat_actions should now be accessible in this scope so you may not need to pass it in. Similarly UatActions::CATEGORY_LIST can just become CATEGORY_LIST.

KatyTaylor:
Looks good. Had a bit of a think about categories but struggled to find good words! V rough ideas below 🤷‍♀️ :

Setting up:
- Study
- Project

Importing samples:
- Manifest
- Plates
- Tubes

Requesting work:
- Submission

Auxiliary data:
- Primer panels
- Tags
- SpikedBuffer tubes

QC:
- Conc
- QC results

Testing tools:
- Int Suite tools
- Plate information

KatyTaylor:
> Maybe something like?

Sure, looks good to me. We can always tweak it in future if people think of better ones!

KatyTaylor (approved):

PR #3979, opened 2023-12-05
DPL-1002 Sample Prep ability to create Working Dilution from Stock RNA Plate

yoldas:
Closes #

- Allow creating Working Dilution from Stock RNA Plate

### DIFF
--- a/db/migrate/20231205095734_add_stock_rna_plate_to_working_dilution_parent_purposes.rb
+++ b/db/migrate/20231205095734_add_stock_rna_plate_to_working_dilution_parent_purposes.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+# This migration adds the 'Stock RNA Plate' plate purpose to the parent plate
+# purposes of the 'Working dilution' plate creator.
+#
+# If the 'Stock RNA Plate' plate purpose does not exist in the database, it is
+# created first.
+#
+# It is assumed that the 'Working dilution' plate creator exists and its plate
+# purposes contain 'Working Dilution' purpose but its parent plate purposes
+# does not contain the 'Stock RNA Plate' purpose.
+#
+# If the migration is rolled back, the 'Stock RNA Plate' purpose is removed from
+# the parent plate purposes of the 'Working Dilution' plate creator. However,
+# the 'Stock RNA Plate' plate purpose is kept in the database.
+#
+class AddStockRnaPlateToWorkingDilutionParentPurposes < ActiveRecord::Migration[6.0]
+  PURPOSE_NAME = 'Stock RNA Plate'
+  CREATOR_NAME = 'Working dilution'
+
+  def up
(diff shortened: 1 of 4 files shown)


db/migrate/20231205095734_add_stock_rna_plate_to_working_dilution_parent_purposes.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+# It is assumed that the 'Working dilution' plate creator exists and its plate
+# purposes contain 'Working Dilution' purpose but its parent plate purposes
+# does not contain the 'Stock RNA Plate' purpose.
+#
+# If the migration is rolled back, the 'Stock RNA Plate' purpose is removed from
+# the parent plate purposes of the 'Working dilution' plate creator. However,
+# the 'Stock RNA Plate' plate purpose is kept in the database.
+#
+class AddStockRnaPlateToWorkingDilutionParentPurposes < ActiveRecord::Migration[6.0]
```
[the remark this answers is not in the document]
yoldas: It was not possible to the record loader for this because it does not allow updates.

I have removed the migration and seeds now. I added a rake task. It is just for adding a record but the benefit is having a test going together.

lib/tasks/support/add_stock_rna_plate_to_working_dilution_parents.rake
```diff
@@ -0,0 +1,21 @@
...
+    # plate purposes contain the 'Working Dilution' purpose, but its parent
+    # plate purposes do not contain the 'Stock RNA Plate' purpose. If the
+    # purpose is already in the parent plate purposes, it is not added again.
+    purpose_name = 'Stock RNA Plate'
+    creator_name = 'Working dilution' # The second initial is lowercase in production.
+    purpose = PlatePurpose.find_or_create_by!(name: purpose_name) { |p| p.stock_plate = true }
+    creator = Plate::Creator.find_by!(name: creator_name)
+    creator.parent_plate_purposes << purpose unless creator.parent_plate_purposes.include?(purpose)
+  end
+end
```
KatyTaylor, line 21: I didn't know (or forgot) this was all controlled by configuration data. Are we talking about the /plates/new page here?

KatyTaylor (approved):

emrojo (approved):
All is good! in my case I would not have added a test, but now that is done and everything is probably good to keep. If it gives problem in future we could disable it.

lib/tasks/support/add_stock_rna_plate_to_working_dilution_parents.rake
yoldas: They can't create Working Dilution plate from Stock RNA plates at the moment. They change their parent Stock RNA Plate to Stock Plate, create the Working Dilution plate, and revert the purpose of the parent back to Stock RNA plate as the workaround. This update allows having a Stock RNA plate as the parent directly. 

Not all the records for all these similar ones were not done using config. Record loader does not allow updates. We need to create one record. I am doing this with a rake task so it is tested.

PR #3994, opened 2023-12-21
DPL-969 [Sequencescape] Support for Chromium chip labware

yoldas:
Closes #

- Add asset shapes to record loading
- Add plate purpose resource
- Use SS API v2 for plate purpose registration
- Create purpose and ChromiumChip 16-well plates
- Test Coordinate module (well description strategy)
- Test Labware view for ChromiumChip plates

### DIFF
--- a/app/controllers/api/v2/plate_purposes_controller.rb
+++ b/app/controllers/api/v2/plate_purposes_controller.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # Provides a JSON API controller for receptacle
+    # See: http://jsonapi-resources.com/ for JSONAPI::Resource documentation
+    class PlatePurposesController < JSONAPI::ResourceController
+      # By default JSONAPI::ResourceController provides most the standard
+      # behaviour, and in many cases this file may be left empty.
+    end
+  end
+end
--- a/app/models/map.rb
+++ b/app/models/map.rb
@@ -18,7 +18,18 @@ class Map < ApplicationRecord
     # TODO: These methods are only valid for standard plates. Moved them here to make that more explicit
     # (even if its not strictly appropriate) They could do with refactoring/removing.
 
-    PLATE_DIMENSIONS = Hash.new { |_h, _k| [] }.merge(96 => [12, 8], 384 => [24, 16])
+    # A hash representing the dimensions of different types of plates.
+    # The keys are the total number of wells in the plate, and the values are
+    # arrays, where the first element is the number of columns and the second
+    # element is the number of rows.
+    #
+    # @note
+    #   - 96 represents a 96-well plate, arranged in 12 columns and 8 rows.
+    #   - 384 represents a 384-well plate, arranged in 24 columns and 16 rows.
+    #   - 16 represents a 16-well Chromium Chip , which has 8 columns and 2 rows.
+    #
+    # @return [Hash{Integer => Array<Integer>}] the dimensions of the plates
+    PLATE_DIMENSIONS = Hash.new { |_h, _k| [] }.merge(96 => [12, 8], 384 => [24, 16], 16 => [8, 2])
 
     # Seems to expect row to be zero-indexed but column to be 1 indexed
     def self.location_from_row_and_column(row, column, _ = nil, __ = nil)
--- a/app/models/map.rb
+++ b/app/models/map.rb
@@ -26,9 +26,9 @@ class Map < ApplicationRecord
     # @note
     #   - 96 represents a 96-well plate, arranged in 12 columns and 8 rows.
     #   - 384 represents a 384-well plate, arranged in 24 columns and 16 rows.
-    #   - 16 represents a 16-well Chromium Chip , which has 8 columns and 2 rows.
-    #     Although a 16-well Chromium Chip does not have 4:3 ratio to be a standard plate,
-    #     the methods here still apply.
+    #   - 16 represents a 16-well Chromium Chip, which has 8 columns and 2 rows.
+    #     Although a 16-well Chromium Chip does not have 3:2 ratio to be a
+    #     standard plate, i.e. it has 4:1 ratio, the methods here still apply.
     # @return [Hash{Integer => Array<Integer>}] the dimensions of the plates
     PLATE_DIMENSIONS = Hash.new { |_h, _k| [] }.merge(96 => [12, 8], 384 => [24, 16], 16 => [8, 2])
 
--- a/app/models/map.rb
+++ b/app/models/map.rb
@@ -27,7 +27,8 @@ class Map < ApplicationRecord
     #   - 96 represents a 96-well plate, arranged in 12 columns and 8 rows.
     #   - 384 represents a 384-well plate, arranged in 24 columns and 16 rows.
     #   - 16 represents a 16-well Chromium Chip , which has 8 columns and 2 rows.
-    #
+    #     Although a 16-well Chromium Chip does not have 4:3 ratio to be a standard plate,
+    #     the methods here still apply.
     # @return [Hash{Integer => Array<Integer>}] the dimensions of the plates
     PLATE_DIMENSIONS = Hash.new { |_h, _k| [] }.merge(96 => [12, 8], 384 => [24, 16], 16 => [8, 2])
 
--- a/app/models/map.rb
+++ b/app/models/map.rb
@@ -281,5 +281,6 @@ class Map < ApplicationRecord
         .order(:row_order)
         .each { |position| yield(position, position.row_order) }
     end
+    alias walk_plate_horizontally walk_plate_in_row_major_order
   end
 end
--- a/app/resources/api/v2/plate_purpose_resource.rb
+++ b/app/resources/api/v2/plate_purpose_resource.rb
@@ -11,7 +11,7 @@ module Api
       # migrate the registration of plate purposes done by the Limber rake
       # task config:generate from API version 1 to API version 2.
 
-      # The following options are sent by Limber for a new plate purpose.
+      # The following attributes are sent by Limber for a new plate purpose.
 
       # @!attribute name
       #  @return [String] the name of the plate purpose
@@ -36,6 +36,12 @@ module Api
       # @!attribute asset_shape
       #  @return [String] the name of the shape of the plates of this purpose
       attribute :asset_shape
+
+      # The following attribute is required by Limber to store purposes.
+
+      # @!attribute [r] uuid
+      #  @return [String] the UUID of the plate purpose
+      attribute :uuid, readonly: true
 
       # Sets the asset shape of the plate purpose by name if given.
       # 'asset_shape' can be given via the Limber purpose configuration and
--- a/app/resources/api/v2/plate_purpose_resource.rb
+++ b/app/resources/api/v2/plate_purpose_resource.rb
@@ -2,7 +2,7 @@
 
 module Api
   module V2
-    # Provides a JSON API representation of Plate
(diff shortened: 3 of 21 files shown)


app/resources/api/v2/plate_purpose_resource.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # Provides a JSON API representation of Plate
```
KatyTaylor, line 5:
```suggestion
    # Provides a JSON API representation of PlatePurpose
```

app/resources/api/v2/plate_purpose_resource.rb
```diff
@@ -0,0 +1,85 @@
...
+    # See: http://jsonapi-resources.com/ for JSONAPI::Resource documentation
+    class PlatePurposeResource < BaseResource
+      model_name 'PlatePurpose'
+
+      # This resource is similar to PurposeResource but it was created to
+      # migrate the registration of plate purposes done by the Limber rake
+      # task config:generate from API version 1 to API version 2.
```
KatyTaylor, line 12: Why was it necessary to create PlatePurposeResource? Did you consider adding to PurposeResource instead?

app/resources/api/v2/plate_purpose_resource.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      # The following attributes are sent by Limber for a new plate purpose.
+
+      # @!attribute name
+      #  @return [String] the name of the plate purpose
+      attribute :name
```
KatyTaylor, line 18: It's probably a personal thing, but I don't find these kind of comments helpful - you're not really saying anything in lines 16 and 17 that isn't obvious from reading line 18 (same for most of the attributes below).

app/models/map.rb
KatyTaylor, line 1: Nice docs thank you.

config/default_records/asset_shapes/default_records.yml
```diff
@@ -0,0 +1,62 @@
+# This file contains the AssetShape and associated Map records to be created
+# in deployment environments by RecordLoader. The loading is triggered by
+# the "post_deploy" task. The records are checked against the database by
+# their names. If the records already exist, they are not created.
+# AssetShape records are created in the "asset_shapes" table, and Map records
+# are created in the "maps" table. Each section in this file starts with the
+# name of the AssetShape record. If the name is defined again in the options,
+# it takes priority. The "horizontal_ratio" and "vertical_ratio" options
```
KatyTaylor, line 8: I didn't know what you meant by "If the name is defined again in the options, it takes priority." 
But I think I figured it out, you mean if there's `name: A different name`, for instance? Might be worth saying that in case it's not clear to anyone else.

config/default_records/asset_shapes/default_records.yml
```diff
@@ -0,0 +1,62 @@
...
+# LRC HT 5p Chip:
+#   :asset_shape: ChromiumChip
+#   :size: 16
+#
+# The information in this file is duplicated in a couple of places in the
+# codebase. When the local development environment is "setup" or "reset", a
+# database "seed" is executed. This results in using the maps hash in
+# PlateMapGeneration. The same hash is used when the RSpec before suite hook
+# calls PlateMapGeneration to create records in the local test environment.
+# Plate sizes and number or rows and columns are also defined separately in
```
KatyTaylor, line 31:
```suggestion
# Plate sizes and number of rows and columns are also defined separately in
```

config/default_records/asset_shapes/default_records.yml
```diff
@@ -0,0 +1,62 @@
...
+# a hash in the Map model, which is used by the nested Coordinate module.
```
KatyTaylor, line 32: Lengthy but all good to know I think! :)

lib/record_loader/asset_shape_loader.rb
```diff
@@ -0,0 +1,28 @@
...
+    # with the same name already exists, it is skipped. If not, a new one is
+    # created. For each size specified in the options, Map records are generated
+    # up to that size, unless they already exist.
+    #
+    # @param name [String] the name of the AssetShape record
+    # @param options [Hash] the options to be used for creating the record
+    # @param options [String] :name If provided, this will override the 'name' parameter
+    # @option options [Integer] :horizontal_ratio The horizontal ratio of the plate
+    # @option options [Integer] :vertical_ratio The vertical ratio of the plate
+    # @option options [String] :description_strategy The strategy for describing the plate
```
KatyTaylor, line 20: I think the descriptions for horizontal_ratio, vertical_ratio and description_strategy are not useful. Either leave them out, or provide more information, or just leave a reference to the extensive docs you made in config/default_records/asset_shapes/default_records.yml ?

lib/record_loader/asset_shape_loader.rb
```diff
@@ -0,0 +1,28 @@
...
+    # @option options [Array<Integer>] :sizes The sizes of the plates to generate Maps for
+    def create_or_update!(name, options)
+      config = { name: name }.merge(options.symbolize_keys)
+      config[:description_strategy] = config[:description_strategy].delete_prefix('Map::')
```
KatyTaylor, line 24: Where does this 'Map::' prefix come from? I don't see it in config/default_records/asset_shapes/default_records.yml ?

KatyTaylor (changes requested):

lib/record_loader/asset_shape_loader.rb
```diff
@@ -0,0 +1,28 @@
...
+    # up to that size, unless they already exist.
+    #
+    # @param name [String] the name of the AssetShape record
+    # @param options [Hash] the options to be used for creating or updating the record
+    # @param options [String] :name If provided, this will override the 'name' parameter
+    # @option options [Integer] :horizontal_ratio The horizontal ratio of the plate
+    # @option options [Integer] :vertical_ratio The vertical ratio of the plate
+    # @option options [String] :description_strategy The strategy for describing the plate
+    # @option options [Array<Integer>] :sizes The sizes of the plates to generate Maps for
+    def create_or_update!(name, options)
```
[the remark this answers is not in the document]
yoldas: The method is required by the [RecordLoader](https://github.com/sanger/record_loader). It is expected to be an instance method with the same signature.

app/controllers/api/v2/plate_purposes_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
```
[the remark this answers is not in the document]
yoldas: I can't find which codeclimate plugin and rule is causing this comment. I am guessing it is about short naming but we cannot change the module V2 to something else. It is part of the routes and fully qualified class names in both Sequencescape and Limber side.

app/resources/api/v2/plate_purpose_resource.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # Provides a JSON API representation of Plate
+    # See: http://jsonapi-resources.com/ for JSONAPI::Resource documentation
+    class PlatePurposeResource < BaseResource
```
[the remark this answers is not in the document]
yoldas: It is the model being wrapped by this resource. We are passing a custom attribute for now, instead of exposing the full asset_shape relationship. I don't see an issue with that.

app/resources/api/v2/plate_purpose_resource.rb
yoldas: Purpose resource is "included" in other resources and used by Limber. I wanted to avoid make changes in Purpose resource. The use-case for having a plate purpose resource at this time is migrating plate purpose creation to SS API v2  to support passing a custom attribute 'asset_shape' as a name. When we start working on DPL-919 and after deciding whether to expose an AssetShape resource instead of passing name, we can think about unifying all Purpose, plate purpose and tube purposes in v2.

app/resources/api/v2/plate_purpose_resource.rb
yoldas: I agree it is very chatty to describe a single attribute but I definitely want to see the type information and a short description. I can remove "@!attribute" lines so repeating names is avoided. It will not be fully compliant but it may serve the purpose. See: https://rubydoc.info/gems/yard/file/docs/Tags.md#attribute

lib/record_loader/asset_shape_loader.rb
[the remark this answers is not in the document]
yoldas: We do not update the records; we only create if they are missing. However, the record loader must have this method signature.

config/default_records/asset_shapes/default_records.yml
yoldas: I have removed the comment and updated the loader as well because it sounds like this is the behaviour of all record loaders, which is misleading. I will push change shortly.

lib/record_loader/asset_shape_loader.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+# RecordLoader handles automatic population and updating of database records
+# across different environments
+# @see https://rubydoc.info/github/sanger/record_loader/
+module RecordLoader
+  class AssetShapeLoader < ApplicationRecordLoader
```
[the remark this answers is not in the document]
yoldas: I have added a similar comment used by other record loaders. I will push the change shortly.

config/default_records/asset_shapes/default_records.yml
yoldas: It documents the duplication and I think it is helpful.

lib/record_loader/asset_shape_loader.rb
yoldas: I have updated 

- the option documentation with

```
    # @option options [Integer] :horizontal_ratio the numerator in the simplest form of the plate width/height ratio
    # @option options [Integer] :vertical_ratio the denominator in the simplest form of the plate width/height ratio
```

- the default_records.yml file with 

```
# record. The "horizontal_ratio" and "vertical_ratio" options are used to define
# the plate's shape. These options represent the simplest form of the ratio
# between the number of columns and rows (i.e. width / height). Specifically,
# "horizontal_ratio" corresponds to the number of columns (numerator), and
# "vertical_ratio" corresponds to the number of rows (denominator). For
# instance, a "Standard" 96-well plate has a horizontal ratio of 3 and a
# vertical ratio of 2 because it has 12 columns (1 to 12) and 8 rows (A to H).
```

I will push changes shortly.

lib/record_loader/asset_shape_loader.rb
yoldas: Now I have added the following comment just before the line to explain.

# PlateMapGeneration expects a non-namespaced constant for
# description_strategy. It adds "Map::" prefix to refer to a nested
# module in the Map class. We remove this prefix in case it is given
# in the records file.

I will push the change shortly.

KatyTaylor: Ah so you're future proofing in case someone follows a different convention to you when using record loader?

KatyTaylor (approved):

lib/record_loader/asset_shape_loader.rb
yoldas: I have reused the PlateMapGeneration class and kept the existing behaviour. I did not want to change how it works. It is also called from the spec_helper before-suit-hook to initialise the test database, and from seeds to initialise the development database. The name-spaced name is written in the asset_shapes table's description_strategy column.

PR #3999, opened 2024-01-09
DPL-975b targeted nanoseq poly metadata ss api v2

andrewsparkes:
Closes #3981


app/models/poly_metadatum.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# A polymetadatum can be assigned to any metadatable record.
```
KatyTaylor, line 3: This makes it sound to me like `metadatable` is the name of an actual model (would have me searching for one called 'metadatable' until I noticed the 'polymorphic' keyword further down and remembered what it meant). Could we add a bit more detail, something like:

A polymetadatum is a key value pair store. It is built to be generic so that it can be associated with multiple different models. It can be linked to any model that has the reverse association set up.

db/migrate/20231204163029_create_poly_metadata.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+#
+# Add polymorphic metadata table to flexibly hold key value pairs
+# See concern metadatable
```
KatyTaylor, line 4: Did there used to be a concern and then you got rid of it?

app/models/request.rb
```diff
@@ -87,6 +87,9 @@ def date_for_state(state)
   # reflect the final product of a request.
   has_many :related_aliquots, class_name: 'Aliquot', inverse_of: :request
 
+  # Can have many key value pairs of metadata
+  has_many :poly_metadata, as: :metadatable
```
KatyTaylor, line 91: Think it makes sense to add `dependent: destroy` here, so if a request is destroyed, so is its metadata.

app/models/poly_metadatum.rb
```diff
@@ -0,0 +1,17 @@
...
+class PolyMetadatum < ApplicationRecord
+  # Associations
+  belongs_to :metadatable, polymorphic: true, optional: false
+
+  # Validations
+  validates :key, presence: true # otherwise nil is a valid key
+  validates :value, presence: true
```
KatyTaylor, line 10: Is it intentional that the `presence: true` is not enforced at the database level?

spec/models/poly_metadatum_spec.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
```
KatyTaylor, line 2: Some matchers I didn't know about here 😃

spec/models/poly_metadatum_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    it 'requires a key' do
+      expect(test_metadatum).to validate_presence_of(:key)
+    end
+
+    it 'requires a value' do
+      expect(test_metadatum).to validate_presence_of(:value)
+    end
+
+    it 'key must be unique' do
+      expect(test_metadatum).to validate_uniqueness_of(:key).scoped_to(:metadatable_id).case_insensitive
```
KatyTaylor, line 19: Does this need to be scoped to metadatable_id and metadatable_type, otherwise if you have Request with id 52 and Receptacle with id 52, who both have a metadatum with key 'comment', you're in trouble?

spec/models/poly_metadatum_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    context 'when creating a new poly_metadatum' do
+      it 'is valid with valid attributes' do
+        expect(test_metadatum).to be_valid
+      end
+
+      it 'is invalid without a key' do
+        test_metadatum.key = nil
+        expect(test_metadatum).not_to be_valid
+        expect(test_metadatum.errors.full_messages).to include('Key can\'t be blank')
+      end
```
KatyTaylor, line 31: Are these tests not duplicating what the `validate_presence_of` tests do above?

spec/models/poly_metadatum_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        expect(test_metadatum).not_to be_valid
+        expect(test_metadatum.errors.full_messages).to include('Metadatable must exist')
+      end
+    end
+  end
+
+  # Tests for associations
+  describe 'associations' do
+    it 'belongs to metadatable' do
+      expect(test_metadatum).to belong_to(:metadatable).required
```
KatyTaylor, line 50: How do you refer to the parent record? Would it be `test_metadatum.metadatable` , to get to the Request record?

app/resources/api/v2/request_resource.rb
```diff
@@ -23,6 +23,7 @@ class RequestResource < BaseResource
       has_one :request_type, always_include_linkage_data: true
       has_one :primer_panel
       has_one :pre_capture_pool
+      has_many :poly_metadata, as: :metadatable, class_name: 'PolyMetadatum', dependent: :destroy
```
KatyTaylor, line 26: What does it do, putting the `dependent: destroy` here in the API resource, rather than in the model?

KatyTaylor (commented):
Left a fair few comments and questions! I think keeping it as a string for now is fine. Looking at custom_metadata, and 'facts' in Samples Extraction, neither of those specify a data type. We could add it as an optional field if needed in future and default to String if not specified?

app/models/poly_metadatum.rb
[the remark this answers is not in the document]
andrewsparkes: I agree it isn't ideal. You're the naming guru, can you think of something better?!
I also don't like 'poly-metadatum' for the polymorphic class which sounds terrible and both plural and singular at the same time.

app/models/request.rb
andrewsparkes: Yes done. Had it in the v2 resource but maybe that's wrong.
What about `, inverse_of: :metadatable`? can't work out if that's needed, or does anything

app/models/poly_metadatum.rb
andrewsparkes: Added a better description with instructions how to implement on other models.

db/migrate/20231204163029_create_poly_metadata.rb
andrewsparkes: For Comments there is a commentable concern, here:
`app/models/concerns/commentable.rb`
I think I was going to make one of those for metadatable, but then ended up only needing one line on the model, so it seemed excessive. What do you think? I don't get the scope code on commentable, and the 2 other methods are specific to comments.

app/resources/api/v2/request_resource.rb
andrewsparkes: Possibly (probably) nothing, have removed it from here and added it to model.

spec/models/poly_metadatum_spec.rb
andrewsparkes: So I think maybe that test is wrong. There is a metadatable_type and a metadatable_id that in combination define the metadatable instance. So in your example one type should be 'requests' and the other 'receptacles' so they shouldn't clash.
Abdullah made the point that this would still be a confusing table to look at if we have many matching keys.

We could enforce unique key names with a join table with a unique index on key, and encourage a model specific prefix. 
e.g. with columns: `key, description, data_type, friendly name`
e.g. example: `'request_comment', 'Holds a text comment from a user', string, 'Comment'`

spec/models/poly_metadatum_spec.rb
andrewsparkes: Yes that works in SS.
It doesn't work from the Limber side, there are some issues with JSONAPI client. 
However, for our use cases we always start with the metadatable instance (e.g. the Request) and are either reading or writing metadata on it. We never start with a metadata that we need to find the owner of. Can't think of a scenario for that.

spec/models/poly_metadatum_spec.rb
andrewsparkes: Yes you are right I will refactor a bit

spec/models/poly_metadatum_spec.rb
andrewsparkes: I'm learning a lot asking code pilot to suggest 'better syntax' for sections of code(!)

spec/models/poly_metadatum_spec.rb
andrewsparkes: Hmm. Getting this odd error
```

PolyMetadatum validations key must be unique for the metadatable
     Failure/Error: expect(test_metadatum).to validate_uniqueness_of(:key).scoped_to(%i[metadatable_type metadatable_id]).case_insensitive
     
     ActiveRecord::SubclassNotFound:
       Invalid single-table inheritance type: Request is not a subclass of Shoulda::Matchers::ActiveRecord::Uniqueness::TestModels::Requesu
     # ./spec/models/poly_metadatum_spec.rb:15:in `block (3 levels) in <top (required)>'
```

app/models/poly_metadatum.rb
andrewsparkes: No, an oversight, have set to not null, good spot

spec/models/poly_metadatum_spec.rb
andrewsparkes: Changed the scope to include both and modified test, seems to behave correctly

app/models/poly_metadatum.rb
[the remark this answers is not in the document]
KatyTaylor: No should have said... I can't think of anything better 😆 and the documentation I saw on polymorphism seems to go with that '-able' convention. 

I actually quite liked 'polymetadata' as a word although I don't think we should let the users hear it! If we end up displaying the metadata in the UI then we should create a unified interface that brings together metadata from the different tables, under a user-friendly heading.

spec/models/poly_metadatum_spec.rb
KatyTaylor: I personally think it's OK - I think if you're reading it in the database you'd probably always do a join so that you have the metadatable type as a column, or you'd be filtering to a specific record or set of records all of the same type.

I like the idea of a 'friendly name' though - that's one of the things I dislike about 'custom metadata', as it looks rubbish in the UI displaying a key that's lowercase with underscores. (although iirc there's nothing preventing you making your key 'friendly')

app/models/request.rb
KatyTaylor: RE `inverse_of` - I wondered this too. Docs here - https://guides.rubyonrails.org/association_basics.html#bi-directional-associations

Think it's unnecessary unless you use `:through` or `:foreign_key`

> Active Record supports automatic identification for most associations with standard names. However, bi-directional associations that contain the :through or :foreign_key options will not be automatically identified.

db/migrate/20231204163029_create_poly_metadata.rb
KatyTaylor: Yeah the scope code is confusing. Looks like it allows you to call e.g. `Plate.with_comments.group` (as well as just `Plate.with_comments` 
(https://stackoverflow.com/questions/39306433/why-is-this-scopes-lambda-followed-by-a-block)

I've had a play in the console and haven't got it to work yet though. Looking through SS & Limber, I can't see `with_comments` used anywhere.

KatyTaylor: In summary, I think I agree, having a concern seems excessive here! We can always add one later if we want to add shared functionality.

PR #4017, opened 2024-02-07
DPL-823 Import PBMC pool plates into Sequencescape

### LINKED ISSUE #3857: DPL-823 Import PBMC pool plates into Sequencescape
**User story**
As a member of the SSR team, who has received a plate from faculty containing thawed, pooled PBMCs, I would like to be able to import them into Sequencescape along with all the required data.

Lesley, Abby, Katy, Liz H

Abby, Liz H

**Acceptance criteria**
- [ ] I can upload the plate as a "LRC PBMC Pools Input" plate in Sequencescape
- [ ] Record pools, with the following information:
    - [ ] The samples that make up the pool
- [ ] For each sample, record:
    - [ ] ~Study~ --> currently has to be the same across the whole manifest
    - [ ] Donor id
    - [ ] Supplier sample name (Vacutainer barcode)
    - [ ] Which well it's in
    - [ ] Unique tag_depth index (assign automatically on upload, not provided by user), to avoid tag clash and enable compound sample creation when sequencing request is made later on
- [ ] Consider whether to insert rows into the stock resource table in the MLWH or not.
- [ ] Update [documentation](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/scRNA+Core+Pipeline+Documentation) - it currently has instructions from DPL-809 (linked below) - add similar instructions regarding this import.

**References**
- https://github.com/sanger/sequencescape/issues/3849 - equivalent but for fresh blood
- https://github.com/sanger/sequencescape/issues/3937 - equivalent but for cDNA

**Additional information**
This will use the normal manifest process - SSRs generate the manifest & barcode labels, send them to faculty, who use the barcodes on their plates and send the manifest back filled in. 

This is new functionality for manifests - existing manifests that import pools (e.g. multiplexed library manifests) do not include the individual sample information.

KatyTaylor:
Closes #3857 

Add initial version of new manifest type
- Plus some formatting / typo fixes

### DIFF
--- a/app/models/sample_manifest.rb
+++ b/app/models/sample_manifest.rb
@@ -6,7 +6,7 @@
 # for the potential samples. It also generates a {SampleManifestExcel}
 # spreadsheet which gets sent to the customer.
 #
-# The labware that gets generate is determined by the {#asset_type} which
+# The labware that gets generated is determined by the {#asset_type} which
 # switches out the {#core_behaviour} module {SampleManifest::CoreBehaviour}.
 # This is concerned with generating {Labware} and {Receptacle receptacles},
 # generating any event specific to the asset type, and setting manifest specific
--- a/app/controllers/sdb/sample_manifests_controller.rb
+++ b/app/controllers/sdb/sample_manifests_controller.rb
@@ -37,6 +37,7 @@ class Sdb::SampleManifestsController < Sdb::BaseController
     @study_id = @sample_manifest.study_id
     @samples = @sample_manifest.samples.paginate(page: params[:page])
   end
+
   def new # rubocop:todo Metrics/AbcSize
     params[:only_first_label] ||= false
     @sample_manifest = SampleManifest.new(new_manifest_params)
--- a/config/sample_manifest_excel/manifest_types.yml
+++ b/config/sample_manifest_excel/manifest_types.yml
@@ -255,6 +255,15 @@ plate_bioscan:
     - :sample_type
     - :donor_id
     - :bioscan_control_type
+plate_scrna_core_pools:
+  heading: "scRNA Core Pools Plate"
+  asset_type: "plate"
+  columns:
+    - :sanger_plate_id
+    - :well
+    - :sanger_sample_id
+    - :collected_by_for_scrna_core
+    - :donor_id_mandatory
 tube_default:
   heading: "Default Tube"
   asset_type: "1dtube"


app/jobs/sample_manifest/generate_wells_job.rb
```diff
@@ -1,25 +1,31 @@
...
+# Generates wells and sample manifest assets, for a plate sample manifest.
 SampleManifest::GenerateWellsJob =
-  Struct.new(:sample_manifest_id, :map_ids_to_sample_ids, :plate_id) do
+  Struct.new(:sample_manifest_id, :map_ids_to_sanger_sample_ids, :plate_id) do
     def perform
       ActiveRecord::Base.transaction do
-        # Ensure the order of the wells are maintained
-        maps = Map.find(map_ids).index_by(&:id)
-        well_data = map_ids_to_sample_ids.map { |map_id, sample_id| [maps[map_id], sample_id] }
+        map_ids_to_sanger_sample_ids.each { |k, v| create_well(k, v) }
```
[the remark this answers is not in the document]
sdjmchattie: @KatyTaylor Would it be worth identifying why CodeClimate feels the need to tell you about this?  I feel it's because there's a flag to warn of single character variable names.  Not that it tells you this in the error message, or suggest how you'd "fix" this.

app/models/sample_manifest.rb
```diff
@@ -128,6 +129,13 @@ def default_filename
     "#{study_id}stdy_manifest_#{id}_#{created_at.to_formatted_s(:dmy)}"
   end
 
+  # Use a default value of 1 for rows_per_well if not set
+  def rows_per_well
+    1
+    # TODO: replace above line with below line when we want to turn the 'rows_per_well' feature on
```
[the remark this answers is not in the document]
sdjmchattie: Needs a reference to an issue where this becomes relevant.

app/models/sample_manifest/plate_behaviour.rb
```diff
@@ -83,65 +74,60 @@ def labware=(labware)
...
+          well_data[plate.id].flat_map do |map_id, sanger_sample_ids|
+            sanger_sample_ids.map do |sanger_sample_id|
+              {
+                barcode: plate.human_barcode,
+                position: plate.maps.find(map_id).description,
+                sample_id: sanger_sample_id
+              }
+            end
+          end
+        end
```
sdjmchattie, line 129: It looks like perhaps the indenting has lost its way here.

spec/models/sample_manifest/generator_spec.rb
```diff
@@ -123,4 +124,24 @@
   it 'does not have a print job if printer name has not been provided' do
     expect(described_class.new(attributes, user, configuration)).not_to be_print_job_required
   end
+
+  context 'with rows_per_well set' do
+    let(:template) { 'pools_plate' }
+
+    # TODO: un-skip when rows_per_well feature is enabled
```
sdjmchattie, line 131: Can we reference an issue for when this is relevant?

spec/models/sample_manifest/generator_spec.rb
```diff
@@ -123,4 +124,24 @@
...
+    skip 'generates a details array with more than one entry per well' do
+      generator = described_class.new(attributes, user, configuration)
+      generator.execute
+      expect(generator.sample_manifest.details_array.size).to eq(4 * 96 * 2)
+    end
+  end
+
+  context 'with rows_per_well not set' do
+    # TODO: un-skip when rows_per_well feature is enabled
```
sdjmchattie, line 140: And a reference here.

sdjmchattie (changes requested):
I think there are a few minor issues here.  Also like was mentioned before, do we have any value in the unit tests that are being skipped via the `xit` command?  If not, we should just remove them.

app/models/sample_manifest/plate_behaviour.rb
KatyTaylor: Hmm, Prettier seems to want it like this. It's because there's a new line after the `=`:

```
@details_array =
        @plates.flat_map do |plate|
```

So the `end` lines up with `@plates`

app/jobs/sample_manifest/generate_wells_job.rb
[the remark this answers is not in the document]
KatyTaylor: Think it's this - https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Variable-Name.md

I've changed it, although the reason I did it was I thought it was fairly obvious what the key and value would be, since `map_ids_to_sanger_sample_ids` on the same line - and it makes it a lot more wordy.

app/models/sample_manifest/plate_behaviour.rb
sdjmchattie: Ah right, I see.  I just saw the `end` statements not being one tab-width apart and thought it was off.  It niggles me when this happens, but you can't fight the code styles unless we change the rules it uses, which is more of a headache usually.

KatyTaylor:
> I think there are a few minor issues here. Also like was mentioned before, do we have any value in the unit tests that are being skipped via the `xit` command? If not, we should just remove them.

Forgot to reply to this. Yes, they test the functionality being added here that is 'flagged off' by the commented out line. When I turn the feature on, I'll also turn the tests on.

app/jobs/sample_manifest/generate_wells_job.rb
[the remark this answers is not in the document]
sdjmchattie: I agree that making it more wordy feels unnecessary and might warrant us looking into whether we're getting value from this check, or if we can prevent it triggering when we don't want it to.  It's hard to get good use out of a tool that fires off too many false positives.

app/jobs/sample_manifest/generate_wells_job.rb
```diff
@@ -4,7 +4,7 @@
   Struct.new(:sample_manifest_id, :map_ids_to_sanger_sample_ids, :plate_id) do
     def perform
       ActiveRecord::Base.transaction do
-        map_ids_to_sanger_sample_ids.each { |k, v| create_well(k, v) }
+        map_ids_to_sanger_sample_ids.each { |map_id, sanger_sample_id| create_well(map_id, sanger_sample_id) }
```
sdjmchattie, line 7: Perhaps the work around for this variable renaming pain is to use that Ruby syntax where you define a method to pass the values to instead of write it out in full like you have here.  I don't know if this works when you have two values coming in, but I know you can push the values through another method when you write it something like:

```suggestion
        map_ids_to_sanger_sample_ids.each &:create_well
```

PR #4050, opened 2024-03-08
DPL-823 [Part 2 - upload] Import PBMC pool plates into Sequencescape

### LINKED ISSUE #3857: DPL-823 Import PBMC pool plates into Sequencescape

KatyTaylor:
Closes #3857 

- Enable 'rows_per_well' feature for manifest generation from https://github.com/sanger/sequencescape/pull/4017 
- Allow upload of pools plates
    - Only thing that needed sorting out was the 'tag clash' when multiple untagged samples / aliquots were saved into the same well
    - To allow this, we set a unique 'tag_depth' on each aliquot in the well
    - This will also be needed for later in the pipeline when these pools become 'compound samples' prior to sequencing and prior to being exported to the warehouse
- It nicely ignores rows that haven't been filled in, so the manifest is generated with excess rows per pool and not all have to be filled
- Added tests in sample_manifest_spec and uploader_spec


spec/models/sample_manifest/uploader_spec.rb
```diff
@@ -302,5 +304,48 @@
...
+
+          wells.each do |well|
+            aliquots = well.aliquots
+
+            # within a well, each aliquot should have its own tag_depth
+            expect(aliquots.map(&:tag_depth).uniq.count).to eq(aliquots.count)
+          end
+        end
+      end
+    end
```
andrewsparkes, line 349: Maybe add tests for where a file is uploaded and some and/or all of the rows for a particular well are not filled in, to check that it works ok and correctly filters out the blanks without losing the rows that do have info.
e.g. num_rows_per_well 3, and A1 has 3 rows, but only 2 are filled in. or A2 has 3 rows and none are filled in.
(Or maybe this is already tested elsewhere as more basic functionality of filtering out empty rows first before we get here?)

andrewsparkes (approved):
Some discussion comments but looks good

app/models/sample_manifest.rb
```diff
@@ -129,11 +129,32 @@ def default_filename
     "#{study_id}stdy_manifest_#{id}_#{created_at.to_formatted_s(:dmy)}"
   end
 
-  # Use a default value of 1 for rows_per_well if not set
+  # Number of rows per well in the manifest file, specified in manifest_types.yml.
```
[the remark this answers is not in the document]
KatyTaylor: It's a good point - I have it on my to do list to ask what the number should be. It kind of has to be high enough, because the spreadsheets are locked down and won't let them insert new rows. This is for a good reason, because some data is created upfront on generation of the manifest (Sanger Sample Ids, Sample Manifest Assets), so if they could insert rows then this data wouldn't exist for them.

spec/models/sample_manifest/uploader_spec.rb
KatyTaylor: OK I've had a look and I think it's alright. There's a test in this file `'will upload a valid partial plate sample manifest'` which only fills out some of the rows. There's also unit testing of the specific bit of code that ignores empty rows - the method called `empty?` in `row.rb` is tested in `row_spec.rb`, test called `'knows if it is empty'`. Think that's enough as I didn't have to add any additional functionality around this.

Issue #4006, opened 2024-01-23
DPL-936-2 [Sequencescape] Update Rails version to 6.1


yoldas:
**Describe the Housekeeping**
As developers we need to update Rails framework from version 6.0 to version 6.1 to enable receiving dependency and security updates as the support for version 6.0 ended on 1st January 2023.

**Blocking issues**
DPL-936-1 [Sequencescape] As developers we need to upgrade JSON API resources to the latest stable version as the current version is broken and it is preventing us upgrading to the long term supported version of Rails 6.1

**Additional context**
See [Upgrading Ruby on Rails](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html)

Do not upgrade Rails defaults to 6.1. That is for another story.

dasunpubudumal:
**References:** 

1. [FastRuby](https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-from-6-0-to-6-1.html)
2. [RailsDiff](https://railsdiff.org/6.0.3.7/6.1.3.2)

dasunpubudumal:
**Documenting some issues faced**:

`ApplicationHelper` is a partially defined module. It's declared in two places.

1. `ApplicationHelper` module in the traditional helpers path: `app/helpers/application_helper.rb`
2. `ApplicationHelper` module in `lib/informatics/lib/informatics/controller_helper.rb` (even though it's named `controller_helper`, the module is named `AppllicationHelper`  )

It seems this confuses Rails' auto-loading module [Zeitwerk](https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html). 

---

**Suggested Fix**: Renaming `ApplicationHelper` module in `lib/informatics/lib/informatics/controller_helper.rb` to `ControllerHelper` and including it in `ApplicationHelper` module in `app/helpers/application_helper.rb`.

dasunpubudumal:
# Issues Captured [Work In Progress]

### Issues with autoloading

1. Partially declared modules are not resolved by the latest Rails version.

`ApplicationHelper` is a partially defined module. It's declared in two places.

1. `ApplicationHelper` module in the traditional helpers path: `app/helpers/application_helper.rb`
2. `ApplicationHelper` module in `lib/informatics/lib/informatics/controller_helper.rb` (even though it's named `controller_helper`, the module is named `AppllicationHelper` )

[...]

PR #4059, opened 2024-03-14
DPL-936-2 sequencescape update rails version to 61

dasunpubudumal:
Closes #

- Upgrades Rails 6.0 to 6.1.


Gemfile
```diff
@@ -42,7 +42,10 @@ group :default do
   # We pull down a slightly later version as there are commits on head
   # which we depend on, but don't have an official release yet.
   # This is mainly https://github.com/resgraph/acts-as-dag/commit/be2c0179983aaed44fda0842742c7abc96d26c4e
-  gem 'acts-as-dag', github: 'resgraph/acts-as-dag', branch: '5e185dddff6563ee9ee92611555cd9d9a519d280'
+  # ----------------------------
+  # NOTE: THIS WAS CHANGED TO payrollhero's master release as it contains a fix for a bug we were
+  # experiencing in Rails 6.1
+  gem 'acts-as-dag', github: 'payrollhero/acts-as-dag'
```
dasunpubudumal, line 48: The reason for this is captured in [this](https://github.com/sanger/sequencescape/issues/4006#issuecomment-1983845226) discussion.

.github/workflows/ruby_test.yml
```diff
@@ -86,21 +86,21 @@ jobs:
...
-          key: ${{ runner.os }}-tests-${{ hashFiles('yarn.lock') }}
-          # If we don't find the specific cache we want, fallback to the last raketests
-          # cache, then finally any cache for this repo.
-          # Github looks for the newest cache beginning with the first entry, before
-          # falling back the the second if none is present.
-          restore-keys: |
-            ${{ runner.os }}-tests-
-            ${{ runner.os }}-
+          node-version: "18"
+#      - name: Cache js
```
dasunpubudumal, line 90: Noticed that _**sometimes**_, `rspec` tests fail when the js cache is enabled. We might have to investigate a little into this.

.github/workflows/ruby_test.yml
```diff
@@ -86,21 +86,21 @@ jobs:
...
-          key: ${{ runner.os }}-tests-${{ hashFiles('yarn.lock') }}
-          # If we don't find the specific cache we want, fallback to the last raketests
-          # cache, then finally any cache for this repo.
-          # Github looks for the newest cache beginning with the first entry, before
-          # falling back the the second if none is present.
-          restore-keys: |
-            ${{ runner.os }}-tests-
-            ${{ runner.os }}-
+          node-version: "18"
+      #      - name: Cache js
```
dasunpubudumal, line 90: Noticed that enabling this cache sometimes lead to failed in `rspec` tests. We might have to investigate this a bit.

.github/workflows/ruby_test.yml
```diff
@@ -86,21 +86,21 @@ jobs:
...
-            node_modules
-          key: ${{ runner.os }}-tests-${{ hashFiles('yarn.lock') }}
-          # If we don't find the specific cache we want, fallback to the last raketests
-          # cache, then finally any cache for this repo.
-          # Github looks for the newest cache beginning with the first entry, before
-          # falling back the the second if none is present.
-          restore-keys: |
-            ${{ runner.os }}-tests-
-            ${{ runner.os }}-
+          node-version: "18"
```
stevieing, line 89: You can refer to the .nvmrc file rather than setting the node version in here which saves changing it every time.

.github/workflows/ruby_test.yml
stevieing: Can we just remove it. Is it needed?

app/models/sample_manifest/uploader.rb
```diff
@@ -28,19 +28,22 @@ def initialize(file, configuration, user, override)
...
-
-      # One of our post processing checks failed, something went wrong, so we
-      # roll everything back
-      raise ActiveRecord::Rollback
-    end
-
-    extract_errors
-    upload.fail
-    false
+    result =
```
stevieing, line 31: what is the reason for the change here? Not entirely sure what is going on.

config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -0,0 +1,67 @@
+# Be sure to restart your server when you modify this file.
```
stevieing, line 1: Is it necessary to keep this? It is all commented out.

config/initializers/permissions_policy.rb
```diff
@@ -0,0 +1,11 @@
+# Define an application-wide HTTP permissions policy. For further
```
stevieing, line 1: Again. Seems no point keeping a file if it is not being used.

lib/accession/accession/submission.rb
```diff
@@ -89,7 +89,11 @@ def close!
     private
 
     def check_sample
-      sample.errors.each { |key, value| errors.add key, value } unless sample.valid?
+      if sample.errors.is_a?(ActiveModel::Errors)
```
stevieing, line 92: Again. What is the reason for the switch? Before we were checking if the sample is valid which is the standard way to do it whereas here you are checking if there are errors. Does sample valid no longer work?

lib/informatics/lib/informatics/controller_helper.rb
```diff
@@ -5,7 +5,7 @@
 require_relative 'view/tabs/list'
 require_relative 'globals'
 
-module ApplicationHelper
+module ControllerHelper
```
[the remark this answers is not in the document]
stevieing: It would be worth fixing this to explain what controller helper is doing.

spec/api/asset_audit_spec.rb
```diff
@@ -44,14 +44,15 @@
...
-            "key": ["Key can only contain letters, numbers or _"]
+            "key": ["can\'t be blank", "Key can only contain letters, numbers or _"]
           }
         }'
       end
       let(:response_code) { 422 }
 
       it 'prevents resource creation' do
         api_request :post, resources, payload
+        p response.body
```
stevieing, line 55: I think you can remove this.

stevieing (changes requested):
Good work to get this done. There are a few things which aren't quite clear.

config/initializers/new_framework_defaults_6_1.rb
dasunpubudumal: Yeah; these files were generated along with the rails upgrade command. I'll remove it so we can add it if we require it.

config/initializers/permissions_policy.rb
dasunpubudumal: yeah; I'll remove it.

app/models/sample_manifest/uploader.rb
dasunpubudumal: This is actually something I find difficult to reason. Take the previous version of the code (that was there in Rails 6.0):

```ruby
 def run!
      return false unless valid? # rubocop:todo Rails/TransactionExitStatement

      return true if process_upload_and_callbacks # rubocop:todo Rails/TransactionExitStatement

      # One of our post processing checks failed, something went wrong, so we
      # roll everything back
      raise ActiveRecord::Rollback
    end

    extract_errors
    upload.fail
    false
  end
```

It appears the line `return true if process_upload_and_callbacks` throws out an exception in Rails 6.1 (not a very helpful exception, I'm afraid):

```
 Failure/Error:
         return false unless valid? # rubocop:todo Rails/TransactionExitStatement
   
         return true if process_upload_and_callbacks # rubocop:todo Rails/TransactionExitStatement
   
         # One of our post processing checks failed, something went wrong, so we
         # roll everything back
         raise ActiveRecord::Rollback
       end
 
 ActiveSupport::DeprecationException:
   DEPRECATION WARNING: Using `return`, `break` or `throw` to exit a transaction block is
   deprecated without replacement. If the `throw` came from
   `Timeout.timeout(duration)`, pass an exception class as a second
   argument so it doesn't use `throw` to abort its block. This results
   in the transaction being committed, but in the next release of Rails
   it will rollback.
    (called from run! at /Users/dp27/Documents/Projects/sequencescape/app/models/sample_manifest/uploader.rb:31)
 # ./app/models/sample_manifest/uploader.rb:31:in `run!'
 # ./app/controllers/sample_manifest_upload_with_tag_sequences_controller.rb:13:in `create'
 # ./lib/capybara_timeout_patches.rb:10:in `synchronize'
 # ./spec/features/sample_manifests/uploader_for_manifests_with_tag_sequences_spec.rb:208:in `block (5 levels) in <top (required)>'
 # ./spec/spec_helper.rb:87:in `block (3 levels) in <top (required)>'
 # ./spec/spec_helper.rb:87:in `block (2 levels) in <top (required)>'
```

I've noticed that this issue occur only when returning true within the transaction scope (returning within a transaction is not recommended and is said to be deprecated in future Rails releases). My solution was to extract out the result from the transaction, and handle it as I've done it in the changes.

@yoldas do you have any suggestions on this?

yoldas: The following is a good quick summary and lists future support issues: https://www.loyalty.dev/posts/returning-from-transactions-in-rails

The existing code is trying to return from the method inside a transaction block. That needs changing. 

As discussed, valid? check can be put outside of the transaction block as an initial guard statement in the method. 

The transaction block can generate a value and it can be assigned to a variable to use in guard statement to skip the calls at the end.

Note that raising a Rollback exception inside the transaction block does not bubble up, it is used as a sign to rollback the current transaction by the block.

There must not be any return statement inside the transaction block; independent from return value. See how using returns inside the transaction block behaves in different Rails versions at the link above.

app/models/sample_manifest/uploader.rb
```diff
@@ -66,6 +69,10 @@ def check_upload
   end
 
   def extract_errors
-    upload.errors.each { |key, value| errors.add key, value }
+    if upload.errors.is_a?(ActiveModel::Errors)
+      upload.errors.each { |error| errors.add error.attribute, error.message }
```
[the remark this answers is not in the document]
yoldas: Can this statement use parenthesis around the arguments of `errors.add` ?

app/models/tag_group/form_object.rb
```diff
@@ -66,7 +66,9 @@ def persist! # rubocop:todo Metrics/AbcSize
     TagGroup.transaction do
       @tag_group = TagGroup.new(name: name, adapter_type_id: adapter_type_id)
       @tag_group.tags.build(parse_oligos_list.each_with_index.map { |oligo, i| { oligo: oligo.upcase, map_id: i + 1 } })
-      return if @tag_group.save # rubocop:todo Rails/TransactionExitStatement
+
+      # return if @tag_group.save
+      @tag_group.save! # TODO: Please test this line thoroughly
```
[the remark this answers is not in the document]
yoldas: I see it tries not using  return in transaction but please don't change the behaviour here. Suggestion:

1. Take the following out of the block and call it after based on the transaction block's return value.
  errors.add(:base, I18n.t('tag_groups.errors.failed_to_save_tag_group'))
  @tag_group.errors.full_messages.each { |msg| errors.add_to_base("TagGroup Error: #{msg}") }
  
 2.  Make sure @tag_group.save is the last statement in the block.
  
  3. Assign the transaction block to a variable. Use it to guard against the statements in (1).

config/initializers/new_framework_defaults_6_1.rb
yoldas: They are useful to see what defaults are available and they should be removed **gradually** by enabling and testing. I think, they should be kept for the next upgrade story (changing the defaults).

config/initializers/permissions_policy.rb
yoldas: Yes, we can remove this.

lib/accession/accession/submission.rb
yoldas: We should remove these is_a? checks. Use the new way of adding errors.

yoldas (commented):
My first batch of review comments.

.nvmrc
```diff
@@ -1 +1 @@
-v16.13.1
+18.0.0
```
stevieing, line 1: should this be v18.0.0. Is this the current LTS version?

.rubocop.yml
```diff
@@ -216,3 +216,7 @@ Rails/ThreeStateBooleanColumn:
 
 RSpec/ExpectInHook:
   Enabled: false
+
+# Remove this after the rubocop fixes
+Lint/RedundantCopDisableDirective:
```
stevieing, line 221: It might be better to put this in rubocop_todo.yml?

stevieing (changes requested):
A couple of questions ...

.github/workflows/ruby_test.yml
```diff
@@ -82,11 +82,15 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
-      # Establish a cache of js modules to improve performance
+      - name: Read .nvmrc
+        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+        id: nvm
       - name: Setup node
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v1
```
stevieing, line 89: I didn't notice this last week but you have downgraded the setup-node version? I checked on some of other repos and we are using v4.

stevieing (changes requested):
1 thing I didn't notice last week. I think using older versions of setup-node can sometimes cause issues.

.github/workflows/ruby_test.yml
dasunpubudumal: I think I've tested this with v1. I made the change to read the node version from `.nvmrc` but forgot to upgrade the actions version. I'll upgrade it to `v4` and do a CI. Thanks @stevieing for pointing this out.

app/models/tag_group/form_object.rb
```diff
@@ -63,13 +63,20 @@ def check_entered_oligos # rubocop:todo Metrics/AbcSize
...
+        )
 
-      errors.add(:base, I18n.t('tag_groups.errors.failed_to_save_tag_group'))
-      @tag_group.errors.full_messages.each { |msg| errors.add_to_base("TagGroup Error: #{msg}") }
-    end
+        # return if @tag_group.save
+        @tag_group.save
+      end
+
+    return unless success
```
yoldas, line 77: Shouldn't it be `return if success`?

yoldas: Thanks for the fix.

PR #4075, opened 2024-04-04
Dev into master 2024 04 04

andrewsparkes:
Includes:
- DPL-936-2 Rails update
- DPL-823 scRNA faculty input sample manifest
- DPL-917 config and uat action changes to support scRNA faculty input submission


KatyTaylor (approved):
There's a lot here and it's all been reviewed on going into develop, so I've just skimmed it

PR #4078, opened 2024-04-09
Y24-052 [BUG] Dropdown list dependencies

### LINKED ISSUE #4076: Y24-052 [BUG] Dropdown list dependencies
**Describe the bug**
Sequencescape has two 'edit' pages for Studies - the standard one, and one for admins. The standard one has logic that restricts the available values in one field based on the value selected in a different list. The admin one does not obey this same logic, which can lead to undesired data combinations.

**To Reproduce**
1. Go to tab 'Studies'
2. Click on link (right hand side) 'Create Study'
3. Fill out the required fields
4. For "What is the data release strategy for this study?", and "How is the data release to be timed?", play with the fields and notice how the value of the first influences the options for the second. 
    - "Open", "Managed" --> "standard", "immediate", "delayed"
    - "Not applicable" --> "never"
6. Save the study. 
7. Click on link 'Study details'
8. Click on link 'Edit'
9. For "What is the data release strategy for this study?", and "How is the data release to be timed?", play with the fields as before and notice that the behaviour is the same as when creating the study.
10. Go to tab 'Admin'
11. Click link 'Study management'
12. Select the study you just created
13. For "What is the data release strategy for this study?", and "How is the data release to be timed?", play with the fields as before.
14. Notice that the value of the first does not influence the options for the second - this is the bug.

**Expected behaviour**
The value of "What is the data release strategy for this study?" should control the possible options for the field "How is the data release to be timed?", consistently in all creation and edit screens, so as to prevent undesired combinations of these two fields.

**Screenshots**
Screenshot showing incorrect behaviour:

Screenshot 2024-04-08 at 11 30 10

**Additional Context**
This was found while researching https://github.com/sanger/sequencescape/issues/4024, and is believed to be the root cause of the issue.

KatyTaylor:
Closes #4076 

- Add server-side validation to enforce correct relationship between data release 'strategy' and 'timing' fields
    - add 'validates' methods on study metadata
    - add constant for 'immediate' timing to study, following existing pattern
    - add tests for the new validation
    - amend existing tests that were creating invalid data
- Feature flagged the above off, because there are some existing database records that would fail the new validation.
- Add some comments around the Content Security Policy - this was due to a difference between the two pages that turned out to be a red herring - but I've added some comments that came out of my learning anyway.
- Change to the JS that the legacy_javascript_tag generates - this is the fix for the dropdown list dependencies in the UI

See PDF below showing differences between the two pages, before and after the UI fix. The rendering of the fields, and the values in the dropdowns, are now dynamic. The content security policy warning and the CSS issue are unchanged.

[Y24-052 [BUG] Dropdown list dependencies - Study Edit pages.pdf](https://github.com/sanger/sequencescape/files/14996035/Y24-052.BUG.Dropdown.list.dependencies.-.Study.Edit.pages.pdf)


config/feature_flags.yml
```diff
@@ -2,3 +2,4 @@
 # dpl202_enable_angry_cat_feature: When enabled this will add an angry cat to all pages that will try to catch the mouse pointer
 # The key should be lower case and snake case, and prefixed with the issue number to allow easy identification
 dpl_395_2_enable_advanced_search_tab: Shows the Search tab in the top navigation bar
+y24_052_enable_data_release_timing_validation: Shows the Search tab in the top navigation bar
```
yoldas, line 5: The feature description is not right. When enabled it validates that if data release strategy is 'not applicable', data release timing must be 'never'.

spec/features/studies/view_study_properties_spec.rb
```diff
@@ -54,6 +54,10 @@
 
     it 'displays HuMFre approval number for Not Applicable (Contact Datasharing) data release strategy' do
       study.study_metadata.data_release_strategy = Study::DATA_RELEASE_STRATEGY_NOT_APPLICABLE
+      study.study_metadata.data_release_timing = Study::DATA_RELEASE_TIMING_NEVER
+      study.study_metadata.data_release_prevention_reason = Study::DATA_RELEASE_PREVENTION_REASONS[0]
+      study.study_metadata.data_release_prevention_approval = Study::YES
+      study.study_metadata.data_release_prevention_reason_comment = 'comment'
```
yoldas, line 60: Unless the feature is enabled for the test, the lines has no effect. I guess they are for completeness.

yoldas (approved):
Looks good. I have only a comment about correcting the feature description.

config/feature_flags.yml
KatyTaylor: Oops thanks! Copy paste error.

spec/features/studies/view_study_properties_spec.rb
KatyTaylor: Good point... I added the feature flag after making this change to the test, so it would have been erroring. I think it's probably OK to leave it like this because this is not testing my feature, it's just something that would have broken when I enabled the feature flag.

yoldas: If the feature (validation) is enabled, the lines will make the test pass; without the lines the test will fail. It documents what the fields must be together. Keeping them is OK.

Issue #4095, opened 2024-04-26
Y24-065 Create a script to backfill the labware table's new retention_instruction column


dasunpubudumal:
**User story**
As a developer I need to backfill the new attribute `retention_instruction` that was created in the `labware` table in the story https://github.com/sanger/sequencescape/issues/4093. 

Retention instructions for existing labware can be extracted using the following query:

```sql
SELECT cm.key, cm.value, cmc.asset_id
FROM <sequencescape-db>.custom_metadata cm INNER JOIN <sequencescape-db>.custom_metadatum_collections cmc ON cm.custom_metadatum_collection_id = cmc.id
WHERE cm.key = 'retention_instruction';
```

The script could be extendable to remove existing `retention_instruction` values from `custom_metadata` table for assets. See https://github.com/sanger/sequencescape/issues/4093 for more details.

This can be tested in training, where dumps from production are being restored.

**Update**: Inform the DBA team of the script as this needs to be run on production after approval. Along with the script, we can remove conditional logic in [`app/views/labware/_asset_summary.html.erb`](app/views/labware/_asset_summary.html.erb) (and in similar code that conditionally checks for `custom_metadata` in search of retention instructions such as [`app/sequencescape_excel/sequencescape_excel/specialised_field/retention_instruction.rb`](https://github.com/sanger/sequencescape/pull/4096/files/4cc00bb3b979abcedd258f561586426df65319b4#diff-98d2c64e6875b5a9dd25d598f7672f377506a43ede905c6fe2db00a1a3d0a33d)) that aren't applicable anymore after the script.

**Who are the primary contacts for this story**
Dasun, Katy

**Who is the nominated tester for UAT**

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] Successfully backfill the labware table's `retention_instruction` column for existing retention instructions in `custom_metadata`.

**Dependencies**
This story is blocked by the following dependencies:
- https://github.com/sanger/sequencescape/issues/4093

References
This story has a non-blocking relationship with:
- https://github.com/sanger/sequencescape/issues/4003

**Additional context**
This story needs to be completed soon after https://github.com/sanger/sequencescape/issues/3676 is done.

dasunpubudumal:
Draft script for backfilling: https://gist.github.com/dasunpubudumal/ce70b7d4d99c64cb8bad91e190457a8e

yoldas:
I think, the script should fill the new column with a default retention instruction value if it does not have a value in custom metadata.

[...]

Issue #4093, opened 2024-04-25
Y24-063 Create a new attribute retention_instruction in labware table and update the way how retention instructions are stored and retrieved


dasunpubudumal:
**User story**

As a developer, I need to update where retention instructions for labware are retrieved from and stored to.

- [x] Create a new attribute `retention_instruction` in `labware` table to store retention instructions for plates and tubes. This attribute can only take the following values: `"Destroy after 2 years"`, `"Return to customer after 2 years"` and `"Long term storage"`.
- [x] Update the current codebase to store retention instructions in `labware` table. Currently, they are stored in `custom_metadata` with  `custom_metadata.key` equal to `retention_instructions`.
- [x] Update the current codebase to retrieve retention instructions from `labware` table and display retention instruction ~in **Metadata** section of the Labware user interface, along with metadata coming from `custom_metadata`. We will introduce a separate page in https://github.com/sanger/sequencescape/issues/3676 to let admin users edit retention instructions.~ in the fields at the top of the Labware page (see the figure below). We will introduce a separate page in https://github.com/sanger/sequencescape/issues/3676 to let admin users edit retention instructions

<img width="746" alt="image" src="https://github.com/sanger/sequencescape/assets/25484917/f226dbe4-ce6c-49e9-88d7-207b7b11ea49">

**Who are the primary contacts for this story**
Dasun, Katy

**Who is the nominated tester for UAT**

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] All retrievals and storage of retention instructions of **new submissions** must occur from/to `labware` table, and not from/to `custom_metadata` table.

**Dependencies**
This story is blocked by the following dependencies:

References
This story has a non-blocking relationship with:
- https://github.com/sanger/sequencescape/issues/4003

**Additional context**

KatyTaylor:
> Update the current codebase to retrieve retention instructions from labware table and display retention instructions in Metadata section of the Labware user interface, along with metadata coming from custom_metadata

I wouldn't do this. It should be easy (if not happen automatically) to add it to the fields at the top of the page, if it's a field on Labware. I think it would be confusing to crowbar it into the Metadata section, since that just pulls from the custom metadata object.

dasunpubudumal:
@KatyTaylor I have updated the description :)

dasunpubudumal:
![image](https://github.com/sanger/sequencescape/assets/25484917/cb25b6b4-a213-493f-a81a-b01ae12ad43e)

PR #4096, opened 2024-04-29
Adding the migration and values for the enum

dasunpubudumal:
Closes #

- 

### DIFF
--- a/app/sample_manifest_excel/sample_manifest_excel/upload/processor/plate.rb
+++ b/app/sample_manifest_excel/sample_manifest_excel/upload/processor/plate.rb
@@ -8,6 +8,8 @@ module SampleManifestExcel
       ##
       # Processor to handle plate manifest uploads.
       class Plate < SampleManifestExcel::Upload::Processor::Base
+        include RetentionKeyHelper
+
         validate :check_for_retention_instruction_by_plate
 
         # For plate manifests the barcodes (sanger plate id column) should be the same for each well from the same
@@ -77,6 +79,8 @@ module SampleManifestExcel
         def non_matching_retention_instructions_for_plates
           return nil, nil unless upload.respond_to?(:rows)
 
+          # Initialize empty plate_retentions and retention_instructions hashes to store retention instructions
+          retention_instructions = {}
           upload
             .rows
             .each_with_object({}) do |row, plate_retentions|
@@ -92,7 +96,7 @@ module SampleManifestExcel
               next unless plate_barcode.present? && sample_id.present?
 
               # check the row retention instruction is valid
-              err_msg = check_row_retention_value(row, plate_barcode, plate_retentions)
+              err_msg = check_row_retention_value(row, plate_barcode, plate_retentions, retention_instructions)
               return row, err_msg if err_msg.present?
             end
           [nil, nil]
@@ -100,7 +104,7 @@ module SampleManifestExcel
 
         # rubocop:enable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity, Metrics/AbcSize, Metrics/MethodLength
 
-        def check_row_retention_value(row, plate_barcode, plate_retentions)
+        def check_row_retention_value(row, plate_barcode, plate_retentions, retention_instructions)
           # if present the column is mandatory
           row_retention_value = row.value('retention_instruction')
           return 'Value cannot be blank.' if row_retention_value.nil?
@@ -113,6 +117,7 @@ module SampleManifestExcel
           else
             # first time we are seeing this plate, add it to plate retentions hash
             plate_retentions[plate_barcode] = row_retention_value
+            retention_instructions[plate_barcode] = find_retention_instruction_key_for_value(row_retention_value)
           end
           nil
         end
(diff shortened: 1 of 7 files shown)


app/views/labware/_asset_summary.html.erb
```diff
@@ -100,5 +100,23 @@
...
+      <%  else  %>
+        <td>Not specified</td>
+      <%  end  %>
+    <!-- If retention instructions are present, display them. -->
+    <%  else  %>
+      <td><%= find_retention_instruction_from_key(asset.retention_instruction.to_sym) %></td>
+    <%  end %>
+    <!-- There shouldn't be a situation where retention instructions are there in both custom_metadata -->
+    <!-- and in labware table's retention instruction attribute -->
+  </tr>
```
yoldas, line 120: I suggest one time data migration instead of doing these checks all the time.

db/migrate/20240429093317_add_retention_instruction_to_labware.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+class AddRetentionInstructionToLabware < ActiveRecord::Migration[6.1]
+  def up
+    add_column :labware, :retention_instruction, :integer
```
yoldas, line 4: Is there a default value for retention_instruction? I think, there should be one and the existing records should be updated in a data migration.

yoldas (changes requested):
I have concerns about keeping the existing custom metadata around. I think it can be addressed by a one time data migration and removing redundant checks for them.

app/sequencescape_excel/sequencescape_excel/specialised_field/retention_instruction.rb
```diff
@@ -7,20 +7,19 @@ module SpecialisedField
...
-        # will have already created the retention instructions field in the labware metadata
-        if labware_metadatum_collection.present?
-          check_and_update_existing_custom_metadatum_collection
-        else
-          asset_labware.custom_metadatum_collection = create_custom_metadatum_collection
-        end
+        # NB: Retention instructions are no longer stored in custom_metadata.
+        # We need to keep this logic in place to ensure that updating existing manifests would update
+        # the retention instructions currently stored in the labware metadata
+        check_and_update_existing_custom_metadatum_collection if labware_metadatum_collection.present?
```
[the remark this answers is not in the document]
dasunpubudumal: Yeah it's captured in https://github.com/sanger/sequencescape/issues/4095.

app/helpers/retention_instruction_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+    return nil if value.include?('Translation missing:')
+    value
+  end
+
+  # Retrieve the I18n key for a given value in the retention_instructions hash
+  def find_retention_instruction_key_for_value(value)
+    key = I18n.t(:retention_instructions).key(value)
+    return key if key
+    nil
+  end
```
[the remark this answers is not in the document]
dasunpubudumal: Yeah, you're correct. I'll update this.

app/sample_manifest_excel/sample_manifest_excel/upload/processor/plate.rb
```diff
@@ -92,27 +95,28 @@ def non_matching_retention_instructions_for_plates
               next unless plate_barcode.present? && sample_id.present?
 
               # check the row retention instruction is valid
-              err_msg = check_row_retention_value(row, plate_barcode, plate_retentions)
+              err_msg = check_row_retention_value(row, plate_barcode, retention_instructions)
               return row, err_msg if err_msg.present?
```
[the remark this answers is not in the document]
dasunpubudumal: This was an existing logic. The change is a mere variable renaming. 

I guess the logic is intentional. When you submit the manifest, it displays the error - if any - of the first instruction.

app/views/labware/_asset_summary.html.erb
dasunpubudumal: Agreed. I just added this until we get the migration story https://github.com/sanger/sequencescape/issues/4093 done. We can remove these checks as part of that story?

db/migrate/20240429093317_add_retention_instruction_to_labware.rb
dasunpubudumal: We have to think about this. I don't know if there is any type of labware that does not relate to a retention instruction. If there is, I think a default value will break things, since when we're creating the column (i.e., doing the migration) it will automatically set the default value to _every_ labware.

I've chosen not to use a default value because we've got another story to build a script to fix the data in https://github.com/sanger/sequencescape/issues/4095. I have made some changes (such as in [`app/views/labware/_asset_summary.html.erb`](https://github.com/sanger/sequencescape/pull/4096/files#diff-c255eec018f41cb51557c417cd5938f01d2ac4d0a1d1e5a487ae2763ae0a8841R103-R120)) to make this story self-contained. That is, this story was made to be deployed independently without side-effects. This story can be deployed independently without breaking existing stuff.

yoldas: A couple of points:

- After deploying the data migration story, this story will need a revisit to remove the redundant checks on availability of retention instruction in metadata.

- A note about validations:
In the schema migration, the null option is not specified, hence it is nullable.
In sample manifest upload processors, if the retention instruction column is present, the value cannot be blank.
Enforcing a retention instruction value is done by adding a manifest column where necessary.

yoldas (approved):
I have added a couple of points in comments about 
- doing the data migration is to be done together or updating this story.
- enforcing retention instruction by adding a column

PR #4106, opened 2024-05-07
Y24-065 Create a script to backfill the labware table's new retention_instruction column

dasunpubudumal:
Closes #

- 

### DIFF
--- a/lib/tasks/retention_instructions.rake
+++ b/lib/tasks/retention_instructions.rake
@@ -17,8 +17,9 @@ namespace :retention_instructions do
         next unless labware.custom_metadatum_collection.present? &&
           labware.custom_metadatum_collection.metadata['retention_instruction'].present?
         labware.retention_instruction = find_retention_instruction_key_for_value(
-          labware.custom_metadatum_collection.metadata['retention_instruction']
-        )
+            labware.custom_metadatum_collection.metadata['retention_instruction']
+        ).to_sym
+        p "Setting retention instruction for labware #{labware.id} to #{labware.retention_instruction}..."
         labware.custom_metadatum_collection.custom_metadata.each do |custom_metadata_record|
           custom_metadata_record.key == 'retention_instruction' && custom_metadata_record.destroy!
         end
--- a/lib/tasks/retention_instructions.rake
+++ b/lib/tasks/retention_instructions.rake
@@ -17,13 +17,14 @@ namespace :retention_instructions do
         next unless labware.custom_metadatum_collection.present? &&
           labware.custom_metadatum_collection.metadata['retention_instruction'].present?
         labware.retention_instruction = find_retention_instruction_key_for_value(
-            labware.custom_metadatum_collection.metadata['retention_instruction']
-          )
-          labware.custom_metadatum_collection.metadata.delete('retention_instruction')
-          labware.custom_metadatum_collection.save!
-          labware.save!
+          labware.custom_metadatum_collection.metadata['retention_instruction']
+        )
+        labware.custom_metadatum_collection.custom_metadata.each do |custom_metadata_record|
+          custom_metadata_record.key == 'retention_instruction' && custom_metadata_record.destroy!
         end
+        labware.save!
       end
     end
+  end
 
 end
--- a/lib/tasks/retention_instructions.rake
+++ b/lib/tasks/retention_instructions.rake
@@ -15,6 +15,8 @@ namespace :retention_instructions do
 
     puts "Backfilling retention instructions with batch size: #{batch_size}..."
 
+    saved_count = 0
+
     ActiveRecord::Base.transaction do
       Labware.where(retention_instruction: nil).find_in_batches(batch_size: batch_size) do |labware_group|
         labware_group.each do |labware|
@@ -23,14 +25,15 @@ namespace :retention_instructions do
           labware.retention_instruction = find_retention_instruction_key_for_value(
             labware.custom_metadatum_collection.metadata['retention_instruction']
           ).to_sym
-          p "Setting retention instruction for labware #{labware.id} to #{labware.retention_instruction}..."
           labware.custom_metadatum_collection.custom_metadata.each do |custom_metadata_record|
             custom_metadata_record.key == 'retention_instruction' && custom_metadata_record.destroy!
           end
-          labware.save!
+          saved_count +=1 if labware.save!
         end
       end
     end
+
+    puts "Backfilled retention instructions for #{saved_count} labware items."
   end
 
 end
--- a/lib/tasks/retention_instructions.rake
+++ b/lib/tasks/retention_instructions.rake
@@ -13,11 +13,10 @@ namespace :retention_instructions do
     puts 'Backfilling retention instructions...'
 
     ActiveRecord::Base.transaction do
-      Labware.where(retention_instruction: nil).find_in_batches(batch_size: 1000) do |labware_group|
-        labware_group.each do |labware|
-          next unless labware.custom_metadatum_collection.present? &&
-            labware.custom_metadatum_collection.metadata['retention_instruction'].present?
-          labware.retention_instruction = find_retention_instruction_key_for_value(
+      Labware.where(retention_instruction: nil).find_each do |labware|
+        next unless labware.custom_metadatum_collection.present? &&
+          labware.custom_metadatum_collection.metadata['retention_instruction'].present?
+        labware.retention_instruction = find_retention_instruction_key_for_value(
             labware.custom_metadatum_collection.metadata['retention_instruction']
           )
           labware.custom_metadatum_collection.metadata.delete('retention_instruction')
@@ -26,6 +25,5 @@ namespace :retention_instructions do
         end
       end
     end
-  end
 
 end
--- a/lib/tasks/retention_instructions.rake
+++ b/lib/tasks/retention_instructions.rake
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+# $ bundle exec rake retention_instructions:backfill
+
+require_relative '../../app/helpers/retention_instruction_helper'
+
+namespace :retention_instructions do
+
+  include RetentionInstructionHelper
+  desc 'Backfill retention instructions'
+
+  task backfill: :environment do
+    puts 'Backfilling retention instructions...'
+
+    ActiveRecord::Base.transaction do
+      Labware.where(retention_instruction: nil).find_in_batches(batch_size: 1000) do |labware_group|
+        labware_group.each do |labware|
+          next unless labware.custom_metadatum_collection.present? &&
+            labware.custom_metadatum_collection.metadata['retention_instruction'].present?
+          labware.retention_instruction = find_retention_instruction_key_for_value(
+            labware.custom_metadatum_collection.metadata['retention_instruction']
+          )
+          labware.custom_metadatum_collection.metadata.delete('retention_instruction')
+          labware.custom_metadatum_collection.save!
+          labware.save!
+        end
+      end
+    end
+  end
+
+end
--- a/lib/tasks/retention_instructions.rake
+++ b/lib/tasks/retention_instructions.rake
@@ -9,21 +9,26 @@ namespace :retention_instructions do
   include RetentionInstructionHelper
   desc 'Backfill retention instructions'
 
-  task backfill: :environment do
-    puts 'Backfilling retention instructions...'
+  task :backfill, [:batch_size] => :environment do |_, args|
+    args.with_defaults(batch_size: 1000)
+    batch_size = args[:batch_size].to_i
+
+    puts "Backfilling retention instructions with batch size: #{batch_size}..."
 
     ActiveRecord::Base.transaction do
-      Labware.where(retention_instruction: nil).find_each do |labware|
-        next unless labware.custom_metadatum_collection.present? &&
-          labware.custom_metadatum_collection.metadata['retention_instruction'].present?
-        labware.retention_instruction = find_retention_instruction_key_for_value(
+      Labware.where(retention_instruction: nil).find_in_batches(batch_size: batch_size) do |labware_group|
+        labware_group.each do |labware|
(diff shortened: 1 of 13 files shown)


spec/lib/retention_instructions_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+                                       create(:custom_metadatum_collection,
+                                              metadata: { 'retention_instruction' => 'Destroy after 2 years' }
+                                       )
+      )
+
+      # Execute
+      run_rake_task
+
+      # Verify
+      expect(labware.reload.retention_instruction).not_to be_nil
```
BenTopping, line 24: Do you want to check the value of the instruction is correct?

spec/lib/retention_instructions_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+                                       create(:custom_metadatum_collection,
+                                              metadata: { 'retention_instruction' => 'Destroy after 2 years' }
+                                       )
+      )
+
+      # Execute
+      run_rake_task
+
+      # Verify
+      expect(labware.reload.custom_metadatum_collection.metadata['retention_instruction']).to be_nil
```
BenTopping, line 40: Can this be included as part of the test above since they are running the same thing?

spec/lib/retention_instructions_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+                                                'other_key' => 'other_value'
+                                              }
+                                       )
+      )
+
+      # Execute
+      run_rake_task
+
+      # Verify
+      expect(labware.reload.custom_metadatum_collection.metadata['other_key']).to eq('other_value')
```
BenTopping, line 78: This one and the above are very similar, is the above one needed if this one checks the key exists, which if true means it cant be nil.

lib/tasks/retention_instructions.rake
```diff
@@ -0,0 +1,44 @@
...
+
+    labware.retention_instruction = find_retention_instruction_key_for_value(
+      labware.custom_metadatum_collection.metadata['retention_instruction']
+    ).to_sym
+
+    labware.custom_metadatum_collection.custom_metadata.each do |custom_metadata_record|
+      custom_metadata_record.key == 'retention_instruction' && custom_metadata_record.destroy!
+    end
+
+    labware.save! ? saved_count + 1 : saved_count
```
BenTopping, line 41: Might need to wrap `custom_metadata_record.destroy!` and `labware.save!` in a begin/rescue as I think it will error on failure instead of returning false.

BenTopping (commented):
Haven't tested but looks sensible.

spec/lib/retention_instructions_spec.rb
dasunpubudumal: Hi Ben, so the test above checks for the `labware.retention_instruction` value, while this test checks whether the scripts removes the retention instruction that is currently stored in `custom_metadata`. So, I think the two tests assert for two different things?

dasunpubudumal: Hi Ben, so the test above checks for the `labware.retention_instruction` value, while this test checks whether the scripts removes the retention instruction that is currently stored in `custom_metadata`. So, I think the two tests assert for two different things?

spec/lib/retention_instructions_spec.rb
dasunpubudumal: So this test intends to check whether the script **does not** remove other metadata (i.e., it checks whether the script removes only the value with the key `retention_instruction` in `custom_metadata` table).

lib/tasks/retention_instructions.rake
dasunpubudumal: Good spot! Thanks Ben.

spec/lib/retention_instructions_spec.rb
BenTopping: Yeah thats seems right, even though they assert different things they are outcomes of the same action so could be tested in the same test. That might be personal preference so happy to go with whatever.

spec/lib/retention_instructions_spec.rb
BenTopping: Again similar to the above, probably different test preferences. If you are testing it does not remove other metadata is that not also testing that metadata is not nil?

lib/tasks/retention_instructions.rake
BenTopping: 👌  Might be worth testing as well but im not sure how easy it is to force validation failures.

spec/lib/retention_instructions_spec.rb
dasunpubudumal: Yeah I guess you're correct. We could move this assertion into the test above, if that's what you're suggesting? 

I guess I've been following this "one-assertion per test" philosophy, which is [quite old](https://stackoverflow.blog/2022/11/03/multiple-assertions-per-test-are-fine/#h2-459b27e85d780). I need to get out of that one :D

spec/lib/retention_instructions_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+
+      # Execute
+      expect { run_rake_task }.not_to raise_error
+
+      # Verify
+      expect(labware.retention_instruction).to be_nil
+    end
+  end
+
+  context 'when a large number or records are given',
```
BenTopping, line 122: If we are skipping this can we remove it? Adding a note in the PR to show you have tested it manually but dont want to add it to the test suite is probably fine.

spec/lib/retention_instructions_spec.rb
BenTopping: Yeah but I see your approach now and both are fine, happy to let you decide.

lib/tasks/retention_instructions.rake
```diff
@@ -0,0 +1,52 @@
...
+    labware.retention_instruction = find_retention_instruction_key_for_value(
+      labware.custom_metadatum_collection.metadata['retention_instruction']
+    ).to_sym
+
+    begin
+      labware.custom_metadatum_collection.custom_metadata.each do |custom_metadata_record|
+        custom_metadata_record.key == 'retention_instruction' && custom_metadata_record.destroy!
+      end
+      labware.save! ? saved_count + 1 : saved_count
+    rescue ActiveRecord::ActiveRecordError => e
```
yoldas, line 44: This exception handling will prevent transaction rollback because the exception will not bubble up to the transaction block. It may be unlikely but If custom_metadata_record.destroy! succeeds and save labware.save! fails, it is a data loss.

dasunpubudumal: Agreed. I updated the to fit this.

app/sequencescape_excel/sequencescape_excel/specialised_field/retention_instruction.rb
KatyTaylor, line 1: Just wondering... if it's just a normal field on labware now, does it even need to be a 'SpecialisedField'? Or could it just work like a normal field?

spec/factories/custom_metadatum_collection_with_asset_factories.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+FactoryBot.define do
+  factory :custom_metadatum_collection_with_asset,  parent: :custom_metadatum_collection do
```
KatyTaylor, line 3: Sorry... stuck record... can we change `asset` terminology to `labware` please? :)

KatyTaylor (commented):
I've added a couple of comments. What's the plan? 

- Deploy and then manually run the rake task on the server?
- Test in training env first - check fixes all the data
- Delete this code afterwards?

lib/tasks/retention_instructions.rake
```diff
@@ -0,0 +1,53 @@
...
+      labware.custom_metadatum_collection.metadata['retention_instruction'].present?
+
+    labware.retention_instruction = find_retention_instruction_key_for_value(
+      labware.custom_metadatum_collection.metadata['retention_instruction']
+    ).to_sym
+
+    begin
+      labware.custom_metadatum_collection.custom_metadata.each do |custom_metadata_record|
+        custom_metadata_record.key == 'retention_instruction' && custom_metadata_record.destroy!
+      end
```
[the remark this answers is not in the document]
dasunpubudumal: Well well well. Not only it is equivalent, but also more efficient 😄 The way I've done it fetches all records (and filters in memory) while the `where` "function" in your suggestion already filters stuff at the database level.

Still getting used to Rails, unfortunately 🚃 😆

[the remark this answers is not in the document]
dasunpubudumal: The only difference is that it should be `find_each` instead of each (typical Rubocop stuff).

lib/tasks/retention_instructions.rake
```diff
@@ -0,0 +1,53 @@
...
+      labware.save! ? saved_count + 1 : saved_count
```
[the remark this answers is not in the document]
dasunpubudumal: Agreed. Will update this.

spec/factories/custom_metadatum_collection_with_asset_factories.rb
dasunpubudumal: This is actually a redundant factory. I'll remove this entire factory.

app/sequencescape_excel/sequencescape_excel/specialised_field/retention_instruction.rb
dasunpubudumal: Interesting. This was already declared as a specialised field. I'm not sure why it was deemed 'specialized' though.

dasunpubudumal:
> I've added a couple of comments. What's the plan?
> 
> * Deploy and then manually run the rake task on the server?
> * Test in training env first - check fixes all the data
> * Delete this code afterwards?

The plan is to actually get this into UAT and to Training, run the task there manually (through SSH) and check how it's going. We'll have to log in to the database and verify whether 1) `retention_instruction` column is populated with the correct data, and whether `custom_metadata` were properly removed. 

In my opinion, we need to export some sort of a restorable database dump in prior to all of these.

lib/tasks/retention_instructions.rake
```diff
@@ -0,0 +1,52 @@
...
+      labware.custom_metadatum_collection.metadata['retention_instruction'].present?
+
+    labware.retention_instruction = find_retention_instruction_key_for_value(
+      labware.custom_metadatum_collection.metadata['retention_instruction']
+    ).to_sym
+
+    begin
+      labware.custom_metadatum_collection.custom_metadata.where(key: 'retention_instruction').find_each(&:destroy!)
+      saved_count += 1 if labware.save!
+      saved_count
```
yoldas, line 41: On failure, labware.save! will raise an exception. On success, it will return true. 

On failure, the last value generated in the block  by the augmented assignment will be nil, but the code jumps to the rescue block. On success, the last value generated in the block will be the result of the augmented assignment.

saved_count statement in the separate line looks redundant.

lib/tasks/retention_instructions.rake
```diff
@@ -0,0 +1,52 @@
...
+    rescue ActiveRecord::ActiveRecordError => e
+      puts "Error processing labware #{labware.id}: #{e.message}"
+      raise e
+    rescue StandardError => e
+      puts "Error processing labware #{labware.id}: #{e.message}"
+    end
```
yoldas, line 47: When can a StandardError happen in this block ? I am asking because we are not rolling back the transaction in this case.

yoldas (changes requested):

lib/tasks/retention_instructions.rake
dasunpubudumal: Yeah I think it should also raise. I added the standard error to make it more generalisable. I didn't actually find a case where it explicitly throw a `StandardError` though.

lib/tasks/retention_instructions.rake
dasunpubudumal: So the idea is to return the updated count after the script is done without errors. So I think it needs to be returned from this function, right? Because only incrementing would not suffice as the saved_count initialized within the transaction is copied into this function's stack. We can return it at the end of the function though.

yoldas (approved):

lib/tasks/retention_instructions.rake
```diff
@@ -0,0 +1,54 @@
...
+    args.with_defaults(batch_size: 1000, limit: nil)
+    batch_size = args[:batch_size].to_i
+    limit = args[:limit]&.to_i
+
+    puts "Backfilling retention instructions with batch size: #{batch_size}..."
+
+    ActiveRecord::Base.transaction do
+      saved_count = 0
+      labwares = Labware.where(retention_instruction: nil)
+      labwares = labwares.limit(limit) unless limit.nil?
```
dasunpubudumal, line 21: @yoldas A `limit` parameter just to test the script in UAT. This is by default `nil` which prompts to fetch all the records.

yoldas: You could add `.order(:id)` to get the labwares in a predictable order to help testing and to quickly know where you would start next time. It is a good idea and it looks good otherwise.

yoldas (approved):

lib/tasks/retention_instructions.rake
yoldas: I see you have now re-throwing the exception in that block similar to the other. I am resolving this convo.

lib/tasks/retention_instructions.rake
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  # rubocop:todo Metrics/AbcSize
+  # rubocop:todo Metrics/MethodLength
+  def process_labware(labware, saved_count)
+    retention_instruction = find_retention_instruction_key_for_value(
+      labware.custom_metadatum_collection.metadata['retention_instruction']
+    )
+
+    # Had to add this because there are some labwares with retention instructions like 'not provided'!
```
dasunpubudumal, line 39: `not provided` thing @yoldas

yoldas (approved):

PR #4108, opened 2024-05-10
Y24 011 pbmc donor pooling required sample volume by study

yoldas:
Closes #

- 

### DIFF
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -130,6 +130,9 @@ class Study < ApplicationRecord # rubocop:todo Metrics/ClassLength
   has_many :sample_manifests
   has_many :suppliers, -> { distinct }, through: :sample_manifests
 
+  # Can have many key value pairs of metadata
+  has_many :poly_metadata, as: :metadatable, dependent: :destroy
+
   # Validations
   validates :name, uniqueness: { case_sensitive: false }, presence: true, latin1: true
   validates :name, length: { maximum: 200 }
--- a/app/models/study.rb
+++ b/app/models/study.rb
@@ -556,6 +556,19 @@ class Study < ApplicationRecord # rubocop:todo Metrics/ClassLength
     broadcast
   end
 
+  # Returns the PolyMetadatum object associated with the given key.
+  #
+  # @param key [String] The key of the PolyMetadatum to find.
+  #
+  # @return [PolyMetadatum, nil] The PolyMetadatum object with the given key,
+  #   or nil if no such PolyMetadatum exists.
+  #
+  # @example
+  #   study.poly_metadatum_by_key("sample_key")
+  def poly_metadatum_by_key(key)
+    poly_metadata.find { |pm| pm.key == key.to_s }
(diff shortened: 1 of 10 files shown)


app/views/shared/poly_metadata/edit/_study.html.erb
```diff
@@ -0,0 +1,30 @@
+<%# This partial contains poly_metadata form fields for a study. It is included
```
andrewsparkes, line 1: So is this polymetadata field entry showing for ALL study edit forms? For an scRNA-only pipeline-specific field? Not sure that's a good idea, studies aren't supposed to know anything about specific pipelines.

yoldas: I can't find a way of finding out which study is aimed for scRNA Core. That is why I created the "Additional Properties" section to separate it from other options.

An alternative could be creating a "Pipeline Options" link on the sidebar and show all Pipeline related options for the study in edit/show in a separate page. Actually, I tried this initially and turned out a bit more complicated (creating new controller, views, routes, etc.) Should we do this instead?

yoldas: One of the reasons I abandoned the idea of "Pipeline Options" is that, it sounds rather limiting instead of just any generic options that can be listed under "Additional Properties".

yoldas: I have added the small text "Following fields are optional and specific to certain studies" to where additional properties appear.

app/models/study.rb
```diff
@@ -553,6 +556,19 @@ def rebroadcast
...
+  #
+  # @param key [String] The key of the PolyMetadatum to find.
+  #
+  # @return [PolyMetadatum, nil] The PolyMetadatum object with the given key,
+  #   or nil if no such PolyMetadatum exists.
+  #
+  # @example
+  #   study.poly_metadatum_by_key("sample_key")
+  def poly_metadatum_by_key(key)
+    poly_metadata.find { |pm| pm.key == key.to_s }
```
andrewsparkes, line 569: just checking, this works without having the metadatable_id (the study_id here) in the find parameters? its finding it through the has_many relationship at the top of the class?

andrewsparkes (approved):
Good comments throughout, thankyou.

Tests look comprehensive too, great.

app/models/study.rb
yoldas: Yes, it is because of the has_many relationship defined as the following. I have copied the same setup with Request metadata.
```
# Can have many key value pairs of metadata
has_many :poly_metadata, as: :metadatable, dependent: :destroy
```

The poly_metadata collection provides access to the PolyMetadatums that belong to the Study only. The poly_metadatum_by_key(key) method finds the one with the specified key. PolyMetadatum has already metadatable_id as the Study's id and metadatable_type as the Study's type from creation.

andrewsparkes (approved):

PR #4113, opened 2024-05-14
Y24-069: Rename filename to combination of template name and time

### LINKED ISSUE #4097: Y24-nnn [BUG] Submission file name too long
**Describe the bug**
During UAT of the scRNA Core pipeline, Abby reported:

> when I created the bulk submission for ~19 tubes at once, I got a really long filename so excel wouldn't let me open it until I renamed it.

This is in reference to the scRNA Core Donor Pooling submission (LRC Bank Seq & Spare tubes).

**To Reproduce**
1. Run scRNA Core Integration Suite test up to the end of the first phase, when the LRC Bank Seq & Spare tubes have been generated.
2. Save the list of tube barcodes that have been generated.
3. Make a bulk submission in Sequencescape using ~19 of these tube barcodes.
4. Download the file and try to open it in Excel
5. Notice the error "Sorry, we couldn't find ... "

**Expected behaviour**
What was expected to happen.

**Screenshots**
image

StephenHulme:
Closes #4097 

- Rename filename to first-to-last barcodes, date, and username
  - `SQPD-9002_SQPD-9003_SQPD-9004_20190101.xlsx` -> `SQPD-9002_to_SQPD-9004_20190101_user.xlsx`
- Also fixes some rubocop style issues

### DIFF
--- a/app/controllers/bulk_submission_excel/downloads_controller.rb
+++ b/app/controllers/bulk_submission_excel/downloads_controller.rb
@@ -3,6 +3,9 @@
 # Generate a bulk submission excel template
 # from basic user provided data
 class BulkSubmissionExcel::DownloadsController < ApplicationController
+
+CONTENT_TYPE = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
+
   def new
     @submission_template = SubmissionTemplate.find_by(id: params[:submission_template_id])
     @input_field_infos = @submission_template&.input_field_infos || []
@@ -50,7 +53,7 @@ class BulkSubmissionExcel::DownloadsController < ApplicationController
   def send_file_to_user(file)
     finder = build_finder
     send_file file.path,
-              content_type: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+              content_type: CONTENT_TYPE,
               filename: build_filename(finder.barcodes)
   end
 
--- a/spec/controllers/bulk_submission_excel/downloads_controller_spec.rb
+++ b/spec/controllers/bulk_submission_excel/downloads_controller_spec.rb
@@ -2,6 +2,8 @@
 require 'rails_helper'
 RSpec.describe BulkSubmissionExcel::DownloadsController, type: :controller do
 
+  subject(:downloads_controller) { described_class.new }
+
   let(:submission) { create :submission }
 
   before do
@@ -40,7 +42,7 @@ RSpec.describe BulkSubmissionExcel::DownloadsController, type: :controller do
         post :create, params: valid_attributes
         expect(
           response.header['Content-Type']
-        ).to include 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
+        ).to include downloads_controller.class::CONTENT_TYPE
       end
 
       it 'generates a new submission Excel file' do
@@ -50,9 +52,7 @@ RSpec.describe BulkSubmissionExcel::DownloadsController, type: :controller do
 
       it 'generates an Excel file with the correct headers' do
         post :create, params: valid_attributes
-        expect(response.headers['Content-Type']).to eq(
-          'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
-        )
+        expect(response.headers['Content-Type']).to eq(downloads_controller.class::CONTENT_TYPE)
         expect(response.headers['Content-Disposition']).to include(
           "#{barcodes.first}_to_#{barcodes.last}_#{Time.current.utc.strftime('%Y%m%d')}_#{session[:user].login}.xlsx"
         )


app/controllers/bulk_submission_excel/downloads_controller.rb
```diff
@@ -33,4 +35,17 @@ def create # rubocop:todo Metrics/AbcSize, Metrics/MethodLength
...
+
+  private
+
+  # Build a filename for the file to be downloaded
+  # Consists of the template name in kebab-case, followed by the current date and time in ISO8601 format
+  # e.g. "submission-template_20190101T120000.xlsx"
+  def build_filename
+    template_name = params.dig(:defaults, :template_name) || 'Submission Template'
+    # convert to kebab-case, catching existing hyphens
+    template_name = template_name.downcase.gsub(/[-\s]+/, '-')
```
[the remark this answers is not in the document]
yoldas: This changes the existing behaviour when a plate is submitted. The first (tube) barcode could be used instead. User might expect seeing a barcode in the filename and they would be most likely to be unique.

However, "Limber-Htp - scRNA Core cDNA Prep" submission is done by the user in the story. The method makes a name like "limber-htp-scrna-core-cdna-prep_20190101T120000.xlsx", which may be indeed useful.

app/controllers/bulk_submission_excel/downloads_controller.rb
```diff
@@ -33,4 +35,17 @@ def create # rubocop:todo Metrics/AbcSize, Metrics/MethodLength
...
+    "#{template_name.dasherize}_#{Time.current.strftime('%Y%m%dT%H%M%S')}.xlsx"
```
yoldas, line 48: Although the comment says it is the current data and time in ISO8601 format, this part has no TZ info. Therefore, could you please add to the comment whether it is local or UTC so the intention is clear?

app/controllers/bulk_submission_excel/downloads_controller.rb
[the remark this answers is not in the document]
KatyTaylor: Why don't we ask the users? You could post in the psd-techadmin channel (which contains the SSRs, who are the users) - present the two options and ask which they'd prefer?

I think I personally prefer Stephen's new format, but I see Abdullah's point - we would be changing the user experience for all submission templates this way.

Another option is a hybrid - keep the submission name in there, but use the first labware barcode instead of / as well as a timestamp?

app/controllers/bulk_submission_excel/downloads_controller.rb
yoldas: I mean this only as a code comment as general when writing a timestamp, but it is OK in the filename too I think.

app/controllers/bulk_submission_excel/downloads_controller.rb
```diff
@@ -9,28 +9,73 @@ def new
...
+  # Build a filename for the file to be downloaded
+  # Follows the format: first barcode_to_last barcode_date_sanger user ID
+  # e.g. "SQPP-1234_to_SQPP-5678_20240521_ec20.xlsx"
+  def build_filename(barcodes)
+    first_barcode = barcodes.first
+    last_barcode = barcodes.last
+    date = Time.current.utc.strftime('%Y%m%d')
+    username = current_user.login
+
+    "#{first_barcode}_to_#{last_barcode}_#{date}_#{username}.xlsx"
```
yoldas, line 78: User asked for adding Study IDs to the name in the Slack channel referenced in the comments. Did that request change?

spec/controllers/bulk_submission_excel/downloads_controller_spec.rb
```diff
@@ -11,23 +13,81 @@
...
+    let(:invalid_attributes) do
+      { bulk_submission_excel_download: { asset_barcodes: 'invalid barcode', submission_template_id: submission.id } }
+    end
+
+    context 'with valid params' do
+      it 'creates a new download and sends a file' do
+        post :create, params: valid_attributes
+        expect(
+          response.header['Content-Type']
+        ).to include 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
```
yoldas, line 43: Can this string be a constant on subject's class so it is not repeated as string?

app/controllers/bulk_submission_excel/downloads_controller.rb
yoldas, line 78: Can this template string be a constant on subject's class so it can be reused with parameters to build a formatted string here and in tests?

yoldas (approved):
Looks good. Suggested minor optional changes. 
I am assuming Study IDs are no longer requested to be in the file name.

yoldas (approved):
Looks good. Thanks for the update.

PR #4125, opened 2024-05-28
DPL-936-3 [Sequencescape] Update Rails config defaults to version 6.1

dasunpubudumal:
Closes #

- 


app/models/tube.rb
```diff
@@ -90,12 +90,28 @@ def self.create_with_barcode!(*args, &block)
...
     # this is to control the order of barcode addition so that it gets set as the 'primary' barcode
     foreign_barcode = attributes.delete(:foreign_barcode)
 
     tube = create!(attributes.merge(sanger_barcode: { prefix: prefix, number: barcode }), &block)
 
+    receptacle = tube.receptacle
+    receptacle.qc_results.build(key: 'molarity', value: concentration, units: 'nM')
+    receptacle.qc_results.build(key: 'volume', value: volume, units: 'ul')
+    target = receptacle.transfer_requests_as_target.build(
+      asset: parent.receptacle,
```
andrewsparkes, line 108: I don't think you can move this into the main Tube class, most tubes will not have QC information or be created from a parent.
Shall we zoom and discuss?

dasunpubudumal: Yeah I just wanted to see whether if it works outside the block. Wanted to locate where the issue actually was. It did confirm that it was the block in `create!` that causes all these havoc. 

I'll arrange a zoom. I'm planning to go through some Rails upgrade documentations to see if they have made changes to the create function and to see what they are. Is it alright if I arrange a Zoom after that?

dasunpubudumal:
This pull request is intended to be kept in an open state until #4138 and #4139 are completed.

PR #4134, opened 2024-06-07
Remove skipped tests

sdjmchattie:
These come up on every pull request we submit for Sequencescape and create noise where real issues might be appearing.  I have kept all the tests that I can, but these were added by Eduardo between 9 months and over a year ago and have not been looked at since.  If they are going to be skipped that long, they don't hold any value and should be removed.


andrewsparkes (approved):

spec/lib/record_loader/request_type_validators_loader_spec.rb
dasunpubudumal, line 1: I think this is an incomplete test. The assertions fail because the test doesn't really populate `RequestType` table. Is this an obsolete use case, or is this tested elsewhere?

dasunpubudumal (approved):

spec/lib/record_loader/request_type_validators_loader_spec.rb
sdjmchattie: I doubt it is tested elsewhere, but it wasn't tested here either with the tests skipped.  I think if it's incomplete, then that's even more reason to throw it away unless someone has time to work on making the test more complete?

dasunpubudumal: Agreed.

PR #4142, opened 2024-06-13
Y24-097 Asset renaming - Barcode model - code references

### LINKED ISSUE #4121: Y24-097 Asset renaming - Barcode model - code references
**Describe the Housekeeping**
A Barcode has a Labware, as shown in the following code in the Barcode model:
```
belongs_to :asset, optional: false, class_name: 'Labware'
```

The name of this relationship is `asset` despite the table and model being called `Labware`.
This story is to add an alias to allow use of the more appropriate relationship name, and update all code references to use the alias.

A future story will be made to rename the database column.

**Tasks**
- [x] Add aliases to the Barcode model to allow use of the relationship name `labware`. Suggested aliases as follows:
```
alias_attribute :labware_id, :asset_id
alias_attribute :labware, :asset
```
- [x] Find all code references to this relationship and alter them to use the above aliases.
- [x] A possible test to see if they have all been found is to rename the `asset_id` column on `barcodes` locally and remove the alias, and run tests to see what errors.
- [x] Write up any learnings as you go, to help with future stories on different models.

**Additional context**
This story came out of the following research story. This has more detail and results of investigations that could be useful to read:
- https://github.com/sanger/sequencescape/issues/4072

dasunpubudumal:
Closes #4121 

- Making `asset` and `labware` synonymous inside `barcode` model.
- Updating references to the `asset` relation in `barcode` model.


app/models/barcode.rb
```diff
@@ -11,9 +11,18 @@ class Barcode < ApplicationRecord
   require 'sanger_barcode_format/legacy_methods'
   extend SBCF::LegacyMethods
 
+  # This association will have to be removed in a subsequent story
   belongs_to :asset, optional: false, class_name: 'Labware'
+  # New association introduced in #4121. This is a temporary measure to allow the renaming
+  # of the asset association to labware. This was created to accommodate for_search_query scope's include() method
+  belongs_to :labware, class_name: 'Labware', optional: false
```
KatyTaylor, line 18: OK so you've actually added a new association - could this be sufficient? Is it necessary to also have `alias_association` and `alias_attribute`?

KatyTaylor: Is it possible to do something like `belongs_to :labware, foreign_key: :asset_id, optional: false` ?

app/heron/factories/plate.rb
```diff
@@ -38,7 +38,9 @@ def save # rubocop:todo Metrics/MethodLength
         ActiveRecord::Base.transaction do
           @plate = purpose.create!
 
-          Barcode.create!(asset: @plate, barcode: barcode, format: barcode_format)
+          # Overriding labware=() function causes issues in the factory
```
KatyTaylor, line 41: What am I meant to take from this comment? It reads like you had a problem that perhaps you have resolved now?

KatyTaylor (commented):
It looks good - you were able to update the code references that use the 'asset' terminology. Just a couple of questions to consider (because we're going to do this many times for different models, I thought it was worth checking if there's a neater way!).

app/heron/factories/plate.rb
dasunpubudumal: I _had_ a problem, and I solved it. Forgot to remove the comment 😄 I'll remove it.

app/models/barcode.rb
dasunpubudumal: Hi Katy, so the `alias_association` adds getters and setters to the `labware` association that wraps the `asset` association. That is, when it calls `barcode.labware` it invokes the getter and when it calls `barcode.labware = ` it invokes the setter. This is how the code references that uses the `asset` association was made interchangeable with `labware` association.

So, I think `alias_association` and `alias_attribute` are necessary to update the code references that uses `asset` to `labware`.

dasunpubudumal: And unfortunately, `belongs_to :labware, foreign_key: :asset_id, optional: false` doesn't really work because it fails the validations when building the model: https://github.com/sanger/sequencescape/actions/runs/9660235857/job/26645343790

KatyTaylor: OK, thanks for trying it and the explanations!

I'm still curious as to 
a) why those validations are failing - you'd think that `asset_id` would have to be populated in the factories due to the existing association `belongs_to :asset, optional: false, class_name: 'Labware'`
b) whether we could add `belongs_to :labware, foreign_key: :asset_id` (without the `optional: false` that is presumably causing the validations to fail) - only useful if we could then get rid of the `alias_attribute` and `alias_association` to simplify things.

Feel free to merge though if you feel like you've spent enough time on it!

KatyTaylor (approved):
Left another comment but approving anyway :)

app/models/barcode.rb
dasunpubudumal: Hi @KatyTaylor, so as far as I've found out, `belongs_to` association adds methods to get the associated object, and assign the associated object (see the image below; it's from Ruby docs. RubyMine has a fantastic feature where you can Cmd + Click on Rails methods and the documentation for that method is nicely displayed above the method 😄)

image

- So, if we get rid of `alias_association`, we will have to keep `asset` terminology for the factories ([`spec/factories/plate_barcode_factories.rb`](https://github.com/sanger/sequencescape/blob/574bf5668156baee56990ccf26027229a428f78a/spec/factories/plate_barcode_factories.rb) for instance). 
```ruby
factory :plate_barcode, class: 'Barcode' do
  format { 'sequencescape22' }
  barcode { "#{configatron.plate_barcode_prefix}-#{generate(:barcode_number)}" }
  asset { build(:labware) }.   # To rename this to labware, we need alias_association
end
```
- We _can_ get rid of the `alias_association` keeping the factories as they are - i.e., _with_ the asset terminology.
- So, if we are to completely get rid of code references that refer to assets without aliasing, it is my opinion that should be followed by a migration. This way it will nicely align with what Rails expects in `ActiveRecord` as it does not allow native support for aliasing relations.
- Aliasing relations is actually a [hot topic in Rails community](https://discuss.rubyonrails.org/t/feature-proposal-alias-association/83439), as [Shopify](https://www.shopify.com/uk) developers submitted a [pretty hefty PR](https://github.com/rails/rails/pull/49801) for Rails main branch to be included in a subsequent Rails release. Spoiler alert: the PR is still open.

app/models/barcode.rb
```diff
@@ -11,9 +11,18 @@ class Barcode < ApplicationRecord
   require 'sanger_barcode_format/legacy_methods'
   extend SBCF::LegacyMethods
 
+  # This association will have to be removed in a subsequent story
   belongs_to :asset, optional: false, class_name: 'Labware'
```
yoldas, line 15: With this declaration staying, we have a duplicate declaration for the same association. One with belongs_to: asset and another with belongs_to :labware . This doesn't look right to me. What happens when you try to save it? Do you save it twice? I think,  belongs_to :asset, optional: false, class_name: 'Labware' could be removed. I see in the comments that it is causing validation failure. What was failing?

dasunpubudumal: Hi @yoldas, it doesn't really save twice; check [`spec/models/barcode_spec.rb`](https://github.com/sanger/sequencescape/blob/95f1fc2cd612a6db675e99ac42cd837570d6977c/spec/models/barcode_spec.rb#L98-L116).

I didn't **remove** `asset` because this story's goal was primarily to add an alias. If we remove that association, validations do fail. Here are some of the failures:

```
...
* library_aliquot - Could not find the inverse association for barcodes (:asset in Barcode) (ActiveRecord::InverseOfAssociationNotFoundError)
* phi_x_aliquot - Could not find the inverse association for barcodes (:asset in Barcode) (ActiveRecord::InverseOfAssociationNotFoundError)
* barcode - undefined method `classify' for nil:NilClass (NoMethodError)
 ....
```

It speaks of an inverse association for barcodes. I suppose that a lot of validations failures occur because of `inverse_of` in [`app/models/labware.rb:22`](https://github.com/sanger/sequencescape/blob/d9b1df66771ba2124209f9411ba2493cecd458a1/app/models/labware.rb#L22). But I could not get it to rename to `labware` as I could not get the factories fixed because I couldn't locate where it was referred in the factories. I didn't go too deep into it because I think the purpose of the story is to change code references where the `barcode` model invokes `asset` association.

If you've got ideas on some updates, maybe we both can pair it?

yoldas: Aren't those because of the asset line in barcode factory? spec/factories/barcode_factories.rb

```
  factory :barcode, aliases: [:external] do
    asset factory: %i[labware]
```

yoldas: If we change the inverse_of to labware in Labware mode and change the factory as well. Doesn't that fix them?
has_many :barcodes, foreign_key: :asset_id, inverse_of: :asset, dependent: :destroy

dasunpubudumal: It _was_ renamed to labware. See [`spec/factories/barcode_factories.rb`](https://github.com/sanger/sequencescape/blob/95f1fc2cd612a6db675e99ac42cd837570d6977c/spec/factories/barcode_factories.rb#L7).

PR #4144, opened 2024-06-14
Y24-126-2 - MultiplexedTube data accessible from v2 API

BenTopping:
This is part of a large story to import multiplexed library tubes into traction. See traction-ui and traction-esrvice pull requests for further information.

-  Adds library and insert_size_to to Aliquot resource
- Adds asset to request resource
- Adds concentration and volume to sample metadata resource

These changes are required for traction ONT multiplexed library import.
The exact request can be found here: https://github.com/sanger/traction-ui/pull/1758/files#diff-f234028dfec548a2c67095326eea8f07b445772797a78ff446d28048ef6206c9R9

Using the asset name in here so checking thats ok with @KatyTaylor as im not sure on the context of the 'asset' work

### DIFF
--- a/app/resources/api/v2/request_resource.rb
+++ b/app/resources/api/v2/request_resource.rb
@@ -24,7 +24,6 @@ module Api
       has_one :primer_panel
       has_one :pre_capture_pool
       has_many :poly_metadata, as: :metadatable, class_name: 'PolyMetadatum'
-      has_one :asset
 
       # Attributes
       attribute :uuid, readonly: true
--- a/app/resources/api/v2/request_resource.rb
+++ b/app/resources/api/v2/request_resource.rb
@@ -24,6 +24,7 @@ module Api
       has_one :primer_panel
       has_one :pre_capture_pool
       has_many :poly_metadata, as: :metadatable, class_name: 'PolyMetadatum'
+      has_one :asset
 
       # Attributes
       attribute :uuid, readonly: true
--- a/app/resources/api/v2/sample_metadata_resource.rb
+++ b/app/resources/api/v2/sample_metadata_resource.rb
@@ -10,6 +10,8 @@ module Api
       attribute :cohort
       attribute :sample_description
       attribute :donor_id
+      attribute :concentration
+      attribute :volume
 
       # set add_model_hint true to allow updates from Limber, otherwise get a
       # 500 error as it looks for resource Api::V2::MetadatumResource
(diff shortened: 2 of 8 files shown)


app/resources/api/v2/sample_metadata_resource.rb
```diff
@@ -10,6 +10,8 @@ class SampleMetadataResource < BaseResource
       attribute :cohort
       attribute :sample_description
       attribute :donor_id
+      attribute :concentration
```
andrewsparkes, line 13: I'm not comfortable with concentration and volume being in sample metadata.
Samples are the abstract sample details, not about physical instances of samples in labwares. Concentration and volume are associated with physical wells and tubes in Sequencescape.

app/resources/api/v2/request_resource.rb
```diff
@@ -24,6 +24,7 @@ class RequestResource < BaseResource
       has_one :primer_panel
       has_one :pre_capture_pool
       has_many :poly_metadata, as: :metadatable, class_name: 'PolyMetadatum'
+      has_one :asset
```
andrewsparkes, line 27: what is asset here? requests are referenced from multiple labware receptacles as the request is passed down from parent to child during transfers, it's not just linked to one receptacle

app/resources/api/v2/shared_behaviour/receptacle.rb
```diff
@@ -35,6 +35,8 @@ module Api::V2::SharedBehaviour::Receptacle
     has_many :transfer_requests_as_source, readonly: true
     has_many :transfer_requests_as_target, readonly: true
 
+    has_one :labware, readonly: true
```
andrewsparkes, line 38: curious why that is required. 
all receptacles have always linked to one labware and it hasn't been needed before

app/resources/api/v2/sample_metadata_resource.rb
andrewsparkes: Discovered the columns are in manifests and get uploaded in specialised fields to both sample_metadata and as a qc result. See app/sequencescape_excel/sequencescape_excel/specialised_field/volume.rb

app/resources/api/v2/request_resource.rb
BenTopping: You are right, this can be removed I think I was using it when experimenting with ways of retrieving the data but its no longer needed.

app/resources/api/v2/shared_behaviour/receptacle.rb
BenTopping: I am using this to access labware and subsequently labware_barcode to get the barcode information. I think I could add a labware_barcode definition in receptacle, same as in labware but not sure thats needed.

sdjmchattie (approved):

andrewsparkes (approved):

PR #4147, opened 2024-06-21
Y24-061: Send Pool XP export message

### LINKED ISSUE #4085: Y24-061: Research the best way to pass Pool XP tube export messages from Limber to RabbitMQ
**Description**
An endpoint needs to be available in Sequencescape for Limber to be able to call so that a Pool XP tube will generate an appropriate message in Avro format and publish it to the RabbitMQ exchange which will be consumed by the TOL Lab Share tool.  See the EPIC for more information.  What we want to know is how we are going to create this endpoint in Sequencescape while limiting the changes we make to the system and keeping this away from existing APIs which are already harder to maintain than we'd like.

Stuart McHattie

Andrew Sparkes
Stephen Inglis
Ben Topping
Thomas Whiteley

This is part of the EPIC: https://github.com/sanger/sequencescape/issues/3754

sdjmchattie:
Closes #4085 

- Extract code for creating compound samples.
- Build an endpoint and job that will create an export message.
- Encode the message with Avro and publish it to RabbitMQ.


app/jobs/export_pool_xp_to_traction_job.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    def fetch(uri_str, limit = 10)
+      raise IOError, 'Too many HTTP redirects' if limit.zero?
+
+      response = Net::HTTP.get_response(URI.parse(uri_str))
+
+      case response
+        when Net::HTTPSuccess     then response
+        when Net::HTTPRedirection then fetch(response['location'], limit - 1)
```
dasunpubudumal, line 31: Hi Stuart, I'm just wondering why this redirection would take place? It looks like it calls the new location upto a certain limit.

app/jobs/export_pool_xp_to_traction_job.rb
```diff
@@ -0,0 +1,125 @@
...
+      }
+    end
+
+    def get_message_schema(subject, version)
+      response = fetch("#{configatron.amqp.schemas.registry_url}#{subject}/versions/#{version}")
+      resp_json = JSON.parse(response.body)
+      resp_json['schema']
+    end
+
+    def avro_encode_message(message, schema)
```
dasunpubudumal, line 71: I think these functions will be re-used; not sure they'll be re-used in sequencescape but surely will in other components (e.g. traction-service). Do you think these functions are worthy of being extracted into a library or something?

app/jobs/export_pool_xp_to_traction_job.rb
sdjmchattie: Servers don't have to serve you the content at the current URL.  They can respond with a 301 and an associated URL which indicates the URL should be updated in future requests as it's permanent.  They can respond with 302 which is when a URL may change again in future as it's temporary.  They can respond 307 which is similar to a 302, but assigned restrictions to the HTTP method being used in the request.  It's quite complicated!

In our case, RedPanda is behind nginx and nginx is performing a redirect to the RedPanda content.  The limit is to stop following a set of redirections that go on forever or in a loop.

app/jobs/export_pool_xp_to_traction_job.rb
sdjmchattie: Not yet, but I agree.  My usual rule of thumb is if you end up with three copies of the same code, it's time to refactor.  Any eariler, and you're introducing technical debt via abstraction for a theoretical future benefit.

See https://stackify.com/premature-optimization-evil/ and https://x.com/KentBeck/status/167620773337505793

app/jobs/export_pool_xp_to_traction_job.rb
```diff
@@ -0,0 +1,125 @@
...
+      case response
+        when Net::HTTPSuccess     then response
+        when Net::HTTPRedirection then fetch(response['location'], limit - 1)
+      else
+        response.error!
+      end
+    end
+
+    def get_message_data(barcode)
+      tube = Tube.with_barcode(barcode).first
```
KatyTaylor, line 38: Think this is equivalent and avoids the need for `.first` ?
```suggestion
      tube = Tube.find_by_barcode(barcode)
```

app/jobs/export_pool_xp_to_traction_job.rb
```diff
@@ -0,0 +1,125 @@
...
+        messageUuid: UUIDTools::UUID.timestamp_create.to_s,
+        messageCreateDateUtc: Time.now.utc,
+        tubeBarcode: tube.human_barcode,
+        library: {
+          volume: 100.0,
+          concentration: 0.0,
+          boxBarcode: "Unspecified"
+        },
+        request: {
+          costCode: project.project_cost_code,
```
KatyTaylor, line 53: Is it worth doing `project&.project_cost_code`... not sure if `tube.projects` is able to return nil. If so actually, would need to do `tube.projects&.first` as well 🤔 

Similar question with `tube.studies.first` and `study.uuid`

app/jobs/export_pool_xp_to_traction_job.rb
```diff
@@ -0,0 +1,125 @@
...
+
+      schema = get_message_schema(subject, version)
+      encoded_message = avro_encode_message(message_data, schema)
+      send_message(encoded_message, subject, version)
+    rescue StandardError => e
+      Rails.logger.error("Error exporting Pool XP tube to Traction: <#{e.message}>")
+      raise
+    end
+
+    def fetch(uri_str, limit = 10)
```
KatyTaylor, line 24: Might be worth giving this a slightly more specific name. `fetch` is a method on a `Hash` in Ruby, so could be a little confusing.

sdjmchattie: Perhaps `fetch_response` since the method on `HTTP` is `get_response`?

app/jobs/export_pool_xp_to_traction_job.rb
sdjmchattie: Hmm this is a tricky one.  costCode is a required field and so is the study UUID.  So if they were `nil` there's not a lot I could do to send the message and it might be better for that to bubble up to create an error log, rather than hide it in safe-chaining.

KatyTaylor: OK gotcha. You could throw an explicit exception for it rather than let the nil pointer happen? So you could say something useful like 'expected data is not present - costcode'.

app/jobs/export_pool_xp_to_traction_job.rb
```diff
@@ -0,0 +1,125 @@
...
+    def fetch_response(uri_str, limit = 10)
+      raise IOError, 'Too many HTTP redirects' if limit.zero?
+
+      response = Net::HTTP.get_response(URI.parse(uri_str))
+
+      case response
+        when Net::HTTPSuccess     then response
+        when Net::HTTPRedirection then fetch_response(response['location'], limit - 1)
+      else
+        response.error!
```
KatyTaylor, line 33: Where does this `error!` method come from, out of interest? Can't see it on https://ruby-doc.org/stdlib-2.7.0/libdoc/net/http/rdoc/Net/HTTPResponse.html - must be a more generic ruby thing?

app/jobs/export_pool_xp_to_traction_job.rb
sdjmchattie: Yes fair.  If we leave it as is, the exception will say something about not being able to get attribute project_cost_code on nil object.  If I do the safe-chaining as you suggest and it gets to the point where it's going to encode the message using Avro, you'll get a useful message back from `fastavro` about the field in the message not being allowed to be null.

app/jobs/export_pool_xp_to_traction_job.rb
sdjmchattie: I took this and tweaked it from a highly uprated Stackoverflow answer.  I agree `error!` doesn't appear in that documentation, but it's in the source code for the `HTTPResponse` class: https://github.com/ruby/ruby/blob/master/lib/net/http/response.rb#L274-L278  It's been marked with `#:nodoc:` which is a bit concerning, but it is a public method.  Perhaps this is a pattern throughout the HTTP code and they didn't want to repeat the documentation everywhere.

config/config.rb
```diff
@@ -77,6 +77,21 @@
...
   configatron.taxon_lookup_url = 'http://eutils.ncbi.nlm.nih.gov/entrez/eutils/'
 
+  configatron.amqp.isg.host = 'localhost'
+  configatron.amqp.isg.tls = false
+  configatron.amqp.isg.vhost = 'tol'
+  configatron.amqp.isg.username = 'admin'
+  configatron.amqp.isg.password = 'development'
+  configatron.amqp.isg.exchange = 'sequencescape'
+
+  configatron.amqp.schemas.registry_url = 'http://redpanda.uat.psd.sanger.ac.uk/subjects/'
```
KatyTaylor, line 87: Do we want local config to be dependent on the UAT instance? Maybe it's not dependent...

sdjmchattie: I had considered this when I made this choice and I don't know that there are any simple and easy answers.  If I make it dependent on a local instance of RedPanda, that adds dependencies to Sequencescape when we already have so many.  Although thinking about it again, you would only be able to run a local instance while on the VPN when you rely on UAT's RedPanda.  Thoughts?

app/controllers/api/v2/bioscan/export_pool_xp_to_traction_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+          Delayed::Job.enqueue ExportPoolXpToTractionJob.new(barcode)
+          render json: {}, status: :ok
+        end
+
+        private
+
+        def preflight_errors(barcode)
+          # Check that the tube exists
+          tube = Tube.with_barcode(barcode).first
```
KatyTaylor, line 25: Not important, but again could change to the following

```suggestion
          tube = Tube.find_by_barcode(barcode)
```

spec/helpers/application_helper_spec.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
-require 'rails_helper'
```
KatyTaylor, line 4: What happened with these deleted 'rails_helper' lines?

config/config.rb
KatyTaylor: If you don't have a connection to UAT's RedPanda, what will happen? If it breaks Sequencescape completely, I think that's not OK. If it just means this particular feature doesn't work, that's not too bad. 

Similarly, I don't want a local RedPanda instance to be a requirement for developing SS locally... but if it was required for just this feature, wouldn't be too bad.

Would either of these work?
1. Skip the check against the schema if one is not available? (in local only)
2. Have a dummy schema in a file in the codebase that it uses in local only

spec/helpers/application_helper_spec.rb
sdjmchattie: I read that you only need `spec_helper` during specs.  `rails_helper` is implicitly part of that `require`.  Correct me if you know otherwise.

KatyTaylor (commented):
Finished reviewing! Only thing that needs working out I think is the schema dependency bit. Everything else was small.

spec/helpers/application_helper_spec.rb
dasunpubudumal: [`rails_helper`](https://github.com/sanger/sequencescape/blob/390611cc6aa1bdc468d78469c0079c175c6a8e9c/spec/rails_helper.rb#L4) imports `spec_helper`.

config/config.rb
sdjmchattie: I've added a cache to the system, which it probably should have had to start with.  This means that you would need to connect to the UAT RedPanda the first time you trigger one of these jobs, but then you'd have a cached file of the schema to use in future calls.  Of course, if you know the schema, you can also just dump it in the appropriately named cache file.  I added some info to the README about this.

README.md
```diff
@@ -419,10 +420,22 @@ As of the time of writing, there are three outcomes to a request made, with resp
...
+### Publishing AMQP Messages
+
+Some API endpoints (such as `/api/v2/bioscan/export_pool_xp_to_traction`) trigger background jobs which are responsible for publishing data to another instance of RabbitMQ.
+In the case of the Bioscan Export Pool XP to Traction job, the message goes to the ISG managed RabbitMQ instance.
+In order to publish a message, the job must get a schema from a registry.
+Under development conditions, you may not have a registry running, hence the default config directs to the UAT instance of PSD's supported RedPanda.
+This means, the first time you publish a message with this schema, you need to be connected to the Sanger network directly or via VPN.
+After the first use, a cached file will be created in `data/avro_schema_cache` so that the registry does not need to be reachable to continue generating messages.
+
+Because this is the first and only job doing this pubishing / RedPanda caching / Avro encoding, etc, there are parts which could be exracted in future if further jobs of this type are created.
```
KatyTaylor, line 432:
```suggestion
Because this is the first and only job doing this pubishing / RedPanda caching / Avro encoding, etc, there are parts which could be extracted in future if further jobs of this type are created.
```

KatyTaylor: Nice docs, thank you.

data/avro_schema_cache/.gitignore
KatyTaylor, line 1: I didn't know you could have .gitignore files in subdirectories 🤯

KatyTaylor (approved):
Addition of caching looks good I think! 
Just pointed out a typo 😆

Issue #4068, opened 2024-03-28
Y24-040 Specify 2 lanes only for NovaSeq X Plus 1.5B flowcell


neilsycamore:
**User story**
As a High Throughput Short Read Sequencing team lead (Tris), I want the system to check the NovaSeqX submitted samples for 'Flow cell Type' and 'Permitted number of lanes/read lengths' before a sequencing batch is created SO THAT the additional monetary expense incurred due to incorrect sample selection can be avoided.

For the NovaSeqX there following flow cell options:

1. 1.5B (Two lane), - _Primary requirement for this story_
2. 10B (Eight lane), 
3. 25B (Eight lane)

**Who are the primary contacts for this story**
Tris

**Who is the nominated tester for UAT**
Tris

**Acceptance criteria**
To be considered successful the solution must allow:

1. If I select requests where the **flow cell requested** is 1.5B, I am only permitted to create a batch for exactly 2 requests i.e. 2 lanes (no more no less).
2. If I select requests where the flow cell requested is something else, behaviour is unchanged i.e. the system allows for 8 lane requests for both 10B and 25B flow cell types.  NB: There was a similar requirement for NovaSeq 6000 added to the story https://github.com/sanger/sequencescape/issues/3770. It would be worth confirming if the change was implemented or still needs to be implemented.
3. If I select requests where the requested **read lengths are different**, I am only permitted to create a batch for the same read lengths. (This is similar to that in NovaSeq 6000 pipeline)
4. If I select requests in a different sequencing pipeline inbox, the behaviour is unchanged.
5. I cannot create a batch containing requests of **different types of flow cells types**(10B, 25B, 1.5B) (NB: Similar to https://github.com/sanger/sequencescape/issues/3819. Need to confirm if can be reused)
6. I can see an updated message prompted on the message banner informing about the (Can reuse similar wording as in NovaSeq 6000 pipeline)
- permitted lanes
- limitations around mixing flow cell types based on users' selection
- limitations around mixing read lengths based on users' selection

![image](https://github.com/sanger/sequencescape/assets/115623404/6c04f92f-4ae8-436f-98cc-36d9e6b4e8e5)

**Additional context**

Comments from Tris and RT raised [RT799336](https://rt.sanger.ac.uk/Ticket/Display.html?id=799336)

1. When selecting requests from the NovaSeq X pipeline inbox the number of lanes permitted for a 1.5B flow cell should be 2 (no more no less).
[...]

PR #4150, opened 2024-06-25
Y24-121 - Refactoring validations for pipelines

### LINKED ISSUE #4135: Y24-121 - As a developer (Ben) I would like there to be fine-grained batch creation validation for sequencing pipelines
**User story**
As a developer (Ben) I would like there to be fine-grained batch creation validation for sequencing pipelines.

Currently batch validation for sequencing pipelines lives in a few places, [default records](https://github.com/sanger/sequencescape/blob/develop/config/default_records/pipelines/default_records.yml), [sequencing pipeline model](https://github.com/sanger/sequencescape/blob/develop/app/models/sequencing_pipeline.rb), [batch model](https://github.com/sanger/sequencescape/blob/develop/app/models/batch.rb). This implementation limits any new validation to be added as global validation to every batch (validation in batch.rb). However with some pipelines we want fine grained batch creation behaviour like [NovaSeqX](https://github.com/sanger/sequencescape/issues/4068)

The aim of this story is to enable custom/individual batch validation for each pipeline. There may be a few ways to do this and it may require some investigation prior to implementation. My suggested implementation:

1. Adding a custom validator class for each  required pipeline. This validator would be referenced in the pipeline config or similar. All existing pipelines could fall back on a 'standard' validator class which contains the existing validation. Then we add a validator class for each pipeline that requires custom validation.
2. We would move the current pipeline/batch.rb validation to the 'standard' validator class.
3. Ensure the batch creation javascript works with the updated methods. https://github.com/sanger/sequencescape/blob/develop/app/views/pipelines/show.html.erb contains some logic to fade out unusable requests when creating a batch and this should still work. Also consider whether this can refactored and allow us to use other logic ontop of batch size.

Ben T, Tris B

Tris B

(May not need UAT as there should be noticeable difference)

dasunpubudumal:
Closes #4135 

- Adding a `validator_class_name` column into `pipelines` table through a migration.
- Updating `batch.rb` model to dynamically import validation classes from the corresponding `pipelines` association.


lib/dynamic_validations.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+# This had to be loaded to call constantize.
+# To call constantize on a string symbol, the class must be eagerly loaded.
+require_relative '../app/validators/nova_seq6000_validator'
```
dasunpubudumal, line 5: Suggestions for this are welcome :)

dasunpubudumal:
@BenTopping I've seen an intermittent failure of tests in CI. The screenshot below shows CI actions that ran for the same commit (one for the `push` event, the other for `pull_request` event).

Any idea why?

image

lib/dynamic_validations.rb
BenTopping: You might be able to eagerly load the app/validators by including it in the either autoload_paths or eagerload_paths in config/application.rb

app/models/batch.rb
```diff
@@ -56,13 +57,8 @@ class Batch < ApplicationRecord # rubocop:todo Metrics/ClassLength
...
 
-  validate :requests_have_same_read_length,
-           :requests_have_same_flowcell_type,
-           :batch_meets_minimum_size,
-           :all_requests_are_ready?,
-           :requests_have_same_target_purpose,
-           on: :create,
-           if: :pipeline
+  # Validations for batches
+  validates_with BatchValidator, on: :create, if: :pipeline
```
BenTopping, line 61: A better name might be BatchCreationValidator if this validation only runs on create.

db/schema.rb
BenTopping, line 1: Lot of changes here - Are these changes because this is the first schema change since rails 6 upgrade and this is the new syntax?

app/validators/nova_seq6000_validator.rb
BenTopping, line 1: If this is just for testing purposes you should be able to build the class in the test file or does that mess with the eager loading?

BenTopping (commented):
Couple comments but overall on the right track.

Just to clarify - this implementation has a base BatchValidator for all pipelines and if you want to include custom validation on top of that you add a validator class and include that in the pipeline validator_class_name. Is there a way to override the default BatchValidator validations if we wanted to?

db/schema.rb
dasunpubudumal: Yeah. I forgot to rollback the other changes. I don't want to introduce the new syntax here, since it should be done in a technical debt story. I'll keep the schema change (i.e., addition of the column) and rollback the other changes in `schema.rb`.

app/validators/nova_seq6000_validator.rb
dasunpubudumal: This is just a class for testing purposes.

dasunpubudumal:
> Couple comments but overall on the right track.
> 
> Just to clarify - this implementation has a base BatchValidator for all pipelines and if you want to include custom validation on top of that you add a validator class and include that in the pipeline validator_class_name. Is there a way to override the default BatchValidator validations if we wanted to?

Interesting point. As we've done it currently, there's no overriding the batch validator. It just adds _additional_ validations on top of the existing ones. 

Do we need to override it? If so, we can think of something that facilitates that.

BenTopping:
> > Couple comments but overall on the right track.
> > Just to clarify - this implementation has a base BatchValidator for all pipelines and if you want to include custom validation on top of that you add a validator class and include that in the pipeline validator_class_name. Is there a way to override the default BatchValidator validations if we wanted to?
> 
> @BenTopping Interesting point. As we've done it currently, there's no overriding the batch validator. This implementation just adds _additional_ validations on top of the existing ones.
> 
> Do we need to override it? If so, we can think of something that facilitates that.

I am not sure, just thinking about how we would implement the [next novaseqX ](https://github.com/sanger/sequencescape/issues/4068) story with this structure.

app/validators/nova_seq6000_validator.rb
andrewsparkes: Yeah don't put anything to do with tests in the app directory. As Ben says try and make it in your spec/lib/dynamic_validations_spec.rb test file as an inline class.

andrewsparkes: e.g.
let(:my_class) do
    Class.new do
      extend CustomValidatorBase
      def validate(record)
        record.errors.add :base, 'NovaSeq6000Validator failed'
      end
    end
  end

dasunpubudumal:
> > > Couple comments but overall on the right track.
> > > Just to clarify - this implementation has a base BatchValidator for all pipelines and if you want to include custom validation on top of that you add a validator class and include that in the pipeline validator_class_name. Is there a way to override the default BatchValidator validations if we wanted to?
> > 
> > 
> > @BenTopping Interesting point. As we've done it currently, there's no overriding the batch validator. This implementation just adds _additional_ validations on top of the existing ones.
> > Do we need to override it? If so, we can think of something that facilitates that.
> 
> I am not sure, just thinking about how we would implement the [next novaseqX ](https://github.com/sanger/sequencescape/issues/4068) story with this structure.

@BenTopping So, if we want to override the batch validation (which is `BatchCreationValidator` in the current implementation) we would have to do it based on the `pipeline`, isn't it? 

One option that I'm thinking of is to add a toggle (a boolean somewhere) for the `batch` model to determine whether use `BatchCreationValidator`, or to use the validator specified in `validator_class_name`. We could add this toggle (say, `use_default_validator`) into the `pipelines` column, with the following outcomes:

1. If `use_default_validator` is true, the batch would use `BatchCreationValidator`, **and** a validator specified in `validator_class_name` (if specified).
2. If `use_default_validator` is false, it would **not** use `BatchCreationValidator`, and would use the validator listed under `validator_class_name`.

What do you think about this approach? We do not necessarily have to add `use_default_validator` in the table per-se; we could add it in a config or somewhere (which I'm not entirely sure whether if it's appropriate).

app/validators/default_validator.rb
BenTopping, line 1: Is this still needed?

lib/custom_validator_base.rb
BenTopping, line 1: Is this needed still?

spec/lib/dynamic_validations_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# rubocop:disable Lint/ConstantDefinitionInBlock
+RSpec.describe 'DynamicValidations' do
+    context 'when added, includes the dynamic validations' do
```
BenTopping, line 7: I would move this whole block to the Batch model spec as it is essentially testing batch behaviour.

BenTopping (changes requested):
Looks good.

Couple comments around testing data and documentation.

spec/lib/dynamic_validations_spec.rb
dasunpubudumal: Moving this again causes an intermittent failure. However, I think this test will become redundant when [Y24-040](https://github.com/sanger/sequencescape/issues/4068) is done because it will have the necessary classes and would not require us to stub any class. So I think we can remove this test after the next story is done.

spec/models/batch_spec.rb
```diff
@@ -76,4 +76,28 @@
       end
     end
   end
+
+  # This test is a flaky test because it leaks state between tests.
```
BenTopping, line 80: Can remove this comment now 🥳

BenTopping (approved):
Looks good to me.

PR #4154, opened 2024-06-26
Y24-135 ANOSPP - sample manifest

andrewsparkes:
Closes #1737

Adds config for ANOSPP stock plate and manifest type

### DIFF
--- a/spec/data/sample_manifest_excel/manifest_types.yml
+++ b/spec/data/sample_manifest_excel/manifest_types.yml
@@ -188,10 +188,28 @@ plate_bioscan:
     - :date_of_sample_collection
     - :sample_taxon_id
     - :sample_common_name
-    - :bioscan_sample_description
+    - :sample_description_specimen_plate_barcode
     - :sample_type
     - :donor_id
     - :bioscan_control_type
+plate_anospp:
+  heading: "ANOSPP Plate"
+  asset_type: "plate"
+  columns:
+    - :sanger_plate_id
+    - :well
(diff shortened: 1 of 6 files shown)


sdjmchattie (approved):
Bear in mind that I don't really know what all the keys do in the pipeline definitions, but it looks neatly done and I'm sure you've tested that the functionality matches the requirements.

spec/data/sample_manifest_excel/manifest_types.yml
```diff
@@ -188,10 +188,28 @@ plate_bioscan:
...
+    - :well
+    - :sanger_sample_id
+    - :supplier_name
+    - :retention_instruction
+    - :cohort
+    - :country_of_origin
+    - :date_of_sample_collection
+    - :sample_taxon_id
+    - :sample_common_name
+    - :sample_description_specimen_plate_barcode
```
yoldas, line 209: Why does ANOSPP needs specimen plate barcode ?

andrewsparkes: Requriements are similar to bioscan, they want that second label with the specimen plate barcode

yoldas: I asked because Bioscan uses it for control sample naming later and there are no control samples for ANOSPP. If it is in requirements, that is OK.

yoldas (approved):
Looks good. Field description is made generic. That is a plus too.

PR #4158, opened 2024-06-27
Y24-107: Fix exceptions in search and other areas that do object deserialisation

### LINKED ISSUE #4122: Y24-107 Sequencescape search function fails to handle exception
When searching 'All' using four digits i.e 1234 Sequencescape returns an exception

Screenshot 2024-05-22 at 10 22 31

An ActionView::Template::Error occurred in searches#index:

  Tried to load unspecified class: HashWithIndifferentAccess
  lib/acts_as_descriptable.rb:49:in `descriptor_hash'

-------------------------------
Request:
-------------------------------

  * URL        : https://sequencescape.psd.sanger.ac.uk/searches?utf8=%E2%9C%93&type=&q=1234
  * HTTP Method: GET
  * IP address : 172.20.242.66
  * Parameters : {"utf8"=>"✓", "type"=>"", "q"=>"1234", "controller"=>"searches", "action"=>"index"}
  * Timestamp  : 2024-05-22 10:22:51 +0100
  * Server : sss0-prod-jammy
    * Rails root : /var/www/sequencescape/releases/20240507130309
  * Process: 3485139

To resolve:
Either:-
a) Sequencescape should handle the exception
b) Helper text should be added to stop users from searching 'All' with four digits

Relates to [RT802971](https://rt.sanger.ac.uk/Ticket/Display.html?id=802971)

UAT
nts

sdjmchattie:
This resolves in issue where this list wasn't being set properly because it had to be wrapped in an after_initialize block.  My assumption is that we were setting this too late and the app wasn't accepting the values.  We were also bypassing the config object which was potentially going to cause further headaches down the road as it's undocumented.

Closes #4122 

- Define YAML column permitted classes as strings instead of classes

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -104,21 +104,17 @@ module Sequencescape
     # end Rails 5 #
 
     # Fix for Psych::DisallowedClass: Tried to load unspecified class
-    # this has to be in "after_initialize" because we need custom classes to be loaded already
-    config.after_initialize do
-      ActiveRecord::Base.yaml_column_permitted_classes = [
-        Symbol,
-        ActiveSupport::HashWithIndifferentAccess,
-        HashWithIndifferentAccess, # rubocop:disable Rails/TopLevelHashWithIndifferentAccess
-        RequestType::Validator::ArrayWithDefault,
-        RequestType::Validator::LibraryTypeValidator,
-        RequestType::Validator::FlowcellTypeValidator,
-        ActionController::Parameters,
-        Set,
-        Range,
-        FieldInfo,
-        Time
-      ]
-    end
+    config.active_record.yaml_column_permitted_classes = Array(config.active_record.yaml_column_permitted_classes) + %w[
+      ActiveSupport::HashWithIndifferentAccess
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -104,21 +104,17 @@ class Application < Rails::Application
...
-        RequestType::Validator::LibraryTypeValidator,
-        RequestType::Validator::FlowcellTypeValidator,
-        ActionController::Parameters,
-        Set,
-        Range,
-        FieldInfo,
-        Time
-      ]
-    end
+    config.active_record.yaml_column_permitted_classes = Array(config.active_record.yaml_column_permitted_classes) + %w[
```
sdjmchattie, line 107: This fix is a combination of:
* Don't set this value in an after_initialize any more (it doesn't work).
* Use strings instead of proper class references to bypass the unloaded classes issue.
* Switch to using `config.active_record` rather than `ActiveRecord::Base` which was an undocumented way of setting the value.

andrewsparkes (approved):

yoldas (approved):

PR #4184, opened 2024-07-02
Y24-166: Add new endpoint on V2 API for creating and querying Volume Updates

### LINKED ISSUE #4191: Y24-166-2 - Add AssetAudits and VolumeUpdates to the V2 API
**User story**
In order to move Asset Audits over to SS API v2, the endpoints it currently uses on v1 need to be added to v2.  This includes AssetAudits and VolumeUpdates.  These should be added in the most restrictive manner to begin with.  That means being able to create them (because that's what Asset Audits app does) and not be able to update them (because we don't need that yet).

Stephen Hulme and Stuart McHattie

PSD

**Acceptance criteria**

- [x] Create AssetAudit records via the API v2.
- [x] Create VolumeUpdate records via the API v2.
- [x] Do not allow either to be modified after creation.

References
- Enables the deployment of https://github.com/sanger/asset_audits/issues/621

sdjmchattie:
Closes #4191 

- Create API V2 Volume Update controller and resource
- Update routing for the new controller


stevieing (approved):

BenTopping (commented):
Changes look fine, do we need tests?

spec/requests/api/v2/volume_update_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    let(:resource_model) { create(:volume_update) }
+    let(:updated_resource_model) do
+      resource_model.update(volume_change: 10.0)
+      resource_model
+    end
+
+    let(:payload) do
+      {
+        'data' => {
+          'id' => resource_model.id,
```
sdjmchattie, line 32: I'm not super comfortable with the way `resource_model` and `updated_resource_model` return the same object.  The second one has side-effects on the first one so it's possble you refer to `resource_model` without realising that it too has the updated `volume_change` value.  I think (by luck and lazy loading of `let` constructs) that because you haven't yet referred to the updated_resource_model in this block, it actually hasn't been modified, but this is a bit of a code smell for me.  I think it would be better to have a single `resource_model` variable and have a method you call when you want to have it updated.  Something like `def update_volume_change(model) { model.update(volume_change: 10.0) }`

sdjmchattie: Yes much better!

Issue #3696, opened 2022-11-02
DPL-554 SS New tag sets for Chromium Library Plate Manifest template


SujitDey2022:
**Title:**
Action to mitigate Tag issues will reduce the number of Sequencing Runs placed on hold, reducing overall sample TAT by days and reduce resources needed for corrective actions (correctly undertaken by SSRs, Data QC and NPG)

**Description:**
There are two types of Manifests used by DNAP to upload library data into LIMS.  Currently, the version used by the Cellgen Faculty contains the tags which must be added as the actual sequence (eg i7 CACCGCACCA and i5 GACTGTCAAT)  Example can be downloaded here: https://sequencescape.psd.sanger.ac.uk/sdb/sample_manifests/20146

If Cellgen were able to use the Chromium Library Plate Manifest version which is tag plate based, then typos and incorrect reverse complementation would be mitigated.   This would result in the correct tags being uploaded to lims, removing the need for the QC team putting runs on hold, SSRs contacting customers for the corrected tag sequences, re-uploading to SS and then requesting NPG to re-run the flowcell analysis.

New tag plates can be added to the Chromium Library Plate Manifest template in SS (listed in acceptance criteria)

The data for these can sets has kindly been provided by Conor Parks in CellGen (look for the tabs that match the names in the list in acceptance criteria):
https://docs.google.com/spreadsheets/d/1KB8vs2vhAkUkAooe9_bjaOyV1CuFDLU0vuzhWslahJc/edit#gid=331796148
The data is correct as of 17/4/24

**Primary contacts for this story:**
Richard C - SSR
Conor Park - CellGen Contact

**Who is the nominated tester for UAT**
Richard C, 
Conor Parks

**Acceptance criteria**
To be considered successful the solution must allow:

- [x] Addition of the following tag groups into the system, along with the tags they are made up of. [The names were changed on 27th June 2024. There will be tag groups only for Workflow A. The names will match the Google Sheet tab names. The names should be searched on the web with exact/direct hits. They will be new tag groups; No renaming of existing tag groups. Adjustments: (i7) and (i5) indexes are shown in parentheses to make them distinguishable and columns are removed after PN numbers to avoid YAML config conflict]
  - [x] PN-1000212 Single Index Kit N, Set A
  - [x] PN-1000215/PN-3000431 Dual Index Kit TT, Set A
    - [x] PN-1000215/PN-3000431 Dual Index Kit TT, Set A (i7)
[...]

PR #4186, opened 2024-07-04
Release DPL-554, Y24-107

KatyTaylor:

[DPL-554](https://github.com/sanger/sequencescape/issues/3696) SS New tag sets for Chromium Library Plate Manifest template:
This adds a new tag set will be added to Chromium Library Plate Manifest: PN-1000212 Single Index Kit N, Set A
 
[Y24-107](https://github.com/sanger/sequencescape/issues/4122) Sequencescape search function fails to handle exception
This resolves errors while searching a short search string with All option in Sequencescape.

[Y24-061](https://github.com/sanger/sequencescape/issues/4085) Research the best way to pass Pool XP tube export messages from Limber to RabbitMQ 


.nvmrc
```diff
@@ -1 +1 @@
-18.0.0
\ No newline at end of file
+22.3.0
```
sdjmchattie, line 1: Could we put a new line at the end of this file?

KatyTaylor: Sure, what's the reason?

sdjmchattie:
This also includes https://github.com/sanger/sequencescape/issues/4085 which hasn't been fully tested yet.

.nvmrc
sdjmchattie: According to the definition of a line in the POSIX standards https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206 it must end with a new line character.  That's why GitHub marks it with a red circle at the end of the file.  Technically this is a file with no lines in it.  I don't think it breaks NVM's reading of the file, but many tools don't handle this well.

KatyTaylor:
> This also includes #4085 which hasn't been fully tested yet.

Yep, is that OK? The code won't be called yet will it? (Although technically could be if anyone had the endpoint)

sdjmchattie:
> > This also includes #4085 which hasn't been fully tested yet.
> 
> Yep, is that OK? The code won't be called yet will it? (Although technically could be if anyone had the endpoint)

I'm not too worried.  I just wanted to point it out.  It's currently in UAT without any of the config it needs and it's not causing issues.

sdjmchattie (approved):

PR #4187, opened 2024-07-04
The Great Prettier Linting

StephenHulme:
When Prettier was updated to version 3, a config was required to specify an plugins, such as https://github.com/prettier/plugin-ruby, that were used. This PR retrospectively adds this change, along with the subsequent linting. 

- Specify `plugin-ruby` in prettier config file
- Add Ruby dependencies for plugin-ruby
- Lint codebase

In some cases Prettier and Rubocop were fighting over who's formatting rules would win. I tried to defer to Rubocop by telling Prettier (STree) to ignore those lines, but in some cases the ignores failed, so I told Rubocop to ignore them instead. I think it is more important that linting is reenabled soon, than the style be completely consistent.

#### Related

See #4089 for the Rubocop equivalent of this PR.

### DIFF
--- a/app/api/core/endpoint/basic_handler/associations/belongs_to.rb
+++ b/app/api/core/endpoint/basic_handler/associations/belongs_to.rb
@@ -32,8 +32,10 @@ module Core::Endpoint::BasicHandler::Associations::BelongsTo
 
       delegate :endpoint_details, to: :@endpoint
 
-      super(node) do |children|
-        self.class.new(@endpoint_helper, children) # prettier-ignore
+      def merge(node)
+        super(node) do |children|
+          self.class.new(@endpoint_helper, children) # prettier-ignore
+        end
       end
 
       def call(object, options, stream)


app/controllers/api/v2/qc_results_controller.rb
```diff
@@ -17,14 +17,22 @@ def create
         end
       end
 
-      def qc_results_params
+      def qc_results_params # rubocop:disable Metrics/MethodLength
```
sdjmchattie, line 20: This is annoying when the tools fight each other isn't it?  Especially over a very restrictive rule.  I would argue that Rubocop is wrong here and the method is the same length as it always was.  You should be able to work around this limitation by declaring the permitted fields in a separate method:

```
def permitted_qc_results_params do
  :barcode, :uuid, :well_location, :key, :value, :units, :cv, :assay_type, :assay_version
end
```

Then using the method in the `permit` call:

```
ActionController::Parameters.new(p.to_unsafe_h).permit(permitted_qc_results_params)
```

app/models/accessionable/base.rb
```diff
@@ -141,7 +142,8 @@ def applies_to?(name)
     # NB: this regexp is defined in <https://www.ebi.ac.uk/ena/browser/api/xml/ERC000011>
     class FieldCollectionDate < FieldSerializer
       # rubocop:disable Layout/LineLength
-      REGEXP = %r{(^[12][0-9]{3}(-(0[1-9]|1[0-2])(-(0[1-9]|[12][0-9]|3[01])(T[0-9]{2}:[0-9]{2}(:[0-9]{2})?Z?([+-][0-9]{1,2})?)?)?)?(/[0-9]{4}(-[0-9]{2}(-[0-9]{2}(T[0-9]{2}:[0-9]{2}(:[0-9]{2})?Z?([+-][0-9]{1,2})?)?)?)?)?$)}
+      REGEXP =
+        %r{(^[12][0-9]{3}(-(0[1-9]|1[0-2])(-(0[1-9]|[12][0-9]|3[01])(T[0-9]{2}:[0-9]{2}(:[0-9]{2})?Z?([+-][0-9]{1,2})?)?)?)?(/[0-9]{4}(-[0-9]{2}(-[0-9]{2}(T[0-9]{2}:[0-9]{2}(:[0-9]{2})?Z?([+-][0-9]{1,2})?)?)?)?)?$)}
```
sdjmchattie, line 146: Simples

app/sample_manifest_excel/sample_manifest_excel/worksheet/test_worksheet.rb
```diff
@@ -135,18 +135,17 @@ def record_plate_samples # rubocop:todo Metrics/AbcSize
...
-          dynamic_attributes[sheet_row][:sanger_plate_id] =
-            if cgap
-              if validation_errors.include?(:sample_plate_id_duplicates)
-                'CGAP-99999'
-              elsif validation_errors.include?(:sample_plate_id_unrecognised_foreign)
-                "INVALID-#{plate_id.to_s.upcase}#{(plate_id % 10).to_s.upcase}"
-              else
-                "CGAP-#{plate_id.to_s(16).upcase}#{(plate_id % 16).to_s(16).upcase}"
-              end
+          dynamic_attributes[sheet_row][:sanger_plate_id] = if cgap
```
sdjmchattie, line 138: Acceptable in Ruby, but I am not at all a fan of the assignment via value returned from an `if`/`elsif`/`else` block.  Please don't try to change this, I'm just saying it's something to avoid if at all possible when writing new code.

app/sequencescape_excel/sequencescape_excel/column.rb
```diff
@@ -23,7 +23,8 @@ class Column
                      defaults: {
                        number: 0,
                        type: :string,
-                       conditional_formattings: {},
+                       conditional_formattings: {
+                       },
```
sdjmchattie, line 27: This feels like a weird change for prettier to want.  Does it still want to make this change if you separate the curly braces with a space or two?  Sometimes it's OK so long as you use `{  }` rather than two curly braces directly next to each other.  Alternatively, you can replace `{}` with `Hash.new`

app/validators/order_compatibility_validator.rb
```diff
@@ -15,9 +15,9 @@ def validate(record) # rubocop:todo Metrics/CyclomaticComplexity
     return if order_request_types.all?(&:not_for_multiplexing?)
 
     record.errors.add(:orders, 'are incompatible') if order_request_types.any?(&:not_for_multiplexing?)
-    unless order_request_types.all? do |request_types|
+    unless order_request_types.all? { |request_types| # stree-ignore # rubocop:disable Style/BlockDelimiters
```
sdjmchattie, line 18: OK now it's just messing with you.  Changing this from `do end` to `{ }` is not right.

lib/tasks/retention_instructions.rake
```diff
@@ -54,5 +56,4 @@ namespace :retention_instructions do
   end
   # rubocop:enable Metrics/MethodLength
   # rubocop:enable Metrics/AbcSize
-
-end
\ No newline at end of file
+end
```
sdjmchattie, line 59: Thank you 🙏🏻

sdjmchattie (changes requested):
Generally a very nice clean up.  Some dubious changes being forced on us by prettier as well though!  They're quite small suggestions though I hope.

sdjmchattie:
Note that GitHub is hiding some of my comments.

app/api/model_extensions/plate.rb
```diff
@@ -35,10 +35,10 @@ def pools
...
       .for_pooling_of(self)
       .each_with_object({}) do |request, pools|
-        pools[request.pool_id] =
-          { wells: request.pool_into.split(','), pool_complete: request.pool_complete == 1 }.tap do |pool_information|
-            request.update_pool_information(pool_information)
-          end unless request.pool_id.nil?
+        pools[request.pool_id] = {
+          wells: request.pool_into.split(','),
+          pool_complete: request.pool_complete == 1
+        }.tap { |pool_information| request.update_pool_information(pool_information) } unless request.pool_id.nil?
```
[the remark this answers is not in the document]
yoldas: code climate is totally annoying in this case. We are not doing type checking here, are we? https://github.com/troessner/reek/blob/master/docs/Nil-Check.md

Maybe `unless request.pool_id.blank?` or `if request.pool_id.present?` is better?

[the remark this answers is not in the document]
sdjmchattie: I wonder if it thinks you should use `if request.pool_id` in place of `unless request.pool_id.nil?`

I'm sure there's logic there somewhere, but the comment from codeclimate is useless unless you understand why the rule was applied.  I'd be 100% happy with it saying what it said if it also told you the fix or linked you to a page that explains the fix.

One of my fears with a PhD viva was to be told "We don't like this bit" without how I can help them to like it.  It was nothing like that.  You get told "We would like to see you clarify this by doing x, y and z".  Actionable.

PR #4190, opened 2024-07-08
Y24-139 - ANOSPP WGS Manifest without H12

BenTopping:

- Adds ability to specify invalid wells in sample_manifests

### DIFF
--- a/app/sample_manifest_excel/sample_manifest_excel/worksheet/data_worksheet.rb
+++ b/app/sample_manifest_excel/sample_manifest_excel/worksheet/data_worksheet.rb
@@ -83,9 +83,15 @@ module SampleManifestExcel
       def create_row(detail)
         axlsx_worksheet.add_row do |row|
           columns.each do |column|
-            style = find_or_create_style(column.style)&.reference
+            # If the row is invalid in the manifest, the cell should be locked and greyed out
+            if sample_manifest.invalid_wells.include?(detail.fetch(:position, nil))
+              row.add_cell column.attribute_value(detail), type: column.type, 
+                           style: styles[[:locked, :string]].reference
+            else
+              style = find_or_create_style(column.style)&.reference
 
-            row.add_cell column.attribute_value(detail), type: column.type, style: style
+              row.add_cell column.attribute_value(detail), type: column.type, style: style
+            end
           end
         end
       end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2024_04_29_093317) do
+ActiveRecord::Schema.define(version: 2024_07_08_091249) do
 
   create_table "aliquot_indices", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC", force: :cascade do |t|
     t.integer "aliquot_id", null: false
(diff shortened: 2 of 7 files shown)


db/schema.rb
```diff
@@ -1329,6 +1329,7 @@
     t.integer "purpose_id"
     t.integer "tube_rack_purpose_id"
     t.integer "rows_per_well"
+    t.text "invalid_wells", size: :medium
```
yoldas, line 1332: The table is not a meta table to configure specific types of manifests. It is holding instances of manifest. It contains 9,575,207 records in production at the time of this writing. I don't think we need to record invalid wells for each manifest uploaded. The type of manifest should enforce it instead. It needs a specialised field and a validation on upload.

yoldas (changes requested):

db/schema.rb
yoldas: This can give some ideas `app/sequencescape_excel/sequencescape_excel/specialised_field/bioscan_control_type.rb`

KatyTaylor: It does feel like a shame / waste to have to add columns to the db for cases like these. I did the same thing with the 'rows_per_well' column, you can see above Ben's one. The problem is that the manifest upload code doesn't actually know what 'type' of manifest it's processing, so it can't pull these things from the manifest_type.yml configuration. It just has access to the saved manifest record. I thought this was a bit odd - it feels intuitive that it should be checking the config - but then I thought, maybe it's intentional. Maybe the reason was to keep it isolated from the config code in case you're re-uploading an old manifest and the manifest_types.yml config has changed since the manifest was generated? We could ask @stevieing in case he remembers the history of the design.

A specialised field feels wrong to me in this case - what would the column be called, and what would the users put in it?

Could have a face to face chat about this when Ben's back?

BenTopping: Yeah happy to have a chat. As Katy mentioned we need some data persisted to verify the manifest isn't using the invalid wells (H12) on upload and we can't get that through the manifest type because we don't know what type it is.
I'll take a look at what the Bioscan one does as well 👍  Didnt realise we already did a similar thing.

stevieing: > It does feel like a shame / waste to have to add columns to the db for cases like these. I did the same thing with the 'rows_per_well' column, you can see above Ben's one. The problem is that the manifest upload code doesn't actually know what 'type' of manifest it's processing, so it can't pull these things from the manifest_type.yml configuration. It just has access to the saved manifest record. I thought this was a bit odd - it feels intuitive that it should be checking the config - but then I thought, maybe it's intentional. Maybe the reason was to keep it isolated from the config code in case you're re-uploading an old manifest and the manifest_types.yml config has changed since the manifest was generated? We could ask @stevieing in case he remembers the history of the design.
> 
> A specialised field feels wrong to me in this case - what would the column be called, and what would the users put in it?
> 
> Could have a face to face chat about this when Ben's back?

We created the configuration for creating the manifests and the processor for upload. One of the requirements at the time was that uploading needed to not be related to the manifest type. That was probably decided by James or Beth.

app/sample_manifest_excel/sample_manifest_excel/worksheet/data_worksheet.rb
```diff
@@ -83,9 +83,16 @@ def add_columns
...
+            # If the row is invalid in the manifest, the cell should be locked
+            if sample_manifest.invalid_wells&.include?(detail.fetch(:position, nil))
+              row.add_cell column.attribute_value(detail),
+                           type: column.type,
+                           style: find_or_create_style([:disabled])&.reference
+            else
+              style = find_or_create_style(column.style)&.reference
+
+              row.add_cell column.attribute_value(detail), type: column.type, style: style
+            end
```
KatyTaylor, line 95: Just a small refactoring to avoid a bit of duplication (haven't tested):

```
columns.each do |column|
  # If the row is invalid in the manifest, the cell should be locked
  if sample_manifest.invalid_wells&.include?(detail.fetch(:position, nil))
    style_name = [:disabled]
  else
    style_name = column.style
  end

  style = find_or_create_style(column.style)&.reference
  
  row.add_cell column.attribute_value(detail), type: column.type, style: style
end
```

KatyTaylor (approved):
Approving but just left a small style suggestion.

app/sample_manifest_excel/sample_manifest_excel/worksheet/data_worksheet.rb
BenTopping: Yeah thats cleaner - I have pushed a change along the same lines.

yoldas (approved):
Looks good.

PR #4209, opened 2024-07-17
Y24-040 - Custom batch creation validation for NovaseqX

### LINKED ISSUE #4068: Y24-040 Specify 2 lanes only for NovaSeq X Plus 1.5B flowcell

BenTopping:
Closes #4068

-  Adds custom batch creation validator (NovaseqxPeValidator) with methods for handling batch size and flowcell_type shared validation

#### Notes

- If you are testing this you will need to regenerate the pipeline config or manually edit the NovaseqX PE pipeline in the rails console to match the updated record. There is no automatic script to do this as I plan to do it manually for each environment.

```
bundle exec rails c
p = Pipeline.find_by(name: "NovaSeqX PE")
p.min_size = 2
p.validator_class_name = 'NovaseqxPeValidator'
p.save
```

### DIFF
--- a/app/validators/novaseqx_pe_validator.rb
+++ b/app/validators/novaseqx_pe_validator.rb
@@ -6,6 +6,11 @@ class NovaseqxPeValidator < ActiveModel::Validator
     batch_size_for_flowcell_type(record)
   end
 
+  # Used in _pipeline_limit.html to display custom validation warnings
+  def self.validation_info
+    'Requests must be selected in pairs for 1.5B flowcells and in groups of 8 for 10B and 25B flowcells.'
+  end
+
   private
 
   def batch_size_for_flowcell_type(record)
--- a/app/validators/novaseqx_pe_validator.rb
+++ b/app/validators/novaseqx_pe_validator.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+# A custom batch creation validator for NovaSeqX PE pipeline
+class NovaseqxPeValidator < ActiveModel::Validator
+  def validate(record)
+    batch_size_for_flowcell_type(record)
+  end
+
+  private
+
+  def batch_size_for_flowcell_type(record)
+    if record.requests.empty? || record.requests.first.request_metadata.nil?
+      # No requests selected or the pipeline doesn't contain metadata to check
+      return true
+    end
+
+    flowcell_type_list = record.requests.filter_map { |request| request.request_metadata.requested_flowcell_type }
+
+    return true if flowcell_type_list.empty?
+
+    # There is a flowcell type uniqueness test in batch creation validator so we don't worry about error messages here
+    return false if flowcell_type_list.uniq.size > 1
+
+    flowcells_match_batch_size(record, flowcell_type_list.first)
+  end
+
+  def flowcells_match_batch_size(record, flowcell_type)
+    case flowcell_type
+    when '1.5B'
+      if record.requests.size != 2
+        record.errors.add(:base, 'You must select exactly 2 requests for 1.5B flowcells')
+        false
+      end
+    when '10B', '25B'
+      if record.requests.size != 8
+        record.errors.add(:base, "You must select exactly 8 requests for #{flowcell_type} flowcells")
+        false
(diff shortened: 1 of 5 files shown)


KatyTaylor:
> I plan to do it manually for each environment

When you say manually do you mean via the rails console? Is it worth at least pasting the code snippet here, for a consistent deployment process, and audit purposes? Thanks

app/validators/novaseqx_pe_validator.rb
```diff
@@ -0,0 +1,46 @@
...
+
+# A custom batch creation validator for NovaSeqX PE pipeline
+class NovaseqxPeValidator < ActiveModel::Validator
+  def validate(record)
+    batch_size_for_flowcell_type(record)
+  end
+
+  # Used in _pipeline_limit.html to display custom validation warnings
+  def self.validation_info
+    'Requests must be selected in pairs for 1.5B flowcells and in groups of 8 for 10B and 25B flowcells.'
```
KatyTaylor, line 11: Nice

spec/models/batch_spec.rb
```diff
@@ -77,28 +77,37 @@
...
+        'NovaseqxPeValidator',
+        Class.new(ActiveModel::Validator) do
+          def validate(_record)
+            true
+          end
+        end
+      )
+
+      expect(batch.valid?).to be true
+      expect(batch.errors[:base]).to be_empty
```
KatyTaylor, line 110: Doesn't feel much point in updating the class name to `NovaseqxPeValidator` since you're still stubbing it out, rather than using the existing class. Perhaps calling it `TestPipelineValidator` would be better to make it explicit that it's not using a real one.

I can see you've added a new test though ('pass' scenario) which is good, thanks.

KatyTaylor (approved):

PR #4213, opened 2024-07-17
BUG - fix to walking and direction select dropdowns in UAT action for tag layout templates

andrewsparkes:
Bugfix

### DIFF
--- a/app/uat_actions/uat_actions/generate_tag_layout_template.rb
+++ b/app/uat_actions/uat_actions/generate_tag_layout_template.rb
@@ -30,7 +30,7 @@ class UatActions::GenerateTagLayoutTemplate < UatActions
              :select,
              label: 'Direction',
              help: 'Direction the tags are laid out by',
-             select_options: -> { TagLayoutTemplatesController::DIRECTIONS },
+             select_options: -> { TagLayout::DIRECTIONS },
              options: {
                include_blank: 'Select a direction...'
              }
@@ -38,7 +38,7 @@ class UatActions::GenerateTagLayoutTemplate < UatActions
              :select,
              label: 'Walking By',
              help: 'Walking by algorithms, will default to TagLayout::WalkWellsOfPlate if left blank',
-             select_options: -> { TagLayoutTemplatesController::WALKING_ALGORITHMS },
+             select_options: -> { TagLayout::WALKING_ALGORITHMS },
              options: {
                include_blank: 'Select a walking by...'
              }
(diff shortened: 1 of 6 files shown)


app/uat_actions/uat_actions/generate_tag_layout_template.rb
```diff
@@ -30,15 +30,15 @@ class UatActions::GenerateTagLayoutTemplate < UatActions
              :select,
              label: 'Direction',
              help: 'Direction the tags are laid out by',
-             select_options: -> { TagLayoutTemplatesController::DIRECTIONS },
+             select_options: -> { TagLayout::DIRECTIONS },
```
yoldas, line 33: `app/controllers/tag_layout_templates_controller.rb` has a similar constant but keys are different and it misses `TagLayout::CombByRows` . The controller makes it available to the view `app/views/tag_layout_templates/new.html.erb` as `@direction_algorithms` . To avoid duplication and for consistency, I think we should also
1. transfer the comments and the examples in the keys from the controller to the TagLayout for DIRECTIONS 
2. remove the constant from the controller
3. assign @direction_algorithms to TagLayout::DIRECTIONS in the controller

andrewsparkes: Ah this needs more thought because I remember now that they were different lists on purpose. We didn't want to expose all of the options through the UAT action. Although I cannot remember why now. Will take another look, thanks for the comment.

andrewsparkes: So we need a kind of 'display' version of the keys, which was the reason for the list in the controller. Agree it's not ideal to duplicate it though.

andrewsparkes: Also, for some reason, the GUI for tag_layout_template/new doesn't allow you to set the walking by option, it fixes it to wells of plate. You can only set Direction. That seems inflexible.

The UAT action isn't using these descriptive keys, I was wrong there.

yoldas (changes requested):
Could you also fix the UI tests in spec/features/tag_layout_template_spec.rb ?
Tests picking the option "InColumns (A1,B1,C1...)"  now should pick "column" for the direction algorithm I think.

yoldas:
I have one more check to do for this PR. In Limber tagging screen, there are two drop-down boxes, one for walking algorithm and one for direction. I will see if the controller changes have any side effect on them.

yoldas:
Thanks for testing with integration suite for the tagging screen.

yoldas (approved):
Fixes the UAT action UI, and provides consistency and avoids duplication throughout.
Tested both the UAT action and the template creation pages.

PR #4217, opened 2024-07-19
Remove code climate

BenTopping:

- Removes code climate in favour of codecov


BenTopping:
I don't think its a big issue but the codecov notification seems to update a few times after cucumber tests finish etc (look at the edits). From the docs it seems as though its suppose to wait to notify until all CI has finished so not sure if we need to tweak something there.

PR #4225, opened 2024-07-23
Y24-190-2: Add barcode printers to v2 API

### LINKED ISSUE #4223: Y24-190-2: Implement V2 endpoints to support Limber's Chromium pipelines
This is a sub-story to support Limber during Chromium pipelines with Sequencescape API V2 endpoints.  These pipelines are defined in the integration suite.

## References

Sibling: https://github.com/sanger/limber/issues/1803

sdjmchattie:
Closes #4223 

- Add new endpoint for Barcode Printers

### DIFF
--- a/app/resources/api/v2/barcode_printer_resource.rb
+++ b/app/resources/api/v2/barcode_printer_resource.rb
@@ -22,6 +22,10 @@ module Api
       attribute :uuid, readonly: true
 
       # @!attribute [r]
+      # @return [String] The service this printer can be instructed to print from.
+      attribute :print_service, readonly: true
+
+      # @!attribute [r]
       # @return [String] The name of the barcode type for this printer.
       attribute :barcode_type, readonly: true
 
--- a/app/resources/api/v2/barcode_printer_resource.rb
+++ b/app/resources/api/v2/barcode_printer_resource.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # Provides a JSON API representation of a barcode printer.
+    # See: http://jsonapi-resources.com/ for JSONAPI::Resource documentation
+    class BarcodePrinterResource < BaseResource
+      immutable
+
+      default_includes :uuid_object
+
+      ###
+      # Attributes
+      ###
+
+      # @!attribute [r]
+      # @return [String] The name of the barcode printer.
+      attribute :name
+
+      # @!attribute [r]
+      # @return [String] The UUID of the barcode printer.
+      attribute :uuid, readonly: true
+
+      # @!attribute [r]
+      # @return [String] The name of the printer type.
+      attribute :type_name
+
+      ###
+      # Getters and Setters
+      ###
+      def type_name
+        @model.barcode_printer_type.name
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


spec/resources/api/v2/barcode_printer_resource_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(resource).not_to have_attribute :id
+    expect(resource).to have_attribute :uuid
+    expect(resource).to have_attribute :name
+    expect(resource).to have_attribute :print_service
+    expect(resource).to have_attribute :barcode_type
+  end
+
+  # Updatable fields
+  it 'allows updating of read-write fields', :aggregate_failures do
+    expect(resource).to have_updatable_field :name
```
BenTopping, line 22: If the resource is immutable is it actually possible to update any of these fields?

app/resources/api/v2/barcode_printer_resource.rb
```diff
@@ -0,0 +1,40 @@
...
+      # Attributes
+      ###
+
+      # @!attribute [r]
+      # @return [String] The name of the barcode printer.
+      attribute :name
+
+      # @!attribute [r]
+      # @return [String] The UUID of the barcode printer.
+      attribute :uuid, readonly: true
```
BenTopping, line 22: Are these readonly by default because the class is immutable?

BenTopping (approved):
Looks good, learnt a few things about json api resources. Just a question about immutability.

app/resources/api/v2/barcode_printer_resource.rb
sdjmchattie: Yes basically.  I've been a bit torn about how to document these as I'm not sure the exported docs we're working on will tell you whether the resource is immutable or not.  We should probably find a way for it to do that.  It might be as simple as putting YARD docs above the class/module that state it.  It would need to be kept in sync by anyone updating these resources of course.

spec/resources/api/v2/barcode_printer_resource_spec.rb
sdjmchattie: No, but that's not managed at the resource level.  The request handler is the only thing that looks at the immutability of the resource itself.  I don't really like the way these tests have to be put together because of that.  I think the request test file does a might nicer job of describing the behaviour of the resource.

BenTopping: Yeah these resources specs seem to be a repetition of alot of the requests test files. Traction doesn't test resources directly, only through the api.

andrewsparkes (approved):
General question/comment: I'm assuming the plan for these API changes is to leave the API v1 endpoints intact for now, and switch everything in our repos across to use these new v2 endpoints. Will we add some kind of logging to catch further uses of v1 endpoints so we know whether they are still in use, and maybe give some clues as to what is calling them?

sdjmchattie:
> General question/comment: I'm assuming the plan for these API changes is to leave the API v1 endpoints intact for now, and switch everything in our repos across to use these new v2 endpoints. Will we add some kind of logging to catch further uses of v1 endpoints so we know whether they are still in use, and maybe give some clues as to what is calling them?

Yes that's right.  We can't take V1 out until we have no signs of it being used.  We have a Confluence document showing us where the usage is and there are code snippets in the associated page about how to pull the latest information about V1 usage: https://ssg-confluence.internal.sanger.ac.uk/display/PSD/API+v1+Usage+Data+-+July+2024?focusedCommentId=222298441#comment-222298441

PR #4228, opened 2024-07-24
add an extra "Contaminated Human Data Access Group" field to Study

sabrine33:


db/schema.rb
```diff
@@ -1928,19 +1929,14 @@
...
     t.text "locale", size: :medium
     t.integer "pipeline_id"
     t.index ["pipeline_id"], name: "index_workflows_on_pipeline_id"
   end
 
-  add_foreign_key "aliquot_indices", "aliquots", name: "fk_aliquot_indices_to_aliquots"
-  add_foreign_key "aliquot_indices", "receptacles", column: "lane_id"
-  add_foreign_key "aliquots", "primer_panels"
-  add_foreign_key "aliquots", "requests"
-  add_foreign_key "assets_deprecated", "plate_types", column: "labware_type_id"
```
yoldas, line 1943: I can't see equivalent statements in create_table blocks. Why are these removed?

config/locales/metadata/en.yml
```diff
@@ -508,6 +508,10 @@ en:
         data_deletion_period:
           label: Data deletion period
 
+        contaminated_human_data_access_group:
+          label: Contaminated Human Data Access Group
+          help: "Allows the specified Unix groups and users access to data separated from out from the main data product as likely contaminated with human - typically rarely used and only for validating the separation process as we may not have the ethical or legal."
+
```
yoldas, line 514: "separated from out from" -> "separated from" ?
"contaminated with human" -> "contaminated with human data" OR "contaminated with human sequence data" ?

sabrine33: I copied past the text in the text, that they requested to show as helper, well done for spotting this

db/schema.rb
sabrine33: This file is being automatically updated; maybe I should not include it in the PR?

app/models/api/study_io.rb
```diff
@@ -99,5 +99,6 @@ def render_class
     map_attribute_to_json_attribute(:hmdmc_approval_number, 'hmdmc_number')
     map_attribute_to_json_attribute(:s3_email_list)
     map_attribute_to_json_attribute(:data_deletion_period)
+    map_attribute_to_json_attribute(:contaminated_human_data_access_group)
```
yoldas, line 102: Could you add an assertion for this in `spec/models/api/study_io_spec.rb` ?

app/models/study.rb
```diff
@@ -256,6 +256,7 @@ class Study < ApplicationRecord # rubocop:todo Metrics/ClassLength
     # External Customers
     custom_attribute(:s3_email_list)
     custom_attribute(:data_deletion_period)
+    custom_attribute(:contaminated_human_data_access_group)
```
yoldas, line 259: Could you add some tests for this?
(Update study_metadata factory or add a variation or trait for this if necessary)
Add test in `spec/models/study_spec.rb` to check it is included in `study.study_metadata`?
Also test that it defaults to NULL as in acceptance criteria.

app/models/api/study_io.rb
sabrine33: I am trying to add some code, Codecov is already complaining about it. Spoiler alert: I am totally new to the Ruby world, so it might take more time than expected.

db/schema.rb
yoldas: db:migrate should handle the updates of the schema. I asked to check if you edited the file directly actually.

app/models/study.rb
sabrine33: I am trying to add some code, but Codecov is already complaining about it.

db/schema.rb
sabrine33: nope, I did not, this is been updated automatically.

sabrine33: nope! I did not have to that

PR #4232, opened 2024-07-25
Y24-181 - Clean up HiSeq submission templates

### LINKED ISSUE #4189: Y24-181 - Clean up HiSeq submission templates
**Describe the Housekeeping**
As an SSR, I would like the HiSeq-related submission templates to be removed from the dropdown list when creating a submission, because the HiSeq instruments are no longer in use, and it will make the list easier to use.

This means all the templates which end with the word sequencing needs to be deleted except the ones which have MiSeq, NovaSeq X & NovaSeq 6000. 
HiSeq is no longer used hence specially all the Hi Seq templates should be deleted. 

Siobhan, Liz C.

Siobhan

**Acceptance criteria**

- [ ] Submission template list accessible from 'Start a Bulk Submission' and 'Start a New Submission' menu options in Sequencescape does not display HiSeq related templates and all other Illumina templates. 
- [ ] Submission template list accessible from 'Generate Template' excel sheet under Bulk Submissions page in Sequencescape does not display HiSeq related templates and all other Illumina templates. 
- [ ] This should also reflect in the excel sheet below.

**Dependencies**
None

**References**
None

**Additional context**

- [_to__20240716_sd31.xlsx](https://github.com/user-attachments/files/16251887/_to__20240716_sd31.xlsx)

- image

- image

- image
- 
image

BenTopping:
Closes #4189

- adds support rake task to control the visibility of hiseq submission templates


lib/tasks/support/disable_hiseq_submission_templates.rake
KatyTaylor, line 1: I think we need a small change here. If you do:
```
SELECT *
FROM submission_templates
WHERE name like '%hiseq%'
ORDER BY `superceded_by_id`;
```

you'll see that there are quite a few superceded ones already. Some are 'superceded by unknown', and some are superceded by actual templates.

I think that in the 'disable' rake task, we should only disable those that are not already disabled (so as to avoid overwriting the info on ones that are superceded by actual templates).

The 'enable' task is a little trickier - you might have to save the list of ids that were disabled, and pass them in?

lib/illumina_c/helper.rb
KatyTaylor, line 1: Where is this module / class used? 🤔

lib/tasks/support/disable_hiseq_submission_templates.rake
BenTopping: Yes thats a good point, I have added a check to ensure the template is not already superceded. I think the enable task matters less because worst case we enable some unused templates which we can manually turn off.

lib/illumina_c/helper.rb
BenTopping: I though it was called by some meta creator but seems its unused. Will remove it.

lib/tasks/support/disable_hiseq_submission_templates.rake
KatyTaylor: It's not as important, because it's unlikely we'll have to run it. But as is, the 'enable' rake task will enable 97 templates that are currently disabled. So, I think we should either remove the 'enable' rake task, or re-write it to accept a list of ids.

lib/limber/helper.rb
```diff
@@ -166,12 +166,17 @@ def each_submission_template
...
+          superceded_by_id = sequencing_request_type&.name&.downcase&.include?('hiseq') ? -2 : -1
+
           yield(
             {
               name: name_for(cherrypick, sequencing_request_type),
               submission_class_name: 'LinearSubmission',
               submission_parameters: submission_parameters(cherrypick, sequencing_request_type),
-              product_line_id: product_line_id,
+              superceded_by_id:,
+              product_line_id:,
```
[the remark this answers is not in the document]
BenTopping: Ruby 3.1 > Shortcut for assigning a hash key to a variable https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/

lib/tasks/support/disable_hiseq_submission_templates.rake
BenTopping: Yeah fair enough, I will remove it.

yoldas (approved):
Looks good

lib/limber/helper.rb
```diff
@@ -166,12 +166,17 @@ def each_submission_template
         sequencing_request_types.each do |sequencing_request_type|
           next if SubmissionTemplate.exists?(name: name_for(cherrypick, sequencing_request_type))
 
+          # Hiseq templates are hidden as of Y24-181 so we set them as superceeded by unknown
+          # This allows us to keep the template in case it needs to be restored but not make it visible
+          superceded_by_id = sequencing_request_type&.name&.downcase&.include?('hiseq') ? -2 : -1
```
andrewsparkes, line 171: Why is codecov whining about tests here and further below when you have a test?

andrewsparkes: ah, because the test is for the rake task rather than this class maybe. Add some small unit tests just to check obvious use cases?

BenTopping: Yeah just added some really basic tests for this as there were none previously.

PR #4235, opened 2024-07-26
change the comments count to include asset comment

wendyyang:
Closes #

- 

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -563,6 +563,10 @@ class Request < ApplicationRecord # rubocop:todo Metrics/ClassLength
   def manifest_processed!
   end
 
+  def all_comments
+    comments.count + asset.comments.count
+  end 
+
   private
 
   def calculate_next_request_type_id
--- a/app/views/shared/_batch.html.erb
+++ b/app/views/shared/_batch.html.erb
@@ -55,8 +55,8 @@
         <% end %>
 
         <td><%= request.state.humanize %></td>
-
-        <td><%= link_to (pluralize request.comments.size, 'comment'),  request_comments_url(request) %></td>
+        
+        <td><%= link_to (pluralize request.all_comments, 'comment'),  request_comments_url(request) %></td>
 
       </tr>
     <% end -%>


app/models/request.rb
```diff
@@ -563,6 +563,11 @@ def product_line
   def manifest_processed!
   end
 
+  def self.get_all_comments(request)
```
dasunpubudumal, line 566: A minor thing, but should this be a class method or an instance method? It's defined on `request` model, and takes in a `request` (very much like `self` in Python and Rust, `this` in Java, TypeScript, and C++) so I would imagine it operates on the given instance, making it a candidate for an instance method?

app/models/request.rb
```diff
@@ -563,6 +563,11 @@ def product_line
...
+    counts = Comment.counts_for_requests([request])
+    counts[request.id]
+  end
```
KatyTaylor, line 569: Handy that this `counts_for_requests` method existed already! Looks like that's used for a similar link on the Pipeline inbox page, e.g. here - https://training.sequencescape.psd.sanger.ac.uk/pipelines/51

spec/models/request_spec.rb
```diff
@@ -494,4 +494,21 @@
       expect(subject[request_type2].started).to eq(1)
     end
   end
+
+  describe '.get_all_comments' do
```
KatyTaylor, line 498:
```suggestion
  describe '#get_all_comments' do
```

I was going to say that this is the convention elsewhere in Sequencescape. But it looks like we've got both!

KatyTaylor (approved):
Looks good to me!

app/models/request.rb
wendyyang: > A minor thing, but should this be a class method or an instance method? It's defined on `request` model, and takes in a `request` (very much like `self` in Python and Rust, `this` in Java, TypeScript, and C++) so I would imagine it operates on the given instance, making it a candidate for an instance method?

if it is an instance method, should it live in the controller?

dasunpubudumal: It looks like it should be in the `request` model as it takes in the current request (current instance). Anyway, I will approve this since it accomplishes what you're trying to do.

dasunpubudumal (approved):
Looks good 👍🏼

spec/models/request_spec.rb
yoldas: '.' is for class methods. '#' for instance methods.

wendyyang: in this case, I should use '.' then?

yoldas: The signature self.get_all_comments(request) shows it is a class method. Test for this method by convention will have a '.' prefix.

wendyyang: OK, thanks! will change back my initial one.

PR #4252, opened 2024-08-02
Y24-177 - scRNA - modify aggregation to simplify submission template name

andrewsparkes:


KatyTaylor (approved):
Looks good to me, haven't checked the Limber changes yet

yoldas (approved):
Yes, the changes make it generic and drops the chromium part from the request type key as well.

PR #4264, opened 2024-08-08
Y24-190: Populate API v2 documentation

sdjmchattie:

- Add overview sections with TODO items for every API V2 resource.
- Add missing controllers and routes for some endpoints that already existed.
- Fix the TubePurpose endpoint so that it only include tube purposes and not plate purposes.


.yardopts
```diff
@@ -1,6 +1,7 @@
 --markup=markdown
 --asset docs/images:images
 --exclude app/controllers/api/v2/[^/]*\.rb
+--exclude app/controllers/api/v2/concerns/.*
```
yoldas, line 4: Does the documentation contain the methods provided by the concerns where they are included?

app/controllers/api/v2/tag_group_adapter_types_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+module Api
+  module V2
+    # Provides a JSON API controller for TagGroup::AdapterType
+    # See: http://jsonapi-resources.com/ for JSONAPI::Resource documentation
+    class TagGroupAdapterTypesController < JSONAPI::ResourceController
+      # By default JSONAPI::ResourceController provides most the standard
+      # behaviour, and in many cases this file may be left empty.
+    end
+  end
+end
```
yoldas, line 12: TagGroupAdapterType is used for filtering TagGroups by adapter type name. A TagGroup has an association with TagGroupAdapterType using adapter_type_id field.

sdjmchattie: Not sure if you're asking for any changes here.  The documentation for these resources is here: https://github.com/sanger/sequencescape/blob/Y24-190-api-v2-documentation/app/resources/api/v2/tag_group_adapter_type_resource.rb

This PR doesn't intend to write specific documentation; only to set up a framework for it to be updated by people who know more about the resources.

app/resources/api/v2/shared_behaviour/labware.rb
```diff
@@ -7,62 +7,75 @@
...
+        included do
+          # Associations:
+          has_one :purpose, readonly: true, foreign_key: :plate_purpose_id, class_name: 'Purpose'
+          has_one :custom_metadatum_collection, foreign_key_on: :related
 
-    has_many :samples, readonly: true
-    has_many :studies, readonly: true
-    has_many :projects, readonly: true
-    has_many :comments, readonly: true
+          has_many :samples, readonly: true
```
dasunpubudumal, line 21: Were these blank lines added by the linters? 🤔

dasunpubudumal: Nope. You've changed the module structure. Got it.

app/resources/api/v2/barcode_printer_resource.rb
```diff
@@ -13,25 +23,26 @@ class BarcodePrinterResource < BaseResource
       # Attributes
       ###
 
-      # @!attribute [r]
-      # @return [String] The name of the barcode printer.
+      # @!attribute [r] name
+      #   @return [String] the name of the barcode printer.
       attribute :name
```
BenTopping, line 28: Does this need to be `readonly: true` to align it with the others since the resource is immutable? I appreciate its not part of the documentation so feel free to ignore.

app/resources/api/v2/labware_resource.rb
```diff
@@ -2,7 +2,18 @@
 
 module Api
   module V2
-    # LabwareResource
+    # @todo This documentation does not yet include a detailed description of what this resource represents.
+    # @todo This documentation does not yet include detailed descriptions for relationships, attributes and filters.
+    # @todo This documentation does not yet include any example usage of the API via cURL or similar.
+    #
+    # @note This resource is immutable: its endpoint will not accept `POST`, `PATCH`, or `DELETE` requests.
```
BenTopping, line 9: Is this immutable?

app/resources/api/v2/plate_purpose_resource.rb
```diff
@@ -13,34 +22,34 @@ class PlatePurposeResource < BaseResource
 
       # The following attributes are sent by Limber for a new plate purpose.
 
-      # @!attribute [rw]
-      # @return [String] The name of the plate purpose.
+      # @!attribute [rw] name
```
BenTopping, line 25: is [rw] the default?

app/resources/api/v2/tag_group_adapter_type_resource.rb
```diff
@@ -2,15 +2,22 @@
...
+    # @note Access this resource via the `/api/v2/tag_group_adapter_types/` endpoint.
+    #
+    # Provides a JSON:API representation of {TagGroup::AdapterType}.
+    #
+    # For more information about JSON:API see the [JSON:API Specifications](https://jsonapi.org/format/)
+    # or look at the [JSONAPI::Resources](http://jsonapi-resources.com/) package for Sequencescape's implementation
+    # of the JSON:API standard.
     class TagGroupAdapterTypeResource < BaseResource
-      model_name 'TagGroup::AdapterType', add_model_hint: false
+      model_name 'TagGroup::AdapterType', add_model_hint: true
```
BenTopping, line 17: Model hint change intended? If so the default is true so it can be removed.

app/resources/api/v2/shared_behaviour/labware.rb
sdjmchattie: The linter said the module had too many lines so I split it in two

app/resources/api/v2/tube_purpose_resource.rb
```diff
@@ -2,30 +2,39 @@
...
+    # @note Access this resource via the `/api/v2/tube_purposes/` endpoint.
+    #
+    # Provides a JSON:API representation of {Tube::Purpose}.
+    #
+    # For more information about JSON:API see the [JSON:API Specifications](https://jsonapi.org/format/)
+    # or look at the [JSONAPI::Resources](http://jsonapi-resources.com/) package for Sequencescape's implementation
+    # of the JSON:API standard.
     class TubePurposeResource < BaseResource
-      model_name 'Tube::Purpose'
+      model_name 'Tube::Purpose', model_hint: true
```
BenTopping, line 17: Is this model_hint correct? should this be add_model_hint?

BenTopping (changes requested):
Definitely a good addition, seems like a tedious task as well so well done for getting through it.

andrewsparkes (approved):

app/resources/api/v2/barcode_printer_resource.rb
sdjmchattie: I found it hard to know what to put in the documentation for immutable resources.  On the one hand, you could write to these attributes because they aren't marked as `readonly: true` but on the other hand, you can't write to the endpoint at all.  I think it doesn't really matter either way to be honest since it's immutable.  However I don't know that we should be using `readonly: true` at all -- see the other PR.  It's definitely not a documentation concern though.

app/resources/api/v2/labware_resource.rb
sdjmchattie: Quite right.  I will remove this line from the YARD documentation.

app/resources/api/v2/plate_purpose_resource.rb
sdjmchattie: I think it is, yeah.  However YARD does say you can put `[rw]` if you want, and I think it's good to have the access modifier listed for all of them rather than only when it's absolutely necessary.

app/resources/api/v2/tag_group_adapter_type_resource.rb
sdjmchattie: Hmmm ... good question, I need to check this again.  I think the endpoint was not working until I changed this to `true`.  If it does need to be `true`, I will remove it as you suggested the default will be `true` anyway.  I'll check it in the morning.

app/resources/api/v2/tube_purpose_resource.rb
sdjmchattie: Again I'm not sure.  I have recollections of it not working as an endpoint without me adding this, but I honestly am not sure what the differences are between all these model names and model hints.  I have read the documentation, and it doesn't explain it well enough.

sdjmchattie: Thanks, so I tried this out this morning and it doesn't need the model hint adding explicitly (and this isn't the right way to add it).  It works without.  Not sure why I felt the need to add it, but I've removed this again.

app/resources/api/v2/tag_group_adapter_type_resource.rb
sdjmchattie: I remove the `add_model_hint: true` because, while it is needed, you don't need to do it explicitly like you said above.

BenTopping (approved):

PR #4268, opened 2024-08-12
Y24-190-2: Support Limber labware metadata in API v2

### LINKED ISSUE #4223: Y24-190-2: Implement V2 endpoints to support Limber's Chromium pipelines

sdjmchattie:
Closes #4223 

- Apply better YARD documentation to CustomMetadatumCollection resource.
- Apply better YARD documentation to User resource.
- Create better tests of the above endpoints.
- Fix the script with generates the Devour config file contents.


app/resources/api/v2/custom_metadatum_collection_resource.rb
```diff
@@ -5,31 +5,41 @@ module V2
...
+      ###
       # Attributes
-      attribute :uuid, readonly: true
-      attribute :metadata
+      ###
 
-      # This is required for POST
+      # @!attribute [r] uuid
+      #   @return [String] The UUID of the collection.
+      attribute :uuid
```
BenTopping, line 16: Should we add `readonly: true` even though its excluded from the creatable and updatable fields. Means you dont need the creatable_fields method.

spec/requests/api/v2/users_spec.rb
```diff
@@ -8,54 +8,103 @@
...
 
-      # check to make sure the right amount of messages are returned
-      expect(json['data'].length).to eq(7)
+      expect(json['data'].length).to eq User.count
     end
 
-    it 'allows filtering of users by user_code with swipecard' do
-      api_get "#{base_endpoint}?filter[user_code]=#{swipecard_code}"
+    describe 'filtering' do
+      let(:user) { users[2] }
```
BenTopping, line 27: test for new uuid filter?

BenTopping (changes requested):

spec/requests/api/v2/users_spec.rb
sdjmchattie: Wow yeah!  I had meant to put that in, but it slipped my radar

app/resources/api/v2/custom_metadatum_collection_resource.rb
sdjmchattie: Either way is legitimate.  The difference is in the error you get from the server.  I preferred the one where you use creatable and updatable fields but I'm happy to take advice.  I think it's better to avoid causing 500 server errors and the error message is much more technical when you get that response.

## Using `creatable_fields`
image

## Using `readonly: true`
image

BenTopping: Ah ok, yeah the first error is nicer. Happy for you to leave as is then.

BenTopping (approved):

Issue #4160, opened 2024-06-28
Y24-170 - Link dual index tag groups


KatyTaylor:
**Background - science of tags**
Tag plates are plates that contain small DNA sequences called tags (or barcodes, indexes...). The contents of these plates are added to samples in order to create libraries that can be sequenced.

Dual index tags are ones where a sample will be linked to two tags, called 'i7' and 'i5' (refers to the direction you're reading the DNA molecule), or sometimes referred to in Sequencescape as 'tag1' and 'tag2'. 

In dual index tag plates, both i7 and i5 oligos are in the same well. i.e. well 'A1' will contain both an i7 and an i5 tag; B1 will contain 2 different ones, etc. They are always combined in the well - they don't come separately.

**Background - tags in Sequencescape & Limber**

Sequencescape represents dual index tag groups as 2 separate tag groups - one that contains the i7 tags, and one that contains the i5 tags. For example, see the following:
- https://training.sequencescape.psd.sanger.ac.uk/tag_groups/227 (i7)
- https://training.sequencescape.psd.sanger.ac.uk/tag_groups/228 (i5)

Tag groups are used in two features in Sequencescape:
1. When creating libraries in SeqOps, using Limber - [see screenshot of page here](https://lucid.app/lucidchart/4d815a3d-a242-4ba7-a305-bb2f7b0d0fc3/edit?viewport_loc=341%2C9%2C2035%2C1267%2C0_0&invitationId=inv_94230715-32a8-42e6-8493-f3e4018fe12f)
2. When importing ready-made libraries, using the Chromium library plate manifest

On the tagging screen linked above, there are two ways of selecting which tags you want. One is to scan a barcode on the actual physical tag plate you're using - see the 'Tag Plate' box. The other is to select the i7 and i5 tag groups that you are using - see the dropdowns further down.

**Research questions**

1. Is there any reason from a user point of view why there should be two tag group dropdowns instead of one, in the above mentioned tagging screen? This is a question for the lab users. Not sure who best to ask, but I think I have talked to Matt J (former bespoke team) about using this screen in the past - maybe a good place to start. Also, Matthew M, Sara W, and Jamie L.
2. If answer to above is no, come up with a design for a combined i7 & i5 dual index tag group - either linking together the existing tag groups (using something more robust than the name), or creating a new model. It should be able to be used in the existing contexts:
    - replacing the two dropdowns on that tagging screen with one
[...]

PR #4273, opened 2024-08-14
Y24-218 - Adds tag set model to link dual index tags

### LINKED ISSUE #4236: Y24-218 - Create a new tag set model to link tag groups
**User story**
As PSD, we would like to create a new data model to link tag groups (indexes) of the same tag sets together in order to support selection of the pair of i7 and i5 tag groups for dual indexed tag sets together and unify access to single and dual indexed tag sets. This data model aims keeping the internal tag group model intact while providing support for a unified access to the tag groups at interfacing. 

Abdullah, Andrew, Katy, Neil

PSD

**Acceptance criteria**
- [ ] Add a TagSet model to link TagGroups.
- [ ] Add name field to TagSet. Ideally, the name will match the kit provider specified names. For example, 
  - Single index tag sets should have a name like 
PN-1000212 Single Index Kit N, Set A . It can link TagGroup with the same name. 
  - Dual index tag sets should have a name like PN-1000215/PN-3000431 Dual Index Kit TT, Set A . It can link TagGroups with names suffixed with (i7) and (i5).
- [ ] Add a tag_group and tag2_group fields (associations). The following are the association names on TagSet to its TagGroups, following the same conventions in Sequencescape.
  - tag_group:
    - the only group for single indexed tag sets and
    - the first group for dual indexed tag sets (i7)
  - tag2_group:
    - null for single indexed tag sets
    - the second group for dual indexed tag sets (i5).
- [ ] TagGroups can be soft-deleted by setting visibility flag on them. A TagSet should be visible if its tag groups are visible.
- [ ] Model allows creating custom tag sets with mix of unrelated or custom tag groups.
- [ ] Validate that adapter type of tag groups of a tag set are equal.
- [ ] Assuming adapter type will be the same for both tag groups of a dual index tag set, provide a method to return the first adapter type of its tag groups.
- [ ] Provide a method to return visibility for a tag set by checking the visibility of its tag groups.
- Associations and integrity. Tag groups are soft-deleted only, which is covered above. The following criteria are to make sure tag sets and tag group records can exist independently. Ensure there is no cascading deletes such as dependent: :destroy 
  - [ ] If the tag set is deleted, its tag group(s) should NOT be deleted
  - [ ] If a tag group is deleted, its tag set(s) should NOT be deleted .

**Dependencies**
- #4160

References
See links in the parent research story.

**Additional context**
This simple model can co-exists and reuses the existing models, records, and operations in Sequencescape and Limber; hence allows changing for interfaces, queries, and UI elements without breaking existing models.

Linking existing TagGroup records in DB is covered in another story. See links in the parent research story.

Not adding a flag to explicitly specify if a tag set is single indexed or dual indexed; Rely on availability of tag_group and tag2_group associations on the tag set for now to keep it simple.

Adapter type can stay as tag group field. For dual indexed tag sets, both tag groups have the same adapter type, which is set when creating tag groups. [Tag set should return the first adapter type of its tag groups with the assumption that the adapters of its tag groups will be the same. Add a validation to enforce that]

Visible column: Tag set will find its visibility by checking the visibility of tag groups [Tag set will be visible if all of its tag groups are visible]

BenTopping:
Closes #4236 

- Adds TagSet table and model

### DIFF
--- a/app/models/tag_set.rb
+++ b/app/models/tag_set.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
+# Links together two related tag groups - i7 and i5 - to represent a dual index tag set
+# It can also be used to represent single index tag sets
+# Background explained in Y24-170 (https://github.com/sanger/sequencescape/issues/4160)
 class TagSet < ApplicationRecord
+  # For dual index tags, tag_group is i7 oligos and tag2_group is i5 oligos
   belongs_to :tag_group, class_name: 'TagGroup', optional: false
   belongs_to :tag2_group, class_name: 'TagGroup', optional: true
 
@@ -9,7 +13,7 @@ class TagSet < ApplicationRecord
   delegate :adapter_type, to: :tag_group
 
   validates :name, presence: true, uniqueness: true
-  validate :tag_group_adapter_types
+  validate :tag_group_adapter_types_must_match
 
   # Dynamic method to determine the visibility of a tag_set based on the visibility of its tag_groups
   def visible
@@ -17,7 +21,7 @@ class TagSet < ApplicationRecord
   end
 
   # Method to determine that both tag groups have the same adapter type
-  def tag_group_adapter_types
+  def tag_group_adapter_types_must_match
     return unless tag2_group && tag_group.adapter_type != tag2_group.adapter_type
     errors.add(:tag_group, 'Adapter types of tag groups must match')
   end
--- a/app/models/tag_set.rb
+++ b/app/models/tag_set.rb
@@ -21,4 +21,12 @@ class TagSet < ApplicationRecord
     return unless tag2_group && tag_group.adapter_type != tag2_group.adapter_type
     errors.add(:tag_group, 'Adapter types of tag groups must match')
   end
+
+  def tag_group_name=(name)
+    self.tag_group = TagGroup.find_by!(name: name)
+  end
+
+  def tag2_group_name=(name)
+    self.tag2_group = TagGroup.find_by!(name: name)
+  end
 end
--- a/app/models/tag_set.rb
+++ b/app/models/tag_set.rb
@@ -5,7 +5,7 @@ class TagSet < ApplicationRecord
   belongs_to :tag2_group, class_name: 'TagGroup', optional: true
 
   # We can assume adapter_type is the same for both tag groups
-  # But tag2_group may be present so we delegate to tag_group
+  # But tag2_group may not be present so we delegate to tag_group
   delegate :adapter_type, to: :tag_group
 
   validates :name, presence: true, uniqueness: true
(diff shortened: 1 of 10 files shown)


app/models/tag_set.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class TagSet < ApplicationRecord
+  belongs_to :tag_group, class_name: 'TagGroup', optional: false
+  belongs_to :tag2_group, class_name: 'TagGroup', optional: true
+
+  # We can assume adapter_type is the same for both tag groups
+  # So we delegate to either group
```
yoldas, line 8: Given that TagSet is allowed to be a single index tag set, having the second tag group (tag2_group) null, the comment is not quite right. We delegate to the first only to cover this case.

Although the validation when TagSet is dual index tag set ensures the adapter types are equal, we still delegate to the first because of the case above.

BenTopping: Good point, ill update the comment

app/models/tag_set.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class TagSet < ApplicationRecord
+  belongs_to :tag_group, class_name: 'TagGroup', optional: false
+  belongs_to :tag2_group, class_name: 'TagGroup', optional: true
+
+  # We can assume adapter_type is the same for both tag groups
+  # But tag2_group may be present so we delegate to tag_group
```
KatyTaylor, line 8:
```suggestion
  # But tag2_group may not be present so we delegate to tag_group
```

app/models/tag_set.rb
KatyTaylor, line 1: Can we add a top-level comment with a bit of background? Something like,

> Links together two related tag groups - i7 and i5 - to represent a dual index tag set.
> Can also be used to represent single index tag sets.
> Background explained in Y24-170 (https://github.com/sanger/sequencescape/issues/4160)

app/models/tag_set.rb
```diff
@@ -0,0 +1,32 @@
...
+  delegate :adapter_type, to: :tag_group
+
+  validates :name, presence: true, uniqueness: true
+  validate :tag_group_adapter_types
```
KatyTaylor, line 12: I think naming convention for validators is more like `validate :tag_group_adapter_types_must_match`

See examples here - https://apidock.com/rails/ActiveRecord/Validations/ClassMethods/validate

app/models/tag_set.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class TagSet < ApplicationRecord
+  belongs_to :tag_group, class_name: 'TagGroup', optional: false
```
KatyTaylor, line 4: Probably worth saying in a comment that, for dual index tag_group links to the 'i7' oligos, and tag2_group links to the 'i5' oligos?

KatyTaylor (commented):
Just a few suggestions, nothing serious!

KatyTaylor (approved):
Thanks for the changes

yoldas (approved):

PR #4274, opened 2024-08-14
Y24-190-3: Support Limber with v2 TransferTemplate and Transfer

### LINKED ISSUE #4244: Y24-190-3: Implement V2 endpoints to support Limber's Labware Creators

## References

sdjmchattie:
Closes #4244 

Note:  This is not a complete implementation.  We are missing tests and I need to build a mechanism for creating `Transfer`s using information stored in a `TransferTemplate`.

- Add endpoints with basic attributes for `Transfer` and `TransferTemplate`.


yoldas (commented):
Overall the changes look good to me. 

A few notes:

I have found the default sti_type filter approach for transfer resources specifically very clever. I could not figure out how more specific types are used in Limber though.

If you have not done so, config:generate task creates a file, for example locally config/settings/development.yml . You will see a list of transfer_templates available to Limber and individual purposes include references to them. Generating the same content may be considered as testable.

In many cases in Limber, we end up creating a custom labware creator and `transfer_material_from_parent!` method is overridden and we pass only the attributes as a hash. V1 `api.transfer_request_collection` is "created" in that case.

If a specific transfer template to be used, we may need to override the `transfer_template` method of Labware creators in Limber if insufficient polymorphic support gets in the way.

app/controllers/api/v2/transfers_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+            context[:polymorphic_type] = template.transfer_class_name
+            attributes[:transfers] = template.transfers if template.transfers.present?
+
+            # Remove the UUID of the transfer template from the attributes.
+            attributes.delete(:transfer_template_uuid)
+          end
+
+          super(operations)
+        end
+      end
```
yoldas, line 32: Assuming that Processor calls create by passing the context in create_resource, this should work.

If resource_klass.new is called directly anywhere, it won't work though because operation's resource_klass is still TransferResource.

sdjmchattie:
OMG.  The tests finally pass!  This is ready for review.

yoldas (approved):
Looks good and tested.

PR #4275, opened 2024-08-14
Y24 234 update missing event history records

seenanair:
Closes #https://github.com/sanger/asset_audits/issues/657

- Adds a rake task to create records in asset_audits table using a csv file
- csv file format is
 barcode,created_by,created_at,message
barcode1,user1@example.com,2023-10-05 12:34:56,Destroying location,
barcode2,user2@example.com,2024-05-05 12:34:56,Destroying labware


spec/lib/add_missing_asset_audit_records_spec.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+RSpec.describe 'asset_audit:add_missing_records', type: :task do
+  let(:file_path) { File.join('spec', 'data', 'asset_audits', 'data.csv') }
+
+  before do
+    Rake.application.rake_require 'tasks/add_missing_asset_audit_records'
+    Rake::Task.define_task(:environment)
+    allow(File).to receive(:exist?).and_return(true) # Stub default value
```
yoldas, line 10: Stubbing `File` like this is too broad and it will have side effects. It may even break the framework itself during the test. Please use fixture files instead.

lib/tasks/add_missing_asset_audit_records.rake
```diff
@@ -0,0 +1,49 @@
...
+          AssetAudit.create!(
+            message: "Process '#{row['message']}' performed on instrument Destroying instrument",
+            created_by: row['created_by'].strip,
+            created_at: row['created_at'].strip,
+            asset_id: asset.id,
+            key: key
+          )
+          puts "Record for asset_id #{asset.id} successfully inserted."
+        rescue ActiveRecord::ActiveRecordError, StandardError => e
+          puts "Error inserting record for asset_id #{asset.id}: #{e.message}"
```
yoldas, line 44: Because you are catching and handling the exception here, the transaction block becomes useless. If we do it like this, we will need to record a log and deal with each individual case separately to fix them. I suppose the intention was rolling back the whole transaction if there is even a single error. Either let the exception bubble up or explicitly rollback the active transaction and break the loop.

spec/data/asset_audits/data.csv
```diff
@@ -0,0 +1,3 @@
+barcode,message,created_by,created_at
+SQPD-1,Destroying location,User1,2021-01-01 12:00:00
+SQPD-2,Destroying labware,User2,2021-01-02 12:00:00
```
yoldas, line 3: Fixture file names or the directory used for grouping them together could show the intention. 

spec/data/asset_audits/add_missing_asset_audit_records/
good_file.csv
badly_formatted.csv
no_data_section.csv
insufficient_columns.csv

etc.

yoldas (changes requested):
I have some comments, mainly about rolling back the transaction and using a fixture file instead of stubbing File.

spec/lib/add_missing_asset_audit_records_spec.rb
seenanair: Ok, Mostly I have used mocking in similar situations in past where CSV.read is mocked to simulate specific conditions which I thought is very useful and easy to test different edge cases associated with parsing and validation. Thought mocks will give a higher level isolation, that tests do not depend on the filesystem.
But, I can see the value in your suggestion  as well of using different files for each case, which provides more realistic testing. I think fixture files also make the tests more readable and understandable. 
I will update my approach based on your suggestion.

seenanair: Ok, I have already removed it in the latest changes with the PR update.

lib/tasks/add_missing_asset_audit_records.rake
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    unless csv_data.headers.length == required_csv_headers.length
+      puts 'Failed to read CSV file: Invalid number of header columns'
+      next
+    end
+
+    # Process the CSV data and create asset audit records data array to insert
+    asset_audit_data = []
+    csv_data.each do |row|
```
yoldas, line 39: We are **skipping** bad rows in this loop, instead of stopping the processing. Is this intended?

yoldas (approved):
Looks good. I have a question about skipping the bad records in CSV file instead of stopping the processing.

lib/tasks/add_missing_asset_audit_records.rake
seenanair: Good question! Now that you've asked it, I think it’s better to stop the processing and return in case of any errors, ensuring uniformity across all validations. Therefore,  for the addition of records to occur, we need to ensure that all rows in the CSV file are error-free. I believe this approach is better.

seenanair: Good question! Now that you've asked it, I think it’s better to stop the processing and return in case of any errors, ensuring uniformity across all validations. Therefore, for the addition of records to occur, we need to ensure that all rows in the CSV file are error-free. I believe this approach is better.
 I will make this change. Thank you.

yoldas (approved):

PR #4277, opened 2024-08-14
add wip_list method to return wip file list

wendyyang:
Closes #

add wip_list method to return wip file list
- 


lib/record_loader/application_record_loader.rb
```diff
@@ -9,5 +10,12 @@ class ApplicationRecordLoader < RecordLoader::Base
     # Uses the standard RailsAdapter
     # @see https://rubydoc.info/github/sanger/record_loader/RecordLoader/Adapter
     adapter RecordLoader::Adapter::Rails.new
+
+    def wip_list
+      wip_files = []
+      wip_files_path = Rails.root.join('config/default_records')
+      Find.find(wip_files_path) { |path| wip_files << path if /\wip\.yml$/.match?(path) }
+      wip_files
+    end
```
yoldas, line 19: This method is supposed to return names but it is returning file paths. The names are intended to list the **features**. For example, it should have an item "011_scrna_core_cdna_prep_submission_templates" instead of a path like "/Users/ay6/webapps/sequencescape/config/default_records/submission_templates/011_scrna_core_cdna_prep_submission_templates.wip.yml"

Ref: https://github.com/sanger/record_loader

config/environments/development.rb
```diff
@@ -71,6 +71,8 @@
     Bullet.bullet_logger = true
     Bullet.rails_logger = true
   end
+
+  config.deploy_wip_pipelines = true
```
yoldas, line 75: This sets a config option similar to Limber but it needs to be used by something. The loader you have written needs to act according to this option for enabling WIPs. At the moment, the loader is trying to enable all WIPS by returning all without any question.

yoldas: In one of our meetings, I think we have talked about a better naming to avoid limiting the semantics to **pipelines** in Sequencescape context.

yoldas: This PR needs to be accompanied by a deployment project PR to set the option value to `true` for UAT environment and `false` for production and training.

The piece of code using this option must also make sure that it does not load features unless it is explicitly set. This is very important to prevent loading WIPs in training and production.

yoldas (changes requested):
I have a few comments.

lib/record_loader/application_record_loader.rb
wendyyang: Thank you! Returning file names works!

config/environments/development.rb
wendyyang: > This sets a config option similar to Limber but it needs to be used by something. The loader you have written needs to act according to this option for enabling WIPs. At the moment, the loader is trying to enable all WIPS by returning all without any question.

We either to set it loading all WIPs, or none of the WIPs, is that right?

yoldas: Yes, if you have to test it on Training. The DB is reset every Saturday from Production DB.

lib/record_loader/application_record_loader.rb
yoldas: It should also return all or none depending on the config option. Where is it done?

yoldas (changes requested):

lib/record_loader/application_record_loader.rb
wendyyang: in dev env, in config/environments/development.rb, I added a line 'config.deploy_wip_pipelines = true' . But i tested, it didn't work. Whatever it set to true or false, all wip files are loaded. any suggestions?

in training/uat/prod, the flag is set in deployment project. Please review the codes if that is the right place.

yoldas: You need to check the config option and return the WIPs if it is true. You need to code that in. The option is new, it has no idea about how to use it. I suppose you will change wip_list method. Have a search in the SS code and see how config options are accessed. Return an empty array from the method or all WIPs based on the option.

yoldas (changes requested):

lib/record_loader/application_record_loader.rb
```diff
@@ -9,5 +10,21 @@ class ApplicationRecordLoader < RecordLoader::Base
...
     # @see https://rubydoc.info/github/sanger/record_loader/RecordLoader/Adapter
     adapter RecordLoader::Adapter::Rails.new
+
+    def wip_list
+      # return a list of WIP files name as features if deploy_wip_pipelines is set to true, or return empty list
+      deploy_wip_pipelines = Rails.application.config.try(:deploy_wip_pipelines) || false
+      return [] unless deploy_wip_pipelines
+
+      wip_files = []
+      wip_files_path = @path
```
KatyTaylor, line 20: Where does this `@path` instance variable come from?

KatyTaylor: I see, it comes from the `RecordLoader::Base` class within the Record Loader gem - https://github.com/sanger/record_loader/blob/master/lib/record_loader/base.rb

This feels like bad practice, because that instance variable is not explicitly exposed / documented so could change in a future version of the gem. But since it's our gem and the alternative is to hardcode the path `config/default_records`, it's probably fine for now! I guess to make it better we'd have to modify the gem itself. 

Can you add a comment above this line explaining where `@path` comes from? Thanks.

spec/data/record_loader/application_record_loader/example.wip.yml
KatyTaylor, line 1: I would personally make the contents of this file simpler, since you're not using the contents, to make it clear it doesn't matter. e.g.
```
# Test record loader records 
---
Test Record:
  test_attribute: "A Value"
```

spec/lib/record_loader/application_record_loader_spec.rb
KatyTaylor, line 1: Nice, covering the important 3 use cases.

KatyTaylor (commented):
Left a couple of suggestions.

KatyTaylor (approved):
Looks good, thanks for making the changes.

lib/record_loader/application_record_loader.rb
yoldas: I would like to record a few notes here.

The "@path" field is initialised from the directory argument passed to the initialiser method of the super class. The initialiser method is inherited just like any other method in ruby. Using the stored "@path" in the child allows testability by passing a different directory in tests. 

The comment says, "@path returns 'config/default_records' directory", which is not quite true. It is whatever the directory argument passed to the initialiser (after a few checks), but it defaults to 'config/default_records' .

As the superclass does not provide a getter or an attribute reader for "@path", the subclass accessed it directly in the subclass. There is no strict private fields in ruby either. It is a bad practice but as Katy said alternative would need modifying the superclass as well.

PR #4284, opened 2024-08-19
“Request type”, “Read length” and “Flowcell type” should not set to default values

sabrine33:


BenTopping (commented):
Tested locally and works correctly 👌 

Is there a way we can add any tests for these changes? I also noticed if you don't select a request type and try to create the request you get a bit of an odd error (attached photo). Can we may the request type field mandatory or improve the error message somehow? 
Screenshot 2024-08-21 at 10 04 27

Issue #4238, opened 2024-07-26
Y24-220 - Create API V2 end point for tag sets. 


yoldas:
**User story**
As PSD, we would like to design an API V2 endpoint to query tag sets. A tag set can be single index or dual index. A single index tag set has an association with one tag group. A dual index tag set has associations with two tag groups.

**Who are the primary contacts for this story**
Abdullah, Andrew, Katy, Neil

**Who is the nominated tester for UAT**
PSD

**Acceptance criteria**
To be considered successful the solution must allow:
- Endpoint:
  - [ ] Listing all tag sets
  - [ ] Reading one tag set by id
  - [ ] Has a filter to return only the tag sets by their tag group adapter types. For dual index tag sets, the model has a validation that the adapter types of its tag groups are equal. Therefore, the model has a method to return the first tag group adapter type as its adapter type.
  - [ ] Includes tag groups and tags of the tag set in the response if requested, (JSON API compound document)
  - [ ] Excludes soft-deleted tag sets i.e. at least one of the tag groups is soft-deleted using visibility flag
  - [ ] See the existing tag group request/response in Additional Context. The new endpoint should attach tag groups to their parent tag sets in the response.
- TagSet resource
  - [ ] Has a name, same as the underlying model
  - [ ] Has a uuid, same as the underlying model, because of the included concern.
  - [ ] Has one tag_group, the only group for single index tag sets and the first group for dual index tag sets
  - [ ] Has one tag2_group, null for single index tag sets and the second group for dual index tag sets
- Client side
  - [ ] convenience methods as in the server model: visible and adapter type

**Dependencies**
This story is blocked by the following dependencies:
- #4160
- #4236

References
This story has a non-blocking relationship with:
See links in the parent research story.

**Additional context**
Initial aim of this story is supporting the Limber UI story for selection of i7 and i5 tag groups for Limber Tagging screen by tag set name only; instead of specifying the groups separately. There will be a Limber UI story for that. For testing the endpoint, similar/modified payloads for LCA Connect PCRXP plate in Cardinal pipeline integration suite test may be useful.

--- 

### Existing tag group listing

Existing tag group API request from Limber retrieves all tag groups by an adapter filter. The same result is listed in both i7 and i5 drop-downs. The following example shows the request and response headers and Preview of a tag group that would belong to a single-index tag set and a pair of tag groups that would belong to a dual-index tag set.

[...]

PR #4286, opened 2024-08-19
Y24-221 - Adds tag set pages

### LINKED ISSUE #4239: Y24-221 - Create Sequencescape pages for tag sets
**User story**
As PSD, we would like to create a page to link tag groups as tag sets and list the tag sets.

Abdullah, Andrew, Katy, Neil

PSD

**Acceptance criteria**

New Tag Sets:
- [ ] Input for the name of the tag set
- [ ] Dropdown for the first tag group
- [ ] Dropdown for the second tag group if dual index tag set
- [ ] Only SSRs and admin to create Tag Sets. [Added after 6th August meeting]

Tag Set Listing
- [ ] List tag sets and its tag groups.

View Tag Set
- [ ] Individual tag set page shows tag set linking
- [ ] The link text to create the tag layout template  is **Create a new tag layout template from this tag set** 
- [ ] The link navigates to the new tag layout template by setting tag_group_id and tag2_group_id URL parameters to initialise the target page with the correct tag groups.

**Dependencies**
- #4160
- #4236

References
See links in the parent research story.

**Additional context**
- Existing UAT New Tag Group page: https://uat.sequencescape.psd.sanger.ac.uk/tag_groups/new
- Existing UAT Tag Group listing page: https://uat.sequencescape.psd.sanger.ac.uk/tag_groups/
- Existing View Tag Group page: https://uat.sequencescape.psd.sanger.ac.uk/tag_groups/1

Update requirements after discussions with users:
- Validation of pairs selected
- Filters based on for example chromium ones or dual ones, adapter type etc.

BenTopping:
Closes #4239 

- Adds index, creation and show pages for TagSets (accessible through the home page sidebar links)


app/resources/api/v2/tag_layout_template_resource.rb
```diff
@@ -23,7 +23,6 @@ class TagLayoutTemplateResource < BaseResource
       #   A relationship for the primary tag group associated with the tag layout template.
       #   @return [Api::V2::TagGroupResource]
       has_one :tag_group
-      has_one :tag2_group
```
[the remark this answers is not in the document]
KatyTaylor: Oh I see sorry, it's duplicated below.

app/controllers/tag_layout_templates_controller.rb
```diff
@@ -23,7 +23,8 @@ def show
   ##
   # Allows for the passing in of tag group id using a link from the tag group show page.
   def new
-    @tag_layout_template = TagLayoutTemplate.new(tag_group_id: params[:tag_group_id])
+    @tag_layout_template =
+      TagLayoutTemplate.new(tag_group_id: params[:tag_group_id], tag2_group_id: params[:tag2_group_id])
```
KatyTaylor, line 27: What happened here, was this a bug?

app/resources/api/v2/tag_set_resource.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # @todo This documentation does not yet include a detailed description of what this resource represents.
+    # @todo This documentation does not yet include detailed descriptions for relationships, attributes and filters.
+    # @todo This documentation does not yet include any example usage of the API via cURL or similar.
```
KatyTaylor, line 7: Would you have a stab at filling these out? It seems like it will be easiest to do right now when we have the model and reasons for creating it fresh in our minds.

KatyTaylor (commented):
Just a couple of comments. Screenshots would also be nice although I am planning to check it out locally and have a look. Thanks!

app/resources/api/v2/tag_set_resource.rb
```diff
@@ -0,0 +1,28 @@
...
+    # For more information about JSON:API see the [JSON:API Specifications](https://jsonapi.org/format/)
+    # or look at the [JSONAPI::Resources](http://jsonapi-resources.com/) package for Sequencescape's implementation
+    # of the JSON:API standard.
+    class TagSetResource < BaseResource
+      # Associations
+      has_one :tag_group
+      has_one :tag2_group, class_name: 'TagGroup'
+
+      # Attributes
+      attribute :name, readonly: true
```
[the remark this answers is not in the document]
sdjmchattie: I don't think you do HAVE to do that, but when you do, you get better error messages from the API.  You get something quite techincal about missing setter methods with a stack trace if you use the `readonly: true` approach.  If you use the methods that indicate which fields are updatable and creatable, the error comes back saying something like `Parameter not allowed: name`

sdjmchattie:
I think you probably want to implement the tests in `spec/requests/api/v2` as well.  Those are actually the more useful tests as they confirm for certain what can and cannot be done on the API.

app/resources/api/v2/tag_layout_template_resource.rb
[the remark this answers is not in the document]
BenTopping: Yep was getting a duplication warning in the console so its unrelated to this PR but still useful to fix.

app/controllers/tag_layout_templates_controller.rb
BenTopping: This is just to facilitate auto populating both the tag_group and tag2_group in a tag layout template when navigating from a TagSet as previously it was only navigated to from a tag group so a tag2_group param wasn't useful.

app/resources/api/v2/tag_set_resource.rb
BenTopping: Yep sure, im not actually sure if I need this resource as I haven't exposed the TagSet in the API, I will check. Should I be exposing TagSets in the API? Is that something I should do in this PR or should that be included if and when required (e.g. needed in a Limber story).

BenTopping:
> I think you probably want to implement the tests in `spec/requests/api/v2` as well. Those are actually the more useful tests as they confirm for certain what can and cannot be done on the API.

Currently TagSets aren't exposed in the API. I will double check if I should be exposing them in this story and if I need the resource for this current implementation, but I will add tests if I do add API support.

app/controllers/admin/tag_sets_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def create
+    @tag_set = TagSet.new(tag_set_params)
+
+    respond_to do |format|
+      if @tag_set.save
+        flash[:notice] = 'Tag Set was successfully created.'
+        format.html { redirect_to admin_tag_set_path(@tag_set) }
+      else
+        format.html { render action: :new }
+      end
```
yoldas, line 35: We are not returning feedback on error. It seems to just go to new TagSet page again. Does it show a banner?

yoldas (commented):
I will continue the review.

app/controllers/admin/tag_sets_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def create
+    @tag_set = TagSet.new(tag_set_params)
+
+    respond_to do |format|
+      if @tag_set.save
+        flash[:notice] = 'Tag Set was successfully created.'
+        format.html { redirect_to admin_tag_set_path(@tag_set) }
+      else
+        format.html { render action: :new }
```
dasunpubudumal, line 34: Looks like when save returns false (I assume that's bad) it just returns to the new action?

Do you think we should respond that the previous response was erroneous or something?

spec/controllers/admin/tag_sets_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      it 'is successful' do
+        expect(response).to have_http_status(:success)
+        expect(response).to render_template('show')
+      end
+    end
+
+    describe '#create' do
+      before { post :create, params: { tag_set: { name: 'test-123', tag_group_id: create(:tag_group).id } } }
+
+      it 'renders the show template' do
```
dasunpubudumal, line 37: Just wondering whether if it's possible for the `save` function to go wrong (e.g. a validation error, maybe?).

app/controllers/admin/tag_sets_controller.rb
sdjmchattie: I don't know if this is what's happening here, but this is usual Rails behaviour.  Save returning false also sets errors on a list of errors on the object.  If the `new` page is set up correctly, it should render any values in that error list so that the first time you go to the page, it doesn't show any.  And if a previous attempt occurred, you see the list of issues generated last time a save was attempted.

app/views/admin/tag_sets/new.html.erb
```diff
@@ -0,0 +1,26 @@
+
+<div class="page-header"><h1>New Tag Set</h1></div>
+
+<%= form_for @tag_set, url: admin_tag_sets_path do |f| %>
+  <%= render_error_messages(@tag_set) %>
```
sdjmchattie, line 5: @yoldas and @dasunpubudumal isn't this where the errors from a failed save attempt appear?

dasunpubudumal: oh yeah apparently this is declared in `application_helper.rb` and adds a `div` with errors. All good!

app/controllers/admin/tag_sets_controller.rb
yoldas: Do we need to add anything to errors collection explicitly or does it just add it automatically?

yoldas: Right, save will run validations. They will add the errors. I think it is OK then.

app/views/admin/tag_sets/index.html.erb
```diff
@@ -0,0 +1,26 @@
+
+<div class="page-header"><h1>Listing Tag Sets</h1></div>
+
+<table class="table">
+  <thead>
+    <th>Name</th>
+    <th>Tag group (i7)</th>
+    <th>Tag group (i5)</th>
```
yoldas, line 8: In later stories, to unify single index and dual index tag sets through the API from Limber, we will need to allow a single tag set with one tag group. I am not sure (i7) and (i5) naming would be correct in this case. @KatyTaylor Do you think this naming is OK? That is, calling the first tag group as (i7) and the second as (i5) always?

app/views/admin/tag_sets/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<dl>
+  <dt>Name</dt><dd><%= @tag_set.name %></dd>
+  <dt>Tag group (i7)</dt><dd><%= link_to @tag_set.tag_group.name, @tag_set.tag_group %></dd>
+  <dt>Tag group (i5)</dt><dd><%= @tag_set.tag2_group ? (link_to @tag_set.tag2_group&.name, @tag_set.tag2_group) : 'None'%></dd>
+</dl>
+
+<br>
+<% if can? :create, TagLayoutTemplate %>
+  <%= link_to 'Create a new tag layout template from this tag set', new_tag_layout_template_path(tag_group_id: @tag_set.tag_group.id, tag2_group_id: @tag_set.tag2_group&.id) %>
+<br>
```
yoldas, line 15: I have added this to criteria to make creating tag layout templates automatically by selecting the correct tag groups in the target page. Do we have a sidebar link to access the readonly view of tag sets? Note that this is different from managing (creating) tag sets.

spec/controllers/admin/tag_sets_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  context 'when a non-admin' do
+    before { session[:user] = create :user }
+
+    describe '#index' do
+      it 'redirects' do
+        get :index
+        expect(response).to redirect_to('/')
+      end
+    end
```
yoldas, line 51: Isn't it OK to view them by non-admin users?

yoldas (commented):
Another round of review with some questions.

app/resources/api/v2/tag_set_resource.rb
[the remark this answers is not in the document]
yoldas: Adding an API end point should not be necessary for this story I think. If you are adding one, please consider the story https://github.com/sanger/sequencescape/issues/4238 , which is On Hold at the moment, along with the Limber UI story, and not prioritised. You can see all TagSet related story at References section of the description of https://github.com/sanger/sequencescape/issues/4160 .

app/resources/api/v2/tag_set_resource.rb
BenTopping: API will be covered by another story so I am going to remove resource references from this one.

spec/controllers/admin/tag_sets_controller_spec.rb
BenTopping: Yeah it should show the new template if the params are bad or save fails, ill add a test.

app/views/admin/tag_sets/show.html.erb
BenTopping: Same reply as below, added a link on home page and made show and index non-admin.

sdjmchattie (approved):

app/views/admin/tag_sets/index.html.erb
yoldas: I have sent an email about this but it does not stop getting the story to UAT.

yoldas (approved):

PR #4293, opened 2024-08-22
Y24 111 set default values to null

sabrine33:


app/views/receptacles/new_request.html.erb
```diff
@@ -9,7 +9,7 @@
     <% instance_variable_or_id_param(:request_type) do |field_name, value| %>
       <div class="form-group  col-xs-12">
         <%= label_tag(field_name, 'Request type', class: 'col-md-4') %>
```
BenTopping, line 11: Optional but it might be good to add the 'required' class here as that adds a star indicator to the users that the label is required.

BenTopping (approved):
Looks good, one optional comment.

app/views/receptacles/new_request.html.erb
sabrine33: Thank you Ben, I don't think they are really required fields, but also I suppose it depends of the type of the request.

app/views/receptacles/new_request.html.erb
```diff
@@ -9,7 +9,7 @@
...
-        <%= select_field_sorted_by_name(field_name, @request_types, value, can_edit) %>
+        <%= select_field_sorted_by_name(field_name, @request_types, value, can_edit, prompt: 'Select a request type', required: true) %>
```
yoldas, line 12: Doesn't this change the default values of all fields? The story is asking for “request additional sequencing” page, the request type, read length and flowcell.

app/views/shared/metadata/edit/_request.html.erb
```diff
@@ -5,7 +5,7 @@
     <%- request.request_metadata.field_infos.each do |field_info| %>
       <!-- todo unify with how submission use field info-->
       <%- if field_info.selection %>
-        <%= metadata_fields.select(field_info.key, field_info.selection) %>
+        <%= metadata_fields.select(field_info.key, field_info.selection, { prompt: "Select a #{field_info.key.to_s.gsub( '_', ' ')}"  }) %>
```
yoldas, line 8: I meant to comment on this one. Doesn't this change the default values of all fields? The story is asking for request type, read length and flowcell.

spec/features/assets/asset_submission_spec.rb
```diff
@@ -15,6 +15,34 @@
...
+
+    describe 'when the user selects a request type' do
+      before { select 'Request Type 1', from: 'Request type' }
+
+      it 'does not set flowcell type to default value' do
+        expect(page).to have_select('Flowcell type', selected: 'Select a requested flowcell type')
+      end
+
+      it 'does not set read length to default value' do
+        expect(page).to have_select('Read length', selected: 'Select a read length')
```
yoldas, line 41: What was causing 150 numeric value on the screenshot in the story?

yoldas: What was causing S2 for flow cell type on the screenshot in the story ?

app/views/receptacles/new_request.html.erb
andrewsparkes: It says in the story request type is required:
_The request type box is mandatory for all requests, and sequencing cannot be requested without this being selected_

andrewsparkes: as are Read length and Flow cell type

spec/features/assets/asset_submission_spec.rb
```diff
@@ -15,6 +15,34 @@
     create(request_factory, study: study, project: project, asset: asset, request_type: original_request_type)
   end
 
+  describe 'The request form does not set default values' do
```
andrewsparkes, line 18: I think also add tests on submit to check the 3 fields must be set with values to continue, or error in GUI if not.
The story is saying they don't want default values set because people are not checking and submitting with the defaults when not appropriate, but that the user must set values.

andrewsparkes (commented):
I don't think this is currently matching the requirements? Fields are required.

PR #4312, opened 2024-08-29
Y24-190-3 Support Limber with API v2 StateChange

### LINKED ISSUE #4244: Y24-190-3: Implement V2 endpoints to support Limber's Labware Creators

sdjmchattie:
Closes #4244 

- Update endpoint for StateChange to be consistent with API v1.


app/resources/api/v2/state_change_resource.rb
```diff
@@ -15,14 +15,87 @@ module V2
...
+      # @!attribute [rw] reason
+      #   @return [String] The previous state of the target before this state change.
+      attribute :reason
+
+      # @!attribute [rw] target_state
+      #   @return [String] The state of the target after this state change.
+      #   @note This attribute is required.
       attribute :target_state
-      attribute :created_at
-      attribute :updated_at
```
sdjmchattie, line 23: @KatyTaylor you added these about 5 years ago.  Other endpoints don't list the created at and updated at dates for resources.  I'm happy to leave these in if you think they're useful, but I wanted to align with the v1 endpoint as much as possible without upsetting anything.

app/resources/api/v2/state_change_resource.rb
```diff
@@ -15,14 +15,87 @@ module V2
...
 
-      has_one :labware
```
sdjmchattie, line 25: This moved down the file and became "target".  I don't know if this causes issues with what was using this endpoint before.

app/models/state_change.rb
```diff
@@ -53,7 +53,7 @@ def record_current_state_of_target
   end
 
   def state_changer
-    target.state_changer
+    target&.state_changer
```
BenTopping, line 56: A target shouldn't able to be nil based on the validations. Was there a specific case this was needed to be conditionally chained?

app/resources/api/v2/state_change_resource.rb
```diff
@@ -15,14 +15,87 @@ module V2
...
+      # @!attribute [w] target_uuid
+      #   This is declared for convenience where the target is not available to set as a relationship.
+      #   Setting this attribute alongside the `target` relationship will prefer the relationship value.
+      #   @param value [String] The UUID of the target labware this state change applies to.
+      #   @return [Void]
+      #   @see #target
+      attribute :target_uuid
+
+      def target_uuid=(value)
```
BenTopping, line 53: Does this belong in here or in the model?

BenTopping (commented):
Looks good, just a few questions.

app/models/state_change.rb
sdjmchattie: This method is called by a validator and causes the model to throw an exception that causes a 500 error when using the API endpoint.  The target cannot be `nil` and the validations pass, but the validations all get run before we can determine that the model is broken.

app/resources/api/v2/state_change_resource.rb
sdjmchattie: This is how we've created other UUID based setters on the API.  Resources let you override the getter and/or setter for attributes, and this one is here to set the actual target on the model.  If we put this in the model, it would be to support the API and I don't think that's the right thing to do.

To be honest, users of the API should really be using the relationships to assign other resources as related, but Limber is very heavily architected around passing the UUID to the v1 API and it would create another hurdle to overcome if we don't have v2 API support the same approach.

app/models/state_change.rb
sdjmchattie: Can we do that?  It would probably make this code less confusing.

sdjmchattie: The validation line currently just says:

```
validates :state_changer, presence: { message: 'target does not have a configured state changer' }
```

BenTopping: Yep we can! https://guides.rubyonrails.org/active_record_validations.html#using-a-symbol-with-if-and-unless

BenTopping (approved):

PR #4314, opened 2024-09-02
Y24-098 - Tube rack banking purpose

BenTopping:

- Adds API support for TubeRack::Purpose

Linked to: https://github.com/sanger/limber/pull/1883

### DIFF
--- a/spec/requests/api/v2/tube_rack_purposes_spec.rb
+++ b/spec/requests/api/v2/tube_rack_purposes_spec.rb
@@ -1,0 +1,203 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require './spec/requests/api/v2/shared_examples/api_key_authenticatable'
+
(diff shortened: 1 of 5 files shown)


spec/requests/api/v2/tube_rack_purposes_spec.rb
```diff
@@ -0,0 +1,203 @@
+# frozen_string_literal: true
```
andrewsparkes, line 1: Not sure this file should be in the spec/requests directory?

andrewsparkes: huh, all the api v2 stuff is in there, I wonder why...

andrewsparkes (approved):
Looks good, thorough tests as well thanks.

General comment about my surprise the specs for api v2 are under requests. Maybe Stuart can refactor the location as part of getting rid of api v1

PR #4315, opened 2024-09-02
Y24-190 Milestone 1: API v1 -> v2 Migration

sdjmchattie:
Milestone 1 for https://github.com/sanger/limber/issues/1772

- Add new/updated API v2 endpoints for:
  - BarcodePrinter
  - CustomMetadatumCollection
  - StateChange
  - TagLayoutTemplate
  - Transfer
  - TransferTemplate
- Minor tweaks to development tool config such as pry.
- Updated documentation stubs for API resources.

Note that these changes were tested by the Integration Suite after deployment to UAT: https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/pipelines/300247/builds


app/resources/api/v2/transfer_template_resource.rb
```diff
@@ -30,6 +30,12 @@ class TransferTemplateResource < BaseResource
       # @!attribute [r] uuid
       #   @return [String] the UUID of the transfer template.
       attribute :uuid, readonly: true
+
+      # @!method filter_uuid
+      #   Filter the transfer templates by UUID.
+      #   @example URL with UUID filter
+      #     https://sequencescape.psd.sanger.ac.uk/api/v2/transfer_templates?filter[uuid]=12345678-1234-1234-1234-123456789012
```
harrietc52, line 37: I think we try to avoid putting the repo urls on GitHub, maybe just put `https://<sequencescape>/api/v2/transfer_templates?filter[uuid]=12345678-1234-1234-1234-123456789012`

sdjmchattie: Good idea.  It's a bit long as it is anyway.

app/resources/api/v2/user_resource.rb
```diff
@@ -15,22 +15,40 @@ module V2
...
+      ###
+
+      # @!method user_code
+      #   A filter to return only users with the given user code.
+      #   The given user code will be compared with the swipecodes and barcodes for users until matches are found.
       filter :user_code, apply: lambda { |records, value, _options| records.with_user_code(*value) }
+
+      # @!method uuid
+      #   A filter to return only users with the given UUID.
+      filter :uuid, apply: lambda { |records, value, _options| records.with_uuid(*value) }
```
harrietc52, line 51: Might this Resource want the `creatable_fields` to not include `uuid` too, if it is created by the system on creation?

spec/requests/api/v2/shared_examples/invalid_post_requests.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  it 'responds with unprocessable_entity' do
+    expect(response).to have_http_status(:unprocessable_entity)
+  end
+
+  it 'specifies which relationship must exist' do
+    expect(json.dig('errors', 0, 'detail')).to eq("#{missing_relationship} - must exist")
+  end
+end
```
harrietc52, line 51: Nice 😄

spec/requests/api/v2/transfers/transfers_spec.rb
```diff
@@ -0,0 +1,241 @@
...
+        expect(json.dig('data', 'type')).to eq('transfers')
+        expect(json.dig('data', 'attributes', 'uuid')).to be_present
+        expect(json.dig('data', 'attributes', 'user_uuid')).to eq(user.uuid)
+        expect(json.dig('data', 'attributes', 'source_uuid')).to eq(source.uuid)
+        expect(json.dig('data', 'attributes', 'destination_uuid')).to eq(destination.uuid)
+        expect(json.dig('data', 'attributes', 'transfers')).to eq(transfer_template.transfers)
+      end
+    end
+
+    context 'when providing an invalid payload' do
```
harrietc52, line 114: Could you use your shared examples here `it_behaves_like 'a POST request with a missing/ wrong attribute’` etc just because the below shared example `it must not contain that attribute` is only being used once which seems similar to those other shared examples but I might be wrong! Just doing a quick scan

spec/resources/api/v2/barcode_printer_resource_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(resource).to have_attribute :barcode_type
+  end
+
+  # Updatable fields
+  it 'allows updating of read-write fields', :aggregate_failures do
+    expect(resource).to have_updatable_field :name
+  end
+
+  it 'disallows updating of read only fields', :aggregate_failures do
+    expect(resource).not_to have_updatable_field :uuid
```
harrietc52, line 26: To not have updatable field `id` too maybe?

spec/resources/api/v2/transfers/transfer_resource_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(resource).not_to have_attribute :transfer_template_uuid
+  end
+
+  # Updatable fields
+  it 'allows updating of read-write fields', :aggregate_failures do
+    expect(resource).to have_updatable_field :source_uuid
+    expect(resource).to have_updatable_field :destination_uuid
+    expect(resource).to have_updatable_field :user_uuid
+    expect(resource).to have_updatable_field :transfers
+    expect(resource).to have_updatable_field :transfer_template_uuid
```
harrietc52, line 28: Quite like the tests of `.not_to have updatable_field` if you wanted to add any here aswell

harrietc52 (approved):
Minor comments, feel free to ignore. Nice work, as always! 👍

spec/factories/transfers.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  factory(:transfer_template) do
+    sequence(:name) { |n| "Transfer Template #{n}" }
+    transfer_class_name { 'Transfer::BetweenPlates' }
+    transfers { { 'A1' => 'A1', 'B1' => 'B1' } }
+
+    factory(:pooling_transfer_template) do
+      transfer_class_name { 'Transfer::BetweenPlatesBySubmission' }
+      transfers { nil }
```
andrewsparkes, line 33: Are these transfers nil because that is correct or because you don't know what they should be? Some comments and or a todo in either case would help.

spec/requests/api/v2/custom_metadatum_collections_spec.rb
```diff
@@ -8,105 +8,282 @@
 
   it_behaves_like 'ApiKeyAuthenticatable'
 
-  context 'with multiple custom_metadatum_collections' do
+  context 'with multiple collection' do
```
andrewsparkes, line 11: collection**s**

app/resources/api/v2/user_resource.rb
sdjmchattie: Bearing in mind that I didn't actually create this endpoint, I only added documentation to it and added the UUID filter.

I have been preferring to use `creatable_fields` in more recent resources as it gives nicer error messages than using `readonly: true` on the attribute itself.  You get an error with a description that you cannot include the attribute instead of getting an error 500 from the server when you try to set an attribute that it read only.  I could go through and update all the existing endpoints, but it will mean more messing around with tests so I've avoided doing it for existing resources.

spec/requests/api/v2/shared_examples/invalid_post_requests.rb
sdjmchattie: I think they can be a double-edged sword.  I've seen them used in a nested way in Limber where one shared examples calls two other shared examples and each of those call two others.  Before you know it, you have no idea what the setup is for a test and they're failing because of a change you made.  You have no idea how to parse where the error is coming from.  But thanks -- this felt like a good use for them.

spec/resources/api/v2/user_resource_spec.rb
```diff
@@ -4,17 +4,24 @@
 require './app/resources/api/v2/user_resource'
 
 RSpec.describe Api::V2::UserResource, type: :resource do
-  subject { described_class.new(resource_model, {}) }
+  subject(:resource) { described_class.new(resource_model, {}) }
```
andrewsparkes, line 7: how does this work? you define the `subject(:resource)` but then it's not referred to in the `it `sections below. Does `is_expected` magically use the subject?

spec/resources/api/v2/barcode_printer_resource_spec.rb
sdjmchattie: I stopped adding `:id` to these simply because I could also check that `:banana` is not an `updatable_field`.  `id` doesn't form part of the resource and so testing for its lack of existence didn't feel like it was adding value to these tests.  These tests are quite poor anyway.  They only test the properties of the `attribute` entries like `readonly`.  They don't confirm whether the resource is immutable and they don't know how to parse the things like `fetchable_fields` and `creatable_fields` methods that are part of the library we're using.  The best tests are in the `requests` spec folder.

andrewsparkes (approved):
Looks good I think, well done. I went a bit blind by the end. A few minor comments only.

spec/resources/api/v2/transfers/transfer_resource_spec.rb
sdjmchattie: Same as above 😄

app/controllers/api/v2/transfers/transfers_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+            # Remove the UUID of the transfer template from the attributes.
+            attributes.delete(:transfer_template_uuid)
+          end
+
+          super(operations)
+        end
+      end
+
+      class BetweenPlateAndTubesController < JSONAPI::ResourceController
```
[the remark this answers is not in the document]
sdjmchattie: Yeah, this is me dealing with polymorphism.  You might want to only GET the BetweenPlates transfers and in order to have a route in the `routes` file, you must have a corresponding controller.  Although they're all implemented in the base class, this gives you methods that handle creating a resource, patching a resource, etc.

spec/factories/transfers.rb
sdjmchattie: Yes fair enough.  I've added a comment.  Basically some transfer types do not have the transfers in the template.  These in particular expect the transfers to be passed in the creation request as they are defined by the user in the Limber UI I believe.  I only have a vague understanding of how it all works.  The production database shows these as nil in the templates.

spec/resources/api/v2/user_resource_spec.rb
sdjmchattie: Yeah it's some sort of rspec magic.  `is_expected` seems to know it needs to use the thing defined as the `subject`.  I didn't know this pattern, but I saw it already existed in our tests.  The only change here was to name the subject, which Rubocop insists on when you remove the exception to the rule like I did here in the rubocop config file.

app/controllers/api/v2/transfers/transfers_controller.rb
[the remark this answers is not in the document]
andrewsparkes: That's useful info, could you add comments to that affect? Just above this BetweenPlateAndTubesController to say 'the following controllers are ...' etc.

app/resources/api/v2/user_resource.rb
harrietc52: Sure, no problem. I thought too it looks much nicer, didn’t know the details but that is definitely better handling. Each resource can be updated as and when they are touched 👍

PR #4316, opened 2024-09-03
Add temporary fix for cucumber tests

yoldas:
Closes #

- I have applied a workaround for the Sequencescape cucumber tests on CI until the Chrome 128 through Selenium is fixed. I got close to fix almost all the tests but could not get JS running properly with headless=new option and disable-search-engine-choice-screen combination. I used the following info and set it to `headless=old` to get the tests pass.

https://github.com/SeleniumHQ/selenium/issues/14438

"Driving Chrome 128 through Selenium on a machine without Chrome installed fails connecting to Chrome, be it headless or not, unless asking for headless=old. The disable-search-engine-choice-screen argument does not change anything about this.

It works only if Chrome is directly installed on the machine when not setting --headless=old."

- The following is explanation of values (new, old, true) passed to headless mode (vs headful):

https://developer.chrome.com/docs/chromium/new-headless

Note that it says the following  in that page but it seems to be still supported:

"We intend to remove the **old** Headless from the Chrome binary and stop supporting this mode"

- The following is the starting point of my investigation, posted by team members in Slack discussion.

https://github.com/SeleniumHQ/selenium/issues/14453

- I have also enabled WIP flags for cucumber environment. It is not directly related to the CI issues but I think it is good to have.

### DIFF
--- a/config/environments/cucumber.rb
+++ b/config/environments/cucumber.rb
@@ -45,4 +45,7 @@ Rails.application.configure do
     ENV_JAVA['https.proxyHost'] = nil
     ENV_JAVA['https.proxyPort'] = nil
   end
+
+  # load WIP features flag
+  config.deploy_wip_pipelines = true
 end
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@
 source 'https://rubygems.org'
 
 group :default do
-  gem 'bootsnap', '1.18.3'
+  gem 'bootsnap'
   gem 'rails', '~> 6.1.7'
 
   # Fix incompatibility with between Ruby 3.1 and Psych 4 (used for yaml)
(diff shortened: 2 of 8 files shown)


config/environments/cucumber.rb
```diff
@@ -45,4 +45,7 @@
     ENV_JAVA['https.proxyHost'] = nil
     ENV_JAVA['https.proxyPort'] = nil
   end
+
+  # load WIP features flag
+  config.deploy_wip_pipelines = true
```
sdjmchattie, line 50: Do we want this?

sdjmchattie (approved):
Apart from the WIP flag, well done finding the responsible feature and the flag to work around it.

I don't know if we need the other flags or not but they look sensible. Are there other browsers we can turn to if Chrome does remove this mode?

config/environments/cucumber.rb
yoldas: It is enabled for test and development environments. I would like to align it with the cucumber environment as well.

sdjmchattie: Doesn't that mean that the Cucumber tests run when creating a production build also uses WIP pipelines?

seenanair (approved):
Great job 👏

config/environments/cucumber.rb
yoldas: Firefox is usually a go to browser for testing but I have not looked into that. It can be started as an alternative browser option. Every chrome and driver version we keep changing the options to initialise them to use through Selenium. It is supposed to be unifying access to different browsers.

PR #4324, opened 2024-09-05
Y24 111 - Add a required constraint to the flowcell type field for NovaSeq requests.

sabrine33:


KatyTaylor (approved):
Looks good I think. 
A few questions:
- Have you manually tested it in the UI? 
- Did you consider adding a unit test for `RequestOptionsValidator`, especially `illumina_htp_novaseq_request?`? (It doesn't look like this is covered in your `asset_submission_spec` test as you use an `instance_double`)

sabrine33:
> Looks good I think. A few questions:
> 
> * Have you manually tested it in the UI?
> * Did you consider adding a unit test for `RequestOptionsValidator`, especially `illumina_htp_novaseq_request?`? (It doesn't look like this is covered in your `asset_submission_spec` test as you use an `instance_double`)

Thank you for having a look 
1) Yes, I performed some manual test
2) I am using instance_double for RequestOptionsValidator class that is receiving a request of type novaseq, which I think it should cover this specific use case, having said that, I am not quite sure now, as I am still quite new to the language

PR #4327, opened 2024-09-06
Y24-190-3 Support Limber TagLayout creation

### LINKED ISSUE #4244: Y24-190-3: Implement V2 endpoints to support Limber's Labware Creators

sdjmchattie:
Closes #4244 

- Add endpoint for TagLayout retrieval and creation.
- Add a custom method to TagLayoutTemplate to be able to create a TagLayout from the template.


sdjmchattie:
I branched off of the work I did on BaitLibraryLayout which hasn't been merged yet so unfortunately those changes appear in this PR as well.

app/models/tag_layout.rb
```diff
@@ -12,9 +12,6 @@ class TagLayout < ApplicationRecord
 
   attr_accessor :tags_per_well
 
-  UnknownDirection = Struct.new(:direction)
-  UnknownWalking = Struct.new(:walking_by)
```
sdjmchattie, line 16: @yoldas and @andrewsparkes these are the changes I made to TagLayout to prevent the exception that was happening in Sequencescape and causing an error 500 response in the API.  I tried a few things in the end but this felt the cleanest.  Effectively I am replacing the useless structs with dummy algorithms whose only job is to give a `nil` direction and a `nil` walking_by when they are called upon via the delegations on lines 71 and 72.  That way, the validation can be done against those attribute names and the error message makes more sense.  Otherwise you have to validate something the API user didn't provide a value for and the error response says something like "direction_algorithm - must not be blank".  With this pattern the error is "direction - must define a valid algorithm" and it handles all values that aren't from the hashes of algorithms above.

app/models/tag_layout.rb
```diff
@@ -65,11 +72,11 @@ class TagLayout < ApplicationRecord
   delegate :walking_by, :walk_wells, :apply_tags, to: :walking_algorithm_helper
 
   def direction=(new_direction)
-    self.direction_algorithm = DIRECTION_ALGORITHMS.fetch(new_direction) { UnknownDirection.new(new_direction) }
+    self.direction_algorithm = DIRECTION_ALGORITHMS.fetch(new_direction, TagLayout::DummyDirectionModule)
```
BenTopping, line 75: Im not sure why direction and walking_by must exist in validation if we are dummying their values? Maybe so other validation can occur? Im not familiar with how TagLayout works so might be a misunderstanding on my part.

app/resources/api/v2/tag_layout_resource.rb
```diff
@@ -0,0 +1,157 @@
...
+      #   @return [String]
+      #   @note This attribute is required.
+      attribute :direction
+
+      # @!attribute [rw] initial_tag
+      #   An offset for the tag set indicating which tag to start with in the layout.
+      #   @return [Integer]
+      attribute :initial_tag
+
+      # @!attribute [w] target_uuid
```
BenTopping, line 33: plate_uuid

BenTopping (commented):
I'm not too familiar with TagLayout but the changes look sensible.

app/models/tag_layout.rb
sdjmchattie: I think it's using a very backwards approach to these values, but you seem to have a similar understanding to mine.  In this case:

- `direction` is assigned a value as an attribute.
- This setter method uses the value to set `self.direction_algorithm` using the attribute value.  If that isn't a valid algorithm, the dummy value is used (you'll see why in a later step)
- The validator on line 63 expects `direction` to be present, BUT
- At this point `direction` is delegated to `direction_algorithm_module` via line 71.
- That delegation tries to get the name of the direction module (which is either a valid module, or the Dummy module from earlier) and constantise its name.
- If we didn't use a Dummy module when the value was invalid, the constantize would generate an exception.
- The `direction` method is called on that module to do the validation.  So we have a method in the Dummy module that returns `nil` which fails the validation of the model.

Convoluted.  Something similar happens for `walking_by`.

app/resources/api/v2/tag_layout_resource.rb
sdjmchattie: D'oh

app/models/tag_layout.rb
BenTopping: Hmm yes rather convoluted.. I think because this story is only replicating the behaviour of the existing API the changes are fine. This does look like something that can be refactored at some point though.

BenTopping (approved):

PR #4331, opened 2024-09-09
Y24-055 Take 2

### LINKED ISSUE #4083: Y24-055 New plate manifest type in SS to reduce the number of incorrect tags
**User story**
As an SSR, I would like there to be a new library plate manifest type that allows the user to select the tag group used in the same manner as the Chromium library plate manifest - with a drop down menu for selecting the plate and wells, rather than having to type out the oligo sequences, in order to avoid typos in the oligo sequence. 

N.B. We can't use the Chromium library plate manifest, because it only works with a particular type of single index 10x tag set in which there are 4 tags per well.

Richard Clark
Conor Parks

Richard Clark
Conor Parks

**Acceptance criteria**
- [x] A new manifest type is available in SS, called "Library plate - dual index tag plate".  
- [x] Above manifest type allows users to add sample tag information via a drop down menu, like the chromium type manifest.
- [x] The existing manifest type "Library type" is renamed to "Library plate - custom tags".
- [x] In the new manifest type, only tag groups relevant to this manifest type are listed in the dropdown (TBC - how to identify these - is it all dual index ones?)
- [x] In the existing Chromium library plate manifest type, only tag plates relevant to this manifest type are listed in the dropdown - these are single index 10x ones that contain 4 tags per well. It currently filters the list using 'adapter type = Chromium', but this is too permissive because that includes Chromium dual index tag sets. Need to investigate:
    - Whether to create more fine-grained adapter types and update the data (is adapter type used for anything else?)
    - Whether to create some other way to filter the tag groups (UPDATE: spin out story https://github.com/sanger/sequencescape/issues/4304)

**Dependencies**
This will be used for dual index tag sets, where i7 and i5 oligos are together in the same well. At time of writing, i7 and i5 tag sets are created separately in Sequencescape, and there is no link between two related sets (apart from the name). 

We could do a quick version where users select the i7 and i5 tag groups separately, however users would prefer to only have to select the tag set once.

This story is blocked by the following research story, to figure out how to implement this:
- https://github.com/sanger/sequencescape/issues/4160
- #4237

harrietc52:
Closes #4083

This is the second attempt at merging in Y24-055 code to `develop`as I merged in the original before it was ready for release. 

Original PR with notes is here https://github.com/sanger/sequencescape/pull/4292


yoldas (approved):
Same as before.

PR #4343, opened 2024-09-13
Remove checks for not raising specific errors

sdjmchattie:
RSpec warns about these while the tests are running and for good reason. We don't expect to raise ANY error, but the test only confirms that a single specific error won't be raised.

### DIFF
--- a/spec/models/study_spec.rb
+++ b/spec/models/study_spec.rb
@@ -663,10 +663,11 @@ RSpec.describe Study do
 
         it 'does not error if the data_release_timing is invalid' do
           study.study_metadata.data_release_timing = 'never'
-          expect { study.save! }.not_to raise_error(
-            ActiveRecord::RecordInvalid,
-            /Study metadata data release timing is not included in the list/
-          )
+
+          # There will be errors, but they shouldn't include anything about data release timing.
+          expect(study.save).to be_falsey
+          expect(study.errors).not_to be_nil
+          expect(study.errors.full_messages).not_to include(/data release timing/)
         end
       end
     end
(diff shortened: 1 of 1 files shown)


KatyTaylor (commented):
~So what's the change? Why will it now raise an error where it didn't before?~

~Edit: oh sorry, read your description! It doesn't raise an error, but you're just changing it for good practice. Seems fine.~

Won't have time to chat before I go on holiday, but here are my more up to date thoughts:

It currently seems to me like the old tests are better. 

```
expect { study.save! }.not_to raise_error(
    ActiveRecord::RecordInvalid,
    /Study metadata data release timing is not included in the list
)
```
This is a very specific unit test, testing that with this specific data setup, and feature flag state, saving the record does not raise this particular error. It doesn't care if it raises other errors or not - we're only testing this specific validation.

The alternative:
```
# There will be errors, but they shouldn't include anything about data release timing.
expect(study.save).to be_falsey
expect(study.errors).not_to be_nil
expect(study.errors.full_messages).not_to include(/data release timing/)
```

I don't like this as much because it is asserting that there are other errors - but I don't want to care if there are or not in this test. We seem to be accidentally testing other validation. If someone updates the other validation so that there are no longer errors, we'd have to change this test to say `.to be_falsey` instead - even though we didn't change the feature this test cares about.

Very happy to discuss! Unfortunately I'm away until the 30th (but this doesn't seem urgent).

Thanks :)

spec/models/study_spec.rb
```diff
@@ -663,10 +663,11 @@
 
         it 'does not error if the data_release_timing is invalid' do
```
dasunpubudumal, line 664: hmm 🤔 Does it assert for raising _any_ error or a specific error? It looks like it's trying to check if the record is valid or not?

sdjmchattie:
As discussed, I've moved the tests away from trying to identify raised errors and instead focusses on messages being generated as error/validation issues.  I don't think `!` update and save method calls make sense in tests as there are so many other reasons why things might throw an error.

harrietc52 (approved):

PR #4353, opened 2024-09-18
4309 bug exception 29082024 at 1042

wendyyang:
Closes #

- 

exception errors: undefined method `labware_id' for nil:NilClass

So add checks if request's asset_id, asset's labware_id is nil to avoid call method on undefined object

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -34,20 +34,14 @@ class Comment < ApplicationRecord
   # @return [Hash] Hash of counts indexed by request_id
   #
   def self.counts_for_requests(requests) # rubocop:todo Metrics/AbcSize
-    begin
-      all_commentables = requests.flat_map do |request|
-        [request, request.try(:asset), request.try(:asset).try(:labware)]    
-      end    
-      counts = where(commentable: all_commentables.compact).group(:commentable_type, :commentable_id).count
+    all_commentables = requests.flat_map { |request| [request, request.try(:asset), request.try(:asset).try(:labware)] }
+    counts = where(commentable: all_commentables.compact).group(:commentable_type, :commentable_id).count
 
-      requests.each_with_object({}) do |request, counter_cache|
-        request_count = counts.fetch(['Request', request.id], 0)
-        receptacle_count = counts.fetch(['Receptacle', request.try(:asset_id)], 0)
-        labware_count = counts.fetch(['Labware', request.try(:asset).try(:labware_id)], 0) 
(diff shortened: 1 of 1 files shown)


app/models/comment.rb
```diff
@@ -34,12 +34,13 @@ def self.counts_for(commentables)
   # @return [Hash] Hash of counts indexed by request_id
   #
   def self.counts_for_requests(requests) # rubocop:todo Metrics/AbcSize
-    all_commentables = requests.flat_map { |request| [request, request.asset, request.asset&.labware] }
+    all_commentables = requests.flat_map { |request| [request, request.try(:asset), request.try(:asset).try(:labware)] }
```
sdjmchattie, line 37: Why use `.try` instead of `&.`?  `&.` protects against the object to its left being `nil` and short circuits to give a `nil` value for the whole method chain.  `.try` attempts to call a method on the object and protects against that method not being implemented.  I think you want to be using `&.` here.  I also assume that having `nil`s all over isn't going to cause issues downstream?

sdjmchattie:
I think we should probably have a test to reproduce the issue so that we can confirm the issue no longer occurs.

wendyyang:
> I think we should probably have a test to reproduce the issue so that we can confirm the issue no longer occurs.

make sense, that is a good reason to have a test.

spec/models/comment_spec.rb
```diff
@@ -18,17 +18,29 @@
...
   describe '#counts_for_requests' do
-    let(:request) { create :sequencing_request, asset: tube }
+    let(:request1) { create :sequencing_request, asset: tube }
     let(:tube) { create :multiplexed_library_tube }
 
-    before do
-      create :comment, commentable: tube, description: 'An excellent tube'
-      create :comment, commentable: tube.receptacle, description: 'A good receptacle'
-      create :comment, commentable: request, description: 'A reasonable request'
+    let(:request2) { create :sequencing_request }
```
sdjmchattie, line 24: You don't need to create a request with a different name this way.  `context` blocks will override values just inside their block if you put the `let` inside there.  So you can leave line 21 called `request` and move this line (24) inside the `context` block starting on line 38 and also call it `request`.  The value will be overwritten, but only for the tests inside that `context`.

That's the whole point of `context`s, to set the fixture values up just for the tests inside its block.

sdjmchattie (changes requested):
Just need to adjust the way you define fixture values for the new test.

sdjmchattie (approved):

PR #4371, opened 2024-09-24
Y24-133 - Upgrade to Zeitwerk autoloader

### LINKED ISSUE #4141: Update Rails 6.1 autoloader to Zeitwerk
**Describe the Housekeeping**

As a developer, I need to complete Rails 6.1 migration by changing Rails default's autoloader to [Zeitwerk](https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html). 

We started Rails upgrade from 5.0 to 6.1 in https://github.com/sanger/sequencescape/issues/4006, and began upgrading the defaults in https://github.com/sanger/sequencescape/issues/4005. However, the defaults upgrade in https://github.com/sanger/sequencescape/issues/4005 was done incrementally, by explicitly defining [`config.autoloader = :classic`](https://github.com/sanger/sequencescape/pull/4125/files#diff-c1fd91cb1911a0512578b99f657554526f3e1421decdb9e908712beab57e10f9). 

This story entails updating the autoloader to `:zeitwerk`, and updating the file structure to match [Zeitwerk requirements](https://github.com/fxn/zeitwerk#file-structure). 

The process can start with a [`zeitwerk:check`](https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html#zeitwerk-check) and update application as necessary.

**Blocking issues**

1. #4138 
2. #4139 

**Additional context**

Related stories:

1. #4006 
3. #4005

BenTopping:
Closes #4141 

- Changes rails autoloader to zeitwerk.
- Removes _most_ references to require_dependency (no longer required with zeitwerk). Some edge cases remain where this is still required.
- Moves heron and warren directories to app/models.
- Updates various class names and references to be zeitwerk compatible.
- Reconfigures lib/label_printer to be zetiwerk compatible.
- Converts cron jobs (volume_check.rb and import_fluidigm_data.rb) to modules to be zeitwerk compatible.

#### Notes for reviewers
- `require_dependency` references have been removed as per: https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html#delete-require-dependency-calls
- Sample Metadata class has been moved to Sample class (where it is initially defined anyway with the `has_metadata` method) because it is doing some order-dependant loading which was not working with zeitwerk.
- `app/controllers/assets` has been removed as it appears it is no longer in use anywhere.
- `lib/bm_plate_creation.rb` has been removed as it does not appear to be useful anymore.
- `lib/exceptions.rb` has been removed as it appears to be redundant.
- `lib/deployed_version.rb` contains the module Deployed and thus the renaming. (Thanks @StephenHulme) 

#### Questions
- Are the cron jobs still used?
- What is lib/informatics?


app/models/heron/factories/sample.rb
```diff
@@ -105,7 +105,7 @@ def params_for_sample_creation
       end
 
       def params_for_aliquot_creation
-        { sample: sample, study: study }.merge(@params.dig(:aliquot) || {})
+        { sample: sample, study: study }.merge(@params[:aliquot] || {})
```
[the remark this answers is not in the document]
BenTopping: These Heron changes are a result of rubocop no longer ignoring the file (because of its destination changes). I believe `dig(:key)` and `[:key]` both return nil if they don't exist. See [rubocop's reasoning here](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/SingleArgumentDig).

lib/label_printer/label/asset_plate_double.rb
```diff
@@ -8,6 +8,7 @@ class AssetPlateDouble < BasePlateDouble
       attr_reader :plates
 
       def initialize(plates)
+        super()
```
stevieing, line 11: ADVISORY: Using ActiveModel::Model would negate the need to do this.

lib/label_printer/label/asset_tube.rb
```diff
@@ -5,6 +5,7 @@ class AssetTube < BaseTube
       attr_reader :tubes
 
       def initialize(tubes)
+        super()
```
stevieing, line 8: ADVISORY: Using ActiveModel::Model would negate the need to do this.

lib/label_printer/label/multiple_labels.rb
```diff
@@ -19,7 +19,7 @@ def create_labels
         end
       end
 
-      def count
+      def count # rubocop:disable Lint/DuplicateMethods
```
stevieing, line 22: ADVISORY: This is a bit of a code smell??

stevieing (commented):
Just a couple of advisories which I am not sure about. Happy to leave as there is enough going on.

Some of the require and require_relative are a bit strange but at this stage not worth worrying about.

Once this is completed and merged to production we can rebase the rails 7 upgrade and decide whether we need another technical debt story to finish it off.

stevieing (approved):

lib/label_printer/label/multiple_labels.rb
BenTopping: Yeah I was forgot to come back and fix this, addressed in https://github.com/sanger/sequencescape/pull/4371/commits/ca3c651b25bab117af2f3be578435ba284b07997

BenTopping:
Ensure deployment occurs on deployment project zeitwerk branch and that it is merged upon this works merge

PR #4378, opened 2024-09-25
Y24-190-3: Support Limber with TransferRequestCollections

sdjmchattie:

- Add an endpoint for TransferRequestCollections.
- Add tests for the endpoint.


spec/factories/transfer_request_collections.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :transfer_request_collection do
+    transient do
+      target_tube_count { 1 }
+      target_assets do
+        Array.new(target_tube_count) { create(:receptacle, labware: create(:stock_multiplexed_library_tube)) }
```
BenTopping, line 8: Do the target assets need to be receptacles? Can they be labware directly?

As a sidenote I think another way of writing this would be:
`target_assets { create_list(:receptacle, target_tube_count, labware: create(:stock_multiplexed_library_tube) }`

spec/factories/receptacles.rb
```diff
@@ -2,5 +2,6 @@
 
 FactoryBot.define do
   factory :receptacle do
+    labware { nil }
```
BenTopping, line 5: I suspect this wasn't required before because its normally referenced the other way round where an asset as a receptacle?

BenTopping (commented):
Looks good, I couldn't really compare to the v1 version by quickly looking at as its quite confusing. Just one question about the target_assets.

spec/factories/transfer_request_collections.rb
sdjmchattie: Taken from the TransferRequest model: `# target_asset and asset are both Receptacle objects, and are the source and target of the transfer request.`.  I used a real TransferRequestCollection to determine how to put one of these together.

I will use your alternative way of defining the list though.  I did know about this method, but I originally created a set of tubes in a separate transient and plucked them one at a time from that list.  Until I moved the `create` inline here and then I should have gone back to `create_list`.

spec/factories/receptacles.rb
sdjmchattie: I don't think this was "necessary" exactly as it works without it, but I felt that it gave a hint to users of the factory that you can assign a labware if you want to.  Which is what I wanted to do.

spec/factories/transfer_request_collections.rb
BenTopping: Cool thanks for the explanation.

BenTopping (approved):

Issue #4126, opened 2024-05-28
Y24-111	Set default fields to null when requesting additional sequencing


neilsycamore:
As team lead (SSRs) currently, when opening the “request additional sequencing” page, the request type, read length and flowcell are filled in by default. These are filled in differently for a Novaseq 6000 request compared to a NovaSeqX request. This can lead to errors in requesting sequencing if the SSR does not change these default selections.

We request that the “Request type”, “Read length” and “Flowcell type” are NOT filled in by default when requesting additional sequencing.

All other boxes on this page can continue to be populated as they are currently.

![RT803569](https://github.com/sanger/sequencescape/assets/299768/b0c16861-e172-4070-986f-769f1e16e9ff)
Acceptance criteria:

·       When the “request additional sequencing” page is opened in SS, the “Request type” “Read length” and “Flowcell type” boxes are empty, but have the same drop down options as currently.

·       The request type box is mandatory for all requests, and sequencing cannot be requested without this being selected – an error message should be displayed “Request type must be selected.”

·       The read length box is mandatory for all sequencing platforms, and sequencing cannot be requested without this being selected – an error message should be displayed “Read length must be selected”.

·       The flowcell type box is mandatory for all NovaSeq sequencing platforms, and sequencing cannot be requested without this being selected – an error message should be displayed “Flowcell type must be selected”.

The SSR stakeholders for this story would be myself and Siobhan (ec20 and slw) and we are happy to be assigned UAT when needed.

PR #4379, opened 2024-09-26
set default fields to null when requesting additional sequencing

sabrine33:
default Request Type, Flowcell Type and Read length fields to null.
add UI validation on flowcell type for novaseq requests


spec/features/assets/asset_submission_spec.rb
```diff
@@ -15,6 +15,63 @@
...
+      select(project.name, from: 'Project')
+      fill_in 'Fragment size required (from)', with: '100'
+      fill_in 'Fragment size required (to)', with: '200'
+      select(selected_read_length, from: 'Read length')
+      click_button 'Create'
+
+      # The JS native validation error 'Please select an item in the list' is being displayed but cannot be inspected.
+      expect(page).not_to have_text 'Created request'
+      expect(page).to have_current_path(current_path)
+    end
```
KatyTaylor, line 72: Would be good to cover a couple more test cases for this validation:
- request type is novaseq, and flowcell type is set - no error
- request type is not novaseq, and flowcell type is not set - no error

KatyTaylor (changes requested):
Hi @sabrine33 , just catching myself up on what happened with this one as it's been a while!

* Original PR - https://github.com/sanger/sequencescape/pull/4293/files
* Got UAT feedback described here - https://github.com/sanger/sequencescape/issues/4126#issuecomment-2315256196
* Fixed here - https://github.com/sanger/sequencescape/pull/4324/files
* Reverted out of develop here - https://github.com/sanger/sequencescape/pull/4342/files - because of Heron integration suite test failing - because the automated submission didn't set flowcell type
* This PR changes the flowcell type validation to be front end (JS) rather than server-side (Rails model), so it shouldn't affect the Heron pipeline

Is that all correct? 

I've taken a look at the code and it looks like it does what's needed. **I've just requested a couple more tests if poss.**

Probably worth testing this one in UAT from branch, rather than merging it into develop - as it's been a bit of a risky one. And running a full int suite to check for breakages!

Thanks

sabrine33:
> * https://github.com/sanger/sequencescape/pull/4342/files

Yes, that is correct!

PR #4382, opened 2024-09-27
Update Rubocop workflow

StephenHulme:
Update this useful, but abandoned workflow.
Should hopefully prevent us from getting too deep into RuboCop techdebt again 🤞 

Resulting PR at #4383 

- Update action versions
- Add `Technical Debt` label to resulting PR


sdjmchattie:
> > Does this have an impact on people like me who will open a PR as soon as the first commit goes into a branch and marks it "Draft"?
> 
> Quite possibly, but I think the impact should be limited to rerunning a the ruby linting and tests. Would this be acceptable?

When I mark it Ready for Review?  To be honest, that should be fine.  It might be hard to avoid having the linting and tests right before marking it ready for review if they don't run while it's draft, but I'm not sure that's what this will do.  Looks like it always runs on Push anyway.

PR #4385, opened 2024-09-27
Y24-304 - Added fixed placement of controls when cherrypicking 

### LINKED ISSUE #4333: Y24-304 - Allow fixed placement of controls when cherrypicking
**User story**
As a user of the cherrypicking pipeline in Sequencescape, I would like there to be an option to add controls into a fixed location on the destination plate(s), so that I can choose where I want my controls.

Existing behaviour:
The cherrypicking pipeline currently has a dropdown called 'Control Plate', which is populated with the existing control plates in the system. A Control Plate is a special kind of labware, which generally has just a few of its wells populated with control samples. When a control plate is selected in the cherrypicking pipeline, these control samples are put into wells on the destination plate(s). The current behaviour is to randomise the location of these samples on the destination plate(s).

Anna G, Adrianne L, Katy

Anna G, Rich C

**Acceptance criteria**
- [x] I am able to choose to place the controls in pre-specified fixed locations on the destination plates (e.g. H10, H11 and H12).
- [x] Existing randomisation behaviour continues to work.
- [x] Control plates are used as the source of the control samples.
- [x] Control samples are added to the same locations on the destination plate as they are in on the control plate.
- [ ] In UAT, get customer to make their own control plate using [this documentation](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/Creating+Control+Plates) (developer needed for final step), and test using that.
- [x] If needed, update [instructions for making Control Plates](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/Creating+Control+Plates) 
- [x] If the control plate and plate template clash, display an error message

**Dependencies**
Original research story - https://github.com/sanger/limber/issues/1864 

**Additional context**
UX & design options are discussed in [a comment](https://github.com/sanger/sequencescape/issues/4333) on this issue.

See code in:
- `cherrypick_task.rb`
- `cherrypick_handler.rb`
- `control_locator.rb`

Quickest way to get to the cherrypicking screen:
- Create stock plate: UAT tab, Create Plate, accept default options, copy resulting plate barcode
- Create cherrypick submission: UAT tab, Test Submission, submission template = 'Cherrypick', Plate barcode = paste in your barcode 
- Create cherrypick batch: Pipelines tab, Cherrypick, refresh page until your plate appears in the table, select that row and hit Submit
- Click on 'Select Plate Template'

SHIV5T3R:
Closes #4333 

- A new type of placement where the original positions on the source plate are used on the destination plate when cherrypicking.
- Made changes to the UI to indicate the type of placement type a control plate will use.
- Control plates can have either a fixed or random placement type, the behaviour of random (the original) placement remains the same.


app/models/cherrypick_task.rb
```diff
@@ -21,9 +21,26 @@ class CherrypickTask < Task # rubocop:todo Metrics/ClassLength
   #
   # @return [CherrypickTask::ControlLocator] A generator of control locations
   #
-  def new_control_locator(batch_id, total_wells, num_control_wells, wells_to_leave_free: DEFAULT_WELLS_TO_LEAVE_FREE)
-    CherrypickTask::ControlLocator.new(batch_id:, total_wells:, num_control_wells:, wells_to_leave_free:)
+
+  # rubocop:disable Metrics/ParameterLists
```
sdjmchattie, line 25: Do we understand why this method is needed?  It seems like the only benefit it's offering is the default value for `wells_to_leave_free`.  We can probably avoid having all these parameters on the method by accepting a single `arguments` parameter instead and then passing the values from that through to `CherrypickTask::ControlLocator.new` after doing `parameters[:wells_to_leave_free] ||= DEFAULT_WELLS_TO_LEAVE_FREE`.  It's always better to try to comply with the Rubocop rules rather than disable them, unless we agree that rules should be turned off.

app/models/cherrypick_task/control_locator.rb
```diff
@@ -40,13 +45,18 @@ class CherrypickTask::ControlLocator
   # @param total_wells [Integer] The total number of wells on the plate
   # @param num_control_wells [Integer] The number of control wells to lay out
   # @param wells_to_leave_free [Enumerable] Array or range indicating the wells to leave free from controls
-  def initialize(batch_id:, total_wells:, num_control_wells:, wells_to_leave_free: [])
+
+  # rubocop:disable Metrics/ParameterLists
+  def initialize(batch_id:, total_wells:, num_control_wells:, template:, control_source_plate:, wells_to_leave_free: [])
```
sdjmchattie, line 50: Again, if we have so many parameters, maybe we should turn them into a structured object.  Either a hash with expected keys for a quick workaround, or a wrapper object with named attributes for a better solution.

sdjmchattie (changes requested):

app/models/cherrypick_task/control_locator.rb
```diff
@@ -29,23 +29,35 @@ class CherrypickTask::ControlLocator
...
-    @total_wells = total_wells
-    @num_control_wells = num_control_wells
-    @wells_to_leave_free = wells_to_leave_free.to_a
-    @available_positions = (0...total_wells).to_a - @wells_to_leave_free
+  # @param params [Hash] A hash containing the following keys:
+  #   - :batch_id [Integer] The id of the batch, used to generate a starting position
+  #   - :total_wells [Integer] The total number of wells on the plate
+  #   - :num_control_wells [Integer] The number of control wells to lay out
+  #   - :wells_to_leave_free [Enumerable] Array or range indicating the wells to leave free from controls
+  #   - :available_positions [Enumerable] Array or range indicating the available positions for controls
```
sdjmchattie, line 49: I don't think this is something you want in the params hash.  It's calculated by removing the `:wells_to_leave` from the full set of wells.

sdjmchattie (changes requested):

sdjmchattie (approved):

PR #4390, opened 2024-10-01
Y24-190-3: Support Limber with TubeFromTubeCreation endpoint

sdjmchattie:

- Add a v2 endpoint under `tube_from_tube_creation` so that Limber can move to API v2.
- Add tests for the endpoint behaviour.


app/resources/api/v2/tube_from_tube_creation_resource.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # @todo This documentation does not yet include a detailed description of what this resource represents.
```
BenTopping, line 5: Since this is a new one can you add a brief description of what its doing in place of these todos.

app/resources/api/v2/tube_from_tube_creation_resource.rb
```diff
@@ -0,0 +1,103 @@
...
+    # of the JSON:API standard.
+    class TubeFromTubeCreationResource < BaseResource
+      ###
+      # Attributes
+      ###
+
+      # @!attribute [w] child_purpose_uuid
+      #   This is declared for convenience where child purpose is not available to set as a relationship.
+      #   Setting this attribute alongside the `child_purpose` relationship will prefer the relationship value.
+      #   @deprecated Use the `child_purpose` relationship instead.
```
BenTopping, line 25: Is this deprecated comment correct? Same in below attributes. If this is a new resource and there is a preferred method of using relationships can we do that instead?

Assuming this would be riskier and mean changing limber requests?

spec/requests/api/v2/tube_from_tube_creations_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+        let(:disallowed_value) { 'child' }
+        let(:payload) do
+          { data: { type: resource_type, relationships: { child: { data: { id: '1', type: 'plates' } } } } }
+        end
+
+        it_behaves_like 'a POST request with a disallowed value'
+      end
+    end
+
+    context 'without a required relationship' do
```
BenTopping, line 281: is child_purpose also a required relationship?

sdjmchattie: Weirdly, no.  The model seems happy without it.

app/resources/api/v2/tube_from_tube_creation_resource.rb
sdjmchattie: Yeah this is an approach I've been using because Limber works with UUIDs.  If I didn't add these "legacy" ways of passing in the relationships, I'd have to either pass the relationships around Limber (more change than I should be doing right now) or I'd have to call another endpoint to find the related object, just to be able to pass it to this endpoint.

app/resources/api/v2/tube_from_tube_creation_resource.rb
sdjmchattie: That's fair.  I guess I was hoping someone more familiar with the model and how it's used could write a better description.  But I admit I did just copy-paste this from other endpoints I've been creating.  I could probably add example usage to the documentation at least, since that's not model specific.

app/resources/api/v2/tube_from_tube_creation_resource.rb
BenTopping: Ah ok, yeah I guess you have to limit the work somewhere. Is the plan to address these legacy methods after v2 is complete?

app/resources/api/v2/tube_from_tube_creation_resource.rb
andrewsparkes: 'Resource for the api endpoint specifically for the creation of a single child tube from a single parent tube.'
maybe?

Looks like a 1:1 transfer only? Can't see anything else particularly special about it.

andrewsparkes (approved):
Looks good.

app/models/tube_creation.rb
```diff
@@ -32,6 +32,6 @@ def create_children!
   end
 
   def record_creation_of_children
-    #    children.each { |child| parent.events.create_tube!(child_purpose, child, user) }
+    # No record is made by this model class.
```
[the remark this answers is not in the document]
sdjmchattie: Hmm yeah I'm not sure I'm afraid.  I know we need to define the method because the base class (`AssetCreation`) calls it in `process_children` without defining it, but that's all I know.  I think perhaps we should implement a stub method in `AssetCreation` so it can be overridden by the classes that actually define some behaviour, but that it's optional to do so.

PR #4416, opened 2024-10-15
Y24-099 scRNA - tube rack banking sequencescape api

andrewsparkes:
Closes #https://github.com/sanger/limber/issues/1696

Adds specific tube rack creation v2 api endpoint


spec/factories/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,35 @@
...
+          tube_rack_barcode: 'TR00000001',
+          tube_rack_purpose_uuid: tube_rack_purpose.uuid,
+          tube_rack_metadata_key: 'tube_rack_barcode',
+          racked_tubes: [
+            {
+              tube_barcode: 'ST00000001',
+              tube_name: 'SEQ:NT1A:A1',
+              tube_purpose_uuid: tube_purpose.uuid,
+              tube_position: 'A1',
+              parent_uuids: [plate.uuid]
```
andrewsparkes, line 24: parent.uuids should be the specific well.uuid

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+  #   },
+  #   etc... more tube racks
+  attr_writer :tube_rack_attributes
+
+  DEFAULT_TARGET_TYPE = 'TubeRack'
+
+  # singular 'parent' getter to stay backwards compatible
+  def parent
+    parents.first
+  end
```
KatyTaylor, line 66: Is this 'backwards compatible' comment relevant here (and below)? I'm guessing it was copied over from the specific tube creation model? 

Maybe still relevant as it inherits from AssetCreation?

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+
+  validates :tube_rack_attributes, presence: true
+
+  has_many :parent_associations,
+           foreign_key: 'asset_creation_id',
+           class_name: 'AssetCreation::ParentAssociation',
+           inverse_of: 'asset_creation',
+           dependent: :destroy
+
+  # NB. assumption that parent is a plate
```
KatyTaylor, line 33: I think making this assumption is OK if it makes things easier. However, just wondering - did you try having it as 'Labware' instead of Plate?

app/models/specific_tube_rack_creation.rb
KatyTaylor, line 1: Did you and Stuart discuss your opinions on whether this is a good pattern for the API to follow?

If so could put a sentence or two under line 5 to explain the pattern.

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+
+  #   @param value [Array<Hash>] Hashes defining the attributes to apply to each tube rack and
+  #     the tubes within that are being created.
+  #     This is used to set custom attributes on the tube racks, such as name. As well as to create
+  #     the tubes within the tube rack and link them together.
+  #   @example [
+  #   {
+  #     :tube_rack_name=>"Tube Rack",
+  #     :tube_rack_barcode=>"TR00000001",
+  #     :tube_rack_purpose_uuid=>"0ab4c9cc-4dad-11ef-8ca3-82c61098d1a1",
```
KatyTaylor, line 45: I remember a conversation related to Stuart's API work where we realised that JSON API requires the passing of 'ids' for each object. And we weren't quite sure why uuids are used instead of just ids. Do you remember that conversation? Wonder if we want to use ids instead here, or use uuids for consistency?

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+  #     :tube_rack_metadata_key=>"tube_rack_barcode",
```
KatyTaylor, line 46: This (`tube_rack_metadata_key`) doesn't feel right being here - this should be app-wide config, not something that varies between tube racks.

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+    belongs_to :specific_tube_rack_creation
+    belongs_to :tube_rack
+  end
+
+  # Allows a many to many relationship between SpecificTubeRackCreations and Purposes.
+  class ChildPurpose < ApplicationRecord
+    self.table_name = 'specific_tube_rack_creation_purposes'
+    belongs_to :specific_tube_rack_creation
+    belongs_to :tube_rack_purpose, class_name: 'Purpose'
+  end
```
KatyTaylor, line 17: Where's the link between rack and purpose? 
Should it be instead that `SpecificTubeRackCreation::ChildTubeRack` has a relationship `belongs_to :purpose`, or just a string with the purpose name or something?

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+  #     :tube_rack_barcode=>"TR00000001",
+  #     :tube_rack_purpose_uuid=>"0ab4c9cc-4dad-11ef-8ca3-82c61098d1a1",
+  #     :tube_rack_metadata_key=>"tube_rack_barcode",
+  #     :racked_tubes=>[
+  #       {
+  #         :tube_barcode=>"SQ45303801",
+  #         :tube_name=>"SEQ:NT749R:A1",
+  #         :tube_purpose_uuid=>"0ab4c9cc-4dad-11ef-8ca3-82c61098d1a1",
+  #         :tube_position=>"A1",
+  #         :parent_uuids=>["bd49e7f8-80a1-11ef-bab6-82c61098d1a0"]
```
KatyTaylor, line 53: What is `parent_uuids`? Should this be `rack_uuid`? But the rack link is already specified. Is it actual parent labware? I thought the parent-child relationships were going on the racks?

KatyTaylor (commented):
Some comments so far. I haven't finished reading it yet, so I might answer my own questions as I go. Hope to finish on Monday! Thanks.

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+  # @return [void]
+  def handle_tube_rack_barcode(tube_rack_barcode, new_tube_rack)
+    existing_barcode_record = Barcode.includes(:asset).find_by(barcode: tube_rack_barcode)
+
+    if existing_barcode_record.nil?
+      create_new_barcode(tube_rack_barcode, new_tube_rack)
+    else
+      redirect_existing_barcode(existing_barcode_record, new_tube_rack, tube_rack_barcode)
+    end
+  end
```
KatyTaylor, line 185: This (handle_tube_rack_barcode) feels like a core method that should probably be elsewhere - on the tube rack model, or purpose perhaps 🤔

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+  #
+  # @return [Barcode] The created Barcode object.
+  def create_new_barcode(tube_rack_barcode, new_tube_rack)
+    barcode_format = Barcode.matching_barcode_format(tube_rack_barcode)
+    if barcode_format.nil?
+      error_message = "The tube rack barcode '#{tube_rack_barcode}' is not a recognised format."
+      raise StandardError, error_message
+    end
+    Barcode.create!(labware: new_tube_rack, barcode: tube_rack_barcode, format: barcode_format)
+  end
```
KatyTaylor, line 206: I thought that this kind of thing was probably replicated in the sample manifest upload code, and then found this comment - https://github.com/sanger/sequencescape/blob/a4a63a2e9b3348a3da62914739240492ed33d2e0/app/sample_manifest_excel/sample_manifest_excel/upload/processor/tube_rack.rb#L154

So looks like there might be 3 places that could do with a refactor!

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+      PolyMetadatum.new(
+        key: metadata_key,
+        value: tube_rack_barcode,
+        metadatable_type: tube_rack.class,
+        metadatable_id: tube_rack.id
+      )
+    return if pm.save
+
+    raise StandardError, "New metadata for tube rack (key: #{metadata_key}, value: #{tube_rack_barcode}) did not save"
+  end
```
KatyTaylor, line 258: This also feels like a core method for tube rack creation, perhaps should be on the tube rack model.

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+  #
+  # @return [void]
+  def create_racked_tubes(racked_tubes, new_tube_rack)
+    # iterate through the rack attributes to create each tube
+    racked_tubes.each { |tube_attributes| create_racked_tube(tube_attributes, new_tube_rack) }
+  end
+
+  def create_racked_tube(tube_attributes, new_tube_rack)
+    tube = create_tube(tube_attributes)
+    link_tube_to_rack(tube, new_tube_rack, tube_attributes[:tube_position])
```
KatyTaylor, line 280: Doesn't feel like `link_tube_to_rack` needs to be a separate method - could just paste contents in here.

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+  #
+  # @raise [StandardError] if the tube barcode is already in use.
+  #
+  # @return [void]
+  def ensure_unique_tube_barcode(tube_barcode)
+    existing_tube_barcode_record = Barcode.includes(:asset).find_by(asset_id: tube_barcode)
+    return if existing_tube_barcode_record.nil?
+
+    error_message = "The tube barcode '#{tube_barcode}' is already in use, cannot continue."
+    raise StandardError, error_message
```
KatyTaylor, line 299: Is this replicating behaviour that would happen already if you tried to insert a tube with a pre-existing barcode?

app/models/specific_tube_rack_creation.rb
```diff
@@ -0,0 +1,436 @@
...
+  #
+  # @return [void]
+  def check_tube_barcode_format(tube_barcode)
+    barcode_format = Barcode.matching_barcode_format(tube_barcode)
+
+    # barcode format should be recognised
+    if barcode_format.nil?
+      error_message = "The tube barcode '#{tube_barcode}' is not a recognised format."
+      raise StandardError, error_message
+    end
```
KatyTaylor, line 322: Some code duplicated between here and `create_new_barcode` method.

app/models/tube_rack.rb
```diff
@@ -4,12 +4,15 @@
...
 class TubeRack < Labware
   include Barcode::Barcodeable
+  include Asset::Ownership::Unowned
 
   self.sample_partial = 'assets/samples_partials/tube_rack_samples'
 
   has_many :racked_tubes, dependent: :destroy, inverse_of: :tube_rack
   has_many :tubes, through: :racked_tubes
   has_many :contained_samples, through: :tubes, source: :samples
+  # TODO: change to purpose_id
```
KatyTaylor, line 14: TODO left in here

app/models/tube_rack/purpose.rb
```diff
@@ -3,11 +3,19 @@
 # The purpose of a tube rack is to hold tubes.
 # Created to hold the size of the tube rack for use when generating manifests.
 class TubeRack::Purpose < Purpose
-  self.default_prefix = 'TR'
```
KatyTaylor, line 6: Was this used anywhere?

app/models/tube_rack/purpose.rb
```diff
@@ -3,11 +3,19 @@
...
-
   has_many :sample_manifests, inverse_of: :tube_rack_purpose, dependent: :restrict_with_exception
 
+  # TODO: change to purpose_id
```
KatyTaylor, line 8: TODO left in

app/resources/api/v2/specific_tube_rack_creation_resource.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # @todo This documentation does not yet include a detailed description of what this resource represents.
+    # @todo This documentation does not yet include detailed descriptions for relationships, attributes and filters.
+    # @todo This documentation does not yet include any example usage of the API via cURL or similar.
```
KatyTaylor, line 7: It feels like we shouldn't be adding new resources with todos in them?

app/resources/api/v2/specific_tube_rack_creation_resource.rb
```diff
@@ -0,0 +1,120 @@
...
+        return if value.nil?
+
+        # Convert ActionController::Parameters into hashes.
+        @model.tube_rack_attributes = value.map(&:to_unsafe_h)
+      end
+
+      # @!attribute [w] user_uuid
+      #   This is declared for convenience where the user is not available to set as a relationship.
+      #   Setting this attribute alongside the `user` relationship will prefer the relationship value.
+      #   @deprecated Use the `user` relationship instead.
```
KatyTaylor, line 73: Adding something that is deprecated already? Am I misunderstanding?

app/resources/api/v2/specific_tube_rack_creation_resource.rb
```diff
@@ -0,0 +1,120 @@
...
+      #   @param value [String] The UUID of the user who initiated the creation of tubes.
+      #   @return [Void]
+      #   @see #user
+      attribute :user_uuid
+
+      def user_uuid=(value)
+        @model.user = User.with_uuid(value).first
+      end
+
+      # TODO: what is this state change uuid?
```
KatyTaylor, line 83: TODO left in here.

Is it a state change uuid? Surely it's the uuid of the 'specific tube rack creation'?

app/resources/api/v2/tube_rack_resource.rb
```diff
@@ -29,17 +29,25 @@ class TubeRackResource < BaseResource
       # Associations:
       has_many :racked_tubes
       has_many :comments, readonly: true
-      has_one :purpose, foreign_key: :plate_purpose_id
+      # TODO: change to purpose_id
```
KatyTaylor, line 32: TODO left in here

app/resources/api/v2/tube_rack_resource.rb
```diff
@@ -29,17 +29,25 @@ class TubeRackResource < BaseResource
...
+      has_one :purpose, foreign_key: :plate_purpose_id, class_name: 'TubeRackPurpose'
+      has_many :parents, readonly: true, polymorphic: true
+      has_many :state_changes, readonly: true
+      has_one :custom_metadatum_collection, foreign_key_on: :related
+      has_many :ancestors, readonly: true, polymorphic: true
+      # NB. no child or descendent associations as tube racks can't have children (tubes have children).
```
KatyTaylor, line 38: I don't think it's true that tube racks can't have children?

KatyTaylor (commented):
There's a lot of work gone into this story!

Lots of comments, sorry. Let's talk them through at some point because I don't want to create a bunch of extra work for you on this one.

app/models/tube_rack.rb
andrewsparkes: it's present for other instances in the codebase where plate_purpose_id is used, it refers to that needing a refactor to labware_purpose_id at some point. requires changes to the DB table.
out of scope for this story

app/models/tube_rack/purpose.rb
andrewsparkes: see above

app/models/tube_rack/purpose.rb
andrewsparkes: It's used for creating new barcodes, which we don't need to do for tube racks. They have etched barcodes and we use those instead.

app/resources/api/v2/specific_tube_rack_creation_resource.rb
andrewsparkes: This is api v1 vs v2 legacy stuff, it will go with api v1 eventually. Copied from the other labwares. Ask Stuart for more info.

app/resources/api/v2/specific_tube_rack_creation_resource.rb
andrewsparkes: removed and comments added

app/resources/api/v2/specific_tube_rack_creation_resource.rb
andrewsparkes: attribute was copied from specific tube creation, comment was for me to check with Stuart, and yes you are correct it is the uuid of the AssetCreation. fixed comment.

app/resources/api/v2/tube_rack_resource.rb
andrewsparkes: yep

app/resources/api/v2/tube_rack_resource.rb
andrewsparkes: we discussed this and decided it was the tubes that have children:
1. parent plate has child tube rack
2. tube rack has racked tubes (not children)
3. tubes have child plate

what would the child of a tube rack be?

spec/factories/specific_tube_rack_creation.rb
andrewsparkes: done

app/models/specific_tube_rack_creation.rb
andrewsparkes: Use case for this endpoint is plate to tube rack creation. Didn't want to get into the complexities of going from a tube to a tuberack (how would that work?) or tuberack to tuberack. Can discuss.

app/models/specific_tube_rack_creation.rb
andrewsparkes: Link between rack and purpose is in the TubeRack model / resource?
This endpoint can make multiple tube racks of different purposes remember.
I copied the pattern from SpecificTubeCreation.
Can discuss.

app/models/specific_tube_rack_creation.rb
andrewsparkes: Think those would be different endpoints.

app/models/specific_tube_rack_creation.rb
andrewsparkes: I kept it consistent. Can discuss with Stuart, think the uuids was api v1 but don't know why they are used.

app/models/specific_tube_rack_creation.rb
andrewsparkes: Can put it in an initializer file like the scRNA pooling constants?

app/models/specific_tube_rack_creation.rb
andrewsparkes: These are the parent plate well uuids for this specific tube. Required to set up the transfer requests.
racked_tubes is nested within a specific tube rack definition which we've supplied the tube rack purpose uuid for. Remember we haven't created the rack or tubes yet so we don't have their uuids.

app/models/specific_tube_rack_creation.rb
andrewsparkes: It was needed, think I had failing tests without. Use case here again is one parent plate to up to 2 tube racks, so will be one plate.

app/models/specific_tube_rack_creation.rb
andrewsparkes: yeah I agree, too much code in this labware creator. can discuss.

app/models/specific_tube_rack_creation.rb
andrewsparkes: Was probably result of rubocop saying a method was too large/complex and me asking copilot to split it up into smaller chunks.

app/models/specific_tube_rack_creation.rb
andrewsparkes: one is for tube and the other for tuberack.
wondering if these should all be in the models extending from a common labware model method.

app/models/specific_tube_rack_creation.rb
KatyTaylor: Yes, or in the TubeRack model or wherever we want to have some generic code related to tube rack creation.

app/resources/api/v2/tube_rack_resource.rb
KatyTaylor: I think in the specific scRNA Core case that's true, but not necessarily in the general case. 

If it was just a simple plate --> tube rack --> plate, then I could see all the parent-child relationships being from the tube rack.

app/models/specific_tube_rack_creation.rb
KatyTaylor: Agreed to move this to a constant on SS side.

app/models/specific_tube_rack_creation.rb
KatyTaylor: @KatyTaylor  Create refactor story to create central tube rack creation code to be shared between this and sample manifest upload code.

app/models/specific_tube_rack_creation.rb
andrewsparkes: done

app/models/specific_tube_rack_creation.rb
andrewsparkes: Might be worth having a discussion about error handling for API endpoints in general, as I'm not very clear on what the best way to do all this is. This endpoint has to make multiple tube racks, each with multiple tubes. If anything goes wrong at any point I want the whole lot rolled back and a suitable human friendly error message passed back through the api response to be displayed to the user.
Does raising an explicit StandardError like this trigger the full rollback? And does that error message make it back to Limber and get displayed? Would I get the same result attempting an insert and that failing with a uniqueness error (given barcode is a different table than labwares / receptacles)? Does it still rollback everything and return a clear message to Limber?

app/models/specific_tube_rack_creation.rb
andrewsparkes: Looking at this again, I think you are right, there is a problem here.
It should be just the one table, not two.

andrewsparkes: I have tried to change it for one table with belongs_to on purpose added but it breaks all the unit tests. I haven't been able to fix it.

app/models/specific_tube_rack_creation.rb
KatyTaylor: Shall we put it on our list to ask Stuart about?

PR #4426, opened 2024-10-17
Y24-190: Fix API v2 transfers

sdjmchattie:

- Sort out options for attribute and relationships across all resources:
    - `writeonly`:  The attribute or relationship can only be written to (not fetched)
    - `readonly`:  The attribute or relationship can only be read from (not created or updated)
    - `write_once`:  The attribute or relationship can only be written to one time (on create but not on update)
- Apply the above to existing resources (I hope I've got this right .. I think although they were named `readonly: true` before they were actually `write_once` but I've kept ones I think should never be writeable as `readonly` anyway).
- Get rid of Transfer sub-classes in the API.
    - It turns out that we don't use them for things like Tubes and Plates and Purposes.  I only just found the code that deals with descendants for this type of resource.  Unfortunately, what I had built previously falls over when you make a transfer a relationship of another resource.


app/resources/api/v2/barcode_printer_resource.rb
```diff
@@ -33,11 +33,11 @@ class BarcodePrinterResource < BaseResource
 
       # @!attribute [r] print_service
       #   @return [String] the service this printer can be instructed to print from.
-      attribute :print_service, readonly: true
+      attribute :print_service, write_once: true
```
BenTopping, line 36: Im not sure if these should be write_once since this is an immutable resource and I don't think printers are created through the API (to my knowledge).

BenTopping (approved):
Nice work, the creatable_fields and updatable_fields default behaviour was well hidden 😆 . 

Approving as I don't think there is anything dangerous in this PR as if anything it will make things more permissive and hopefully any incorrect attribute types will be caught in the int suite.

app/resources/api/v2/barcode_printer_resource.rb
sdjmchattie: I'm happy to change this back to `readonly` if you believe that's what we need.  Bear in mind that before I changed this, `readonly` actually meant `write_once` (bad naming in the past).  So by changing this to `write_once` this is a "no actual change made" amendment.  Anything I've left as `readonly` is now LESS permissive than before.

BenTopping: Okay, happy to go with your decision. Does write_once still do anything when the resource is immutable?

sdjmchattie: Technically yes, but `immutable` makes it impossible to hit the endpoint with a method that modifies resource.  When you put `immutable` in a resource, the `jsonapi_resources` entry in `routes.rb` only sets up a GET endpoint and you cannot submit a `POST` or `PATCH` method request.

spec/resources/api/v2/barcode_printer_resource_spec.rb
```diff
@@ -4,14 +4,16 @@
...
-  # Test attributes
-  it { is_expected.to have_readonly_attribute :uuid }
-  it { is_expected.to have_readonly_attribute :print_service }
-  it { is_expected.to have_readonly_attribute :barcode_type }
+  # Model Name
+  it { is_expected.to have_model_name 'BarcodePrinter' }
 
+  # Attributes
+  it { is_expected.to have_readonly_attribute :barcode_type }
   it { is_expected.to have_readwrite_attribute :name }
```
BenTopping, line 16: should name be readonly?

Issue #4406, opened 2024-10-09
Y24-373 - Data migration to update sample_metadata country_of_origin


neilsycamore:
Two country_of_origin of options ("not applicable: control sample" and "not applicable: sample group") have been identified and [Y24-372](https://github.com/sanger/sequencescape/issues/4405) created to change these.

The samples in the Sequencescape and MLWH databases requires updating to reflect the changes
not applicable: control sample => missing: control sample
not applicable: sample group => missing: sample group

neilsycamore:
```
def update_country_of_origin(old_id,new_id)
  ActiveRecord::Base.transaction do
    data = Sample::Metadata.where(country_of_origin: old_id)
    data.each {|sm| sm.update!(missing: new_id)
  end
end
```

BenTopping:
Confirmed data migrated correctly on UAT and warehouse records were updated.

PR #4433, opened 2024-10-18
Y24-372 - Correct country of origin values for submissions

### LINKED ISSUE #4405: Y24-372 - [BUG] incorrect country_of_origin options
When setting a sample's country_of_origin (in a manifest or via the sample edit view) the options "not applicable: control sample" and "not applicable: sample group" are not correct and prevents a sample from being accessioned

The correct options are:
"missing: control sample" and "missing: sample group"
this is listed in the ERC000011 [checklist](https://www.ebi.ac.uk/ena/browser/view/ERC000011)

These will need changing.

There will be a data migration story required

Update @sw32 when closing this

BenTopping:
Closes #4405 

- Sets Indsc::Country `not applicable: control sample` and `not applicable: sample group` to be invalid.
  - This means they will not appear in new manifests as options but does not prevent existing manifests with these types to be uploaded.  
- Creates Indsc::Country `missing: control sample` and `missing: sample group`

There is [another story to migrate existing data](https://github.com/sanger/sequencescape/issues/4406). However this story ensures all new manifests will only have the valid countries visible whilst also allowing existing manifests to use the old/invalid countries so as not break any behaviour.

### DIFF
--- a/app/models/accessionable/base.rb
+++ b/app/models/accessionable/base.rb
@@ -95,6 +95,8 @@ class Accessionable::Base
       MISSING_DATA_AGGREEMENT_PRE2023 = 'missing: data agreement established pre-2023'
       MISSING_ENDANGERED_SPECIES = 'missing: endangered species'
       MISSING_HUMAN_IDENTIFIABLE = 'missing: human-identifiable'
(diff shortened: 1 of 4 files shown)


neilsycamore:
ok

neilsycamore (approved):

PR #4438, opened 2024-10-21
Tweak newer RSpec Rubocop rules

sdjmchattie:
i.e. disable the MultipleMemoizedHelpers rule and set the maximum NestedGroups to 6 which matches our existing maximum nesting and prevents us creating more complicated nesting in the future.


sdjmchattie:
Whoops ... sorry @StephenHulme .  It was starting to get on my nerves for each new resource I was adding.

PR #4439, opened 2024-10-21
Y24-190-3: Support Limber with a V2 endpoint for PlateCreations

sdjmchattie:

- Add an endpoint for PlateCreations on API v2.
- Add tests for the new endpoint.
- Fix naming of shared examples 🤦🏻‍♂️ 

### DIFF
--- a/app/resources/api/v2/plate_creation_resource.rb
+++ b/app/resources/api/v2/plate_creation_resource.rb
@@ -1,0 +1,130 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # Provides a JSON:API representation of {PlateCreation} for creation of a child plate with a given purpose which
+    # is linked to a parent plate as one of its children.
+    # Creation of this resource via a `POST` request will initiate the child plate creation.
+    # The child plate is returned by this endpoint under the {#child} relationship.
+    #
+    # @note This resource cannot be modified after creation: its endpoint will not accept `PATCH` requests.
+    # @note Access this resource via the `/api/v2/plate_creations/` endpoint.
+    #
+    # @example POST request with arguments specified by UUID (deprecated)
+    #   POST /api/v2/plate_creations/
+    #   {
+    #     "data": {
+    #       "type": "plate_creations",
+    #       "attributes": {
+    #         "parent_uuid": "11111111-2222-3333-4444-555555666666",
+    #         "child_purpose_uuid": "22222222-3333-4444-5555-666666777777",
+    #         "user_uuid": "33333333-4444-5555-6666-777777888888"
+    #       }
+    #     }
+    #   }
+    #
+    # @example POST request with arguments specified by relationship
+    #   POST /api/v2/plate_creations/
+    #   {
+    #     "data": {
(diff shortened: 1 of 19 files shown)


SHIV5T3R (approved):

app/resources/api/v2/plate_creation_resource.rb
```diff
@@ -0,0 +1,130 @@
...
+      # @!attribute [w] user_uuid
+      #   This is declared for convenience where the {User} is not available to set as a relationship.
+      #   Setting this attribute alongside the `user` relationship will prefer the relationship value.
+      #   @deprecated Use the `user` relationship instead.
+      #   @param value [String] The UUID of the {User} who initiated this plate creation.
+      #   @return [Void]
+      #   @see #user
+      attribute :user_uuid, writeonly: true
+
+      def user_uuid=(value)
```
BenTopping, line 95: I wonder if this user_uuid= is something we can move to the base resource as it always has the same implementation. Might be an annoying bit of meta programming though 😆

BenTopping (approved):
Looks good.

app/resources/api/v2/plate_creation_resource.rb
sdjmchattie: Yeah I think you're probably right.  Except there will be one resource that doesn't want that and then you're looking at including it as a module.  I think it's OK to get this all done with the repetition for now.  It was always repeated in v1.

spec/resources/api/v2/plate_creation_resource_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  it { is_expected.to have_writeonly_attribute :parent_uuid }
+  it { is_expected.to have_writeonly_attribute :user_uuid }
+  it { is_expected.to have_readonly_attribute :uuid }
+
+  # Relationships
+  it { is_expected.to have_a_readonly_have_one(:child, 'Plate') }
+  it { is_expected.to have_a_writable_have_one(:child_purpose, 'PlatePurpose') }
+  it { is_expected.to have_a_writable_have_one(:parent, 'Plate') }
+  it { is_expected.to have_a_writable_have_one(:user, 'User') }
+end
```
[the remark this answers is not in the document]
sdjmchattie: @StephenHulme 

> Does this specificy rule out children that are tubes or tube-racks?

This is the plate_creation endpoint, so it is intended to create plates.  I don't know why this was built so specifically that way, but there it is.

> To confirm, write-only is set but not read, readonly is read but not set, and writable is is read and write?

Not quite true:

- Write-only is set only when the resource is created, and not when you update (PATCH) it.
- Read-only is readable but can never be set via the API.
- Writable is able to be written in both POST and PATCH requests, but says nothing about the readability.

The `have_one` and `have_many` matchers only check for readonly or writable relationships because that's how we've been using them.  In the matchers themselves, you can see that we do checks like `have_attribute` which is from an external library and actually means that the attribute is readable and not that it was defined.  You can define the attribute as readonly and the `have_attribute` returns `false`.  I've tried to be more explicit in my naming.

PR #4450, opened 2024-10-24
Y24-111 Set default fields to null when requesting additional sequencing

sabrine33:
From @KatyTaylor :

Hi @sabrine33 , just catching myself up on what happened with this one as it's been a while!

Original PR - https://github.com/sanger/sequencescape/pull/4293/files
Got UAT feedback described here - https://github.com/sanger/sequencescape/issues/4126#issuecomment-2315256196
Fixed here - https://github.com/sanger/sequencescape/pull/4324/files
Reverted out of develop here - https://github.com/sanger/sequencescape/pull/4342/files - because of Heron integration suite test failing - because the automated submission didn't set flowcell type
This PR changes the flowcell type validation to be front end (JS) rather than server-side (Rails model), so it shouldn't affect the Heron pipeline
Is that all correct?

I've taken a look at the code and it looks like it does what's needed. I've just requested a couple more tests if poss.

Probably worth testing this one in UAT from branch, rather than merging it into develop - as it's been a bit of a risky one. And running a full int suite to check for breakages!

Thanks


dasunpubudumal (commented):
Changes look okay to me. Have we run a full pass of the int suite for this?
_the plan is to do branch deploy and run the int suite, probably I can do now. good point_ 

I ran it locally and it seems to do what is required.

image

spec/features/assets/asset_submission_spec.rb
```diff
@@ -7,12 +7,124 @@
...
+    before do
+      allow(selected_request_type).to receive(:validator_for).with('requested_flowcell_type').and_return(
+        request_flowcell_type_validators
+      )
+      login_user user
+      visit labware_path(asset)
+      click_link 'Request additional sequencing'
+    end
+
+    it 'displays an error if request type is of type illumina_htp_novaseq and Flowcell Type is not set' do
```
KatyTaylor, line 66: Hi @sabrine33 

Am I right in saying this doesn't actually use a 'illumina-HTP NovaSeq request', it uses a test request type made by the factory - but you're mocking the 'validator' behaviour against that request type so it should behave in the same way?

I think it's OK but could you change the wording so it doesn't say that you're using illumina-HTP NovaSeq requests, and add a comment to explain what you're doing?

Thanks 🙏

PR #4456, opened 2024-10-24
Y24 335 barcode for norm plate tests

dasunpubudumal:
Closes #

- Create a new request type and a submission for integrating the norm plate which is under the sequencing stage (not library prep). This new request type is used in Bespoke MX HiC pipeline config in https://github.com/sanger/limber/pull/2023.

We tried to do this without creating a new submission (i.e., with no new request types) but was unsuccessful as the Norm plate is an intermediary plate used in the ~sequencing~ multiplexing stage and library types aren't really used after the library pass step (i.e., after library prep stage) to filter pipeline configs. Even if you can declare a `library_type` in the pipeline config in Limber it won't be used to filter the configs in sequencing stages of the pipeline configs (i.e., stages after library pass).

Int suite PR: https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/196/diffs

--- 
Edit: "... intermediary plate used in the sequencing stage" should be corrected to " .. intermediary plate used in the **multiplexing** stage"


KatyTaylor:
Thanks for the explanation, really helpful. 

> is an intermediary plate used in the sequencing stage

--> I would say "is an intermediary plate used in the **multiplexing** stage"
(sequencing happens even later - after the final LB Lib Pool Norm tube is created)

config/default_records/submission_templates/007_novaseqx_sequencing.yml
```diff
@@ -118,6 +118,13 @@ Limber-Bespoke - PCR - NovaSeqX paired end sequencing:
     product_line_name: Bespoke
     product_catalogue_name: GenericPCR
     request_type_keys: ["limber_pcr_bespoke", "limber_multiplexing", "illumina_htp_novaseqx_paired_end_sequencing"]
+Limber-Bespoke - PCR HiC - NovaSeqX paired end sequencing:
+  name: "Limber-Bespoke - PCR HiC - NovaSeqX paired end sequencing"
+  submission_class_name: "LinearSubmission"
+  related_records:
+    product_line_name: Bespoke
+    product_catalogue_name: GenericPCR
+    request_type_keys: ["limber_pcr_bespoke", "limber_multiplexing_hic", "illumina_htp_novaseqx_paired_end_sequencing"]
```
KatyTaylor, line 127: Did you get a response from Danni as to whether this 'PCR HiC' naming made sense to her? I'm wondering whether just 'HiC' would be better.

KatyTaylor: Also could consider not propagating the 'Bespoke' terminology, since the Bespoke team doesn't exist any more.

KatyTaylor: And finally...! Do they want to do any other types of sequencing or just NovaseqX?

Normally we'd create a set of submission templates, including:
- One without any sequencing
- One for each sequencing instrument (e.g. NovaSeq 6000 and MiSeq)

KatyTaylor (commented):
Left some comments and questions.

config/default_records/submission_templates/007_novaseqx_sequencing.yml
andrewsparkes: PCR came from the variant of the pipeline it's copied from, there were PCR and no PCR variants. I think it's important info, but Danni would know what's acceptable.
Same for Bespoke in that it came from the original pipeline name. This is the HiC variant so same name with HiC in it.

config/default_records/request_types/004_limber_high_throughput.yml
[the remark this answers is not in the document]
andrewsparkes: Can't do that as they are shared files with lots of other active stuff in them that is already live.
Could move the new sections into new files? But then they aren't together with other similar configs.
Bit of a flaw with this record loader design.
Be nice if we could flag individual sections within files as WIP somehow.

andrewsparkes (commented):
Consider moving new sections into their own files and WIP flagging them. As Katy says, then safe to merge into develop.

config/default_records/submission_templates/007_novaseqx_sequencing.yml
dasunpubudumal: Yeah I have requested some info on the submission name [here](https://github.com/sanger/limber/issues/1958#issuecomment-2435467328) but I still haven't got any reply for it.

andrewsparkes (approved):

PR #4463, opened 2024-10-25
Y24-190: Use stricter RSpec testers on v2 resources

sdjmchattie:

- Switch tests for `has_one` and `has_many` over to new matchers that also assess the readability and writeability of the relationship.


app/resources/api/v2/tube_rack_resource.rb
```diff
@@ -14,33 +14,29 @@ module V2
...
-      attribute :number_of_rows, write_once: true
-      attribute :number_of_columns, write_once: true
       attribute :name, write_once: true
-      attribute :tube_locations
+      attribute :number_of_columns, write_once: true
+      attribute :number_of_rows, write_once: true
+      attribute :size
+      attribute :tube_locations, writeonly: true
+      attribute :updated_at, readonly: true
+      attribute :uuid, readonly: true
```
sdjmchattie, line 34: @andrewsparkes I don't know if any of the changes I've made here are going to conflict with what you're doing.  Sorry if they do.  The newer test file will hopefully help to explain what to expect from one of these resources, hence the changes.

sdjmchattie:
> Looks good, a much needed tidy-up.
> 
> What is the difference between `write_once` and `readonly`?

It's detailed in the BaseResource (https://github.com/sanger/sequencescape/blob/develop-Y24-190/app/resources/api/v2/base_resource.rb#L26-L33).

- `readonly` means that the attribute or relationship is returned in any JSON response, but you can't change its value via the API.
- `write_once` means that the attribute can be written to during a create (POST) request, but not in an update (PATCH) request.  Although its name doesn't explain this, it is also readable so it gets included in responses.
- `writeonly` allows you to write the attribute or relationship in either POST or PATCH requests, but you cannot read it (i.e. it's excluded from JSON responses).

sdjmchattie:
> Thanks @sdjmchattie for the extra detail 👌

Knowing what I know now about what we can do with chaining matchers, I'd probably opt for a rework of the permissions setters for `attribute` and `has_one`/`has_many`.  Instead of `readonly` and `writeonly` which already existed and hence why I went down this path.  I'd replace them all with `no_create: true`, `no_update: true` and `no_fetch: true`.  This aligns perfectly with the underlying implementation for the permissions and allows you to hit all eight 3-bit combinations for any field.  The matchers would then change to `have_an_attribute`, `have_a_has_one` and `have_a_has_many` with chain operations for `not_creatable`, `not_updatable` and `not_fetchable` so you can write your test something like:

`it { is_expected.to have_an_attribute(:uuid).not_creatable.not_updatable` and that would be equivalent to checking for a read-only field.

PR #4468, opened 2024-10-28
Y24-014 - Ability to re-run RVI BCL pipeline

BenTopping:

- Updates limber_reisc to include RVI acceptable purpose and library type

#### Notes

Relates to Limber PR: https://github.com/sanger/limber/pull/2030


andrewsparkes (approved):
Just check after deployment in the SS server rails console that the acceptable purposes was updated. I think it does work to update and add new ones, but check to be sure.
```
rt = RequestType.find_by(name: 'request type name')
rt.acceptable_purposes
```

Issue #4306, opened 2024-08-28
Y24-286 - [BUG] RVI Bait Capture Library Preparation requests not being updated.


BenTopping:
**Describe the bug**
**NOTE:** The `Limber-Htp - RVI - Bait Capture Library` submission is being superseded by the `RVI Limber-Htp - RVI Bait Capture Library - NovaSeq 6000 paired end sequencing` but the issue still applies to both.

The `Limber-Htp - RVI - Bait Capture Library` submission template creates 2 request groups, `Bait Capture Library Preparation`, `Limber Multiplexing`. These requests should update automatically whilst going through the RVI BCL Pipeline in Limber. However the `Bait Capture Library Preparation` requests are not getting updated.

This does not appear to affect the pipeline but we should fix this so it does get updated correctly and update the existing records so the requests are correctly passed.

**Suspected cause**
The requests should be started during the RVI Cherrypick -> RVI RT plate transfer / bed verification. This is controlled by [the transfer request logic here](https://github.com/sanger/sequencescape/blob/develop/app/models/transfer_request.rb#L296) and the transition states in Limber's robot file for this bed verification. However the state update is not getting fired because the RVI RT does not go into the started state.

**Acceptance criteria**
- [x] Update the RVI-BCL pipeline so that it starts it's `Bait Capture Library Preparation` requests when transferring from RVI Cherrypick to RVI RT. This may require updating the state machine in sequencescape transfer_request and the states in the Limber Robots file for the RVI-BCL pipeline.
Completed in: https://github.com/sanger/sequencescape/pull/4482

- [ ] Update historical `Bait Capture Library Preparation` requests so that they have the correct states via a rake task.
  - [ ] Should be 'Passed' if the limber multiplexing requests are passed.
  - [ ] Should be 'Started' if the limber multiplexing requests have not yet been started.
  - [ ] Check whether the state change requires a correct time stamp.
  - [ ] Check whether the request state is sent to the warehouse, if so update the rake task accordingly.

**To Reproduce**
Steps to reproduce the behaviour:
1. Run the RVI BCL integration test suite to completion.
2. Go to the created `Limber-Htp - RVI - Bait Capture Library` submission 
3. See `Bait Capture Library Preparation` requests are not started or passed.

**Screenshots**
![Screenshot 2024-08-28 at 14 00 35](https://github.com/user-attachments/assets/418b6824-b305-4886-9271-0b500f59122c)

**Additional context**
[...]

PR #4482, opened 2024-11-04
Y24-286-1 - Fix RVI BCL Library prep requests

BenTopping:
Part 1 of https://github.com/sanger/sequencescape/issues/4306

- Updates transfer request to start sibiling requests when going from pending to processed_1

### DIFF
--- a/app/models/transfer_request.rb
+++ b/app/models/transfer_request.rb
@@ -80,7 +80,7 @@ class TransferRequest < ApplicationRecord # rubocop:todo Metrics/ClassLength
     end
 
     event :process_1 do
-      transitions to: :processed_1, from: [:pending]
+      transitions to: :processed_1, from: [:pending], after: :on_started
     end
 
     event :process_2 do
--- a/spec/models/state_changer/standard_plate_spec.rb
+++ b/spec/models/state_changer/standard_plate_spec.rb
@@ -21,12 +21,25 @@ RSpec.describe StateChanger::StandardPlate do
     context 'when the plate is the initial plate in the pipeline' do
       include_context 'a limber target plate with submissions', 'pending'
 
-      let(:target_state) { 'started' }
+      context 'when the target state is started' do
+        let(:target_state) { 'started' }
 
-      it 'starts the requests', :aggregate_failures do
-        # Requests are started and we create one event per order.
-        expect { state_changer.update_labware_state }.to change(BroadcastEvent::LibraryStart, :count).by(1)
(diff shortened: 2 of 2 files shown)


spec/models/state_changer/standard_plate_spec.rb
```diff
@@ -21,12 +21,25 @@
...
-        expect { state_changer.update_labware_state }.to change(BroadcastEvent::LibraryStart, :count).by(1)
-        expect(library_requests).to all(be_started)
+        it 'starts the requests', :aggregate_failures do
+          # Requests are started and we create one event per order.
+          expect { state_changer.update_labware_state }.to change(BroadcastEvent::LibraryStart, :count).by(1)
+          expect(library_requests).to all(be_started)
+        end
+      end
+
+      # RVI BCL uses the processed_1 state as the first state for the the intial plate
```
andrewsparkes, line 34: typo intial -> initial

andrewsparkes: My first question would be why doesn't RVI used the 'started' state as the first state? So it sticks to the same convention as others. I forget the details of this pipeline...

My second question is, what happens if a pipeline uses both 'started' and 'processed_1' states? It doesn't trigger LibraryStart twice does it?

app/models/transfer_request.rb
```diff
@@ -80,7 +80,7 @@ class TransferRequest < ApplicationRecord # rubocop:todo Metrics/ClassLength
     end
 
     event :process_1 do
-      transitions to: :processed_1, from: [:pending]
+      transitions to: :processed_1, from: [:pending], after: :on_started
```
andrewsparkes, line 83: what does that change do? what is on_started ?

andrewsparkes (commented):
I had questions!

spec/models/state_changer/standard_plate_spec.rb
BenTopping: Good question. The RVI RT plate (the second plate in the pipeline) has 4 operations on it which require bed verifications so it needs to move through a lot of states. The best way to do this was to move it through the various processed states (I think we added processed_3 and processed_4 for this).

In hindsight moving to 'started' first and then the other states might have made more sense, but started is not a required first state. The on_started method is also called when moving from 'pending' to 'passed' which most pipelines do every plate transition.

No pipeline has both started and processed_1 but if they did nothing would happen, once the requests are started they cannot be started again.

PR #4489, opened 2024-11-06
Update Flipper to v1

StephenHulme:
Flipper was causing dependency lock with Sinatra (see [Depfu](https://depfu.com/repos/github/sanger/sequencescape?project_id=6076)).

- Update Flipper to v1 branch
  - Removes warning `DEPRECATION WARNING: Bolding log text with a positional boolean is deprecated and will be removed in Rails 7.2.`
- Update flipper config
- Disable feature-gate/flag [preloading](https://www.flippercloud.io/docs/optimization#preloading)
- Update usage of flipper
- Add usage details to readme
- Fix breaking tests

### DIFF
--- a/config/initializers/flipper.rb
+++ b/config/initializers/flipper.rb
@@ -12,7 +12,8 @@ Rails.application.configure do
   ## Flipper preloads all features before each request, which is recommended if:
   ##   * you have a limited number of features (< 100?)
   ##   * most of your requests depend on most of your features
-  ##   * you have limited gate data combined across all features (< 1k enabled gates, like individual actors, across all features)
+  ##   * you have limited gate data combined across all features (< 1k enabled gates,
+  ##     like individual actors, across all features)
   ##
   ## For more info, see https://www.flippercloud.io/docs/optimization#preloading
   # We only have a handful of features, so no need to preload them for all requests.


Gemfile
```diff
@@ -121,9 +121,9 @@ group :default do
   gem 'cancancan'
 
   # Feature flags
-  gem 'flipper', '~> 0.25.0'
-  gem 'flipper-active_record', '~> 0.25.0'
-  gem 'flipper-ui', '~> 0.25.0'
+  gem 'flipper'
+  gem 'flipper-active_record'
+  gem 'flipper-ui'
```
sdjmchattie, line 126: Would it be preferable to lock the versions again to the newly identified ones during dependency resolution?  i.e. 1.3.1 for each of these.

I know this can cause dependency lockout in the future, but it stops these randomly upgrading in the future and causing issues because of a regression or because of breaking changes in a version update.

README.md
```diff
@@ -508,3 +504,15 @@ open erd.pdf
 ```
 
 The command uses the [rails-erd](https://github.com/voormedia/rails-erd) gem.
+
+### Updating the table of contents
+
+To update the table of contents after adding things to this README you can use the [markdown-toc](https://github.com/jonschlinkert/markdown-toc)
+node module. To install it, make sure you have installed the dev dependencies from yarn. To update
+the table of contents, run:
```
sdjmchattie, line 512: It's better to keep each sentence on its own line.  This makes diffs in future edits much easier to parse and relies on line breaks being after logical units of text and not because of some abstract column count.  Git relies on line breaks not moving around too much for its diffs to be useful.

```suggestion
To update the table of contents after adding things to this README you can use the [markdown-toc](https://github.com/jonschlinkert/markdown-toc) node module.
To install it, make sure you have installed the dev dependencies from yarn.
To update the table of contents, run:
```

db/migrate/20241106150235_change_flipper_gates_value_to_text.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class ChangeFlipperGatesValueToText < ActiveRecord::Migration[6.1]
+  def up
+    # Ensure this incremental update migration is idempotent
+    return unless connection.column_exists? :flipper_gates, :value, :string
```
sdjmchattie, line 6: Is this really needed?  What would happen if the column were not already containing a string value, other than spending a bit longer than strictly needed doing the migration?  If you do think it's needed, should it also be used in the `down` method?

db/migrate/20241106150235_change_flipper_gates_value_to_text.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    remove_index :flipper_gates, %i[feature_key key value] if index_exists? :flipper_gates, %i[feature_key key value]
+    change_column :flipper_gates, :value, :text
+    add_index :flipper_gates, %i[feature_key key value], unique: true, length: { value: 255 }
+  end
+
+  def down
+    change_column :flipper_gates, :value, :string
```
sdjmchattie, line 14: Does this also need to do the index modifications to be a true reversal of the above?

sdjmchattie (changes requested):

db/migrate/20241106150235_change_flipper_gates_value_to_text.rb
sdjmchattie: Another alternative is to mark it as non-reversible as we've done in some previous migrations (I think I've seen it here anyway)

PR #4494, opened 2024-11-08
Y24-190: Support Limber with v2 plates

sdjmchattie:

- Ensure V2 endpoints can support Limber with plates
- Update tests to match


app/controllers/api/v2/qc_files_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+module Api
+  module V2
+    # Provides a JSON API controller for QcFile
+    # See: http://jsonapi-resources.com/ for JSONAPI::Resource documentation
+    class QcFilesController < JSONAPI::ResourceController
+      # By default JSONAPI::ResourceController provides most the standard
+      # behaviour, and in many cases this file may be left empty.
+    end
+
+    class QcFileProcessor < JSONAPI::Processor
```
sdjmchattie, line 12: In case you're wondering why I overrode this, it's so that we can get the incoming contents persisted to disk before a new QcFile gets created and we can reference the file in the new QcFile record.  This has to be ready before the new instance is created and the Processor is responsible for asking the `Resource` class to generate the new record.

This varies from the way I did it in `Transfer`s where I did some modifications of the operation parameters and moving data into the context.  I think the approach here is cleaner as it doesn't pollute the context and is makes sure we only do any parsing of inputs for a `create_resource` operation while leaving all the other operations functioning as normal.

app/controllers/api/v2/qc_files_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+        Tempfile.open(filename) do |file|
+          file.write(contents)
+          return file
+        end
+      end
+
+      # Override the default behaviour for a JSONAPI::Processor when creating a new resource.
+      # We need to parse the filename and contents attributes so that we can generate a temporary file for the new
+      # QcFile model object. Replacing the fields with these values after the new QcFile was created does not generate
+      # the file correctly in the database. The CarrierWave library needs these values sooner.
```
yoldas, line 40: I vaguely remember when you upload a file with the same name, the library is storing the file with (number) suffix. I am trying to understand if that would be an issue here. The filename in request header and filename we are dealing with might end up different if the same file is uploaded.

app/models/request.rb
```diff
@@ -156,7 +156,7 @@ def date_for_state(state)
                   'INNER JOIN uuids ON uuids.resource_id=requests.submission_id AND uuids.resource_type="Submission"'
                 ]
             )
-            .group('uuids.external_id')
+            .group('pool_id')
```
yoldas, line 159: can you group by selection alias here?

app/resources/api/v2/plate_resource.rb
```diff
@@ -21,26 +47,74 @@ class PlateResource < BaseResource
       # labware
       include Api::V2::SharedBehaviour::Labware
 
-      # Constants...
-
+      # TODO: {Y24-213} Possibly this line doesn't actually do anything and could be removed.
       default_includes :uuid_object, :barcodes, :plate_purpose, :transfer_requests
```
yoldas, line 51: Its name is misleading as well. In older versions of the library, it was used for pre-loading associations. It sounds like the include parameter to make a compound document, but it is not.

app/models/request.rb
sdjmchattie: As far as I know it's valid SQL syntax and it does give the same output.  Plus it makes it clearer what you're actually grouping by.

app/controllers/api/v2/qc_files_controller.rb
sdjmchattie: I couldn't spot any problems, but then it's hard to test access by multiple people at the same time.  Creating a `TempFile` does put your file in a folder with a unique number, so I don't think any issues are going to happen.

yoldas: I meant CarrierWave opens a file for the upload and reads the content from request body into the file, so it does not read into memory. I meant that file's name might be suffixed with a number and I suspected a mismatch but as I can't remember and find the details, ignore that.

app/models/request.rb
yoldas: I think aliasing happens after grouping. I just made up the below and it gives an error with group by pool_id
`Error Code: 1054. Unknown column 'pool_id' in 'group statement'`
but it works with uuids.external_id . 


```
select "uuids.external_id AS pool_id,
              GROUP_CONCAT(DISTINCT pw_location.description ORDER BY pw.map_id ASC SEPARATOR \",\") AS pool_into,
              SUM(requests.state = 'passed') > 0 AS pool_complete,
              MIN(requests.id) AS id,
              MIN(requests.sti_type) AS sti_type,
              MIN(requests.target_asset_id) AS target_asset_id,
              MIN(requests.submission_id) AS submission_id,
              MIN(requests.request_type_id) AS request_type_id"
from requests  
JOIN receptacles AS pw ON requests.asset_id=pw.id
JOIN well_links ON well_links.source_well_id=requests.asset_id
join maps AS pw_location ON pw.map_id=pw_location.id
join  uuids ON uuids.resource_id=requests.submission_id AND uuids.resource_type="Submission"
group by pool_id
limit 10;
```

app/resources/api/v2/submission_pool_resource.rb
```diff
@@ -2,10 +2,33 @@
...
+
+      # @!attribute [r] plates_in_submission
+      #   @return [Integer] The number of plates in the submission pool.
       attribute :plates_in_submission, readonly: true
-      attribute :used_tag2_layout_templates, readonly: true
-      attribute :used_tag_layout_templates, readonly: true
+
+      # @!attribute [r] tag_layout_templates
+      #   @return [Array<TagLayoutTemplateResource>] The tag layout templates for this submission pool.
+      has_many :tag_layout_templates, readonly: true
```
yoldas, line 31: Removing the attributes will break the TagCollection.used method in Limber app/models/labware_creators/tagging/tag_collection.rb .

app/controllers/api/v2/qc_files_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+        errors += JSONAPI::Exceptions::ParameterMissing.new('contents').errors if contents.nil?
+
+        [filename, contents, errors]
+      end
+
+      # Create a temporary file with the contents.
+      def create_tempfile(filename, contents)
+        # The filename for a Tempfile is passed as an array with the basename and extension.
+        # e.g. [ 'file', '.csv' ]
+        filename = [File.basename(filename, '.*'), File.extname(filename)]
```
harrietc52, line 30: Maybe one of the two `filename` variables could be renamed. I was just a bit confused reading the comment on like 28, thinking it was refering to the filename parameter, instead of the variable created on line 30

yoldas (commented):
It looks good as far as I understand it. I left a few comments.

app/resources/api/v2/qc_file_resource.rb
```diff
@@ -0,0 +1,126 @@
...
+      ###
+
+      # @!method create_with_tempfile
+      #   Create a new QcFile resource with the uploaded data from a temporary file. This is called by the controller
+      #   when a create request for a QcFile is made. It ensures the contents of the file have been written to a
+      #   new TempFile instance.
+      # @param context [Hash] The context for the request.
+      # @param tempfile [Tempfile] A temporary file containing the uploaded data.
+      # @param filename [String] The filename for the uploaded data.
+      # @return [QcFileResource] The new QcFile resource.
```
harrietc52, line 119:
```suggestion
      # @return [QcFile] The new QcFile resource.
```
Is it returning a QcFile resource, as apposed to a QcFileResource resource?

config/routes.rb
```diff
@@ -61,6 +62,7 @@
       jsonapi_resources :specific_tube_creations, except: %i[update]
       jsonapi_resources :state_changes, except: %i[update]
       jsonapi_resources :studies
+      jsonapi_resources :submission_pools
```
harrietc52, line 65: Does any `except: ` want to be added here too, or is `immutable` in the resource enough?

spec/requests/api/v2/shared_examples/requests.rb
```diff
@@ -34,28 +50,28 @@
   end
 end
 
-shared_examples 'a GET request including a has_one relationship' do
+shared_examples 'a GET request including a has_one relationship' do |related_name|
```
harrietc52, line 53: Nice

spec/requests/api/v2/submission_pools_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    it 'finds no route for the method' do
+      expect { api_patch "#{base_endpoint}/1", payload }.to raise_error(ActionController::RoutingError)
+    end
+  end
+
+  describe '#POST a create request' do
+    it 'finds no route for the method' do
+      expect { api_post base_endpoint, {} }.to raise_error(ActionController::RoutingError)
+    end
+  end
```
harrietc52, line 82: Could add a DELETE too

harrietc52 (approved):

app/models/request.rb
sdjmchattie: I just ran the following query (which Rails generated from the code shown) and it works fine:

```
SELECT uuids.external_id AS pool_id,
              GROUP_CONCAT(DISTINCT pw_location.description ORDER BY pw.map_id ASC SEPARATOR ",") AS pool_into,
              SUM(requests.state = 'passed') > 0 AS pool_complete,
              MIN(requests.id) AS id,
              MIN(requests.sti_type) AS sti_type,
              MIN(requests.target_asset_id) AS target_asset_id,
              MIN(requests.submission_id) AS submission_id,
              MIN(requests.request_type_id) AS request_type_id FROM `requests` INNER JOIN well_links ON well_links.source_well_id=requests.asset_id INNER JOIN receptacles AS pw ON well_links.target_well_id=pw.id AND well_links.type="stock" INNER JOIN maps AS pw_location ON pw.map_id=pw_location.id INNER JOIN uuids ON uuids.resource_id=requests.submission_id AND uuids.resource_type="Submission" WHERE `pw`.`labware_id` = 13821 AND `requests`.`submission_id` = 1463 AND `requests`.`state` != 'cancelled' GROUP BY pool_id
```

app/controllers/api/v2/qc_files_controller.rb
sdjmchattie: Good catch.  I agree.

app/resources/api/v2/submission_pool_resource.rb
sdjmchattie: It's OK ... I've updated that, but it's a good point.  When deploying this, I can't just do Sequencescape and then do Limber at some later point in time.

app/resources/api/v2/qc_file_resource.rb
sdjmchattie: It is actually a `QcFileResource`.  I've found that confusing about these resources as well.  This is a class method which is creating an instance of the class.  The first `new` below creates the resource instance and passes it a `QcFile.new` and the `context` as parameters.  The first is then assigned to `@model` for use by instance methods that read or modify the underlying model object.

config/routes.rb
sdjmchattie: Other resources are not using `except` when they're immutable.  You only need it if the resource is not immutable and you don't want to be able to update existing resources.  It's interesting that you mention `delete` though.  We haven't been protecting against deletion and we haven't been testing those requests either.  Rather than handle it here, I think that's a bigger issue that could be considered for a future PR.

spec/requests/api/v2/submission_pools_spec.rb
sdjmchattie: Yeah I agree it probably should be, but we haven't set up a schema for what we do and don't want to do for each endpoint.  I don't want to wrap handling delete for all the endpoints into this PR so I think it's something we should address later.

yoldas (approved):

PR #4509, opened 2024-11-20
fix: don't cache classes by default in development

StephenHulme:
Whenever I make a change to ruby code I need to restart the rails server...

- In development only, disable `config.cache_classes` by default

I assume this is the best way to achieve this?

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -14,7 +14,10 @@ Rails.application.configure do
   # since you don't have to restart the web server when you make code changes.
   config.cache_classes = ENV.fetch('CACHE_CLASSES', 'false') == 'true'
 
-  # Do not eager load code on boot.
+  # Eager load code on boot. This eager loads most of Rails and
+  # your application in memory, allowing both threaded web servers
+  # and those relying on copy on write to perform better.
+  # Rake tasks automatically ignore this option for performance.
   config.eager_load = true
 
   # Show full error reports.


config/environments/development.rb
```diff
@@ -12,7 +12,7 @@
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = ENV.fetch('CACHE_CLASSES', 'true') == 'true'
+  config.cache_classes = ENV.fetch('CACHE_CLASSES', 'false') == 'true'
```
sdjmchattie, line 15: I think the defaults are supposed to be safe to deploy to production so that we're not relying on the config in production to ensure the behaviour we want.  So a better solution to this would be to change the default in the environment dot file so that it sets `CACHE_CLASSES` to `false` for development.

sdjmchattie: Oh sorry, I've just noticed this file is development specific

sdjmchattie (approved):

config/environments/development.rb
sdjmchattie: Weird, because the comment above the line kind of hints that it does the opposite of what is happening.

sdjmchattie: Rather than comparing to a specific string `true`, I wonder if there's any way to check that the environment variable was set at all.  Not having any value set on it should be like saying `false` while having any value set should be like `true`.  IMO anyway

sdjmchattie: Ahh well that explains the comment!  Must have been copy pasted without updating when the logic was inverted.  I guess check with the team why we would be doing caching by default on Sequencescape.  Perhaps it can cause issues.  Don't want to break everyone's development environment and then they need to set an environment variable.

sdjmchattie (approved):

BenTopping (approved):

PR #4510, opened 2024-11-21
Y24-236 - Fix for concurrent sample manifest uploads causing duplicates

### LINKED ISSUE #4266: [BUG] Uploading the same manifest concurrently can cause duplicate samples/aliquots to be created.
**Describe the bug**
When uploading the same manifest twice and the second upload occurs before the first one has finished, duplicate records for samples and aliquots are created. A breakdown of this bug is in the additional context story linked below. Essentially it is a concurrency error caused by users being allowed to upload a manifest even if a version of that manifest is currently being processed.

**RT Ticket Number**
If applicable

**To Reproduce**
1. Create a sample manifest and fill it in.
2. Upload the sample manifest and upload it again in quick succession.
Note: The larger the manifest the longer it takes to upload, this means the uploads dont have to be in quick succession but the second one has to occur before the first one has finished,

**Expected behaviour**
When users upload the same manifest twice, at similar times, the second upload is rejected and users are shown an appropriate error message.

A suggested implementation for this is to 'lock' a manifest. When uploading a manifest we could add a check to see if the manifest is currently being processed (this should be fairly trivial as manifests already use a state machine for state management). There should then be an appropriate error message for the user telling them to wait.

Other possible implementations can be seen in the attached document in the story in 'additional context'.

**Additional context**
This is a follow on story from https://github.com/sanger/sequencescape/issues/4193

BenTopping:
Closes #4266 

- Add uploader validation to prevent upload if it is already uploading.

- To test locally create a fairly large manifest > 3 plates. (this just makes it easier to replicate concurrent uploads as small ones upload quickly).
- Update config/puma and `workers 1` so you can handle concurrent requests locally.
- You many need to increase db pools in config/database.yml to handle all the connections.
- Upload 2 manifests within 5-10 seconds of one another.

#### Notes

@yoldas managed to create duplicate samples, even with these changes, by actively trying to bypass validation by uploading two manifests at the exact same time. There is a theoretical window of a few milliseconds where the validation passes in the first manifest and second manifest at the same time before both setting the manifest state to processing.
We discussed possible solutions but agreed this solution is good enough as it is very unlikely this level of concurrency happens in reality. If this problem continues we will need to use record locking.

### DIFF
--- a/app/models/sample_manifest/uploader.rb
+++ b/app/models/sample_manifest/uploader.rb
@@ -56,7 +56,7 @@ class SampleManifest::Uploader
   def process_upload_and_callbacks
     return false unless upload.process(tag_group)
 
-    upload.finished!
+    upload.sample_manifest.finished!
     upload.broadcast_sample_manifest_updated_event(user)
     upload.register_stock_resources
     upload.trigger_accessioning
--- a/app/models/sample_manifest/uploader.rb
+++ b/app/models/sample_manifest/uploader.rb
@@ -74,6 +74,8 @@ class SampleManifest::Uploader
   end
 
   def check_sample_manifest_upload
+    # Refetch the sample manifest to ensure it is in it's newest state
+    upload.sample_manifest.reload
     return true unless upload.sample_manifest&.state == 'processing'
 
     errors.add(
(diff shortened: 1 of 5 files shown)


yoldas (approved):
Good enough to cover most cases. There is a small time window before two uploads set the state of the manifest but it is hard to reproduce.

PR #4530, opened 2024-11-29
style(submission): fixes form layout

BenTopping:

- Fixes style regression from rails 7 upgrade. Some changes to class handling in `form_for` in rails 7.

Before:
Screenshot 2024-11-29 at 11 02 20

After:
Screenshot 2024-11-29 at 11 01 25

### DIFF
--- a/app/views/submissions/_orders.html.erb
+++ b/app/views/submissions/_orders.html.erb
@@ -3,7 +3,7 @@
 <%= vite_javascript_tag('submissions') %>
 
 
-  <%= form_for(:submission, class: 'form-inline') do |f| -%>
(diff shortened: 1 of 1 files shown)


BenTopping:
> I guessing this is related to a poor implementation of inline forms by SS? https://getbootstrap.com/docs/4.6/components/forms/#inline-forms
> 
> It certainly looks a lot better now though!

Yes there is some css media query that looks for `.form-inline .form-group` and applies some flex that breaks the view. This wasn't broken in 6.1 because the `form_for class:'form-inline'` wasn't actually applying styling.

PR #4550, opened 2024-12-10
adding a new plate purpose for multiplexed sequencing submissions, to avoid FC

sabrine33:


dasunpubudumal (approved):

config/default_records/plate_purposes/005_limber_purposes.yml
```diff
@@ -77,3 +77,7 @@ LTN Lib PCR XP:
 RVI Lib PCR XP:
   type: PlatePurpose
   stock_plate: false
+
+# A new plate purpose for multiplexed sequencing submissions,
+# avoiding FC on library quant submission.
+Library Stock Plate: *limber_input_96
```
KatyTaylor, line 83: what does FC stand for?

sabrine33: Fake Cherrypicking - It is used on the Confluence page, so I thought it was a common abbreviation—maybe not after all

KatyTaylor: ah, fake cherrypick. I think a better description might be something like:

```
# A plate provided as ready-made libraries.
# These used to be imported as Stock Plates and then 'fake cherrypicked' onto 'LB-Lib-PCR-XP', which were allowed down the sequencing pipeline.
# Now, they can be imported and submitted for sequencing as this more accurate plate purpose.
```

sabrine33:
closing this - as merging with develop went messy

PR #4555, opened 2024-12-11
Y24-190: Support Limber with Submission creation

sdjmchattie:

- Add support for creating Submissions from a list of Orders and then submitting them.
- Update tests for this new functionality.


app/resources/api/v2/submission_resource.rb
```diff
@@ -2,53 +2,168 @@
...
-      # Custom methods
-      # These shouldn't be used for business logic, and a more about
-      # I/O and isolating implementation details.
-      def lanes_of_sequencing
-        _model.sequencing_requests.size
+      ###
+      # Post-create state changer
+      ###
+
+      after_replace_fields :submit!
```
dasunpubudumal, line 157: Nice and clean!

spec/requests/api/v2/submissions_spec.rb
```diff
@@ -2,44 +2,385 @@
...
+      it 'returns all the resources' do
+        expect(json['data'].count).to eq(resource_count)
+      end
     end
 
-    # Check filters, ESPECIALLY if they aren't simple attribute filters
+    describe '#filter' do
+      let(:target_resource) { resources.sample }
+
+      shared_examples 'it filters the resources correctly' do
```
dasunpubudumal, line 33: This shared example seems to declared in `spec/requests/api/v2/qc_files_spec.rb`, `spec/requests/api/v2/submissions_spec.rb` and in `spec/requests/api/v2/shared_behaviour/labware_shared_behaviour_spec.rb`. 

And all of them seem to be doing the following:

1. Checking if the response is an HTTP succes.
2. Checking if the call returns only one resource.
3. Checking if the call returns the correct resource.

Maybe we can move all these declarations into somewhere in `shared_behavior` directory and reuse it?

sdjmchattie: In case @StephenHulme picks this up (please feel free), I've implemented something like this in my Qcable work.  So probably best to reuse that and go through existing tests, looking to replace this pattern with the shared example.

Shared example:
https://github.com/sanger/sequencescape/blob/Y24-190-support-limber-v2-qcable/spec/requests/api/v2/shared_examples/requests.rb#L138-L150

Usage:
https://github.com/sanger/sequencescape/blob/Y24-190-support-limber-v2-qcable/spec/requests/api/v2/qcables_spec.rb#L30-L57

app/resources/api/v2/submission_resource.rb
```diff
@@ -2,53 +2,168 @@
...
+      end
+
+      # @!attribute [rw] name
+      #   @return [String] the name of the {Submission}.
+      #      This can only be written once on creation.
+      attribute :name, write_once: true
+
+      # @!attribute [w] order_uuids
+      #   This is declared for convenience where the {Order} resources are not available to set as a relationship.
+      #   Setting this attribute alongside the `orders` relationship will prefer the relationship value.
```
yoldas, line 88: I could not understand the doc. 
Setting the order_uuid, finds the order with uuid and overwrites the existing. So, the orders relationship is used. There is no preference.

app/resources/api/v2/submission_resource.rb
```diff
@@ -2,53 +2,168 @@
...
+
+      # @!attribute [rw] used_tags
+      #   @return [String] the tags that were used in this {Submission}.
+      #      This can only be written once on creation.
+      attribute :used_tags, write_once: true
+
+      # @!attribute [w] user_uuid
+      #   This is declared for convenience where the {User} is not available to set as a relationship.
+      #   Setting this attribute alongside the `user` relationship will prefer the relationship value.
+      #   @deprecated Use the `user` relationship instead.
```
yoldas, line 112: I could not understand the doc.
whatever is set last wins. user directly, or indirectly by setting the uuid, which finds the user and sets it. There is no preference.

app/resources/api/v2/submission_resource.rb
yoldas: Where does the after_replace_fields callback come from? I am particularly interested in seeing how it is triggered.

app/resources/api/v2/submission_resource.rb
sdjmchattie: As a user of the endpoint, you could give both a list of UUIDs and a relationship in your creation request.  For example, if you made a request like so:

```json
    {
      "data": {
        "type": "submissions",
        "attributes": {
          "order_uuids": [
            "uuid1",
            "uuid2"
          ]
        },
        "relationships": {
          "orders": [
            { "data": { "type": "orders", "id": "123" } },
            { "data": { "type": "orders", "id": "234" } }
          ]
        }
      }
    }
```

Then the orders on the model at the end of the process will be those defined as relationships.  The `JSONAPI::Resources` gem works through the attributes before it works through the relationships and both are setting the same value on the model, so the relationships always win.

app/resources/api/v2/submission_resource.rb
sdjmchattie: See above

app/resources/api/v2/submission_resource.rb
sdjmchattie: It's a standard Rails callback mechanism implemented by the `JSONAPI::Resources` gem:

It's declared here:
https://github.com/sanger/jsonapi-resources/blob/develop/lib/jsonapi/resource.rb#L21

Generically set up with Rails here:
https://github.com/sanger/jsonapi-resources/blob/develop/lib/jsonapi/callbacks.rb#L45

And gets called via these lines:
https://github.com/sanger/jsonapi-resources/blob/develop/lib/jsonapi/resource.rb#L111
https://github.com/sanger/jsonapi-resources/blob/develop/lib/jsonapi/resource.rb#L51-L53

config/routes.rb
```diff
@@ -64,7 +64,7 @@
       jsonapi_resources :studies
       jsonapi_resources :submission_pools
       jsonapi_resources :submission_templates
-      jsonapi_resources :submissions
+      jsonapi_resources :submissions, except: %i[update]
```
andrewsparkes, line 67: Curious about that restriction and also state being read only.
Wondering if we need to be able to directly update submission state. Or if's all done indirectly through requests...

spec/requests/api/v2/shared_examples/requests.rb
```diff
@@ -51,7 +51,7 @@
 end
 
 shared_examples 'a GET request including a has_one relationship' do |related_name|
-  before { api_get "#{base_endpoint}/#{resource.id}?include=#{related_name}" }
+  before { api_get "#{base_endpoint}/#{resource.id}?include=#{related_name}&fields[#{resource_type}]=#{related_name}" }
```
andrewsparkes, line 54: what's going on with this change (and the one below)?

andrewsparkes (approved):
Looks good, thorough tests again.
Couple of minor questions.

config/routes.rb
sdjmchattie: This pull request takes an endpoint that could not previously be created or updated (i.e. it was `immutable`) and adds a create POST endpoint while still leaving the update PATCH endpoint turned off.  If we need to update state, it's not needed so far, but could be enabled carefully in future.  `immutable` in the resource effectively adds `, only: %i[show]` to the route without explicitly writing it.  I think it's a poor choice by the gem creators and they should expect you to just add that annotation to the routes file instead so it's explicit.  As it is, you cannot tell from the routes alone which are immutable and which are full access.

spec/requests/api/v2/submissions_spec.rb
sdjmchattie: I have now done this!

PR #4562, opened 2024-12-16
Y24-190: Refactor JSONAPI::Resources Processors

sdjmchattie:

- Clean up for the processors to ensure we're keeping them as simple as possible.
- Use callbacks instead of overridden base class methods.
- Move logic to the resource class where it makes sense.


yoldas (approved):
Definitely simpler and easier to read. Very good use of the context hash and callback methods.

Question: 
When you raise JSONAPI::Exceptions, does it return a json output with 4xx code rather than 500? 
I could not figure out what we were doing with the collected errors in the previous code.

sdjmchattie:
> Definitely simpler and easier to read. Very good use of the context hash and callback methods.
> 
> Question: When you raise JSONAPI::Exceptions, does it return a json output with 4xx code rather than 500? I could not figure out what we were doing with the collected errors in the previous code.

It uses whatever code is associated with the types of errors you raised.

It is handled here: https://github.com/sanger/jsonapi-resources/blob/develop/lib/jsonapi/acts_as_resource_controller.rb#L261-L278
Which eventually hits this line and uses the status of the first error in the list: https://github.com/sanger/jsonapi-resources/blob/develop/lib/jsonapi/acts_as_resource_controller.rb#L218

The built in exceptions (which I am using) mostly have a status code of `:bad_request`: https://github.com/sanger/jsonapi-resources/blob/develop/lib/jsonapi/exceptions.rb

PR #4565, opened 2024-12-17
Populate 'sample_public_name' column values with 'sanger_sample_id' column values for RVI-prefixed samples

sabrine33:


sabrine33:
> I'm not as familiar with the data and linkages, but from a rails point of view this looks fine.
> 
> Would you please update the PR title to be more human readable?

I updated the title, thanks

dasunpubudumal (commented):
Do you think that this is more suitable to be a Rake task than a migration? 

It does not involve any schema updates, and it only has data updates.

sabrine33:
> Do you think that this is more suitable to be a Rake task than a migration?
> 
> It does not involve any schema updates, and it only has data updates.

I agree. I initially created it as a Rake task, but later I discovered that the data was being updated/populated through a migration. So, I updated it to follow the convention of doing things.

PR #4578, opened 2025-01-02
Y24-402 PR-2 Migration for number of pools

yoldas:
Closes #

- Add migration to rename the column number_of_samples_per_pool to number_of_pools in request_metadata
- Remove the rake task setting the old option on submissions (no longer necessary; it was an emergency measure in case we didn't get the original bulk submission story done in time for R&Ds 'end to end test')
- Update the tube submission UAT action with the new option (Assume single study-project).
- Rename the attribute number_of_samples_per_pool to number_of_pools in request_metadata_resource

The number_of_pools option is specified for a study-project now. Without the migration, we would get an error "can’t write unknown attribute `number_of_pools`" at the cDNA Prep submission.

The request metadata update makes Limber to receive the expected attribute.

Note that Asset Group plays an important role in bulk submission to set different values of number_of_pools and cells_per_chip_well options for different study-projects. 


app/resources/api/v2/request_metadata_resource.rb
```diff
@@ -28,9 +28,9 @@ class RequestMetadataResource < BaseResource
       # Attributes
       ###
 
-      # @!attribute [r] number_of_samples_per_pool
-      #   @return [Int] the number_of_samples_per_pool.
-      attribute :number_of_samples_per_pool, write_once: true
+      # @!attribute [r] number_of_pools
+      #   @return [Int] the number_of_pools.
```
KatyTaylor, line 32: Bit more detail?

```suggestion
      #   @return [Int] the number_of_pools requested in the Submission. As used in the scRNA Core pipeline, it is specified at the Study-Project level: it will have the same value for all Requests that share the same Study and Project. It is used in the pooling algorithm.
```

KatyTaylor (approved):
Thanks very much for spotting and fixing this. I added one suggestion to add a bit of documentation.

app/resources/api/v2/request_metadata_resource.rb
yoldas: Thanks for the suggestion. I have now updated the comments of both `number_of_pools` and `cells_per_chip_well` attributes.

PR #4586, opened 2025-01-08
Update sample_public_name for RVI Program - Bait Capture study

sabrine33:
Populate the sample_public_name field with sanger_sample_id values for samples associated with the 'RVI Program - Bait Capture' study,  and where the sanger_sample_id starts with "RVI".

### DIFF
--- a/lib/tasks/update_sample_public_name_for_rvi_study.rake
+++ b/lib/tasks/update_sample_public_name_for_rvi_study.rake
@@ -1,0 +1,20 @@
+namespace :samples do
+  desc "Update sample_public_name for RVI Program - Bait Capture study"
+  task update_public_name: :environment do
+    ActiveRecord::Base.connection.execute('SET autocommit = 0')
+
+    ActiveRecord::Base.connection.execute(<<~SQLQUERY)
+      UPDATE sample_metadata
+      JOIN samples ON sample_metadata.sample_id = samples.id
+      JOIN study_samples ON sample_metadata.sample_id = study_samples.sample_id
+      JOIN studies ON study_samples.study_id = studies.id
+      SET sample_metadata.sample_public_name = samples.sanger_sample_id
+      WHERE studies.name = 'RVI Program - Bait Capture' AND sanger_sample_id LIKE 'RVI%';
+    SQLQUERY
+
+    ActiveRecord::Base.connection.execute('COMMIT')
+    ActiveRecord::Base.connection.execute('SET autocommit = 1')
+
+    puts "Sample public names updated successfully for RVI Program - Bait Capture study."
+  end
+end


lib/tasks/update_sample_public_name_for_rvi_study.rake
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+namespace :samples do
+  desc 'Update sample_public_name for RVI Program - Bait Capture study'
+  task update_public_name: :environment do
+    ActiveRecord::Base.connection.execute('SET autocommit = 0')
```
KatyTaylor, line 5: Could you explain the benefit of having autocommit turned off, please?

I was reading the docs - https://dev.mysql.com/doc/refman/8.4/en/innodb-autocommit-commit-rollback.html - and it looks like the benefit would come if there were multiple SQL statements - but we only have one here?

lib/tasks/update_sample_public_name_for_rvi_study.rake
```diff
@@ -0,0 +1,21 @@
...
+
+    ActiveRecord::Base.connection.execute(<<~SQLQUERY)
+      UPDATE sample_metadata
+      JOIN samples ON sample_metadata.sample_id = samples.id
+      JOIN study_samples ON sample_metadata.sample_id = study_samples.sample_id
+      JOIN studies ON study_samples.study_id = studies.id
+      SET sample_metadata.sample_public_name = samples.sanger_sample_id
+      WHERE studies.name = 'RVI Program - Bait Capture' AND sanger_sample_id LIKE 'RVI%';
+    SQLQUERY
```
KatyTaylor, line 14: Did you consider whether to check if `sample_public_name` already has a value, to avoid overwriting it if it has been intentionally set already?

lib/tasks/update_sample_public_name_for_rvi_study.rake
KatyTaylor, line 1: Using SQL directly means not using the ORM... which I think might mean you're bypassing some useful functionality. When a sample (and hopefully sample metadata) is updated, Sequencescape automatically updates the corresponding value in the MLWH (via the RabbitMQ queue). If you directly use SQL, I don't think this will happen.

KatyTaylor (commented):
The SQL looks fine, but I've left a few questions about the approach.

KatyTaylor:
Oh and finally - 

What's your plan for testing? You could add some unit tests, or just test it in training (should definitely do this anyway). If just manual tests in training, you could put a couple of SQL queries here that you intend to run to check the data? 

Thanks!

lib/tasks/update_sample_public_name_for_rvi_study.rake
sabrine33: You are right; it is only one SQL statement, but it runs on all the samples of the RVI studies (quite a lot). Turning off auto-commit ensures the atomicity of the operation.

lib/tasks/update_sample_public_name_for_rvi_study.rake
sabrine33: Good point!

PR #4635, opened 2025-01-22
Y24-213 - Removed `default_includes` from v2 API Resources

### LINKED ISSUE #4233: Y24-213 - Remove use of `default_includes` in API v2 resources and the underlying support for it
# Summary

A decent number of our JSONAPI::Resources use `default_includes` for things like `:uuid_object` and this probably worked 7 years ago when it was added, but this is no longer true.

## What are we doing?

In many resources, one example of which is [PlateTemplateResource](https://github.com/sanger/sequencescape/blob/develop/app/resources/api/v2/plate_template_resource.rb#L14), we are using a method which we [added to BaseResource](https://github.com/sanger/sequencescape/blob/develop/app/resources/api/v2/base_resource.rb#L32-L36) to record which related records we should include when a request is made for the resource.  These default includes are then recorded [via an `inclusions` method](https://github.com/sanger/sequencescape/blob/develop/app/resources/api/v2/base_resource.rb#L38-L40) and used to modify the behaviour of [`apply_includes`](https://github.com/sanger/sequencescape/blob/develop/app/resources/api/v2/base_resource.rb#L42-L49) which JSONAPI::Resources is supposed to call when it wants to find out which related resources to include in a response.

## How do we know it's no longer needed?

As can be seen in the [upgrade checker rake task for JSONAPI::Resources](https://github.com/cerebris/jsonapi-resources/blob/e92afc690f5c53da6f4fa5e91ca72a50873390be/lib/tasks/check_upgrade.rake#L34-L41), they no longer call the `apply_includes` method which is the core of our `default_includes` functionality.  So our implementation does nothing.

## What needs to happen?

* Strip out all the code from BaseResource that handles the storing of `default_includes` and directs it through to the override of `apply_includes`.  We should not be overriding that deprecated method any more.
* Remove `default_includes` from any resources that are specifying it.
* Check every one of those resources to ensure that the response is unchanged after the above changes.

SHIV5T3R:
Closes #4233 

- Removed underlying support and all references to `default_includes` because it is no longer required.

### DIFF
--- a/app/resources/api/v2/aliquot_resource.rb
+++ b/app/resources/api/v2/aliquot_resource.rb
@@ -14,8 +14,6 @@ module Api
     # or look at the [JSONAPI::Resources](http://jsonapi-resources.com/) package for Sequencescape's implementation
     # of the JSON:API standard.
     class AliquotResource < BaseResource
-      default_includes :tag, :tag2
-
       # Associations
       has_one :study
       has_one :project
--- a/app/resources/api/v2/asset_resource.rb
+++ b/app/resources/api/v2/asset_resource.rb
@@ -16,8 +16,6 @@ module Api
     class AssetResource < BaseResource
       attributes :uuid
 
-      default_includes :uuid_object
-
       has_one :custom_metadatum_collection
       has_many :comments, readonly: true
 
--- a/app/resources/api/v2/barcode_printer_resource.rb
+++ b/app/resources/api/v2/barcode_printer_resource.rb
@@ -17,8 +17,6 @@ module Api
     class BarcodePrinterResource < BaseResource
       immutable
 
-      default_includes :uuid_object
-
       ###
       # Attributes
       ###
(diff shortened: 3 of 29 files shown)


stevieing (approved):

dasunpubudumal (commented):
Nice. We need to make sure that the JSONAPO responses haven't been affected by removing this. How do we test it? Is it through the integration suite or do we plan to manually the endpoints pertaining to these resources?

SHIV5T3R:
> Nice. We need to make sure that the JSONAPI responses haven't been affected by removing this. How do we test it? Is it through the integration suite or do we plan to manually test the endpoints pertaining to these resources?

Just finished testing, there's one response that has changed, but it's very insignificant. I can give some extra info, but it was just one tag group where the order of the tags was slightly changed because of duplicate indexes. It's hard to tell if it's a problem with the code or not, because it persists throughout different environments. Aside from that, it should be ready to deploy to UAT.

dasunpubudumal (approved):

PR #4638, opened 2025-01-23
4579-y25-002---bug-duplicate-samples-per-pool-validations

wendyyang:
Closes #

##### remove duplicate error messages display
- 


app/models/submission/validations_by_template_name.rb
```diff
@@ -147,19 +146,10 @@ def process_rows(rows)
...
     plate = Plate.find_from_any_barcode(barcodes.uniq.first)
-    return if plate.nil?
-
-    wells = plate.wells.for_bulk_submission.located_at(well_locations)
-    total_number_of_samples_per_study_project = wells.map(&:samples).flatten.count.to_i
-    number_of_pools = rows.pluck(headers.index(HEADER_NUMBER_OF_POOLS)).uniq.first.to_i
-
-    validate_samples_per_pool(rows, total_number_of_samples_per_study_project, number_of_pools)
+    nil if plate.nil?
   end
```
yoldas, line 152: `validate_for_plates` is not doing any validation. What is the purpose of it?

app/models/submission/validations_by_template_name.rb
```diff
@@ -169,12 +159,8 @@ def validate_for_plates(barcodes, well_locations, rows)
...
   # @return [void]
-  def validate_for_tubes(barcodes, rows)
-    tubes = find_tubes(barcodes)
-    total_number_of_samples_per_study_project = calculate_total_samples(tubes)
-    number_of_pools = extract_number_of_pools(rows)
-
-    validate_samples_per_pool(rows, total_number_of_samples_per_study_project, number_of_pools)
+  def validate_for_tubes(barcodes)
+    find_tubes(barcodes)
   end
```
yoldas, line 164: `validate_for_tubes` is not doing any validation. What is the purpose of it?

app/views/bulk_submissions/_upload.html.erb
```diff
@@ -3,7 +3,6 @@
   <strong>none</strong> of them will be processed. You will get feedback about any problems.</p>
 <% end %>
 <%= semantic_form_for @bulk_submission, html: { multipart: true } do |form|  %>
-  <%= form.semantic_errors :base %>
```
yoldas, line 6: Note: This removes the duplicate listing of the same error messages at the form level because they are already listed at the input control level.

yoldas (changes requested):
I think there is some redundant calls, which look like validations but they are not. There are no error messages added to the errors collection.

app/models/submission/validations_by_template_name.rb
wendyyang: I think it does validate if the plates existing or not, it maybe still need to keep?

app/models/submission/validations_by_template_name.rb
wendyyang: the same as the plates, validate if the tubes exist or not.

app/models/submission/validations_by_template_name.rb
yoldas: A validator adds an error if it is not valid. Where does that happen?

wendyyang: so maybe need to add the error if the plate or tube not exist.

yoldas: Maybe you could check if it is done already in another validation.

You have removed the call to `validate_samples_per_pool` (because it is a duplicate validation); but the preceding lines are only preparing arguments for the removed call. So, the method `validate_for_plates` does not have any purpose now. Do you think the purpose of the method was checking the presence of the plate as well?

wendyyang: The main purpose was definitely not for checking if plate or tube exist or not. Just when I want to remove the whole method, and was wondering if it is safe to remove them all.

wendyyang: I have another look at the code, you are right, in validate_scrna_core_cdna_prep_feasibility, it does find tube/plate, so it doesn't need to have a separate place for checking tube or plate.

PR #4665, opened 2025-02-05
DPL-1047 - Replace jest with vitest

### LINKED ISSUE #4368: Y24-334 - Replace Jest with Vitest
**Describe the Housekeeping**
Replace Jest with Vitest, like in Limber https://github.com/sanger/limber/pull/1848

**Blocking issues**
None

**Additional context**
This will assist with keeping Sequencescape and Limber using common dependencies and allow old and troublesome dependencies like Babel to be removed

BenTopping:
Closes #4368

- Replaces jest with vitest.
- Removes babel.
- Updates Eslint to latest version and updates config format.


BenTopping:
> Looks good, nay great!
> 
> Thanks for putting this together.
> 
> Only thoughts are to check that the updated config files are similar to their Limber counterparts, just to reduce the maintenance burden between the two applications.

Thanks. Limber hasn't moved over to the new ESLint config format yet so the files look a bit different but they are essentially the same.

dasunpubudumal (approved):

PR #4667, opened 2025-02-07
Y24 391 anospp autosubmission for gbs

andrewsparkes:
Closes #https://github.com/sanger/limber/issues/1996

- Added order role to submission template record loader
- Created linear submission template for ANOSPP plus GBS


lib/record_loader/submission_template_loader.rb
```diff
@@ -24,23 +24,35 @@ def generate_derived_options(related_records)
...
 
+    def find_order_role(role)
+      if Rails.env.production?
+        OrderRole.find_by!(role:)
+      else
+        # In development mode or UAT we don't care so much
+        OrderRole.find_by(role:) || UatActions::StaticRecords.order_role
+      end
+    end
+
```
KatyTaylor, line 55: Totally get that you're just copying the pattern below here, so this is not a blocker, but -

I feel like this pattern is setting us up for unexpected failures upon deployment to production:
1. Deploy to UAT, forget to include the creation of the new Order Role / Project / Study
2. Deploy to training, still works
3. Deploy to prod, all of a sudden the deployment fails - didn't get any warning in previous environments.

Perhaps we should enforce their presence in training?

KatyTaylor (approved):
Nice simple solution. Just left one (non-blocking) comment.

lib/record_loader/submission_template_loader.rb
andrewsparkes: Ah good spot, that's not right is it. Idea was for it to be optional. Will fix and add a test.

PR #4674, opened 2025-02-11
4409 y24 379  study setup fields to align to the ena database fields improvement

wendyyang:
Closes #

make the new fields only required for new study
for existing studies, when update, no need to fill in the new fields.
To achieve this,  validate only when create a new study; in the view, add class to label for these new fields when create a new record to show the require mark * 
- 


app/models/study.rb
```diff
@@ -218,10 +218,10 @@ class Study < ApplicationRecord # rubocop:todo Metrics/ClassLength
...
     custom_attribute(:study_name_abbreviation)
 
-    # add ebi library strategy
-    custom_attribute(:ebi_library_strategy, required: true, in: EBI_LIBRARY_STRATEGY_OPTIONS)
-    custom_attribute(:ebi_library_source, required: true, in: EBI_LIBRARY_SOURCE_OPTIONS)
-    custom_attribute(:ebi_library_selection, required: true, in: EBI_LIBRARY_SELECTION_OPTIONS)
+    # add ebi library strategy, ebi library source, ebi library selection
+    custom_attribute(:ebi_library_strategy, in: EBI_LIBRARY_STRATEGY_OPTIONS)
+    custom_attribute(:ebi_library_source, in: EBI_LIBRARY_SOURCE_OPTIONS)
+    custom_attribute(:ebi_library_selection, in: EBI_LIBRARY_SELECTION_OPTIONS)
```
yoldas, line 224: label_option is set to contain `required` for these fields, but it is not enforced here. Is the removal of `required: true` intentional?

app/models/study.rb
```diff
@@ -318,9 +318,9 @@ class Study < ApplicationRecord # rubocop:todo Metrics/ClassLength
                 allow_blank: true
               }
 
-    validates :ebi_library_strategy, inclusion: { in: EBI_LIBRARY_STRATEGY_OPTIONS }
-    validates :ebi_library_source, inclusion: { in: EBI_LIBRARY_SOURCE_OPTIONS }
-    validates :ebi_library_selection, inclusion: { in: EBI_LIBRARY_SELECTION_OPTIONS }
```
yoldas, line 323: Doesn't removal of inclusion validation allow arbitrary values to be passed?

app/views/shared/metadata/edit/_study.html.erb
```diff
@@ -16,9 +16,10 @@
...
       <%= group.text_area(:study_abstract) %>
       <%= group.radio_select(:study_sra_hold, Study::STUDY_SRA_HOLDS) %>
-      <%= metadata_fields.select(:ebi_library_strategy, Study::EBI_LIBRARY_STRATEGY_OPTIONS) %> 
-      <%= metadata_fields.select(:ebi_library_source, Study::EBI_LIBRARY_SOURCE_OPTIONS) %> 
-      <%= metadata_fields.select(:ebi_library_selection, Study::EBI_LIBRARY_SELECTION_OPTIONS) %> 
+     
+      <%= metadata_fields.select(:ebi_library_strategy, options_for_select([['Please select...', nil]] + Study::EBI_LIBRARY_STRATEGY_OPTIONS, metadata_fields.object.ebi_library_strategy), (metadata_fields.object.new_record? ? { required: true } : {})) %>
+      <%= metadata_fields.select(:ebi_library_source,  options_for_select([['Please select...', nil]] + Study::EBI_LIBRARY_SOURCE_OPTIONS, metadata_fields.object.ebi_library_source), (metadata_fields.object.new_record? ? { required: true } : {})) %> 
+      <%= metadata_fields.select(:ebi_library_selection, options_for_select([['Please select...', nil]] + Study::EBI_LIBRARY_SELECTION_OPTIONS, metadata_fields.object.ebi_library_selection), (metadata_fields.object.new_record? ? { required: true } : {})) %>
```
yoldas, line 22: What is the purpose of the nil option? The field is required, `*_OPTIONS` are populated from the ena_requirements_fields.yml and there is a 'Please select...'  placeholder option.

yoldas (changes requested):
- I can't understand how it would work with existing studies. If user edits an existing study for something else, do they have to set values for the EBI_Library_strategy, _source, and _selection now?
- I have a few questions inlined as well.

app/models/study.rb
wendyyang: actually, with the dropdown list provided, there won't be arbitrary values selected, this inclusion maybe redundant.

app/models/study.rb
wendyyang: Yes, it is intentional. Not setting 'required' true in model level allows existing study not having these fields. The label_options show the 'required' mark only in creating page, not in edit page. Does it make sense?

yoldas: Yes, the edit form is used for both creating a study and modifying it. You are applying the validation only when creating  a new study because of `on: :create` .

app/models/study.rb
yoldas: You can tamper with the form though. My question was about removing the existing server side validation. The dropdown list of options does not enforce it.

yoldas (changes requested):
It doesn't list the fields in study details I think.
`app/views/shared/metadata/show/_study.html.erb`

app/models/study.rb
wendyyang: So you think this inclusion need to keep?

wendyyang: I have added the inclusion validation now. Please have a look! Many thanks!

yoldas: As we discussed, we are adding a check if the field is changed now:

```
validates :ebi_library_strategy,
              inclusion: {
                in: EBI_LIBRARY_STRATEGY_OPTIONS
              },
              if: -> { ebi_library_strategy_changed? }

    validates :ebi_library_source,
              inclusion: {
                in: EBI_LIBRARY_SOURCE_OPTIONS
              },
              if: -> { ebi_library_source_changed? }

    validates :ebi_library_selection,
              inclusion: {
                in: EBI_LIBRARY_SELECTION_OPTIONS
              },
              if: -> { ebi_library_selection_changed? }
```

yoldas (approved):
Handles both old and new studies after validation changes.

yoldas (approved):
Thanks. It looks fixed after adding the new EBI fields to the test XML response.

PR #4682, opened 2025-02-13
calculate allowance band for scrna core cdna prep bulk submission

sabrine33:


app/models/order.rb
```diff
@@ -172,15 +172,29 @@ def multiplier_for(request_type_id)
     (request_options.dig(:multiplier, request_type_id.to_s) || 1).to_i
   end
 
+  # Determines and returns calculated request metadata based on the request type key.
+  # This is to cover calculated metadata that does not require user input
+  def calculated_request_metadata_by_request_key(request_type_key)
```
BenTopping, line 177: Is this necessary? If we add allowance_band to https://github.com/sanger/sequencescape/blob/develop/app/models/pbmc_pooling_customer_request.rb does that achieve the same effect?

app/models/bulk_submission.rb
```diff
@@ -161,6 +161,12 @@ def process # rubocop:todo Metrics/CyclomaticComplexity
       # Apply any additional validations based on the submission template name
       apply_additional_validations_by_template_name unless errors.count > 0
 
+      # Calculates the allowance band based on the submission template name.
+      # If the submission template name matches `SCRNA_CORE_CDNA_PREP_GEM_X_5P` and all required headers are present,
+      # the allowance band is calculated for each study and project combination.
+      # Otherwise, an empty hash is assigned.
+      @allowance_band = calculate_allowance_band
```
BenTopping, line 168: Does this need to be calculated here? Could we calculate [here](https://github.com/sanger/sequencescape/pull/4682/files#diff-565afe05f44c168c516703612ee97a04bb1bb874a9b42d23662d33db6d7288fdR369) instead, keeps the logic closer together and removes the need to check the submission name in the calculate_allowance_band method as it has already been checked.

BenTopping (commented):
Couple small suggestions, haven't tested with the limber changes yet. Worth getting a review from Katy too.

app/models/order.rb
sabrine33: good point, it does! but also adding it as a custom_attribute, trigger it to show in the UI as a user input field - which is not

app/models/bulk_submission.rb
sabrine33: We can, but the calculation will run redundantly. The allowance_band is calculated per pool (by project - study), which may result with multiple requests.

app/models/submission/scrna_core_cdna_prep_feasibility_calculator.rb
```diff
@@ -67,4 +96,54 @@ def calculate_total_cells_in_300ul(count_of_samples_in_pool)
...
+    calculate_resuspension_volume(number_of_samples)
+  end
+
+  # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+  def determine_allowance(rows)
+    number_of_cells_per_chip_well = rows.first[headers.index(HEADER_CELLS_PER_CHIP_WELL)].to_i
+    final_volume = calculate_final_volume(rows)
+    case final_volume
+    when ->(v) { v >= calculate_full_allowance(number_of_cells_per_chip_well) }
+      'Full allowance'
```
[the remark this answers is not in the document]
sabrine33: I will do that, thank you for pointing this out, I am also checking your PR, I think I also need to need to return nil for when v < '1 attempt, 1 count' , as for now, I am returning 'no allowance'

app/resources/api/v2/request_metadata_resource.rb
```diff
@@ -43,6 +43,13 @@ class RequestMetadataResource < BaseResource
       #     pooling.
       attribute :cells_per_chip_well, write_once: true
 
+      # @!attribute [r] cells_per_chip_well
+      #   @return [String] the allowance_band requested in the Submission. As
+      #     used in the scRNA Core pipeline, it is specified at the Study-Project
+      #     level: it will have the same value for all Requests that share the
+      #     same Study and Project.
+      attribute :allowance_band, write_once: true
```
KatyTaylor, line 51: I guess this does not currently have to be writeable over the API, as it's calculated internally - so perhaps it should be read only?

app/models/submission/scrna_core_cdna_prep_feasibility_calculator.rb
```diff
@@ -67,4 +103,52 @@ def calculate_total_cells_in_300ul(count_of_samples_in_pool)
...
+    chip_loading_volume + (2 * scrna_config[:volume_taken_for_cell_counting]) + scrna_config[:wastage_volume]
+  end
+
+  # @param number_of_cells_per_chip_well [Integer] the number of cells per chip well from the bulk submission
+  # @return [Float] the "1 pool attempts, 1 count" volume
+  def calculate_one_attempt_one_count(number_of_cells_per_chip_well)
+    # "1 pool attempts, 1 count" =  "Chip loading volume" + "Volume taken for cell counting" + "Wastage volume"
+    chip_loading_volume = calculate_chip_loading_volume(number_of_cells_per_chip_well)
+    chip_loading_volume + scrna_config[:volume_taken_for_cell_counting] + scrna_config[:wastage_volume]
+  end
```
KatyTaylor, line 153: Did you consider consolidating these methods together, something like:

```
  def calculate_volume_needed(number_of_cells_per_chip_well, number_runs, number_cell_counts)
    chip_loading_volume = calculate_chip_loading_volume(number_of_cells_per_chip_well)
    (number_runs * chip_loading_volume) + (number_cell_counts * scrna_config[:volume_taken_for_cell_counting]) + scrna_config[:wastage_volume]
  end
```

and call it like:

```
    case final_volume
    when ->(v) { v >= calculate_volume_needed(number_of_cells_per_chip_well, 2, 2) }
      ALLOWANCE_BANDS[:two_pools_two_counts]
    when ->(v) { v >= calculate_volume_needed(number_of_cells_per_chip_well, 2, 1) }
      ALLOWANCE_BANDS[:two_pools_one_count]
    when ->(v) { v >= calculate_volume_needed(number_of_cells_per_chip_well, 1, 2) }
      ALLOWANCE_BANDS[:one_pool_two_counts]
    when ->(v) { v >= calculate_volume_needed(number_of_cells_per_chip_well, 1, 1) }
      ALLOWANCE_BANDS[:one_pool_one_count]
    end
```

app/models/submission/scrna_core_cdna_prep_feasibility_calculator.rb
```diff
@@ -22,6 +36,28 @@ def calculate_full_allowance(number_of_cells_per_chip_well)
...
+  # Calculates the allowance band for each study and project combination.
+  # This method evaluates the final resuspension volume for each study-project group
+  # and assigns an appropriate allowance band based on predefined thresholds.
+  # @return [Hash] A hash where the keys are hashes containing `:study` and `:project`,
+  #   and the values are the corresponding allowance bands as strings.
+  # Example output:
+  # {
+  #   { study: "Study A", project: "Project X" } => "2 pool attempts, 2 counts",
+  #   { study: "Study B", project: "Project Y" } => "1 pool attempt, 2 counts"
+  # }
```
KatyTaylor, line 48: Nice comment, thank you.

KatyTaylor (commented):
Looks good I think, thank you. Left a few comments for consideration.

app/models/submission/scrna_core_cdna_prep_feasibility_calculator.rb
```diff
@@ -3,25 +3,71 @@
...
 module Submission::ScrnaCoreCdnaPrepFeasibilityCalculator
-  # This method calculates the full allowance volume (in microlitres) for the
-  # specified number of cells per chip well, which is typically specified in
-  # in a bulk submission per study and project. It uses the pooling settings
-  # from the scRNA config. It first calculates the chip loading volume for the
-  # given number of cells per chip well, and then the full allowance for that
-  # chip loading volume.
+  SCRNA_CORE_CDNA_PREP_GEM_X_5P = 'Limber-Htp - scRNA Core cDNA Prep GEM-X 5p'
+  # Constants for the headers in the bulk submission for scRNA core cDNA prep.
+  HEADER_BARCODE = 'barcode' unless defined?(HEADER_BARCODE)
```
BenTopping, line 8: Instead of checking for defined you can use `or-equals` e.g. `HEADER_BARCODE ||= 'barcode'`.

app/models/submission/scrna_core_cdna_prep_feasibility_calculator.rb
```diff
@@ -67,4 +113,26 @@ def calculate_total_cells_in_300ul(count_of_samples_in_pool)
...
+    number_of_samples = calculate_number_of_samples_in_smallest_pool(rows)
+    calculate_resuspension_volume(number_of_samples)
+  end
+
+  # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+  def determine_allowance(rows)
+    number_of_cells_per_chip_well = rows.first[headers.index(HEADER_CELLS_PER_CHIP_WELL)].to_i
+    final_volume = calculate_final_volume(rows)
+    case final_volume
+    when ->(v) { v >= calculate_volume_needed(number_of_cells_per_chip_well, 2, 2) }
```
BenTopping, line 127: This looks like it can be refactored further slightly as they are all checking for volume >= and calling the same func with different params. COpilot might have a clever suggestion.

BenTopping (approved):
Looks good, couple advisories. We will need to wait before merging to see if additional changes are needed after going over the int suite issues.

app/models/bulk_submission.rb
```diff
@@ -352,6 +358,18 @@ def extract_request_options(details)
     end
   end
 
+  # Adds extra request options based on the submission template name.
+  # This method checks the 'template name' from the provided details and
+  # adds additional request options accordingly. Currently, it supports
+  # the `SCRNA_CORE_CDNA_PREP_GEM_X_5P` template by including an
+  # `allowance_band` value.
+  # The allowance_band values are grouped by study and project name.
+  def calculated_request_options_by_template_name(details)
```
yoldas, line 367: Could you give some info about the details argument in the comment? Is it available for each (study, project) pair in our case?

app/models/submission/scrna_core_cdna_prep_feasibility_validator.rb
```diff
@@ -312,7 +311,7 @@ def validate_scrna_core_cdna_prep_full_allowance
       number_of_samples_in_smallest_pool = calculate_number_of_samples_in_smallest_pool(rows)
       number_of_cells_per_chip_well = rows.first[headers.index(HEADER_CELLS_PER_CHIP_WELL)].to_i
       final_resuspension_volume = calculate_resuspension_volume(number_of_samples_in_smallest_pool)
-      full_allowance = calculate_full_allowance(number_of_cells_per_chip_well)
+      full_allowance = calculate_volume_needed(number_of_cells_per_chip_well, 2, 2)
```
yoldas, line 314: number_runs and number_cell_counts arguments are hardcoded values 2  and 2. Where do 2s come from ?

db/migrate/20250214120128_add_allowance_bands_to_request_metadata.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddAllowanceBandsToRequestMetadata < ActiveRecord::Migration[7.0]
+  def change
+    add_column :request_metadata, :allowance_band, :string
```
yoldas, line 4: A comment on the scope and format of allowance_band would be useful to see the purpose of the column. It is per study-project I suppose.

yoldas (changes requested):
I have a few comments.

app/models/bulk_submission.rb
yoldas, line 168: We could call the calculation inside apply_additional_validations_by_template_name in when SCRNA_CORE_CDNA_PREP_GEM_X_5P so we check the template name only once and isolate it from other templates completely.

I see it is added as request_options['allowance_band']. I can't understand the instance variable @allowance_band. Maybe it meant to be named plural because it contains multiple bands per (study, project) .

app/models/submission/scrna_core_cdna_prep_feasibility_validator.rb
sabrine33: @KatyTaylor requested to hard code the number of runs and the number of runs and the number of the cell counts

app/models/bulk_submission.rb
sabrine33: > Could you give some info about the details argument in the comment? Is it available for each (study, project) pair in our case?

I will add some comments for more visibility, but yes we are reading the template_name from details fro validation, also yes it is by study - project pair

db/migrate/20250214120128_add_allowance_bands_to_request_metadata.rb
sabrine33: > A comment on the scope and format of allowance_band would be useful to see the purpose of the column. It is per study-project I suppose.

I don't think it is necessary, at this stage, as the allowance_band is attached to the request

app/models/bulk_submission.rb
sabrine33: > We could call the calculation inside `apply_additional_validations_by_template_name` in `when SCRNA_CORE_CDNA_PREP_GEM_X_5P` so we check the template name only once and isolate it from other templates completely.
> 
> I see it is added as request_options['allowance_band']. I can't understand the instance variable @allowance_band. Maybe it meant to be named plural because it contains multiple bands per (study, project) .

Good point to rename it to plural. 
I did not add the calculation method inside the validation file - where apply_additional_validations_by_template_name is defined, for more visibility and better readability purposes, I suppose it is better to not include it into the validation file

app/models/submission/scrna_core_cdna_prep_feasibility_validator.rb
yoldas: Could you do one of the following please? 
- add them as defined constants or 
- add them in scrna_config in initialisers as options or
- add some comments about what 2s are.

sabrine33: The comments about the static values form the runs and cell count are well explained `calculate_volume_needed` method definition

app/uat_actions/uat_actions/tube_submission.rb
```diff
@@ -221,6 +223,28 @@ def custom_request_options
...
+  # for the scRNA Core testing suite.
+  #
+  # Allowance bands are only supported for bulk submissions using the
+  # Limber-HTP scRNA Core cDNA Prep GEM-X 5p template.
+  # In bulk submissions, allowance_band is determined by the
+  # `calculate_allowance_bands` method.
+  #
+  # The allowance_band is not included in the form, as it is a calculated value (does not require user input).
+  #
+  # @todo: Implement allowance_band support for standard tube submissions.
```
sabrine33, line 236: This is only a quick , 'dirty' fix, to prevent the breakage of the scran core testing suite.
We have a couple of related stories in place:
- Story for supporting normal submission for scRNA Core (we can remove this, once this story is done) 
- Story for bulk submission UAT action

yoldas: Thanks for this. Adding story number to the todo comments may be useful in the future. I think [Y25-153](https://github.com/sanger/sequencescape/issues/4703) in this case.

sabrine33: > Thanks for this. Adding story number to the todo comments may be useful in the future. I think [Y25-153](https://github.com/sanger/sequencescape/issues/4703) in this case.

Thank you very much for adding the story number! I can't agree more!

SHIV5T3R (approved):

sabrine33:
@yoldas, @KatyTaylor I’ve updated the PR by setting `desired_chip_loading_concentration` to a float. Since this value is used in the `final_volume` calculation, keeping it as an integer resulted in incorrect values.

In Ruby on Rails, when dividing two integers, it performs integer division, discarding the decimal part and returning only the whole number. This was causing an incorrect `allowance_band`, which @BenTopping and I suspect led to faulty testing with Abby.

PR #4684, opened 2025-02-13
Revert "Reapply "4409 y24 379  study setup fields to align to the ena databas…"

wendyyang:
Reverts sanger/sequencescape#4640


db/schema.rb
sabrine33, line 1: The schema needs to be rollbacked using rails migration

sabrine33 (changes requested):

db/schema.rb
sabrine33: in SS, you can do : 
`rails db:migrate:status   ` => this is will return all the migration, identify the migration version to rollback, and then run 
`rails db:migrate:down VERSION=your_version_number   `

sabrine33 (approved):

PR #4699, opened 2025-02-20
remove huMFre_code from sample_metadata column using migration

sabrine33:
**_Background history_**:
Initially, the column was removed using a Rails rollback, followed by manually deleting the migration file. However, after deploying the project, it became apparent that the column was not removed from the table in the production environment. Additionally, this process caused confusion regarding the release rollback procedure.

This PR addresses the issue by restoring the migration file and properly removing the column through a new migration.


KatyTaylor (approved):
Looks good I think! Thanks for doing this.

PR #4710, opened 2025-03-03
Y25-171 - Update QC Results page styling

### LINKED ISSUE #5004: Y25-381 - Update styling of QC Results page
**Describe the Housekeeping**
Update styling of QC Results page, to match modern web standards and increase accessibility

**Blocking issues**
None

**Additional context**
Extracted from [Y25-171 - Show plate volumes display](https://github.com/sanger/sequencescape/issues/4712) but without the additional current volume display

StephenHulme:
Closes #5004

- Restyles existing plate displays to meet modern web standards
- Adjusts colours for accessibility and to better integrate with Sequencescape

*Before*
Screenshot 2025-03-03 at 15 17 43

*After*

Screenshot 2025-04-16 at 17 00 57
Screenshot 2025-04-16 at 17 01 35

### DIFF
--- a/spec/features/plate_qc_display_spec.rb
+++ b/spec/features/plate_qc_display_spec.rb
@@ -12,7 +12,7 @@ describe 'Plate QC display' do
     it 'displays an empty table' do
       login_user user
       visit plate_path(plate)
-      %w[concentration volume current_volume quantity_in_nano_grams loci_passed rin].each do |qc_result|
+      %w[concentration volume quantity_in_nano_grams loci_passed rin].each do |qc_result|
         within("##{qc_result}") do
           expect(page).to have_css('td', count: 126)
           expect(page).to have_css('tr', count: 10)
@@ -34,7 +34,6 @@ describe 'Plate QC display' do
         well.qc_results << [
           build(:qc_result_concentration),
           build(:qc_result_volume),
-          build(:qc_result_current_volume),
           build(:qc_result_loci_passed),
           build(:qc_result_rin)
         ]
@@ -48,7 +47,7 @@ describe 'Plate QC display' do
     end
 
     it 'displays the correct data' do
-      %w[concentration volume current_volume quantity_in_nano_grams loci_passed rin].each do |qc_result|
+      %w[concentration volume quantity_in_nano_grams loci_passed rin].each do |qc_result|
         within("##{qc_result}") do
           plate.wells.each_with_index do |well, index|
             expect(page).to have_css("tr[#{1 + index}]/td[2]", text: well.qc_result_for(qc_result))


app/views/plates/show.html.erb
```diff
@@ -1,18 +1,33 @@
...
 
 <h2>QC Results</h2>
 
-<h3>Concentration</h3>
-<%= render partial: "qc_table", locals: { plate: @plate, qc_type: "concentration" } %>
+<h3>
+Concentration (ng/µl)
+<small class="text-muted">Measured from QC File upload</small>
+</h3>
+<%= render partial: "plate_schematic", locals: { plate: @plate, well_values: ['qc_result_for', "concentration"] } %>
```
KatyTaylor, line 9: Probably not a big deal... but is having the method name as a string like this bad security practice? Could allow you to call any method on Well? 🤔 

I get what you're trying to do - it is neater like this, especially as you've added more types of data that come from 'well attribute'.

app/views/plates/show.html.erb
```diff
@@ -1,18 +1,33 @@
+<a href="<%= labware_path(@plate) %>">&lt;&lt; Back to Labware <%= @plate.human_barcode %></a>
 
 <h2>QC Results</h2>
 
-<h3>Concentration</h3>
-<%= render partial: "qc_table", locals: { plate: @plate, qc_type: "concentration" } %>
+<h3>
+Concentration (ng/µl)
+<small class="text-muted">Measured from QC File upload</small>
```
KatyTaylor, line 7: I would personally go for less capitalisation....

`Measured from QC file upload`
`Calculated after bed verification`

I think we have a tendency to promote phrases we say a lot ('Bed Verification') to official proper nouns 💪

app/views/shared/_show_plate.html.erb
```diff
@@ -1,32 +1,39 @@
-
-<table id="plate" width="100%">
+<%# Note: very similar to app/views/plates/_plate_schematic.html.erb %>
```
KatyTaylor, line 1: Then what's the difference? Why are we creating another version? :)

KatyTaylor (approved):
Haven't reviewed in detail but happy for you to merge - looks like some good improvements from the screenshots. Just a few questions to consider.

app/models/well.rb
```diff
@@ -220,13 +220,8 @@ def qc_results_by_key
...
   def qc_result_for(key)
-    result =
-      if key == 'quantity_in_nano_grams'
-        well_attribute.quantity_in_nano_grams
-      else
-        results = qc_results_by_key[key]
-        results.first.value if results.present?
-      end
+    results = qc_results_by_key[key]
+    result = results.first.value if results.present?
```
yoldas, line 224: `quantity_in_nanograms` was returning the computed value `estimated_volume * concentration` after checking nil and negative values. Now the change is returning the value directly. Is this correct?

app/views/plates/_plate_schematic.html.erb
```diff
@@ -0,0 +1,51 @@
...
+<%# Note: very similar to app/views/shared/_show_plate.html.erb %>
+
+<% descriptor = well_values[1].to_s %>
+
+<table class="plate mb-3" id="<%= descriptor %>">
+  <thead>
+    <tr>
+      <% num_rows = plate.height -%>
+      <% num_columns = plate.width -%>
+      <th></th>
```
yoldas, line 14: Does this header column need a css class?
Does it need to specify colspan attribute?

app/views/plates/_plate_schematic.html.erb
```diff
@@ -0,0 +1,51 @@
...
+      <% num_columns.times do |column| -%>
+        <th><%= column + 1 %></th>
+      <% end -%>
+      <th class='border'></th>
```
yoldas, line 18: Similar to above, does this header column need colspan attribute?

app/views/shared/_show_plate.html.erb
```diff
@@ -1,32 +1,39 @@
...
+        <% end -%>
+      <td class='border'><%= row %></td>
     </tr>
   <% end -%>
+  <tr>
+    <th></th>
+    <% plate.plate_columns.each do |header| %>
+      <th><%= header %></th>
+    <% end %>
+    <th></th>
```
yoldas, line 37: does this th need styling and colspan ?

yoldas (commented):
I have left a few comments.

app/views/plates/_current_volume.html.erb
```diff
@@ -0,0 +1,11 @@
+
+<% if value %>
+  <% if value < 10 %>
+    <% colour = "red" %>
+  <% elsif value < 13 %>
+    <% colour = "orange" %>
+  <% else %>
```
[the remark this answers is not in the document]
StephenHulme: Unfortunately history doesn't relate their sources as they are present in the original SS import from 15 years ago.

app/views/plates/_plate_schematic.html.erb
StephenHulme: I don't think so. This is intentionally an empty cell with no specific formatting, rather than a place-holder or space-filler.

app/views/plates/_plate_schematic.html.erb
StephenHulme: The empty column here is intentional: a colspan would imply that the '12' applies to the row labes (A-H) as well as the cell-values, which is not the case.
Screenshot 2025-06-13 at 12 21 09

app/views/shared/_show_plate.html.erb
StephenHulme: No styling or colspan required in this case.

app/views/plates/_plate_schematic.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    <% num_rows.times do |row| -%>
+      <tr>
+        <% rowchar = (65 + row).chr %>
+        <td class='border'><%= rowchar %></td>
+        <% num_columns.times do |column| -%>
+          <% well_name = "#{rowchar}#{column + 1}" %>
+          <% well = plate.find_well_by_name(well_name) %>
+          <% if well %>
+            <% value = begin
+              well.send(well_values[0].to_s, well_values[1].to_s)
```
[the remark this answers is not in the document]
StephenHulme: Good point, yes that is correct.
I've updated f884327c8648fb88ac1693c3d54cfc6164f9b169 to better reflect their usage and include examples.

app/models/well.rb
StephenHulme: Good catch, thanks for checking.

Inspecting `well_attribute.quantity_in_nano_grams` it appears that it handles all the checks that are in `qc_result_for`, specifically nil, negative values, and decimal rounding (only returns integers). It seems that all these scenarios are covered.

```rb
  def quantity_in_nano_grams
    return nil if estimated_volume.nil? || concentration.nil?
    return 0 if estimated_volume < 0 || concentration < 0

    (estimated_volume * concentration).to_i
  end
```

app/views/plates/_current_volume.html.erb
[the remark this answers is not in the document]
yoldas: Yes, this is the problem with magic numbers. After 15 years passed we don't know what they are. Thanks for checking.

app/views/plates/_plate_schematic.html.erb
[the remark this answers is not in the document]
yoldas: Thanks for this change and comments, which make it very easy to read now.

app/models/well.rb
yoldas: Thanks for explanation. Yes, I confirm that the computation estimated_volume * concentration happens in well_attribute.quantity_in_nano_grams and checks are there too.

yoldas (approved):

PR #4748, opened 2025-03-17
4706 y25 157 tol bioscan register lbsn 96 lysate plates in mlwh

wendyyang:
Closes #

add an API endpoint register_stock_for_plate in PlatesController

- 


app/models/plate_creation.rb
```diff
@@ -39,10 +39,17 @@ def create_children!
...
+  attr_accessor :register_stock
 
   private
 
   def record_creation_of_children
     parent.events.create_plate!(child_purpose, child, user)
+    register_stock! if register_stock
+  end
+
+  def register_stock!
```
andrewsparkes, line 51: I don't think it should have an exclamation mark unless it throws exceptions.

maybe `register_stock_for_plate`

andrewsparkes (approved):
Looks good

config/routes.rb
```diff
@@ -41,6 +41,9 @@
       jsonapi_resources :plate_purposes, except: %i[update]
       jsonapi_resources :plate_templates
       jsonapi_resources :plates, except: %i[update]
+      jsonapi_resources :plates, only: [] do
```
andrewsparkes, line 44: not sure you can have 2 `jsonapi_resources :plates` routes?

harrietc52: Yes by adding `only: [] do` I’m not sure if this would reject any request, and only takes in the actions (get/ update etc)

spec/controllers/api/plate_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
harrietc52, line 3: I think the test file wants to be within the v2 directory

spec/controllers/api/plate_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
```
harrietc52, line 4: Alot of the v2 tests live within this https://github.com/sanger/sequencescape/tree/e75001f3e3a487e98cde90dbfe4b8508af63617d/spec/requests/api/v2 folder too, so it might be worth adding tests there aswell

app/controllers/api/v2/plates_controller.rb
```diff
@@ -7,6 +7,25 @@ module V2
     class PlatesController < JSONAPI::ResourceController
       # By default JSONAPI::ResourceController provides most the standard
       # behaviour, and in many cases this file may be left empty.
+      # API endpoint for registering stock for a plate
```
harrietc52, line 10: Thank you for adding comments 👍

app/models/plate_creation.rb
dasunpubudumal: It seems like the bang methods are declaring that the method is [modifying its receiver](https://docs.ruby-lang.org/en/3.0/syntax/methods_rdoc.html#:~:text=The%20bang%20methods,other%20Ruby%20libraries.). If there is no mutation taking place, I think the bang (`!`) is not necessary. For example, in `'hello'.downcase!` `uppercase!` mutates the receiver `'hello'`. There's a non-bang counterpart `downcase` that returns a new string `HELLO` without mutating the original.

dasunpubudumal (approved):
Looks okay, but please make sure that this route does not conflict with the resources generated by `jsonapi_resources :plates` in line number 43. Better to have it tested in a branch before merging it in?

spec/controllers/api/plate_controller_spec.rb
wendyyang: moved

andrewsparkes (approved):

PR #4752, opened 2025-03-18
update manifest conditional format by asset type

sabrine33:


app/models/sample_manifest/generator.rb
```diff
@@ -105,6 +119,7 @@ def attributes
   end
 
   def asset_type
+    return unless configuration.manifest_types.find_by(params[:template])
```
harrietc52, line 122: I think this could be refactored e.g., 

```
def asset_type
  configuration.manifest_types.find_by(params[:template])&.asset_type
end
```

app/models/sample_manifest/column_conditional_format_updater.rb
```diff
@@ -0,0 +1,81 @@
...
+#   conditional_updater = ConditionalFormatUpdater.new(columns: columns, asset_type: 'library')
+#   conditional_updater.update_by_asset_type
+# SampleManifest::Generator
+class SampleManifest::ColumnConditionalFormatUpdater
+  # Predefined conditional formats for required fields.
+  REQUIRED_COLUMNS_CONDITIONAL_FORMATS = [
+    SequencescapeExcel::ConditionalFormatting.new(
+      name: 'empty_mandatory_cell',
+      options: {
+        'type' => :cellIs,
```
harrietc52, line 15: What is `cellIs`

app/models/sample_manifest/column_conditional_format_updater.rb
```diff
@@ -0,0 +1,81 @@
...
+# This class is responsible for updating the conditional formatting for columns in the manifest.
+# The formatting is applied based on the asset type, specifically targeting columns related to library assets.
+#
+# Example Usage:
+#   conditional_updater = ConditionalFormatUpdater.new(columns: columns, asset_type: 'library')
+#   conditional_updater.update_by_asset_type
+# SampleManifest::Generator
+class SampleManifest::ColumnConditionalFormatUpdater
+  # Predefined conditional formats for required fields.
+  REQUIRED_COLUMNS_CONDITIONAL_FORMATS = [
```
harrietc52, line 11: Messaged re an alternative/ additional way of doing this :)

app/models/sample_manifest/generator.rb
sabrine33: this check is added to not break the unit tests, as not all the test contains a 'valid' template - the check is added for the template. I can not image a valid use case for this check in prod.

app/models/sample_manifest/column_conditional_format_updater.rb
sabrine33: > Messaged re an alternative/ additional way of doing this :)

The proposed solution is to modify the columns.yml config file. This will result in modifying the column format for all templates where the columns in question are present. Here, the column background is set to red, which may cause confusion, as the columns may be present without necessarily having the "not empty" constraint. For example, in the multiplex library manifest types, the library_type column is present but not required to be filled in.

app/models/sample_manifest/generator.rb
```diff
@@ -28,8 +28,22 @@ def initialize(params, user, configuration)
...
+  # the asset type requires specific formatting rules (e.g., for library assets).
+  #
+  # @return [Array<Column>] The columns associated with the specified template, with conditional formatting
+  #         updated if any.
+
   def columns
     @columns ||= configuration.columns.find(params[:template])
+    return unless asset_type
+    conditional_updater = SampleManifest::ColumnConditionalFormatUpdater.new(columns: @columns, asset_type: asset_type)
+    conditional_updater.update_column_formatting_by_asset_type
```
dasunpubudumal, line 46:
```suggestion
    SampleManifest::ColumnConditionalFormatUpdater.new(columns: @columns, asset_type: asset_type).update_column_formatting_by_asset_type
```

dasunpubudumal: Also, do we need to instantiate an instance of `ColumnConditionalFormatUpdater` every time `column` is called? Might `update_column_formatting_by_asset_type` a class method with `@columns` and `asset_type` as inputs? 

It's fine either way, but seems like it's some work for Ruby GC each time `columns` is invoked 😄

sabrine33: > Also, do we need to instantiate an instance of `ColumnConditionalFormatUpdater` every time `column` is called? Might `update_column_formatting_by_asset_type` a class method with `@columns` and `asset_type` as inputs?
> 
> It's fine either way, but seems like it's some work for Ruby GC each time `columns` is invoked :)

`ColumnConditionalFormatUpdater` is instanced only once with all the columns for the manifest to generate and the manifest asset_type, it's its internal method `update_columns_formatting` that loops through the columns that need formatting.

harrietc52 (commented):
Sorry it's taken me a while to get back to this. 
I’m just wondering whether there is a way to split up the `app/models/sample_manifest/column_conditional_format_updater.rb` file to make it as general as the name suggests, as it is currently quite specific to the library/asset code. 
Maybe the library/ asset specific code could be passed in as a sort of validation, and then this file could be reused and generic, as I think its a good layer of abstract to have between the column config not being coupled with every manifest. I’m sure there is a design pattern for this which I’ll try dig out!
Equally, this works so ok if you would like to keep it as is thats fine. Maybe worth seeing what other members of the team say in code reviews. I’m happy to try help design it out if we decide to go down that route

sabrine33:
> Sorry it's taken me a while to get back to this. I’m just wondering whether there is a way to split up the `app/models/sample_manifest/column_conditional_format_updater.rb` file to make it as general as the name suggests, as it is currently quite specific to the library/asset code. Maybe the library/ asset specific code could be passed in as a sort of validation, and then this file could be reused and generic, as I think its a good layer of abstract to have between the column config not being coupled with every manifest. I’m sure there is a design pattern for this which I’ll try dig out! Equally, this works so ok if you would like to keep it as is thats fine. Maybe worth seeing what other members of the team say in code reviews. I’m happy to try help design it out if we decide to go down that route

No worries at all! I see what you mean about making column_conditional_format_updater.rb more generic. asset_type is indeed more generic than the manifest, as different manifests can share the same asset type.

I agree that decoupling the column formatting logic from the manifest could be beneficial, though currently, we rely on the manifest type to generate the spreadsheet and apply formatting. I’ve kept the class general and extensible to accommodate similar future requirements, but for now, it only checks for asset_type = library.

app/models/sample_manifest/column_conditional_format_updater.rb
KatyTaylor: Where is the code that makes `library_type` required for one manifest type but not the other?

I would prefer a different approach, as this adds a new place where columns are configured - it's all quite neat at the moment, in `columns.yml`.

Can't find the words to explain this properly, but I also don't think that 'by asset type' will turn out to be a common thing to control the requiredness. 

Another thing we can do is, if we want a column to behave differently in one manifest type vs another, we make a new column definition for it - e.g. see `collected_by_for_cardinal`, `collected_by_for_controls` and `collected_by_for_scrna_core`, which all save to the 'collected_by' column, but have different configuration, to be used in different manifest types.

KatyTaylor (changes requested):
Hey, haven't reviewed the code fully but I have a concern about the approach - see other comment. Thanks.

sabrine33:
> Hey, haven't reviewed the code fully but I have a concern about the approach - see other comment. Thanks.

Thank you for the pointing out to the right direction - it makes more sense and it is more straightforward !

app/models/sample_manifest/column_conditional_format_updater.rb
sabrine33: > Where is the code that makes `library_type` required for one manifest type but not the other?
> 

It is happening here: 
```
 def update_column_formatting_by_asset_type
    if library_asset?
      update_columns_formatting(%w[library_type insert_size_from insert_size_to], REQUIRED_COLUMNS_CONDITIONAL_FORMATS)
    else
      @columns
    end
  end
```

> I would prefer a different approach, as this adds a new place where columns are configured - it's all quite neat at the moment, in `columns.yml`.
> 
> Can't find the words to explain this properly, but I also don't think that 'by asset type' will turn out to be a common thing to control the requiredness.
> 
> Another thing we can do is, if we want a column to behave differently in one manifest type vs another, we make a new column definition for it - e.g. see `collected_by_for_cardinal`, `collected_by_for_controls` and `collected_by_for_scrna_core`, which all save to the 'collected_by' column, but have different configuration, to be used in different manifest types.

Thank you for pointing to right direct, this is seems to put a simpler approach!

PR #4769, opened 2025-03-24
BGE change to multiplexing config for BGE to create BGE Cap Lib Pool Norm tube

andrewsparkes:
Rather than the generic LB Lib Pool Norm


yoldas (approved):
Yes, this has a new request type to customise the target_purpose_name to BGE Cap Lib Pool Norm.

request filter for BGE ISC MX pipeline config in Limber needs changing too. There is a similar request type needed for the BGE PCR Free MX part.

PR #4779, opened 2025-03-25
Y24-365 - RVI GSLP Pipeline config

### LINKED ISSUE #4394: As an R&D scientist I would like to have a separate submission and library type required for GSLP RVI project
**User story**
As an R&D scientist I would like to have a separate submission and library type required for GSLP RVI project.

Di R, Scott T, Ben T, Andrew S

Di R, Scott T

**Acceptance criteria**
- [ ] Add new submission template for GSLP RVI: `Limber-Htp - GSLP RVI Bait Capture Library - NovaSeq 6000 Paired end sequencing`. Confirm with stakeholders the name and whether a sequencing request is needed.
- [ ] Add a new library type for the GSLP RVI pipeline. This should be confirmed with NPG
- [ ] Add a new request type for the GSLP RVI pipeline.

**Dependencies**
This story will most likely need to be done in tandem with the Limber pipeline configuration: https://github.com/sanger/limber/issues/1975

**Additional context**
This story will require similar changes to those of the original RVI BCL story. The [relevant PR is here](https://github.com/sanger/sequencescape/pull/3939/files).

BenTopping:
Closes #4394 

- Adds RVIG Submission
- Adds RVIG-BCL library type
- Adds RVIG Request type.
- Adds ability to ReISC RVIG-BCL pipeline

### DIFF
--- a/config/default_records/request_types/004_limber_high_throughput.yml
+++ b/config/default_records/request_types/004_limber_high_throughput.yml
@@ -38,6 +38,7 @@ limber_reisc:
     - LB Lib PCR-XP
     - LTN Lib PCR XP # for Targeted NanoSeq ReISC
     - RVI Lib PCR XP # for RVI Bait Capture Library
+    - RVIG Lib PCR XP # for RVIG Bait Capture Library
   library_types:
     - Agilent Pulldown
     - Twist Pulldown
@@ -46,6 +47,7 @@ limber_reisc:
     - Targeted NanoSeq Pulldown Agilent
     - BGE
     - RVI-BCL
+    - RVIG-BCL
 limber_pcr_free:
   <<: *limber_htp_library
   name: Limber PCR Free
--- a/config/default_records/plate_purposes/005_limber_purposes.yml
+++ b/config/default_records/plate_purposes/005_limber_purposes.yml
@@ -78,6 +78,11 @@ RVI Lib PCR XP:
   type: PlatePurpose
   stock_plate: false
 
+# RVIG BCL plate for reISC
+RVIG Lib PCR XP:
+  type: PlatePurpose
+  stock_plate: false
+
 # A plate provided as ready-made libraries.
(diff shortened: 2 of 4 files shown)


config/default_records/request_types/004_limber_high_throughput.yml
BenTopping, line 1: Wondering if these changes need to be in a separate PR as RVIG Lib PCR XP and RVIG-BCL won't exist in production because of WIP, so could fail. Ill do some testing

andrewsparkes: Yes I've had that issue before and had to use a separate wipped file. It's the only solution with record loader atm I think.

config/default_records/request_types/004_limber_high_throughput.yml
```diff
@@ -38,6 +38,7 @@ limber_reisc:
     - LB Lib PCR-XP
     - LTN Lib PCR XP # for Targeted NanoSeq ReISC
     - RVI Lib PCR XP # for RVI Bait Capture Library
+    - RVIG Lib PCR XP # for RVIG Bait Capture Library
```
andrewsparkes, line 41: it may be that record loader won't automatically try to update the acceptable purpose added in this file, in which case you could separate the RVIG Lib PCR XP purpose into a wip file.
On the other hand, what's the harm if you define that purpose and have it in here if there's no way to use that plate type.
I think the PR is good as is.

andrewsparkes: You will want to check what it does on deployment, in case you need to manually add that acceptable purpose via server rails console.

andrewsparkes (approved):
I think this is fine as is, creating that extra purpose does no harm.
You may need to manually add it to the acceptable purposes for re-isc in rails console after deploy.

config/default_records/request_types/004_limber_high_throughput.yml
BenTopping: Works fine locally, it creates both the library type and RVIG Lib PCR XP plate. I think that should be fine but I will also test on UAT.

PR #4801, opened 2025-03-28
Update column formatting for required fields in library manifests

sabrine33:


sabrine33:
> Thank you, much simpler. One more uncertainty:
> 
> How did you choose which manifest types to update? You've updated 5 out of the 8 that include the library_type column (didn't try searching for the other fields).
> 
> On a similar note, I'm wondering whether we need both `library_type` AND `library_type_required` - or whether we just need one (required) version.
> 
> Thanks.

I run through some manual testing - the multiplex manifests does not require the library type (present but not required) input, that also should explain the reason behind defining the library_type_required column type

harrietc52 (approved):

Issue #4730, opened 2025-03-11
edit title here


seenanair:
**User story**

Following [Y24-222](https://github.com/sanger/limber/issues/1814), the following questions were raised in internal discussions (between Katy, Abdullah, and Andrew). 
Katy and Seena have gone through existing production data, and these were our findings:

1. Are the lab users still using the 'tag group' dropdowns? – Yes, all the time (for the CustomTaggedPlate page).
3. Which pipelines are they using the above in? – Bespoke Chromium and scRNA core.
5. Which tag groups are they using? – See the following list

**Who are the primary contacts for this story**
Katy, Seena

**Who is the nominated tester for UAT**
Matt J

e.g. John S (don't include surnames in public repos)

**Acceptance criteria**

Consider if these features can be _featured flagged_ to decouple testing and deployment.
To be considered successful the solution must allow:
- [ ] Create tagset entries for all tag_groups using record loader. (eg : config/default_records/tag_sets/001_chromium_single_index_tag_sets.yml)

1.  10X_PlateTT Set A i7
2.  10X_PlateTT Set A i5
3.  TS_pWGSC_UDI96_i7
4.  TS_pWGSC_UDI96_i5
5.  Dual Index Kit TN Set A 10xgenomics i7 - ColumnWise
6.  Dual Index Kit TN Set A 10xgenomics i5_a ColumnWise
7.  TS_pWGSD_UDI96_i7
8.  TS_pWGSD_UDI96_i5
9.  TS_pWGSB_UDI96_i7
10.  TS_pWGSB_UDI96_i5
11.  pWGSA_UDI_tempi7s
12.  pWGSA_UDI_tempi5s
13.  Chromium Genome (Tagset for this is be already updated as part of [Y24-285](https://github.com/sanger/sequencescape/issues/4304)
14.  Sanger_168tags - 10 mer tags
15.  Test Bespoke PCR Free (This is used a dual index one according to query results. Need to be checked.

- [ ] Modify tag_set_loader to fail quietly in development environment and noisily in prod if the corresponding tag_groups mentioned for a tagset creation in file were not found in tag_groups table
- [ ] Check whether any tag groups above has got  identical tag list. If, so check whether any can be deprecated after communicating with Katy and the users 

**Dependencies**

This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
https://github.com/sanger/limber/issues/1814
https://github.com/sanger/sequencescape/issues/4238

[...]

PR #4807, opened 2025-03-31
Y25 185 tagsets creation for all taggroups in custom tagged plate page

seenanair:
Closes https://github.com/sanger/sequencescape/issues/4730

- The following is a list of tag groups used on the CustomTaggedPlate page since 2020-01-01 
  - Entries marked in red are those used only once or twice, which appear to be invalid or created by error. This has been verified with Katy and Matt.
    **Tagsets have been created only for the green-marked entries, as they are used more frequently.**
   - Entries marked in red are those used only once or twice, which appear to be invalid or created by error. This has been verified with Katy and Matt.
   
Screenshot 2025-03-31 at 14 55 59

- Additionally, the following tag group entries contain similar tags (oligos). After verification, these need to be set as "visible": false in the production data.
[visible_tag_groups_with_duplicate_tags.csv](https://github.com/user-attachments/files/19534205/visible_tag_groups_with_duplicate_tags.csv)

The SQL used to identify this is as follows
```sql
WITH tag_group_data AS (
    SELECT 
        t.tag_group_id,
        g.name AS tag_group_name,
        GROUP_CONCAT(oligo ORDER BY oligo) AS tag_list
    FROM 
        tags t
    JOIN 
        tag_groups g ON t.tag_group_id = g.id
	WHERE
		g.visible = true
    GROUP BY 
        tag_group_id
),
groups_with_number_duplicates AS (
    SELECT
        tag_list, 
        GROUP_CONCAT(tag_group_name) AS tag_group_list,
        Count(*) AS number_groups
    FROM 
        tag_group_data
    GROUP BY 
        tag_list
)
SELECT 
    *
FROM 
    groups_with_number_duplicates
HAVING 
    number_groups > 1;

```

- The tag_set_loader is updated to fail silently in development environment and noisly in other environments

### DIFF
--- a/config/default_records/tag_sets/001_tag_sets_for_tag_groups.yml
+++ b/config/default_records/tag_sets/001_tag_sets_for_tag_groups.yml
@@ -1,0 +1,61 @@
+10X_Plate TT Set A:
+  name: 10X_Plate TT Set A
+  tag_group_name: 10X_Plate TT Set A i7
+  tag2_group_name: 10X_Plate TT Set A i5
+384plex_11mers:
+  name: 384plex_11mers
+  tag_group_name: 384plex_11mers
+Dual Index Kit TN Set A 10Xgenomics - ColumnWise:
+  name: Dual Index Kit TN Set A 10Xgenomics - ColumnWise
+  tag_group_name: Dual Index Kit TN Set A 10Xgenomics i7 - ColumnWise
+  tag2_group_name: Dual Index Kit TN Set A 10Xgenomics i5 - ColumnWise
+Haplotag 96:
+  name: Haplotag 96
+  tag_group_name: Haplotag 96 i7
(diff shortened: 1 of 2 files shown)


KatyTaylor (approved):
Looks good. 

Will probably fail in UAT after the db drop and reseed is done (or maybe even before) - so that might bring up the discussion again about whether to add them to the seed data, or ignore errors in UAT too!

seenanair:
> Looks good.
> 
> Will probably fail in UAT after the db drop and reseed is done (or maybe even before) - so that might bring up the discussion again about whether to add them to the seed data, or ignore errors in UAT too!

Yes, or may be while creating tag sets , we can create tag groups if missing in develop or UAT. What do you think?

KatyTaylor:
How about if we make it fail silently in development + UAT (because we don't expect realistic data), and noisily in training + prod (because we do).

PR #4660, opened 2025-02-03
4579-y25-002---bug-duplicate-samples-per-pool-validations_v2

wendyyang:
Closes #

#### Changes proposed in this pull request
remove duplicate validation
remove duplicate error messages display
- 

#### Instructions for Reviewers

_[All PRs] - Confirm PR template filled_  
_[Feature Branches] - Review code_  
_[Production Merges to `main`]_  
 &nbsp; &nbsp; \- _Check story numbers included_  
 &nbsp; &nbsp; \- _Check for debug code_  
 &nbsp; &nbsp; \- _Check version_

Issue #4579, opened 2025-01-03
Y25-002 - [BUG] Duplicate Samples per Pool validations for scRNA cDNA Prep Submission


yoldas:
**Describe the bug**
Both of the stories below implement the "Sample per Pool" validation for scRNA cDNA Prep Submission with different error messages. Pool sizes for a study-project are found by the division number_of_samples / number_of_pools. The size of some of the pools (or all) will be equal to the quotient. The size of some of the pools will be the quotient plus one if there is a remainder.

- Y24-402 - Specify 'no. pools' rather than 'no. samples per pool' in submission #4423
Excerpt from acceptance criteria: "Number of samples per pool' (to be calculated - see additional context) must be between 5 and 25 (inclusive)"
Error messages:
There is a separate error message for each possible pool. They are numbered as pool number 1, pool number 2, pool number 3, and so on.
Comments: 
There is no need to check the range and repeat the same error for each possible pool. There are only two possible sizes (quotient and quotient plus one), independent of how many pools there are (one to eight). 
The number of error messages can be as high as 8 while only the "pool number" is different in them.
Screenshot: See the red rectangle below

- Y24-429 - Check feasibility of cDNA Prep submission #4460
Excerpt from acceptance criteria: "The number of pools requested must be feasible given the number of samples"
Error messages:
There is a separate error message only for the smallest and biggest pools. If the smallest and biggest pools are the same or the number_of_pools is one, the error is reported only for the smallest pool.
Comments:
Only the smallest and biggest pools sizes are checked and reported.
Screenshot: See the blue rectangle below

The following screenshot shows an example when the number of samples is 10 and number of pools is 3. 

number_of_samples/number_of_pools = 10/3
Quotient = 3 
Remainder = 1
Pool sizes: 3, 3, 4
Smallest pool size: 3
Biggest pool size: 4

<img width="461" alt="Screenshot 2025-01-03 at 00 59 42" src="https://github.com/user-attachments/assets/c5e0682a-3a8c-4d9e-a6e9-4baf12700540" />

**RT Ticket Number**
N/A

**To Reproduce**
Steps to reproduce the behaviour:
See the notes for manual testing at https://github.com/sanger/sequencescape/issues/4460#issuecomment-2568560391 .
Change the number of pools in the spreadsheet to make the quotient of the division number_of_samples / number_of_pools or the quotient plus one outside of the range 5 to 25.

**Expected behaviour**
- [ ] Remove the duplicated validation of the same condition.
[...]

PR #4843, opened 2025-04-14
Y25-179 - Bulk submission upload error messages

### LINKED ISSUE #4717: Y25-179 - [BUG] uploading bulk submission file fails without true reason
When  uploading bulk submission file, where samples have not been accessioned, this fails with
'There was a problem when building your submissions'

https://sequencescape.psd.sanger.ac.uk/bulk_submissions/new

Reported in [SR-23137](https://sanger.freshservice.com/a/tickets/23137)

Should report the error rescued

(additional info: SSR fed back that the samples did not have accessions and this was due to the study not having been accessioned)

Note: Testing the accessioning service requires this to be turned on in UAT/Training envs which at present it isn't

Changes to deployment project required (`@ay6` in Slack @psd-user-requests 05/03/2015 13:00) 

config template: `roles/deploy_sequencescape/templates/config.rb.j2`

`accession_url` and `accession_view_url` for different environments: 
[same for all environments and `disable_accession_check: true` for both UAT and Training. We might change URLs to EBI test URLs to enable it for all environments to ensure isolation from production]
`roles/deploy_sequencescape/vars/main.yml`

credentials: 
[same used for all environments, we might change them per environment to ensure isolation from production]
```
environments/uat/group_vars/all/vault.yml
environments/training/group_vars/all/vault.yml
environments/production/group_vars/all/vault.yml
```

BenTopping:
Closes #4717 

- Changes location for new bulk submission RecordInvalid error handling.
- Restores `form.semantic_errors` which was removed from here: https://github.com/sanger/sequencescape/pull/4660
  - This is required for errors on base (bulk submission) to appear on the form.
  - This was originally removed to fix a duplicate error message bug: https://github.com/sanger/sequencescape/issues/4579. I have confirmed this change does not break this behaviour and users only see 1 set of error messages.
- Adds test to bulk_submission model spec for accessioning checks.

#### Additional notes
- There are no test cases for bulk submission controller tests but I have added one to the model that covers the logic change.
- To test this locally you need to set `configatron.disable_accession_check = false`

### DIFF
--- a/app/models/bulk_submission.rb
+++ b/app/models/bulk_submission.rb
@@ -206,6 +206,8 @@ class BulkSubmission # rubocop:todo Metrics/ClassLength
               ] = "Submission #{submission.id} built (#{submission.orders.count} orders)"
             rescue Submission::ProjectValidation::Error => e
               errors.add :spreadsheet, "There was an issue with a project: #{e.message}"
+            rescue ActiveRecord::RecordInvalid => e
+              errors.add :base, e.message
             end
           end
         end
--- a/app/controllers/bulk_submissions_controller.rb
+++ b/app/controllers/bulk_submissions_controller.rb
@@ -33,8 +33,7 @@ class BulkSubmissionsController < ApplicationController
       render action: 'new'
     end
   rescue ActiveRecord::RecordInvalid => e
-    flash.now[:error] = 'There was a problem when building your submissions'
-    @bulk_submission.errors.add(:base, e.message)
+    flash.now[:error] = "There was a problem when building your submissions: #{e.message}"
     render action: 'new'
   end
 
--- a/app/controllers/bulk_submissions_controller.rb
+++ b/app/controllers/bulk_submissions_controller.rb
@@ -33,7 +33,7 @@ class BulkSubmissionsController < ApplicationController
       render action: 'new'
     end
   rescue ActiveRecord::RecordInvalid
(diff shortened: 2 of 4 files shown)


app/models/bulk_submission.rb
```diff
@@ -206,6 +206,8 @@ def process # rubocop:todo Metrics/CyclomaticComplexity
               ] = "Submission #{submission.id} built (#{submission.orders.count} orders)"
             rescue Submission::ProjectValidation::Error => e
               errors.add :spreadsheet, "There was an issue with a project: #{e.message}"
+            rescue ActiveRecord::RecordInvalid => e
+              errors.add :base, e.message
```
dasunpubudumal, line 210: `e.message` prints a bit of weird messages like "Validation failed: Orders request options read length for the LO flow cell kit the user can select a Read Length of 150". I don't know if we need to sanitise them but I guess they are comprehendable.

dasunpubudumal (approved):

app/models/bulk_submission.rb
BenTopping: Yeah I agree its a bit ugly. I will have a look see if there is a quick win but I think this is a symptom of how we/rails handle error messages generally.

dasunpubudumal: Yeah I feel like unless we override something it would be a bit difficult to get around with. But I don't know if we should be overriding anything.

PR #4873, opened 2025-04-25
add actions to element aviti batches

sabrine33:
Dev testing 
 Run  `bundle exec rake application:post_deploy` to create the tasks
1. Create Element Aviti submissions.
2. From the Element Aviti pipeline page, create a batch.
3. On the batch page, under the Action section, there should be two rows: 'Confirm Element PhiX used' and 'Loading', as shown in the image below:
Screenshot 2025-04-25 at 12 24 59


lib/record_loader/task_loader.rb
```diff
@@ -0,0 +1,49 @@
...
+    #
+    # @return [Task, nil] The created or updated Task, or `nil` if the Workflow is not found in specific environments.
+    # @raise [ActiveRecord::RecordNotFound] If the Workflow is not found in environments other than
+    # `development`, `staging`, or `cucumber`.
+    def create_or_update!(name, options)
+      workflow_name = options.delete('workflow')
+      workflow = find_workflow!(workflow_name, name)
+      return nil if workflow.nil?
+      options[:pipeline_workflow_id] = workflow.id
+      Task.find_or_initialize_by(name: name, pipeline_workflow_id: workflow.id).tap { |task| task.update!(options) }
```
yoldas, line 31: This will always issue an update on the task at post_deploy. Is it intended? Can it be done using find_or_create_by?

lib/record_loader/task_loader.rb
```diff
@@ -0,0 +1,49 @@
...
+    #
+    # @param name [String] The name of the Task.
+    # @param options [Hash] The options for creating or updating the Task.
+    #
+    # @return [Task, nil] The created or updated Task, or `nil` if the Workflow is not found in specific environments.
+    # @raise [ActiveRecord::RecordNotFound] If the Workflow is not found in environments other than
+    # `development`, `staging`, or `cucumber`.
+    def create_or_update!(name, options)
+      workflow_name = options.delete('workflow')
+      workflow = find_workflow!(workflow_name, name)
```
yoldas, line 28: Note: workflow record is created by the pipeline loader.

lib/record_loader/task_loader.rb
sabrine33: I used `_create_or_update_` as I needed to update some bits while working on it  - I agree probably best to use `find_or_create_by`

Issue #4630, opened 2025-01-21
Y25-037 - At a TL (Tris B) I would like to the actions on the batch page for Element Aviti sequencing batches to represent the workflow for these sequencers.


TWJW-SANGER:
**User story**

At a TL (Tris B) I would like to the actions on the batch page for Element Aviti sequencing batches to represent the workflow for these sequencers.

**Who are the primary contacts for this story**
Tris B
Mike Q
Liz C
See our Confluence project page for "Element Aviti" for a link to Mockups and flowcell details.

**Who is the nominated tester for UAT**
Tris B ?

**Acceptance criteria**
Consider if these features can be _featured flagged_ to decouple testing and deployment.

To be considered successful the solution must allow, on the /batches/xxxxx pages for Element Aviti submissions:

- [x] The Actions list to contain only ~"Confirm Element PhiX used" and~ "Loading"

**Additional context**
After discussing with the NPG team and the Data QC team, we agreed to remove the 'Confirm Element PhiX used' task action.

PR #4874, opened 2025-04-25
loading action descriptors for Element Aviti pipeline

sabrine33:
**Dev testing notes:**
This is should be deployed after [Y25-037](https://github.com/sanger/sequencescape/issues/4630)
1. Run `bundle exec rake application:post_deploy` to create the descriptors 
2. Create an element aviti batch
3. On the batch view page, actions section, click on Loading
4. The loading descriptors should be displayed successfully, as below:
Screenshot 2025-04-25 at 15 32 58
5. User should be able to save and edit the task details values (Saving is performed by clicking on "Next Step" button)

### DIFF
--- a/lib/tasks/record_loader/descriptor.rake
+++ b/lib/tasks/record_loader/descriptor.rake
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+# This file was automatically generated via `rails g record_loader`
+namespace :record_loader do
+  desc 'Automatically generate Descriptor through DescriptorLoader'
+  task descriptor: :environment do
+    RecordLoader::DescriptorLoader.new.create!
+  end
+end
+
+# Automatically run this record loader as part of record_loader:all
+# Remove this line if the task should only run when invoked explicitly
+task 'record_loader:all' => 'record_loader:descriptor'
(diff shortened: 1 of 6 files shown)


lib/tasks/record_loader/descriptor.rake
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# This file was automatically generated via `rails g record_loader`
+namespace :record_loader do
+  desc 'Automatically generate Descriptor through DescriptorLoader'
+  task descriptor: :environment do
+    RecordLoader::DescriptorLoader.new.create!
```
yoldas, line 7: It looks like Descriptor depends on the Task. Since it comes first alphabetically by directory name, record loading may fail. The dependencies of the rake task can be listed after the environment, see for example lib/tasks/record_loader/pipeline_request_information_type.rake

yoldas (changes requested):

lib/tasks/record_loader/descriptor.rake
sabrine33: Thank you @yoldas I was not aware of this, I will check 'pipeline_request_information_type' file

PR #4885, opened 2025-04-29
Experimenting with sync asset creations

dasunpubudumal:
Closes #

- Makes the assert link creation synchronous for transfers between plates and tubes


andrewsparkes (approved):
Approving your change, but don't understand why it was asynchronous before. Performance reasons I'd assume, but even it makes 96 tubes from a plate (and 96 asset links) does that take long?
Bit nervous we're not understanding why it was originally done this way.

Also, if we're not doing it asynchronously, do we even need the AssetLink::Job classes? Should the code e.g. be on Labware?

PR #4892, opened 2025-05-01
Y25-258 - Enable 50PE Read Length Option in NovaSeqX Submissions

wendyyang:
…ncing

Closes #

### add test cases to spec/models/submission/submission_creator_spec.rb & spec/models/bulk_submission_spec.rb
- 


app/models/nova_seq_x_sequencing_request.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+class NovaSeqXSequencingRequest < SequencingRequest
+end
```
yoldas, line 3: For NovaSeqX, The `illumina_htp_novaseqx_paired_end_sequencing` request type specifies the `request_class_name` as `HiSeqSequencingRequest` . 

This change creates a new model, `NovaSeqXSequencingRequest` and it is not referred anywhere, except the test written in the PR. How is this new model used in the code?

I think, the tests are not addressing the change in the read length options of the request type validator `ReadLengthRequestedNovaseqX` correctly.

yoldas (changes requested):

app/models/nova_seq_x_sequencing_request.rb
wendyyang: Thank you for the comments. I have made the changes: 
I  use 'HiSeqSequencingRequest' for illumina_htp_novaseqx_paired_end_sequencing, and remove the new model 'NovaSeqXSequencingRequest'.
I also add a new test for RequestType::Validator for the read length changes.
I keep the other tests as I though they test if the read length options to include 50 in the submission form.
Can you review again please?

yoldas (approved):
Looks good but I suspect the post_deploy command will not update the existing record in the target database. As deployment notes:

Verify that valid_options column shows the correct YAML list after deployment:

```sql
select * from request_type_validators rtv where rtv.key = 'ReadLengthRequestedNovaseqX';
```

Update valid_options in the database either by writing a rake task or using the sequencescape console on the server:
```ruby
rtv = RequestType::Validator.find_by(key: 'ReadLengthRequestedNovaseqX')
rtv.valid_options = [50, 100, 150]
rtv.save!
```

stevieing (approved):
I think you are correct that the test is arbitrary but in this case it does test the correct thing so it looks fine.

PR #4906, opened 2025-05-08
adding element aviti pipeline tasks

sabrine33:
Closes #

- 

### DIFF
--- a/lib/record_loader/task_loader.rb
+++ b/lib/record_loader/task_loader.rb
@@ -35,7 +35,8 @@ module RecordLoader
     def find_workflow!(workflow_name, task_name)
       Workflow.find_by!(name: workflow_name)
     rescue ActiveRecord::RecordNotFound
-      raise ActiveRecord::RecordNotFound, "Task '#{task_name}' creation or update failed because there was no workflow named '#{workflow_name}'"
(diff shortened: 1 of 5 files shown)


lib/record_loader/task_loader.rb
```diff
@@ -0,0 +1,48 @@
...
+      Task.create_with(options).find_or_create_by!(name: name, pipeline_workflow_id: workflow.id)
+    end
+
+    private
+
+    def find_workflow!(workflow_name, task_name)
+      return unless workflow_name
+      Workflow.find_by!(name: workflow_name)
+    rescue ActiveRecord::RecordNotFound
+      message = "Task '#{task_name}' creation or update failed because there was no workflow named '#{workflow_name}'"
```
yoldas, line 39: The exception is caught and logged but it just continues with something broken. If workflow is missing, the deployment will continue with a logging. 

Exception handling and environment based logging is redundant. Just let it fail.

There is already a dependency setup on the pipeline task to make sure the workflow is created before. All what is done above is just for sanity anyway.

yoldas (changes requested):
Please remove redundant exception handling in the task loader. Let all environments behave the same and let the deployment fail.

It shouldn't because you already defined pipeline dependency that creates the workflow record.

PR #4908, opened 2025-05-08
adding loading task descriptors to the Element Aviti workflow

sabrine33:


lib/tasks/record_loader/descriptor.rake
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# This file was automatically generated via `rails g record_loader`
+# The descriptor creation should follow the task creation
+# Tasks creation should follow the workflow creation — workflows are created within the Pipeline loader
+#
+namespace :record_loader do
+  desc 'Automatically generate Descriptor through DescriptorLoader'
+  task descriptor: [:environment, 'record_loader:pipeline', 'record_loader:task'] do
```
yoldas, line 9: Note: This dependency already makes sure that they are executed first. As long as the loaders are idempotent, this is enough to cover the associations queried.

yoldas (changes requested):

PR #4921, opened 2025-05-14
[bug] Adjust role form spacing and options sorting

StephenHulme:
The form needed 5 minutes of TLC. It feels better now.

- Sorted studies and projects list alphabetically
- Adjust spacing of Roles checkboxes and form fields

Before:
Screenshot 2025-05-14 at 14 52 48

After:
Screenshot 2025-05-14 at 14 51 53

### DIFF
--- a/app/views/admin/users/_add_role.html.erb
+++ b/app/views/admin/users/_add_role.html.erb
@@ -3,11 +3,11 @@
 <h3><%= role_class_name %></h3>
 <div class='text-center'>
   <%= form_tag [:grant_user_role_admin,@user], remote: true, data: {success: "#role_list"}, class: 'form-inline remote-form' do -%>
-    &nbsp; <label for="role_authorizable_name"><%= role_class_name %> role</label> &nbsp;
-    <%= select "role", "authorizable_name", @all_roles, { prompt: true }, {class: 'form-control'} %>
+  <label for="role_authorizable_name"><%= role_class_name %> role</label>
+    <%= select "role", "authorizable_name", @all_roles, { prompt: true }, {class: 'form-control-sm mx-2'} %>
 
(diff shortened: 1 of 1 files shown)


app/views/admin/users/_add_role.html.erb
```diff
@@ -3,12 +3,16 @@
 <h3><%= role_class_name %></h3>
 <div class='text-center'>
   <%= form_tag [:grant_user_role_admin,@user], remote: true, data: {success: "#role_list"}, class: 'form-inline remote-form' do -%>
-    <label for="role_authorizable_name"><%= role_class_name %> role</label>
+    &nbsp; <label for="role_authorizable_name"><%= role_class_name %> role</label> &nbsp;
```
BenTopping, line 6: Can we use padding / margin instead of non breaking spaces?

app/helpers/bootstrap_helper.rb
```diff
@@ -92,8 +92,8 @@ def pagination(collection)
   end
 
   # <div class="col-md-size form-group sqs-form"></div>
-  def form_group(&)
-    tag.div(class: 'form-group row sqs-form', &)
+  def form_group(classes = '', &)
```
BenTopping, line 95: Would this be better making `classes = ' '` -> `class = 'form-group row sqs-form'` and then `:class` in the tag.div. Makes it fully overrideable.

BenTopping (commented):
Nice improvement.

app/helpers/bootstrap_helper.rb
BenTopping: Yes, I guess it depends on how customisable we want it to be. Fine to leave as is 👍

PR #4938, opened 2025-05-21
DPL-903 Bioscan - Add automatic library prep submissions

andrewsparkes:
Closes #https://github.com/sanger/limber/issues/1381

Modifies Bioscan configuration to allow automatic library prep submissions.
Removes no longer needed Additional Input class


app/uat_actions/uat_actions/test_submission.rb
```diff
@@ -323,12 +348,8 @@ def default_purpose_name
 
   # Any helper methods
 
-  def study
-    UatActions::StaticRecords.study
-  end
-
   def project
-    UatActions::StaticRecords.project
+    Project.find_by(name: project_name) || UatActions::StaticRecords.project
```
yoldas, line 352: Similar to study, if project_name is explicitly specified, it needs a validation.

config/default_records/plate_purposes/009_bioscan_purposes.yml
```diff
@@ -12,11 +12,18 @@ LILYS-96 Stock:
...
 LBSN-96 Lysate:
-  type: PlatePurpose::AdditionalInput
+  stock_plate: false
+  cherrypickable_target: false
+# This is the lysed material plate, ready for the Library prep pipeline.
+# This version of the Lysate plate is created by the faculty team manually,
+# and input via manifest.
+LBSN-96 Lysate Input:
+  type: PlatePurpose::Input
+  input_plate: true
```
yoldas, line 27: Separate type attribute is redundant if input_plate is set. See app/models/plate_purpose.rb

config/default_records/plate_purposes/009_bioscan_purposes.yml
```diff
@@ -12,11 +12,18 @@ LILYS-96 Stock:
...
-# This is the lysed material plate, ready for the Library prep pipeline. These plates
-# are either prepared by the faculty team manually, or are created on the Beckman robot
-# from the LILYS-96 Stock parent.
-# These plates include controls.
+# This is the lysed material plate, ready for the Library prep pipeline.
+# This version of the Lysate plate is created on the Beckman robot from
+# the LILYS-96 Stock parent.
+# These plates include 2 added controls.
 LBSN-96 Lysate:
-  type: PlatePurpose::AdditionalInput
```
yoldas, line 20: Explicitly showing `input_plate: false ` would be useful to see the contrast with the `Input` version of this purpose.

config/default_records/submission_templates/005_bioscan_submission_templates.yml
```diff
@@ -7,11 +7,13 @@ Limber-Htp - Bioscan Lysate Prep:
     request_type_keys: ["limber_bioscan_lysate_prep"]
     product_line_name: Bioscan
     product_catalogue_name: Bioscan
-# Library preparation from the Lysate plate
+# Library preparation from the Lysate Input plate
 # NB. PacBio long read Sequencing happens in Traction for Bioscan
-Limber-Htp - Bioscan Library Prep:
```
yoldas, line 12: The old template needs superceded_by (note the spelling) to hide it from use, either (SUPERCEDED_BY_UNKNOWN_TEMPLATE = -2) or the id of the new submission. It could be added to deployment notes.

yoldas (changes requested):
I have suggested some changes for consideration.

spec/uat_actions/test_submission_spec.rb
```diff
@@ -228,4 +258,170 @@
...
+    before do
+      allow(UatActions::GeneratePlates).to receive(:default).and_return(generator)
+      allow(generator).to receive(:plate_purpose_name=)
+      allow(generator).to receive(:well_count=)
+      allow(generator).to receive(:well_layout=)
+      allow(generator).to receive(:number_of_samples_in_each_well=)
+      allow(generator).to receive(:study_name=)
+      # rubocop:disable RSpec/SubjectStub
+      allow(test_submission).to receive(:default_purpose_name).and_return(default_purpose_name)
+      # rubocop:enable RSpec/SubjectStub
```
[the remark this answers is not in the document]
andrewsparkes: I couldn't see how to write the test without doing this.

spec/uat_actions/test_submission_spec.rb
```diff
@@ -228,4 +258,170 @@
...
+      it 'configures the generator with the provided values' do
+        test_submission.send(:setup_generator)
+
+        expect(generator).to have_received(:plate_purpose_name=).with('Custom Purpose')
+        expect(generator).to have_received(:well_count=).with(48)
+        expect(generator).to have_received(:well_layout=).with('Random')
+        expect(generator).to have_received(:number_of_samples_in_each_well=).with(3)
+        expect(generator).to have_received(:study_name=).with('Test Study')
+      end
+      # rubocop:enable RSpec/ExampleLength
```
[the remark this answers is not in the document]
andrewsparkes: split into small tests

app/uat_actions/uat_actions/test_submission.rb
andrewsparkes: added

config/default_records/plate_purposes/009_bioscan_purposes.yml
andrewsparkes: yes, added

config/default_records/submission_templates/005_bioscan_submission_templates.yml
andrewsparkes: Yes ok, will add to deployment notes

yoldas: `superseded_by_id` to be set to either `-2` or the **`id` of the new template**. I gave the SS constant name to show where the number comes from.

config/default_records/submission_templates/005_bioscan_submission_templates.yml
KatyTaylor, line 1: I see you've got Deployment Notes on the story corresponding to your 'default records' changes - thanks. Not sure whether we should encourage using SQL directly though as it bypasses the validation that sits at the Rails model level.

config/default_records/submission_templates/005_bioscan_submission_templates.yml
```diff
@@ -7,11 +7,13 @@ Limber-Htp - Bioscan Lysate Prep:
...
-  submission_class_name: "LinearSubmission"
```
KatyTaylor, line 13: They definitely don't want an option to do an SSR/manual submission here?

config/default_records/submission_templates/005_bioscan_submission_templates.yml
```diff
@@ -7,11 +7,13 @@ Limber-Htp - Bioscan Lysate Prep:
     request_type_keys: ["limber_bioscan_lysate_prep"]
     product_line_name: Bioscan
     product_catalogue_name: Bioscan
-# Library preparation from the Lysate plate
+# Library preparation from the Lysate Input plate
```
KatyTaylor:
```suggestion
# Library preparation from the Lysate or Lysate Input plate
```

config/default_records/submission_templates/005_bioscan_submission_templates.yml
```diff
@@ -7,11 +7,13 @@ Limber-Htp - Bioscan Lysate Prep:
...
-Limber-Htp - Bioscan Library Prep:
-  submission_class_name: "LinearSubmission"
+Limber-Htp - Bioscan Library Prep - Automated:
+  name: "Limber-Htp - Bioscan Library Prep - Automated"
+  submission_class_name: "AutomatedOrder"
   related_records:
     request_type_keys: ["limber_bioscan_library_prep"]
     product_line_name: Bioscan
     product_catalogue_name: Bioscan
+    project_name: "BIOSCAN"
```
KatyTaylor, line 19: Does it set the Study correctly, using 'autodetect_studies=true`, using the Study that the samples are linked to in the sample manifest?

app/models/state_changer/standard_plate.rb
```diff
@@ -76,13 +77,18 @@ def transfer_requests
...
     # Note: Do *NOT* go through labware here, as you'll pull out all requests
     # not just those associated with the wells in the 'contents' array
+    # e.g. contents might be just the wells you are failing.
     def associated_requests
       # uniq is present here in case more than one well shares the same request
       # e.g. a parent well was split into multiple child wells
-      receptacles.flat_map(&:aliquot_requests).uniq
+      aliquot_requests = receptacles.flat_map(&:aliquot_requests).uniq
+      requests_as_source = receptacles.flat_map(&:requests_as_source).uniq
+      aliquot_requests + requests_as_source
```
KatyTaylor, line 113: (Maybe redundant if you agree with comment below)

In the interests of changing one thing at a time -

Consider refactoring out `aliquot_requests` and `requests_as_source` out into separate methods, and making `pending_orders` call `aliquot_requests` instead and therefore be unchanged.

I'm uncertain whether this is correct and whether perhaps `library_start` events might already be being called in weird places... but probably best to tackle that as a separate issue?

KatyTaylor: I'm not sure about this change, it seems potentially a bit far-reaching. It's treating requests coming into the well in the same breath as requests coming out of the well . What about if we are passing the plate rather than failing? Then it would try to pass the request coming out AND the one coming in? The Limber UI may avoid this by only showing the automated submission button when the plate is already passed, but that doesn't seem watertight. I think we don't normally want to apply the same state change to both requests at once - possibly only when failing.

config/default_records/submission_templates/005_bioscan_submission_templates.yml
andrewsparkes: So is it better to do these as migration updates?

andrewsparkes: Use rails console commands

config/default_records/submission_templates/005_bioscan_submission_templates.yml
andrewsparkes: Ask SSR

config/default_records/submission_templates/005_bioscan_submission_templates.yml
andrewsparkes: set BIOSCAN study here?

app/models/state_changer/standard_plate.rb
andrewsparkes: Agree, reverting change

Issue #4817, opened 2025-04-02
Investigate impact of cherrypicking bug


KatyTaylor:
**Description**
Extract some data to be used for the investigation into the cherrypicking bugs, which are detailed in https://github.com/sanger/sequencescape/issues/4413

**Acceptance Criteria**
The following outputs from the research are produced:
- [ ] Find all cherrypicks that happened in timeframe **to be provided by Liz C**, where the picked sample volume was less than 5ul
- [ ] Find all cherrypicks that happened in same timeframe, where the picked buffer volume was exactly 5ul

After the above, I think we should show the data to the stakeholders, and ask what they want to do next. One idea mentioned was assessing how many of these samples failed library prep.

**Who the primary contacts are for this work**
Katy

**Knowledge or Stake holders**
Liz C, Scott T, Nat S, Ruth B

KatyTaylor:
Liz confirmed the date when they starting using 5ul as the robot minimum picking volume was September 2023.

andrewsparkes:
MySQL Queries:
==================================
Where picked volume < 5

```
WITH cherrypick_reqs AS (
	SELECT r.id, 
           r.state,
           r.submission_id,
           r.user_id,
           r.asset_id,
           r.target_asset_id,
           req_types.request_class_name
	FROM requests r
    JOIN request_types req_types ON req_types.id = r.request_type_id
    WHERE req_types.request_class_name like '%Cherrypick%'
	-- this uses all request types with cherrypick in their class names, which may be incorrect 
	-- Includes:
	--   CherrypickForFluidigmRequest
	--   CherrypickForPulldownRequest
	--   PooledCherrypickRequest
	-- changeover date for the 5ul volume was 23rd Sept 2024
	AND r.created_at > '2023-09-01 00:00:01'
    AND r.state = 'passed'
),
requests_as_source AS (
	SELECT samp.id AS sample_id,
		samp.sanger_sample_id,
		samp_meta.supplier_name AS sample_supplier_name,
		cherrypick_reqs.id AS request_id,
        cherrypick_reqs.request_class_name,
        cherrypick_reqs.state AS request_state,
		cherrypick_reqs.submission_id,
        subm.state AS submission_state,
        subm.created_at AS submission_created_at,
        subm.updated_at AS submission_last_updated_at,
		cherrypick_reqs.user_id,
		labw.id AS plate_id,
[...]

PR #4939, opened 2025-05-21
Y24-383 - Cherrypicking Improvements

### LINKED ISSUE #4413: Y24-383 - Cherrypicking improvements (minimum volume)
**User story**
As SSR and Sample Management stakeholders, I would like the following behaviour changes to the cherrypicking pipeline.

Stephen, Katy

Liz C, Lynda

**Acceptance criteria**
- [x] If required volume < robot minimum picking volume, round up to the robot minimum picking volume*
- [ ] If source volume < robot minimum picking volume, warn the user (but allow them to continue)
- [x] Change the robot minimum picking volume to only apply to sample (V) not buffer (B), in amount and concentration picks**
- [x] Change the final destination volume to equal the actual picked volume (V+B), in amount and concentration picks
- [x] Prevent the user from entering negative amounts when cherrypicking, at the form level.
- [x] Rename fields (in UI) as follows (for clearer user experience):
    - [x] Volume Required (µl) --> Target Volume Required (µl)
    - [x] Concentration Required (µl) --> Target Concentration Required (µl)
    - [x] Minimum Volume (µ/l) --> Minimum Target Volume (µl)
    - [x] Maximum Volume (µ/l) --> Maximum Target Volume (µl)
    - [x] Volume (µl) --> Volume to pick (µl)

**Dependencies**
- https://github.com/sanger/sequencescape/issues/4412

**References**
- https://github.com/sanger/sequencescape/issues/4410
- https://github.com/sanger/sequencescape/issues/4411

**Additional Context**

Screenshot of where the renaming of fields will appear:
Image

*this is because liquid handlers struggle to pick up very small volumes. If you want it to pick up 1μl, the robot is more successful at picking it up if it is programmed to pick up 5μl (because it builds up more negative pressure). Based on experiments Scott T (R&D) did many years ago.

**this is because the viscosity of the liquid is important. We don't know what the viscosity of the sample will be, so we make the safer assumption that it will be high (and take precautions - minimum picking volume). We know the buffer will have low viscosity so the robot can pick up small volumes more easily.

borhara:
Closes #4413 

- Modified label names in cherrypicking forms
- Input validation for negative values
- The final volume is rounded up if it's less than the minimum robot picking volume
- Excluded the minimum picking volume from the buffer calculation in volume and concentration picks
- The current volume calculation now includes the buffer volume


andrewsparkes:
Hi @borhara ,
I noticed when doing the other cherrypicking story (https://github.com/sanger/sequencescape/issues/4817), that the various volume columns in the well_attributes table did not 'add up' in many cases or some were null. e.g. picked_volume 50 when buffer volume was 168 and current_volume nil or other nonsense values. Does this story improve that situation? Or do you think we'd need another story to address that, to maybe have all calculations that set those columns go through the same module or something?

app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
```diff
@@ -76,9 +76,10 @@ def volume_to_cherrypick_by_nano_grams_per_micro_litre(
       source_volume < robot_minimum_pick_vol ? source_volume : source_volume_to_tell_robot_to_pick
 
     well_attribute.picked_volume = source_volume_to_tell_robot_to_pick
-    well_attribute.buffer_volume =
-      calculate_buffer_volume(final_volume_desired, source_volume_it_will_actually_pick, robot_minimum_pick_vol)
+    well_attribute.buffer_volume = calculate_buffer_volume(final_volume_desired, source_volume_it_will_actually_pick)
+
     well_attribute.robot_minimum_picking_volume = robot_minimum_pick_vol
+    well_attribute.current_volume = robot_minimum_pick_vol if well_attribute.current_volume < robot_minimum_pick_vol
```
KatyTaylor, line 82: As we were discussing yesterday -
The `well_attribute.current_volume` (volume on the target well) is set in 2 places - here and earlier on line 50, where it sets it to the desired target volume.

It would have previously been incorrect in the case where the required buffer volume was v low (<5), and had to be rounded up to 5 because of the robot minimum pick vol - taking the target well volume over the desired target volume.

I can't think of an edge case where the current approach would be incorrect anymore, now that we've removed the rounding up of buffer to the robot minimum pick vol.

However, I still think we should modify it to set the target well volume to the source picked + the buffer picked - because:
a) it matches the acceptance criteria on the story
b) some other change in future might mean that this approach breaks again, and
c) it would make the code more consistent with the other equivalent files for picking by amount and volume.

Could you make that change, please?

spec/models/well_spec.rb
```diff
@@ -422,6 +422,19 @@
...
+    #   measured_conc: 1,
+    #   measured_vol: 1,
+    #   min_vol: 100,
+    #   max_vol: 100,
+    #   min_pick_vol: 0,
+    #   source_pick_vol: 1,
+    #   buffer_vol: 99,
+    #   final_src_vol: 0,
+    #   final_dest_vol: 100
+    # }
```
StephenHulme: Should this be commented?

spec/models/well_spec.rb
```diff
@@ -705,7 +719,22 @@
...
+        source_volume: 0,
+        robot_minimum_pick_volume: 0,
+        source_volume_picked: 0,
+        buffer_volume_picked: 100,
+        destination_volume: 100,
+        source_volume_remaining: 0
       }
+      # Given a conc scenario where you want 100 total final volume with 20 source volume, it will pick 20 from source
+      # and add 80 buffer, giving a total of 100. if you have 0 source volume, and you desire to pick 20 from source
+      # you will get 0 volume and 100 buffer for a total of 100
```
StephenHulme: Can we put a summary of this in the scenario field? `scenario: 'Y24-382: Test 2 - pick a total of 100 ul comprising of 0 vol and 100 buffer',`

spec/models/well_spec.rb
borhara: I think both tests should be commented out. I just used them to investigate Scott's hypothetical. Although there'd be no harm in having them both in, it's just extra coverage.

borhara:
> I'm not sure if there's an issue with the patch coverage report, but would you please double-check that all new code is coveraged by tests?

It should be. The lines I've changed have forced changes to the tests because they were failing, so I'm not sure how the lines aren't covered.

app/models/cherrypick/volume_by_nano_grams_per_micro_litre.rb
```diff
@@ -79,7 +74,11 @@ def volume_to_cherrypick_by_nano_grams_per_micro_litre(
     well_attribute.buffer_volume = calculate_buffer_volume(final_volume_desired, source_volume_it_will_actually_pick)
 
     well_attribute.robot_minimum_picking_volume = robot_minimum_pick_vol
-    well_attribute.current_volume = robot_minimum_pick_vol if well_attribute.current_volume < robot_minimum_pick_vol
+
+    # Set current volume to the sum of the actual picked source volume and buffer volume.
+    # In most cases, this matches the requested final volume, except in rare cases where
+    # the robot's minimum pick causes more buffer to be added than intended.
```
KatyTaylor: This comment isn't valid anymore, as we've changed it so that the robot minimum pick does not apply to the buffer.

PR #4941, opened 2025-05-21
generate a Sample Sheet to load into the Element Aviti sequencer

sabrine33:
Closes #

- 

### DIFF
--- a/app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
+++ b/app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+# This module is responsible for generating a sequencing run manifest file for the Element Aviti machines.
+# The run manifest format can support different configurations, but this implementation is tailored to
+# the specific template used in Sanger's lab.
+#
+# The generated file includes structured sections:
+# - A [SETTINGS] section with static sequencing parameters used in the lab.
+# - A [SAMPLES] section containing static PhiX control sample entries.
+# - A dynamic sample section populated from the batch object.
+#
+# The sequencing settings section is static, defining the consistent parameters applied in our lab.
+# The PhiX control samples section is also static, as no metadata about the actual PhiX controls is stored.
+# Element recommends using a consistent sample name that clearly identifies the control sequences.
+#
+# Allowed characters in the output file name include: letters, numbers, dashes, dots, parentheses, and underscores.
+
+module AvitiSampleSheet::SampleSheetGenerator
+  def self.generate(batch)
+    Generator.new(batch).generate
+  end
+  class Generator
+    def initialize(batch)
+      @batch = batch
+    end
+
+    def generate
+      CSV.generate(row_sep: "\r\n") do |csv|
+        append_settings_section(csv)
+        append_phix_controllers(csv)
+        append_samples_section(csv)
+      end
+    end
+
+    # Appends the [SETTINGS] section to the manifest.
+    # These settings define parameters for Bases2Fastq execution and are fixed for lab usage.
+    # The 'Lane' value of '1+2' indicates that the setting applies to both lanes.
+    # Users may manually update this section if necessary
+    def append_settings_section(csv)
+      csv << ['[SETTINGS]']
+      csv << %w[SettingName Value Lane]
+      csv << ['# Replace the example adapters below with the adapter used in the kit.']
+      csv << %w[R1Adapter AAAAAAAAAAAAAAAAAAA 1+2]
+      csv << %w[R1AdapterTrim FALSE 1+2]
+      csv << %w[R2Adapter TTTTTTTTTTTTTTTTTTT 1+2]
+      csv << %w[R2AdapterTrim FALSE 1+2]
+    end
+
+    # Appends static PhiX control sample rows to the manifest.
+    # This section is not dynamically generated, as control metadata is not stored.
+    # Users may manually update this section if necessary.
+    def append_phix_controllers(csv)
+      csv << ['[SAMPLES]']
+      csv << %w[SampleName Index1 Index2]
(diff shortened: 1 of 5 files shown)


app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    # Appends batch-specific sample rows to the manifest.
+    # Each line represents a sample, including its tags, lane and the study associated with.
+    # requests are filtered to exclude failed ones.
+    def append_samples_section(csv)
+      csv << [
+        '# Fill in the correct sample schema associated with the Adept Preparation Workflow for all sequenced samples.',
+        '',
+        '',
+        'Lane',
```
nerdstrike, line 69: Please hoist Lane and Project up to the section column name. They're no use in a comment row, and NPG have to guess which column is Project to correctly permission the data in archive. It's much safer for us if these columns are correctly labelled at the start of the sample section.

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    # Appends the [SETTINGS] section to the manifest.
+    # These settings define parameters for Bases2Fastq execution and are fixed for lab usage.
+    # The 'Lane' value of '1+2' indicates that the setting applies to both lanes.
+    # Users may manually update this section if necessary
+    def append_settings_section(csv)
+      csv << ['[SETTINGS]']
+      csv << %w[SettingName Value Lane]
+      csv << ['# Replace the example adapters below with the adapter used in the kit.']
+      csv << %w[R1Adapter AAAAAAAAAAAAAAAAAAA 1+2]
```
yoldas, line 43: Check if there needs to be between R1Adapter, sequence, and lane number specification.

yoldas: Please comment on where these sequences came from. Consider moving them to visible constants.

app/controllers/batches_controller.rb
```diff
@@ -350,6 +351,11 @@ def find_batch_by_barcode
     end
   end
 
+  def generate_sample_sheet
+    csv_string = AvitiSampleSheet::SampleSheetGenerator.generate(@batch)
+    send_data csv_string, type: 'text/plain', filename: "batch_#{@batch.id}_run_manifest.csv", disposition: 'attachment'
```
yoldas, line 356: Even if the content disposition is attachment, showing 'text/plain' is browser dependent usually. It will navigate and show in window or inline it if it can.

app/models/presenters/batch_submenu_presenter.rb
```diff
@@ -56,6 +56,19 @@ def build_submenu
...
+
+    # This is used to determine if we need to display the Aviti run manifest option
+    # in the batch submenu.
+    # @return [Boolean] true if the batch is released and has Element Aviti requests
+    def aviti_run_manifest?
+      @batch.released? && aviti_requests?
+    end
+
+    def aviti_requests?
+      @batch.requests.any? { ElementAvitiSequencingRequest }
```
yoldas, line 71: How does this work? Is this checking the type of the request or request class?

app/models/presenters/batch_submenu_presenter.rb
```diff
@@ -56,6 +56,19 @@ def build_submenu
       add_submenu_option pluralize(@batch.comments.size, 'comment'), batch_comments_path(@batch)
       load_pipeline_options
       add_submenu_option 'NPG run data', "#{configatron.run_data_by_batch_id_url}#{@batch.id}"
+      return unless aviti_run_manifest?
+      add_submenu_option 'Download Sample Sheet', id: @batch.id, controller: :batches, action: :generate_sample_sheet
```
yoldas, line 60: Note: the generate_sample_sheet action added to the batches controller is Element Aviti specific. generate_sample_sheet works with AvitiSampleSheet::SampleSheetGenerator only at the moment.

yoldas (commented):
I have left a few comments. I can revisit once all the tests pass and discussions are over.

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
```diff
@@ -0,0 +1,72 @@
...
+  def self.generate(batch)
+    Generator.new(batch).generate
+  end
+  class Generator
+    def initialize(batch)
+      @batch = batch
+    end
+
+    def generate
+      CSV.generate(row_sep: "\r\n") do |csv|
```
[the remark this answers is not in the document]
sabrine33: > Are these line endings something to be checked? Example templates downloaded from Elembio have new line character only; there is no carriage return. https://docs.elembio.io/docs/run-manifest/seq-prepare-manifest/
> 
> Maybe not important, they also suggest editing via Excel and saving it. For example, https://docs.elembio.io/docs/run-manifest/seq-prepare-manifest/#create-an-original-file

Technically there is no much a difference between using `\r\n` abd `\n` but "`\r\n`" is safest when sharing for compatibility with Excel on Windows, older systems, or when you're distributing files to users with Windows machines. The Element Aviti machine, it reads csv files, but if the users needs to update same date, they need to open with excel and edit the file. converting from csv to excel is always safer than from excel to csv.

app/models/presenters/batch_submenu_presenter.rb
yoldas: It is not the same as `is_a?` . It means, check for any requests, the block returns true; truthy; other than nil or false.

app/controllers/batches_controller.rb
sabrine33: This is been updated accordingly : https://github.com/sanger/sequencescape/pull/4941/files#diff-0c168ea0ddaf911b3c467a04c5a8fe78f8e4c1eb964d4f747a9b380e5b364c9bR357

PR #4953, opened 2025-05-23
Work completion for tube racks

dasunpubudumal:
Closes #

- 

### DIFF
--- a/app/models/work_completion/tube_rack_completion.rb
+++ b/app/models/work_completion/tube_rack_completion.rb
@@ -17,6 +17,7 @@ class WorkCompletion::TubeRackCompletion < WorkCompletion::LabwareCompletion
   #
   # @example
   #   connect_requests
+  #     # => Connects upstream requests to the target tubes' receptacles
   def connect_requests
     target_tubes.each do |target_tube|
       detect_upstream_requests.each { |upstream| pass_and_link_up_requests(target_tube.tube.receptacle, upstream) }


app/models/tube_rack/purpose.rb
```diff
@@ -12,10 +14,10 @@ def self.standard_tube_rack
     TubeRack::Purpose.find_by(name: 'TR Stock 96')
   end
 
-  def create!(*args, &block)
+  def create!(*args, &)
```
andrewsparkes, line 18: I struggle enough to understand what methods like this do (with blocks) without the linting further shortening it to an ampersand!

app/models/work_completion.rb
```diff
@@ -42,8 +42,28 @@ class WorkCompletion < ApplicationRecord
...
+  #   - If the target labware responds to `wells`, it uses `PlateCompletion`.
+  #   - Otherwise, it defaults to `TubeCompletion`.
   def pass_and_attach_requests
-    processing_class = target.respond_to?(:wells) ? PlateCompletion : TubeCompletion
+    processing_class =
+      if target.respond_to?(:wells)
+        PlateCompletion
+      elsif target.respond_to?(:racked_tubes)
+        TubeRackCompletion
+      else
```
andrewsparkes, line 64: I'm not sure I like how the existing code just assumes it must be a tube if it's not a plate with wells. Seems a bit risky to me, would rather it positively checked it's a tube and throw exception if target doesn't match something in the list.
Not sure what there is unique about a tube though... a tube is also a receptacle maybe.

app/models/work_completion/tube_rack_completion.rb
```diff
@@ -0,0 +1,61 @@
...
+class WorkCompletion::TubeRackCompletion < WorkCompletion::LabwareCompletion
+  # Connects upstream requests to the target labware's racked tubes.
+  #
+  # Iterates through each upstream request and links it to the receptacle
+  # of each tube in the target labware's racked tubes.
+  #
+  # @return [void]
+  #
+  # @example
+  #   connect_requests
```
andrewsparkes, line 19: what's the point of these example comments when they just have the method name?

app/resources/api/v2/request_resource.rb
```diff
@@ -2,17 +2,73 @@
...
-    # Provides a JSON:API representation of {Request}.
+    # Access this resource via the `/api/v2/requests/` endpoint.
+    #
+    # @note This resource supports `GET` and `PATCH` requests. It allows fetching requests by their attributes
+    #       and updating them.
+    #
+    # @example GET request to retrieve all requests
+    #   GET /api/v2/requests/
+    #
+    # @todo Figure out how to send a POST for a request with request metadata association. Currently,
```
andrewsparkes: Is this todo something that needs figuring out to complete this story?

app/resources/api/v2/request_resource.rb
```diff
@@ -21,27 +77,90 @@ class RequestResource < BaseResource
...
+
+      # @note `always_include_linkage_data` is currently broken. It is supposed to ensure that
+      #   linkage data is always included in the response. However, it is not working as expected.
+      #   To ensure linkage data is always included, explicitly include the relationship in the request.
+      #   For example, use `?include=order` in the request URL to include the `submission` relationship.
+      #   See Y25-164 for more details and a possible fix.
+      # @!attribute [r] order
+      #   The order associated with this request.
       has_one :order, always_include_linkage_data: true
-      # NB. adding always_include_linkage_data: true caused a problem in the JSON Api client in Limber in that it has
```
andrewsparkes: It seems inconsistent to have `always_include_linkage_data: true` on the submission and order on the right, with comments saying it doesn't work, but then not have it on this request_type (and also remove the comment saying why it's not there).
If it's not working (as the comments on the right say), then maybe it shouldn't be present on submission and order, and maybe all 3 fields should have comments saying it was removed and what to do instead?

andrewsparkes (commented):
I think this looks very good. And well commented.

I left a few points I think I'd like to discuss before I can approve.

PR #4969, opened 2025-05-29
Y25-292: Kinnex Pipeline

dasunpubudumal:
Closes https://github.com/sanger/limber/issues/2358

- The resource and the model for creating plates out of tube wells.

Refer to the comment https://github.com/sanger/sequencescape/pull/4969#issuecomment-2945475117.

### DIFF
--- a/config/default_records/request_types/023_limber_kinnex_prep.wip.yml
+++ b/config/default_records/request_types/023_limber_kinnex_prep.wip.yml
@@ -4,7 +4,7 @@ kinnex_prep:
   name: Kinnex Prep
   asset_type: Well
   order: 1
-  request_class_name: IlluminaHtp::Requests::StdLibraryRequest
+  request_class_name: CustomerRequest
   for_multiplexing: false
   billable: true
   product_line_name: Short Read


dasunpubudumal:
**Couple of problems with the current implementation**

1. Because we are creating tubes out of wells - and the well has a single request - the two tubes that are created are attached to the same request. 

For instance, if we create a Kinnex submission for a `LRC GEM-X 5p cDNA Input` plate with 2 wells having samples, and manually transfer to the four tubes created, the following will happen:

- For both tubes, receptacles will be created through the transfers.
- For Limber's  [`v2_labware.in_progress_submission_uuids`](https://github.com/sanger/limber/blob/c8f05ef2f72c8aaef610783ab288a275ab94cf33/app/models/state_changers.rb#L108) call, both tubes will return the same UUIDs. These are in-progress requests.
- When we invoke [`Sequencescape::Api::V2::WorkCompletion.create!`](https://github.com/sanger/limber/blob/c8f05ef2f72c8aaef610783ab288a275ab94cf33/app/models/state_changers.rb#L111) in Limber, it invokes [`app/models/work_completion/tube_completion.rb`](https://github.com/sanger/sequencescape/blob/2c1b828f6e63869a97ab900a0e14c3f6b4812a2d/app/models/work_completion/tube_completion.rb) in Sequencescape. This will detect the upstream requests by filtering requests using `target_labware.aliquots.pluck(:request_id)`. Note that this invocation does not return anything for the second tube, as the request is only associated with the first tube.
- Therefore, linking up does not occur due to an empty array returned for `detect_upstream_requests` method.

2. Hence, the user would only see passed requests in Sequencescape Labware page (for tubes) for the first tube created for each pair.

app/controllers/api/v2/tube_from_plate_creations_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # Provides a JSON API controller for Tube From Tube Creations
```
KatyTaylor:
```suggestion
    # Provides a JSON API controller for Tube From Plate Creations
```

app/resources/api/v2/tube_from_plate_creation_resource.rb
```diff
@@ -0,0 +1,132 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # Provides a JSON:API representation of {TubeFromPlateCreation} for the creation of a single child tube from a
+    # single parent tube.
```
KatyTaylor:
```suggestion
    # single parent plate.
```

app/resources/api/v2/tube_from_plate_creation_resource.rb
```diff
@@ -0,0 +1,132 @@
...
+    # Creation of this resource via a `POST` request will create the new child tube with the parent as
+    # an ancestor.
+    # The created child will have the specified purpose and can be accessed via the {#child} relationship.
+    #
+    # @note Access this resource via the `/api/v2/tube_from_tube_creations/` endpoint.
```
KatyTaylor:
```suggestion
    # @note Access this resource via the `/api/v2/tube_from_plate_creations/` endpoint.
```
?

app/resources/api/v2/tube_from_plate_creation_resource.rb
```diff
@@ -0,0 +1,132 @@
...
+    #         }
+    #       }
+    #     }
+    #   }
+    #
+    # @example GET request for all TubeFromTubeCreation resources
+    #   GET /api/v2/tube_from_tube_creations/
+    #
+    # @example GET request for a TubeFromTubeCreation with ID 123
+    #   GET /api/v2/tube_from_tube_creations/123/
```
KatyTaylor: Above 2 examples reference 'tube from tube' rather than 'tube from plate' 4 times.

app/resources/api/v2/tube_from_plate_creation_resource.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      # @!attribute [w] user_uuid
+      #   This is declared for convenience where the user is not available to set as a relationship.
+      #   Setting this attribute alongside the `user` relationship will prefer the relationship value.
+      #   @deprecated Use the `user` relationship instead.
+      #     See [Y25-236](https://github.com/sanger/sequencescape/issues/4812).
+      #   @param value [String] The UUID of the user who initiated the creation of this tube from a parent tube.
+      #   @return [Void]
+      #   @see #user
+      attribute :user_uuid, writeonly: true
```
KatyTaylor, line 95: Do we need to include this - it's deprecated - and we're adding a new endpoint here

app/resources/api/v2/tube_from_plate_creation_resource.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      def user_uuid=(value)
+        @model.user = User.with_uuid(value).first
+      end
+
+      # @!attribute [r] uuid
+      #   @return [String] The UUID of the tube from tube creation.
```
KatyTaylor:
```suggestion
      #   @return [String] The UUID of the tube from plate creation.
```
😆

config/default_records/request_types/023_limber_kinnex_prep.wip.yml
```diff
@@ -0,0 +1,13 @@
+# Request types for Kinnex Prep pipeline.
+---
+kinnex_prep:
+  name: Kinnex Prep
+  asset_type: Well
+  order: 1
+  request_class_name: IlluminaHtp::Requests::StdLibraryRequest
```
KatyTaylor: I don't think it should be a library request - we're not making libraries here, we're simply transferring liquid from a plate into tubes.

config/default_records/submission_templates/018_kinnex_prep_submission_templates.wip.yml
```diff
@@ -0,0 +1,9 @@
+# This submission template is associated with the Limber KInnex Prep pipeline.
```
KatyTaylor:
```suggestion
# This submission template is associated with the Limber Kinnex Prep pipeline.
```

app/resources/api/v2/tube_from_plate_creation_resource.rb
dasunpubudumal: Consequences of copy-pasting tubes_from_tubes code 😆

config/default_records/request_types/023_limber_kinnex_prep.wip.yml
dasunpubudumal: oh, I took this from Seena's PR. What should it be? 🤔

KatyTaylor: Think you can just use CustomerRequest.

app/resources/api/v2/tube_from_plate_creation_resource.rb
dasunpubudumal: Sorry I forgot about this comment.

I think it a _little_ complicated. We can pass the user, but I think we've been passing UUIDs throughout our [creation controllers](https://github.com/sanger/limber/blob/develop/app/controllers/creation_controller.rb#L84-L86). I'm not sure if we want that change to be in this pull request, given that we already have one for it [Y25-236](https://github.com/sanger/sequencescape/issues/4812)? If it does, I'm happy to make the change.

We can use the UUID, send a request to Sequencescape and fetch the user at this resource-level but I think  it's probably the wrong place to do that? I'm not sure.

KatyTaylor: Makes sense if we've already got a story to cover it. Thanks.

KatyTaylor (approved):

Issue #4905, opened 2025-05-07
Deliver Element Aviti flowcell messages to unified warehouse and write into iseq_flowcell table


yoldas:
**User story**

As PSD Developer (Abdullah) I would like the Element Aviti flowcell messages delivered to unified warehouse and written into iseq_flowcell table so these can be used for tracking correctly.

**Who are the primary contacts for this story**

Abdullah

**Who is the nominated tester for UAT**

PSD

e.g. John S (don't include surnames in public repos)

**Acceptance criteria**

Consider if these features can be _featured flagged_ to decouple testing and deployment.
To be considered successful the solution must allow:
- [ ] There is a dedicated `ElementAvitiSequencingPipeline` class as a subclass of `SequencingPipeline`.
- [ ] The `sti_type` field of Element Aviti record is updated with the class name in the record file `002_element_aviti_pipelines`
- [ ] `ElementAvitiSequencingPipeline` overrides the method `post_release_batch` to change the root argument to `'iseq_flowcell'`
- [ ] The flowcell root in the JSON message is changed to `iseq_flowcell` as in `{"iseq_flowcell": {...}, "lims": "SQSCP"}`
- [ ] Releasing an Element Aviti batch creates records in the`iseq_flowcell` table in the warehouse database.

**Dependencies**

This story is blocked by the following dependencies:
- #4630
- #4632
- sanger/unified_warehouse#715

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

**Deployment notes**

If the pipelines record is already created in the sequencescape database in the target environment, you need to manually change the sti_type field of the row in the pipelines table. Verify that the `sti_type` reads the correct class name, i.e. `ElementAvitiSequencingPipeline`.

```sql
select * from pipelines where name = 'Element Aviti';
```

Issue #4731, opened 2025-03-12
Y25-188 - Create ESeq flowcell IO or reuse the existing IO for Element Aviti bulk sequencing to update MLWH


yoldas:
**User story**

As a developer, Abdullah, I would like to create a new flowcell IO or reuse the existing flowcell IO for Element Aviti bulk sequencing to update MLWH eseq_flowcell table.

**Who are the primary contacts for this story**
Tom W.
Abdullah Y.

**Who is the nominated tester for UAT**
PSD and/or NPG

**Acceptance criteria**

Consider if these features can be _featured flagged_ to decouple testing and deployment.
To be considered successful the solution must allow:
- [ ] Sequencing batch generates message describing a flowcell for Element Aviti and they are recorded in eseq_flowcell table in the warehouse database.

**Dependencies**

This story is blocked by the following dependencies:
- sanger/sequencescape#4649
- sanger/unified_warehouse#715

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

This story has been extracted from sanger/unified_warehouse#715 because we need to implement a new flowcell IO or reuse the existing to generate messages, however we need more understanding of sequencing requests for Element Aviti.

**Deployment notes**

If eseq_flowcell table in unified_warehouse needs additional migrations, the unified warehouse release will need to be deployed before sequencescape.

SujitDey2022:
@yoldas can we confirm if we need this story or close it. Regards,

SujitDey2022:
Closing based on discussion on 17Mar2026

PR #4972, opened 2025-05-29
4905-y25-305---write-element-aviti-flowcell-messages-to-unified-warehouse-eseq_flowcell-table

wendyyang:
Closes #

- 


app/models/element_aviti_sequencing_pipeline.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# Specialized sequencing pipeline for Element Aviti
+class ElementAvitiSequencingPipeline < SequencingPipeline
+  def post_release_batch(batch, _user)
+    # Same logic as the superclass, but with a different Messenger root
+    batch.assets.compact.uniq.each(&:index_aliquots)
+    Messenger.create!(target: batch, template: 'FlowcellIo', root: 'eseq_flowcell')
```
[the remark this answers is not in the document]
yoldas: This copilot suggestion is partially good (avoiding strings), but makes maintenance harder with constants. Instead, we could extract the template name and root into the superclass as attributes and override only those rather than modifying the method in the subclass.

spec/models/element_aviti_sequencing_pipeline_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe ElementAvitiSequencingPipeline, type: :model do
+  describe '#post_release_batch' do
+    let(:pipeline) { create(:element_aviti_sequencing_pipeline) }
+    let(:batch) { create(:batch) }
+
+    it 'calls Messenger with eseq_flowcell root' do
+      allow(Messenger).to receive(:create!)
```
yoldas, line 10: This stubbing is OK for checking that the overridden method calls the Messenger with correct arguments but it would be good to have an additional test on the rendered flowcell message using the flowcell IO template. The superclass has a similar test. It will give us a chance to make assertions on the content.

app/models/element_aviti_sequencing_pipeline.rb
sabrine33, line 8: The MLWH is not accepting the messages because the wrong template is being used for the iseq_flowcell table. You need to create a new template that matches the structure of the iseq_flowcell table.
The message currently constructed using [flowcell_io](https://github.com/sanger/sequencescape/blob/develop/app/models/api/messages/flowcell_io.rb)

yoldas: @sabrine33 we are reusing the template for now. In the **additional context** of the [story](https://github.com/sanger/sequencescape/issues/4905), it is explained how to change the root to make the warehouse handle the message. On the warehouse side, you need to add the unrelated columns into ignore list on the EseqFlowcell model if attributes are not related to ElementAviti. 

If you want to work on the IO template (if you need extra attributes or customisations which cannot be handled by the existing template), there is a story for that which needs prioritising. https://github.com/sanger/sequencescape/issues/4731

app/models/api/messages/eseq_flowcell_io.rb
```diff
@@ -0,0 +1,281 @@
...
+        map_attribute_to_json_attribute(:library_type, 'pipeline_id_lims')
+        with_association(:sample) { map_attribute_to_json_attribute(:uuid, 'sample_uuid') }
+        with_association(:study) { map_attribute_to_json_attribute(:uuid, 'study_uuid') }
+        # map_attribute_to_json_attribute(:library_id, 'legacy_library_id')
+        # with_association(:library) { map_attribute_to_json_attribute(:external_identifier, 'id_library_lims') }
+        map_attribute_to_json_attribute(:control_aliquot_type, 'entity_type')
+      end
+    end
+  end
+
```
yoldas: This IO template is not part of this story.

app/models/api/messages/eseq_flowcell_io.rb
```diff
@@ -0,0 +1,281 @@
+# frozen_string_literal: true
+# Generates warehouse messages describing a flowcell. While flowcells are not
+# modeled directly in Sequencescape they can be approximated by a sequencing
+# {Batch}
+class Api::Messages::EseqFlowcellIo < Api::Base
```
yoldas: This class is not in the scope of this story.

app/models/api/messages/eseq_flowcell_io.rb
wendyyang: yes, i am going to remove the class.

yoldas (approved):
Looks good.

PR #4982, opened 2025-06-04
Update PhiX controls for Aviti and add an additional descriptor.

sabrine33:


app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
```diff
@@ -34,16 +34,16 @@ class Generator
...
       ['[SAMPLES]'],
       %w[SampleName Index1 Index2 Lane Project],
-      %w[Adept_CB1 ATGTCGCTAG CTAGCTCGTA],
-      %w[Adept_CB2 CACAGATCGT CACAGATCGT],
-      %w[Adept_CB3 GCACATAGTC GACTACTAGC],
-      %w[Adept_CB4 TGTGTCGACA TGTCTGACAG]
+      %w[PhiX_Third ATGTCGCTAG CTAGCTCGTA],
+      %w[PhiX_Third CACAGATCGT ACGAGAGTCT],
+      %w[PhiX_Third GCACATAGTC GACTACTAGC],
+      %w[PhiX_Third TGTGTCGACA TGTCTGACAG]
```
yoldas, line 46: 1. OK: Sequences given in Index1 and Index2 match the ones in the section "PhiX Control Library, Adept" of page "PhiX Control Library Index Sequences" at https://docs.elembio.io/docs/run-manifest/index-sequences/index-phix/
2. The comment above could include the info that Adept names are renamed to PhiX_Third .
3. Do we need to use suffixes after PhiX_Third to distinguish SampleNames like in the previous version of the code? e.g., Adept_CB1 -> PhiX_Third_CB1 .
4. Lane and Project columns are specifying the headers. Do Sample rows specified need any values for Lane and project?
5. Since controls are assumed to be predetermined values; not scanned; and they are not represented in Sequencescape, unified warehouse eseq_flowcell table will not include control rows.

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
```diff
@@ -102,6 +102,35 @@ def group_samples_by_identity
...
+    # The PSD-assisted software supports sample index lengths of either 8 bp or 10 bp.
+    # The default PhiX control indexes are 10 bp long.
+    #
+    # This method ensures that PhiX control sample indexes are truncated to match the actual
+    # sample index length in the current batch to maintain compatibility and prevent sequencing issues.
+    def adjusted_phix_indexes
+      PHIX_SECTION.map do |row|
+        if row.length > 2
+          row = row.dup
+          row[1] = row[1][0, sample_index_length] if row[1]
```
yoldas, line 127: If the index1 is ATGTCGCTAG and index length is 8, the index is changed to ATGTCGCT. Is this matching the reality? Do they have different controls with shorter indexes?

config/default_records/descriptors/002_element_aviti_descriptors.wip.yml
```diff
@@ -40,52 +40,59 @@ PhiX Expiry:
...
+  required: false
+  sorter: 6
 PhiX %:
   name: PhiX %
   task: Loading
   workflow: Element Aviti
   kind: Text
   required: false
-  sorter: 6
+  sorter: 7
```
yoldas, line 56: Changing attributes of the existing records may require a custom deployment step to change them in the database. Please include them to deployment notes if there are custom steps and post-deployment checks in the story.

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
sabrine33: yes, we discussed this, and Kieron mentioned we need to truncate the existing phix indexes (10bp) to match the samples, in case they have 8 length indexes

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
sabrine33: > 1. OK: Sequences given in Index1 and Index2 match the ones in the section "PhiX Control Library, Adept" of page "PhiX Control Library Index Sequences" at https://docs.elembio.io/docs/run-manifest/index-sequences/index-phix/
> 2. The comment above could include the info that Adept names are renamed to PhiX_Third .
> 3. Do we need to use suffixes after PhiX_Third to distinguish SampleNames like in the previous version of the code? e.g., Adept_CB1 -> PhiX_Third_CB1 .
> 4. Lane and Project columns are specifying the headers. Do Sample rows specified need any values for Lane and project?
> 5. Since controls are assumed to be predetermined values; not scanned; and they are not represented in Sequencescape, unified warehouse eseq_flowcell table will not include control rows.

1. We agreed to use `Cloudbreak Freestyle PhiX Control, Third Party` all the phix contorl sample should have the same `PhiX_Third`, the indexes are set as in the official documentation for the `PhiX_Third`
2. We are not renaming here, we are using a different control type `PhiX_Third`, also, I don't see the point of adding this in the comment section, as the pipeline is not on production yet. The comment is already mentioning that the controls are set as in the official documentation
3. Nope, we agreed all the controls should be set under the same name - NPG would like consider this as a single  sample  in the QC process.
4. I think so, but I am not entirely sure - better to check with NPG team

yoldas (approved):

PR #4984, opened 2025-06-05
Add additional columns to scrna_core template

sabrine33:


config/sample_manifest_excel/manifest_types.yml
```diff
@@ -830,6 +830,13 @@ scrna_core:
     - :volume
     - :retention_instruction
     - :sample_description
+    - :date_of_sample_collection
+    - :country_of_origin
+    - :phenotype
+    - :gender
+    - :sample_ebi_accession_number
+    - :sample_taxon_id
+    - :sample_common_name
```
StephenHulme, line 839: I note a couple differences between these column names and the names in the original story, but there seem to be many references to the variations above in SS, so I trust that the translation to column names happens somewhere else.

PR #4987, opened 2025-06-06
NPG actions endpoints - improve error message

KatyTaylor:
Closes #

- 

### DIFF
--- a/app/controllers/npg_actions/assets_controller.rb
+++ b/app/controllers/npg_actions/assets_controller.rb
@@ -31,7 +31,7 @@ class NpgActions::AssetsController < ApplicationController
       if @last_event.present?
         # If we already have an event we check to see its state. If it matches,
         # we just continue to rendering, otherwise we blow up.
-        raise NPGActionInvalid, 'NPG user run this action. Please, contact USG' if @last_event.family != state
+        raise NPGActionInvalid, already_completed_message(@last_event.family, state) if @last_event.family != state
       else
         generate_events(state)
       end
@@ -88,5 +88,10 @@ class NpgActions::AssetsController < ApplicationController
 
   def rescue_error_bad_request(exception)
     render xml: "<error><message>#{exception.message.split("\n").first}</message></error>", status: :bad_request
+  end
+
+  def already_completed_message(existing_state, requested_state)
+    "The requests on this lane have already been completed with state: '#{existing_state}'. "\
+    "Unable to set them to new state: '#{requested_state}'."
   end
 end


app/controllers/npg_actions/assets_controller.rb
```diff
@@ -31,7 +31,7 @@ def action_for_qc_state(state)
       if @last_event.present?
         # If we already have an event we check to see its state. If it matches,
         # we just continue to rendering, otherwise we blow up.
-        raise NPGActionInvalid, 'NPG user run this action. Please, contact USG' if @last_event.family != state
+        raise NPGActionInvalid, already_completed_message(@last_event.family, state) if @last_event.family != state
```
harrietc52: Took me a second to realise that `family` returns “pass” or “fail”. And that completed means either “pass” or “fail”. But once I figured that out, this refactor made a bit more sense! Not sure whether its worth adding that to a comment in the `already_completed_message`, that the two completed states are either pass or fail.

app/controllers/npg_actions/assets_controller.rb
```diff
@@ -40,6 +40,15 @@ def action_for_qc_state(state)
     end
   end
 
+  # Does a variety of things to do with closing off sequencing requests & batches, and recording an audit trail:
```
harrietc52:
```suggestion
  # Does a variety of things to do with completing sequencing requests & batches, and recording an audit trail:
```

Maybe use `completing`, just to have consistent lingo, if that is the correct case?

app/controllers/npg_actions/assets_controller.rb
KatyTaylor: Makes sense, thank you. I'll add a few more comments.

KatyTaylor:
@harrietc52 Thanks for your comments. I did a bit more refactoring to try to make things more readable and clearer... does it help?

app/controllers/npg_actions/assets_controller.rb
```diff
@@ -28,10 +28,10 @@ def pass
 
   def action_for_qc_state(state)
     ActiveRecord::Base.transaction do
-      if @last_event.present?
-        # If we already have an event we check to see its state. If it matches,
-        # we just continue to rendering, otherwise we blow up.
-        raise NPGActionInvalid, 'NPG user run this action. Please, contact USG' if @last_event.family != state
+      if endpoint_previously_called?
+        # If the state provided in the API call (`state`) matches the one previously recorded,
```
harrietc52: Sorry! I don’t quite follow this bit "If the state provided in the API call (`state`)". 

Is the state here taken from the API route `pass_qc_state` or `fail_qc_state`? Maybe this comment suggests its from and API endpoint called `state`.

app/controllers/npg_actions/assets_controller.rb
```diff
@@ -89,4 +100,23 @@ def rescue_error(exception)
...
+  # If there is already a 'pass' or 'fail' event recorded,
+  # it implies this endpoint has already been called for this request.
+  def endpoint_previously_called?
+    @last_event.present?
+  end
+
+  # Returns the state previously recorded for the request,
+  # if the endpoint has been called before.
+  # Could be 'pass' or 'fail'.
+  def existing_state
```
harrietc52, line 116: It's being pernickety, but you could also rename `@last_event` to `@last_pass_or_fail_event` too, as that confused me initially. But that might be a bit more involved and not worth the effort either tests etc. Looking at the `find_last_event` method code, it is clear its Events with the pass and fail scope.

KatyTaylor:
I checked with NPG whether this change would affect them (since it is changing the response from our endpoint), but they said it wouldn't - so don't need to factor that into the release.

app/controllers/npg_actions/assets_controller.rb
```diff
@@ -10,28 +10,31 @@ class NpgActions::AssetsController < ApplicationController
...
 
+  # POST /npg_actions/assets/:asset_id/pass
   def fail
     action_for_qc_state('fail')
   end
 
+  # POST /npg_actions/assets/:asset_id/fail
   def pass
     action_for_qc_state('pass')
   end
```
StephenHulme, line 27: @KatyTaylor Just stumbled across this while debugging an error message, are the pass and fail comments meant to be swapped?

KatyTaylor: oops yes!

StephenHulme: No worries, thanks! Will swap back round :)

PR #4993, opened 2025-06-09
[dev] Replace prettier-ruby with rubocop

StephenHulme:
_One ring to rule them all, one ring to find them, One ring to bring them all and in the darkness bind them..._

- Replaces prettier-ruby with rubocop, since rubocop now handles layout formatting
- Configure cops
- Autocorrect layout differences from prettier-ruby

Now all Ruby linting and formatting is controlled by Rubocop only:

```sh
bundle exec rubocop -a 
```

To format non-ruby files, run Prettier as before:

```sh
yarn prettier --write .
```

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -76,3 +76,15 @@ c3490ce690b41a7b0fcde0daee1e0b3cf03182b9
 
 # More rubocop
 ea76275305e87b62ac8bb092f4d445c576a33922
+
+# The Great Prettier Linting
+fa00b74a668b3623d51ad97641d8a35b552a8c0f
+
+# Upgrade_prettier
+be261f9d4813aefb8c53d81042c6fe4792c419d6
+
+# Update prettier cops
+df35dbf84c47e19f0ce46610bf2d323305e4f17b
+
+# Experiment with prettier
+5f59ed67b135ec78fdc29ec1facadd3faba5a1c6


.rubocop_todo.yml
```diff
@@ -1,11 +1,57 @@
...
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+# Offense count: 10
+# This cop supports safe autocorrection (--autocorrect).
+Layout/ClosingHeredocIndentation:
+  Exclude:
+    - 'db/migrate/20190829125351_update_view_aliquots.rb'
```
BenTopping, line 13: I dont think we should be including migrations or seeds in rubocop

Gemfile
```diff
@@ -43,7 +43,7 @@ group :default do
   gem 'will_paginate-bootstrap'
 
   # Provides eg. error_messages_for previously in rails 2, now deprecated.
```
BenTopping, line 45: Can we just remove this gem? Rails 2 😆

app/models/robot/verification/source_dest_control_beds.rb
```diff
@@ -16,12 +16,12 @@ def layout_data_object(data_object)
   def valid_plate_locations?(params, batch, robot, expected_plate_layout) # rubocop:todo Metrics/MethodLength
     return false unless super
     unless valid_control_plates_on_robot?(
-             params[:control_bed_barcodes],
-             params[:control_plate_barcodes],
-             robot,
-             batch,
-             expected_plate_layout
-           )
+      params[:control_bed_barcodes],
```
BenTopping, line 19: Do you know which rule this is? I think either syntax is fine so feels a bit overkill to format it one way or the other.

BenTopping (commented):
All looks fine from a brief skim. Couple minor comments.

.rubocop_todo.yml
StephenHulme: I was considering this too... but it's still production code?
What's your thinking about why not?

Gemfile
StephenHulme: Out of scope ;) but the `error_messages_for` method was previously in rails 2, now it's a separate gem and the method is still used in our code. Potentially create story to remove and replace with something else?

.rubocop_todo.yml
BenTopping: We don't really touch the seeds file and migrations are historical records of things that have occured and are likely to never run again so there is no point formatting them.

app/models/robot/verification/source_dest_control_beds.rb
StephenHulme: Yes, it's Layout/FirstArgumentIndentation. I've tried all the various configurations and the default is the most similar to our codebase. I'm happy to leave these in todo if you would prefer?

.rubocop_todo.yml
StephenHulme: Yeah, that's why I've left them in todo, but not excluded the rule.
My thinking here is I have no issues with the existing code, but if new migrations or seeds are added, then at a preference, they should have the same style as the rest of our code.

app/controllers/api/v2/bait_library_layouts_controller.rb
```diff
@@ -64,11 +64,11 @@ def preview_records
 
         user =
           User.with_uuid(param_hash[:user_uuid]).first ||
-            record_errors.append("The User record identified by UUID '#{param_hash[:user_uuid]}' cannot be found")
+          record_errors.append("The User record identified by UUID '#{param_hash[:user_uuid]}' cannot be found")
```
yoldas, line 67: It is easier to spot the continuation of the line if there is indent here. What rule is this?

yoldas:
I found having the indent less on the left helpful to keep the line more visible in the editor and having space between methods is useful too. 

Are there README changes to explain which command (if any) will be executed instead of prettier?
Are there changes for the workflow files?

app/controllers/api/v2/bait_library_layouts_controller.rb
StephenHulme: This is set by Layout/MultilineOperationIndentation, but neither the `aligned` or `indented` EnforcedStyle replicates our existing formatting.

I'm happy to disable this cop if you would prefer?

yoldas: Thanks for checking. I thought about this again. It is indenting with respect to the left hand side of the assignment. The expression on the right is divided and indented/aligned under the first line. It reads better now especially in favour of shortening the line lengths. No change needed; keep it please.

PR #4995, opened 2025-06-10
add search by project cost code

sabrine33:
For more visibility, I am copy pasting @KatyTaylor's comment here:

`Discussed with Liz whether they'd select a value from the dropdown in top right or leave it saying 'All' - she's happy to do either as long as they've got a way to search, but I think the ideal solution was thought to be that the cost code search should work on 'All', and also to have a new option 'Cost code' in the list, which would make it very clear you could use that to search. We could also enhance the 'Project' drop down to allow searching by Cost code.`

=> The code change adds a 'cost code' option to the search box, the cost code search works on 'All' too.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require 'aasm'
 
+# rubocop:disable Metrics/ClassLength
 class Project < ApplicationRecord
   # It has to be here, as there are has_many through: :orders associations in modules
   has_many :orders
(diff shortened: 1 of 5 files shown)


dasunpubudumal (commented):
Nice. Do we need to cover it with a test?

sabrine33:
> Nice. Do we need to cover it with a test?

Good point, the search tests does not cover case by case option, but also it is build on a locked data, so I don't think that is much added value to it - I am happy to add some though, if you convince me otherwise ")

app/models/project.rb
```diff
@@ -57,7 +58,15 @@ def self.states
   validates :name, :state, presence: true
   validates :name, uniqueness: { on: :create, message: "already in use (#{name})", case_sensitive: false }
 
-  scope :for_search_query, ->(query) { where(['name LIKE ? OR id=?', "%#{query}%", query]) }
+  scope :for_search_query,
+        ->(query) do
+          joins(project_metadata: :project).where(
+            'projects.name LIKE ? OR projects.id = ? OR project_cost_code LIKE ?',
+            "%#{query}%",
```
BenTopping, line 65: `"%#{query}%"` does this need to be included twice?

BenTopping (approved):
Nice, tested locally and works well.

app/models/project.rb
sabrine33: > `"%#{query}%"` is to replace the `?` placeholders, the first is to replace the first `? ` placeholder, `query` is for the second and the last `"%#{query}%"`is for the last placeholder

BenTopping: Ah I see, all good then. Thanks for explaining.

PR #5010, opened 2025-06-17
support for samples with no tags

sabrine33:
Support for samples with no tags:
If all pooled samples do not have a tag, we need to remove the tags from the PhiX samples as well.

We are checking for the longest sample tag and trying to match the PhiX tag to that length, based on the assumption that pooled sample tags should vary between 8 bp and 10 bp.

For the first phase, all pooled samples should (or will) have the same tag length.

In the second phase, on rare occasions, we might have pools with samples that have different tag lengths. Tris is looking into whether and how we can pad the shorter tags with adapter nucleotides — this is why I updated the code to check for the longest tag length."
doc of how padding tags: https://docs.elembio.io/docs/run-manifest/samples/#reconciling

### DIFF
--- a/app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
+++ b/app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
@@ -14,7 +14,7 @@
 # Element recommends using a consistent sample name that clearly identifies the control sequences.
 #
 # Allowed characters in the output file name include: letters, numbers, dashes, dots, parentheses, and underscores.
-# todo: support use cases where the samples in the pool have different index lengths(https://docs.elembio.io/docs/run-manifest/samples/#reconciling)
+# In the next phase, we will add support for use cases where the samples in the pool have different index lengths(https://docs.elembio.io/docs/run-manifest/samples/#reconciling)
 
 module AvitiSampleSheet::SampleSheetGenerator
   def self.generate(batch)
@@ -36,12 +36,15 @@ module AvitiSampleSheet::SampleSheetGenerator
       %w[R2AdapterTrim FALSE 1+2]
     ].freeze
 
+    SAMPLE_SECTION_HEADERS = [
+      ['[SAMPLES]'],
+      %w[SampleName Index1 Index2 Lane Project]
+    ].freeze
+
     # The values are set according to the official documentation from Element Biosciences.
     # This section is static because metadata for control samples is not tracked in the Aviti pipeline.
     # Users can manually modify this section if necessary.
     PHIX_SECTION = [
-      ['[SAMPLES]'],
-      %w[SampleName Index1 Index2 Lane Project],
       %w[PhiX_Third ATGTCGCTAG CTAGCTCGTA],
       %w[PhiX_Third CACAGATCGT ACGAGAGTCT],
       %w[PhiX_Third GCACATAGTC GACTACTAGC],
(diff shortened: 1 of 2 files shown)


app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
```diff
@@ -103,25 +106,47 @@ def group_samples_by_identity
...
+        truncated_phix_indexes(sample_index_length)
+      else
+        PHIX_SECTION
+      end
+    end
+
+    # Returns only the PhiX sample names (first column)
+    def remove_phix_control_tags
+      PHIX_SECTION.map { |row| [row[0]] }
+    end
```
yoldas, line 150: This produces `[["[SAMPLES]"], ["SampleName"], ["PhiX_Third"], ["PhiX_Third"], ["PhiX_Third"], ["PhiX_Third"]]` for the `PHIX_SECTION` constant. I think, the headers need separate treatment.
```
    PHIX_SECTION = [
      ['[SAMPLES]'],
      %w[SampleName Index1 Index2 Lane Project],
      %w[PhiX_Third ATGTCGCTAG CTAGCTCGTA],
      %w[PhiX_Third CACAGATCGT ACGAGAGTCT],
      %w[PhiX_Third GCACATAGTC GACTACTAGC],
      %w[PhiX_Third TGTGTCGACA TGTCTGACAG]
    ].freeze
```

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
```diff
@@ -14,6 +14,7 @@
 # Element recommends using a consistent sample name that clearly identifies the control sequences.
 #
 # Allowed characters in the output file name include: letters, numbers, dashes, dots, parentheses, and underscores.
+# todo: support use cases where the samples in the pool have different index lengths(https://docs.elembio.io/docs/run-manifest/samples/#reconciling)
```
yoldas: This todo does not have a story number.

yoldas (changes requested):

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
sabrine33: Thanks for the review @yoldas, yes` [["[SAMPLES]"], ["SampleName"], ["PhiX_Third"], ["PhiX_Third"], ["PhiX_Third"], ["PhiX_Third"]]` is the desired result , we only need to keep the phix control names

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
sabrine33: there is no story number for it yet, Tris is happy with the the manifest outcome for this phase, the support of the pooled sample with different index length is planned for the next phase, I remove the todo, if it is bringing confusion

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
yoldas: The header row `SampleName,Index1,Index2,Lane,Project` becomes `SampleName`.

app/controllers/aviti_sample_sheet/sample_sheet_generator.rb
yoldas: It is very useful to add important points as comments. When it is a todo though, the PR becomes unfinished without a reference story number.

PR #5021, opened 2025-06-23
Y25-221 - Tubes in location reports

### LINKED ISSUE #4776: Y25-221 - Add tubes to the labware that is returned by the plate location report
As Team Lead Long Read (Danni) I would like to have both tubes AND plates returned by the report, presently this only returns plates

https://sequencescape.psd.sanger.ac.uk/location_reports

dw15

tested by dev.

location lw-y25-221plate-tube-test-3833 has been created for UAT testing

this contains
SQPU-86259-J
TRAC-2-172
NT1846739B

presently the report, when searching by location or by the barcodes, will only return the plate SQPU-86259-J

**Acceptance criteria**
Reports should return plates and tubes

BenTopping:
Closes #4776 

- Renames `Plate location report` to `Labware location report`.
- Includes tubes in location report when searching by labwhere location.
- Includes tubes in location report when searching by selection.

Local testing go to: http://localhost:3000/location_reports
Or navigate to `pipelines > sample management > labware location reports`
You will also need labwhere running.


app/models/location_report.rb
```diff
@@ -135,32 +135,32 @@ def generate_report_rows # rubocop:todo Metrics/MethodLength
 
   #######
 
-  def plates_list
-    @plates_list ||=
+  def labware_list
```
andrewsparkes, line 138: Have you tested this with Tube Racks? You may have to filter out anything that's not a tube or a plate still.

andrewsparkes (approved):
Looks good to me but you might need to either add Tube Racks to the tests and check they behave (not sure what you do with them tbh, fetch current racked-tubes in them to add to the report?) or filter them out.

app/models/location_report.rb
BenTopping: Yes good point, ill take a look tomorrow.

BenTopping: Attempted to add tube racks but the main sticking point came with study search. Tube racks currently don't have 'studies' other than via tubes. This is a different approach to other labware that use receptacles so the shared search methods wouldnt work without changes. 
Since this is out of scope for this story i decided to just filter Labware report results to just be plate and tube as you suggested.

yoldas (approved):
Looks good; specifically checked:
- clean conversion from plate-only report to labware report
- using Labware to include Tubes; no redundant additional code
- Purpose rather than PlatePurpose is used in filtering
- scopes have guards to protect against empty parameters
- no-labware-found case is handled
- unknown labware does not cause exception
- date beginning and end of the day in filtering is correct
- tube-racks are excluded, which is not in the scope of the story
- tube-rack in the same freezer, for example, is ignored

PR #5040, opened 2025-07-02
[test] Skip statistical distribution tests

StephenHulme:

- Skip stastical distribution analysis in unit tests
- Reduce batch tests to just a single batch

### DIFF
--- a/spec/models/cherrypick_task/control_locator_spec.rb
+++ b/spec/models/cherrypick_task/control_locator_spec.rb
@@ -98,7 +98,7 @@ RSpec.describe CherrypickTask::ControlLocator do
     end
 
     # Test the basics for a range of batches
-    1.upto(100) do |batch_id|
+    1.upto(1) do |batch_id| # SKIP batches 2-100 to reduce unneeeded tests
       context "when batch is #{batch_id} and we have a 96 well plate with no wells free" do
         let(:batch_id) { batch_id }
         let(:total_wells) { 96 }
@@ -137,6 +137,8 @@ RSpec.describe CherrypickTask::ControlLocator do
     end
 
     context 'when over a range of batches' do
+      skip 'This analysis is not required to be run every time, so we skip it by default'
+
       let(:range) { (1...1000) }
       let(:control_positions) do
         range.map do |batch_id|


spec/models/cherrypick_task/control_locator_spec.rb
```diff
@@ -98,7 +98,7 @@
     end
 
     # Test the basics for a range of batches
-    1.upto(100) do |batch_id|
+    1.upto(1) do |batch_id| # SKIP batches 2-100 to reduce unneeeded tests
```
borhara, line 101: could we keep a few iterations of this? it doesn't need to be many, but there's a method that uses the `batch_id` to generate a seed, so i'm thinking it could be worth keeping a few?

StephenHulme: Ah I must have missed that, could you send me a link to the method please? I'm interested in how it is actually used.

borhara: https://github.com/sanger/sequencescape/blob/aa43f06c23af5550f008227ddcd1b1717b40ea93/app/models/cherrypick_task/control_locator.rb#L104-L106

i'm not sure if testing more than one batch makes sense, i wonder why it was added originally, maybe for this?

StephenHulme: Yeah, it looks like it was to prevent repeated patterns from (I'm guessing) a retangular plate layout or hi-frequencency pseudo-randon numbers. I'd agree that probably only a single instance is required to test the functionality

borhara (approved):
lgtm

spec/models/cherrypick_task/control_locator_spec.rb
KatyTaylor: Yes, it's not clear to me why the iteration is needed. As Shiv said, the `seed_for` method would return different things for different batch ids, so that's worth testing. But it seems like a unit test for that method would be more efficient. Actually, it may be covered adequately by `context 'when over a range of batches'` - because that calls `control_location` 1000 times - which is a method that ends up calling `seed_for`.

So, haven't looked in great detail, but I think I agree that the iteration is probably not needed here.

PR #5106, opened 2025-07-23
y25-409 scRNA - removed retention instruction column

andrewsparkes:
Closes #y25-40

Remove retention instruction column
-


yoldas (approved):
Looks good. It removes the retention instruction column from scrna_core manifest only. the RetentionInstruction specialised column won't be used.

PR #5110, opened 2025-07-24
Y25-405: set supplier name on the compound sample

sabrine33:
- The initial intuition when starting to work on the story was to add the supplier_name update within the `SampleCompoundAliquotTransfer` `transfer_aliquots_into_compound_sample_aliquots` method, which essentially updates the records immediately after creating them.
A more efficient approach that I adopted, it was to set the supplier_name before creating the compound sample aliquots.
- The rake task is for setting supplier names for previously created sample compounds


app/helpers/compound_sample_helper.rb
```diff
@@ -20,9 +20,22 @@ def find_compound_sample_for_component_samples(component_samples)
...
+    return unless supplier_names.length == 1
+
+    supplier_names.first
+  end
+
   # Generates the compound sample
   def create_compound_sample(study, component_samples)
+    supplier_name = extract_consistent_supplier_name_from_source_aliquots
     study.samples.create!(
+      supplier_name: supplier_name,
```
dasunpubudumal, line 37: If the `extract_consistent_supplier_name_from_source_aliquots` function returns due to the condition `supplier_names.length` being not equal to 1, would we invoke `create!` with `supplier_name` equal to `nil`?

app/helpers/compound_sample_helper.rb
```diff
@@ -20,9 +20,22 @@ def find_compound_sample_for_component_samples(component_samples)
     compound_samples.find { |compound_sample| compound_sample.component_samples == component_samples }
   end
 
+  # Returns the supplier name shared by all source aliquots, if they are consistent.
+  #
+  # @return [String, nil] the supplier name if all source aliquots share the same one; nil otherwise
+
```
dasunpubudumal: Not sure how YARD treats new lines between the docstring and the method but would it probably worth removing this newline so that there's no separation between the method and the docstring that documents the function?

.rubocop_todo.yml
```diff
@@ -1365,6 +1365,7 @@ Rails/HelperInstanceVariable:
   Exclude:
     - 'app/helpers/tag_substitution_helper.rb'
     - 'app/helpers/workflows_helper.rb'
+    - 'app/helpers/compound_sample_helper.rb'
```
dasunpubudumal, line 1368: Would it be worth refactoring the helper class to fit the linter rules so that we don't further increase the todos?

lib/tasks/assign_compound_sample_supplier_name.rake
dasunpubudumal, line 1: Nice!

app/helpers/compound_sample_helper.rb
sabrine33: Yes, we should still create the sample compound event even when supplier_name is nil. That’s also the current behaviour.

.rubocop_todo.yml
sabrine33: Codecov is flagging the use of instance variables in the helper, but since CompoundSampleHelper is a specific helper for the CompoundSample class, I don't see an issue with using instance variables in this context. The tight coupling seems appropriate here, given the helper’s dedicated scope.

lib/tasks/assign_compound_sample_supplier_name.rake
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+namespace :compound_sample do
+  desc 'Set the supplier_name on compound samples if all component samples share the same supplier_name'
+  task set_consistent_supplier_name_for_compound_samples: :environment do
```
yoldas, line 4: The criteria says " (just for 007c compound samples)". This rake task however does not have any filter to apply only to relevant compound samples. There should be a submission/request based filter I think.

yoldas (changes requested):
Looks good but the rake task needs to operate on scRNA Core related samples only, according to the criteria.

lib/tasks/assign_compound_sample_supplier_name.rake
sabrine33: Thank you, @yoldas — I’ve been looking for a way to establish this, and it does not look fairly straightforward.

scRNA requests are mainly for library prep. For sequencing requests, they can basically choose any platform — although in the Lucid chart, it’s mentioned that they tend to use the NovaSeqX.

Maybe I’m missing something, but do you know how to filter samples coming from scRNA requests? I’m not sure how to isolate only the 007c compound samples.

On the other hand, the code change covers all sequencing requests, so in my opinion, for the sake of data consistency, it would be better to update all the compound samples accordingly.

yoldas (approved):
Thanks for checking how many samples will be processed.

PR #5115, opened 2025-07-29
Y25 003: add check for number_of_pools eq 0, to prevent divide by zero errors

sabrine33:


app/models/bulk_submission.rb
```diff
@@ -165,7 +165,7 @@ def process # rubocop:todo Metrics/CyclomaticComplexity
       # If the submission template name matches `SCRNA_CORE_CDNA_PREP_GEM_X_5P` and all required headers are present,
       # the allowance band is calculated for each study and project combination.
       # Otherwise, an empty hash is assigned.
-      @allowance_bands = calculate_allowance_bands
+      @allowance_bands = calculate_allowance_bands unless errors.count > 0
 
       raise ActiveRecord::RecordInvalid, self if errors.count > 0
```
StephenHulme: Should the `unless errors.count > 0` be inside the calculate_allowance_bands function? I see the same check is made on the line below where an error is raised?

spec/models/scrna_core_cdna_prep_feasibility_validator_spec.rb
```diff
@@ -709,4 +709,24 @@ def scrna_config
...
+      it 'adds the error message' do
+        error_message =
+          I18n.t(
+            'errors.number_of_pools_exists',
+            scope: i18n_scope
+          )
+        expect { bulk_submission.process }.to raise_error(ActiveRecord::RecordInvalid)
+        expect(bulk_submission.errors[:spreadsheet]).to include(error_message)
+      end
+      # rubocop:enable RSpec/MultipleExpectations, RSpec/ExampleLength
```
StephenHulme: Could this be split into individual examples? Maybe put the `.to raise_error(ActiveRecord::RecordInvalid)` check in it's own example called `it 'raises an invalid record error' do`?

yoldas (changes requested):
Number of pools is specified per study-project group. User does not specify the "total" number of pools across all groups. The screenshots in the story are showing the problems of empty or zero values using a single study-project group. If there is a problem, the feedback should specify at least which study project group has a problem if not which row. See the validate_unique_values example method in app/models/submission/validations_by_template_name.rb as an example user feedback.

Could you please check the logic in the PR again according to this information?

spec/models/scrna_core_cdna_prep_feasibility_validator_spec.rb
```diff
@@ -709,4 +709,27 @@ def scrna_config
       # rubocop:enable RSpec/MultipleExpectations, RSpec/ExampleLength
     end
   end
+
+  describe '#validate_scrna_core_cdna_prep_total_number_of_pools_is_not_zero?' do
```
yoldas: Note: 
This test file aims testing `ScrnaCoreCdnaPrepFeasibilityValidator`. The description at the top:
```
# This spec tests the Submission::ScrnaCoreCdnaPrepFeasibilityValidator module
# included by the BulkSubmission class through the ValidationsByTemplateName
...
```
By convention, `#` prefix is used for showing that this individual test example is for an instance method.

The feasibility validator has no such method as `validate_scrna_core_cdna_prep_total_number_of_pools_is_not_zero?` and even the BulkSubmission class does not have such a method after including other modules.

[updated: there is no such method anywhere]

PR #5124, opened 2025-07-29
Y25-450 - API Allow list parameter

### LINKED ISSUE #5107: Y25-450 - API key allowlist
**User story**

Implement an allowlist of endpoints where use of an API key is not required. See discussion here - https://github.com/sanger/sequencescape/issues/5094#issue-3248655438

Katy

This can be tested within the team.

**Acceptance criteria**

- [ ] Certain endpoints (URL + http method) can be added to an 'allowlist'
- [ ] Endpoints on this list do not require an API key to be passed 
- [ ] It is quick and easy for a developer to add/remove an endpoint to/from the 'allowlist' (e.g. single line change + deployment)

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

If there's a way to mark an endpoint in the 'routes' file, this would be an intuitive place to store it. Otherwise, a separate config file. They would be referred to from `app/controllers/api/v2/concerns/api_key_authenticatable.rb` - this is where it currently checks the API key.

**Deployment notes**

BenTopping:
Closes #5107 

- Adds ability to include 'permissive' as a path_parameter for a given route in order to skip API key checks.

### DIFF
--- a/app/controllers/api/v2/concerns/api_key_authenticatable.rb
+++ b/app/controllers/api/v2/concerns/api_key_authenticatable.rb
@@ -74,7 +74,8 @@ module Api
         #
         # A route is considered permissive if the 'permissive' path parameter is present in the request
         # and the HTTP request method is 'GET'.
-        # Path parameters are defined next to the route in routes.rb e.g. jsonapi_resources :samples, permissive: true
+        # Path parameters can be defined via defaults next to the route in routes.rb
+        # e.g. jsonapi_resources :samples, defaults: { permissive: true }
         #
         # @return [Boolean] true if the route is permissive, false otherwise.
         def permissive_route
--- a/app/controllers/api/v2/concerns/api_key_authenticatable.rb
+++ b/app/controllers/api/v2/concerns/api_key_authenticatable.rb
@@ -3,6 +3,7 @@ module Api
   module V2
     module Concerns
       # Provides the tools needed to confirm that a valid API key was provided for the header X-Sequencescape-Client-Id.
+      # - Skips the check if the endpoint specifies permissve GETs
       # - Where the API key is in the request and exists in among ApiApplication, allow the request to be served.
       # - Where the API key is in the request but does not exist, log the attempt and render an unauthorized response.
       # - Where the API key is not in the request, respond normally (for now) and log the system that made the request.
@@ -12,6 +13,9 @@ module Api
         included { prepend_before_action :authenticate_with_api_key }
 
         def authenticate_with_api_key
+          # Check if the route requires an API key
+          return if permissive_route
+
           http_env_api_key = 'HTTP_X_SEQUENCESCAPE_CLIENT_ID'
 
           if request.env.key? http_env_api_key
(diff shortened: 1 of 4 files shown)


app/controllers/api/v2/concerns/api_key_authenticatable.rb
```diff
@@ -65,6 +69,18 @@ def render_unauthorized
                    ]
                  } and return
         end
+
+        # Checks if the current request is a permissive route.
+        #
+        # A route is considered permissive if the 'permissive' path parameter is present in the request
+        # and the HTTP request method is 'GET'.
+        # Path parameters are defined next to the route in routes.rb e.g. jsonapi_resources :samples, permissive: true
```
StephenHulme: Just want to double-check this wording, the route is permissive if the path parameter is present in the _route definition_, rather than the request headers themselves?

spec/controllers/api/v2/concerns/api_key_authenticatable_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  # Dummy controller to include the concern for testing
+  controller(ApplicationController) do
+    include Api::V2::Concerns::ApiKeyAuthenticatable # rubocop:disable RSpec/DescribedClass
+  end
+
+  describe '#permissive_route' do
+    context 'when permissive param is present and request is GET' do
+      it 'returns true' do
+        allow(controller.request).to receive_messages(path_parameters: { permissive: true }, get?: true)
+        expect(controller.send(:permissive_route)).to be true
```
StephenHulme: [nitpick] I think the tests would be clearer if the `allow` statement was in the context rather than the expectation (x3)

app/controllers/api/v2/concerns/api_key_authenticatable.rb
BenTopping: Yes slightly confusing, path_parameters form part of the request but are derived from the route definition.

StephenHulme: Sounds fair, so there is no way to put `permissive=true` in the request header to get past the check?

BenTopping: The documentation is pretty poor but I believe that is the case, see: https://github.com/rails/rails/blame/33beb0a38db1c058123a8e3cc298cad918adfe32/actionpack/lib/action_dispatch/http/parameters.rb#L67. 

I could use the [better documented defaults](https://guides.rubyonrails.org/routing.html#defining-default-parameters) which explicitly say are not overrideable to be on the safer side...

StephenHulme: I'm encouraged by the `delete_header("action_dispatch.request.parameters")` but agree that the lack of documentation is concerning... it might imply that it could be silently removed in a future update? The defaults might be better?

I would feel better if we had a test-driven spec to confirm that it's not overridable, but that might be hard to prove if we aren't sure if it can be overridden...

BenTopping: Ill do some testing and see what comes up.

StephenHulme: Thanks 👌

BenTopping: Looks like defaults vs no defaults are synonymous from some testing. I will update to use defaults explicitly as they are better documented.

PR #5130, opened 2025-07-31
Y24-415:Ultima library types

sabrine33:


andrewsparkes (approved):
I do wonder why they want 'High Throughput' in the library type like that, seems superfluous and isn't in all the other high throughput pipeline library types.
edit: I see that they want a 'Low Throughput' version later. We used to name those with 'Bespoke'.

PR #5131, opened 2025-07-31
Y25-410:ultima plate purpose

sabrine33:


config/default_records/plate_purposes/016_ultima_purposes.wip.yml
```diff
@@ -0,0 +1,6 @@
+---
+# 96 well input plates
```
andrewsparkes, line 2: Given the name you probably need to add:
`cherrypickable_target: true`

andrewsparkes: Doesn't hurt to explicitly say these too, although they are the defaults:
```
target_type: Plate
size: 96
```

andrewsparkes (changes requested):
add cherrypickable_target true (See PF Cherrypicked)

dasunpubudumal (approved):

PR #5137, opened 2025-08-01
Y25-418 - Ultima Tag sets

### LINKED ISSUE #5062: Y25-418 - Support New tags for Ultima sequencing submissions
**User story**
As a Laboratory Automation Specialist (Abdulrahman), I would like to have new tag layout to support Ultima sequencing submissions.

Abdulrahman T

This is can be tested by the developer.

The customer is using gatekeeper to create the tag plate, please have a look if we need to do any extra work in gatekeeper to support this.

**Acceptance criteria**

The feature should be feature-flagged, to allow decoupling of testing and deployment.
- [ ] create new tags and tagsets and tag layout templates 

Optional: To fully test this in the integration suite, update the tagset for the PCR Free pipeline to use Ultima P1/2 for the tag plate. The integration suite should still run successfully.

References
https://github.com/sanger/sequencescape/issues/5041

**Additional context**
The tag set files are available on the Confluence page.
The tag group and  the tag layout name both under the names of ‘Ultima P1’ and ‘Ultima P2’
directions: columns
walking algorithm: Walk wells of plate

BenTopping:
Closes #5062

- Adds Ultima 1 and Ultima 2 tag groups.
- Adds Ultima 1 and 2 layout templates.

### Additional Context
- No Tag Set created - unsure if this is needed.


config/default_records/tag_groups/005_ultima.yml
```diff
@@ -0,0 +1,199 @@
+Ultima P1:
+  visible: true
```
BenTopping, line 2: No adapter_type given

config/default_records/tag_layout_templates/limber.dev.yml
BenTopping, line 1: .dev files so will need manual creation in training/production

KatyTaylor: Why not put them in a new file that isn't `.dev.yml`, so we don't have to create them manually?

config/default_records/tag_groups/005_ultima.yml
```diff
@@ -0,0 +1,199 @@
...
+    92: CAAGCATATCTAGAT
+    93: CAATGACAGCTCGAT
+    94: CTGATGAGCTTGATGAT
+    95: CATATGACCTGAGAT
+    96: CAGCCGATATCAGAT
+
+Ultima P2:
+  visible: true
+  tags:
+    97: CAGTCAGTTGCAGAT
```
KatyTaylor: I think these integer keys correspond to `map_id` on the `Tag` model. It makes sense to me to start from 1 for the first tag in a tag group - and looks like this is what most existing ones do. Was it intentional to start from 97? Thanks.

See
```
SELECT 
  MIN(map_id) AS first_map_id, 
  MAX(map_id) AS final_map_id, 
  MAX(map_id) - MIN(map_id) + 1 AS num_tags_assuming_sequential
FROM tags
GROUP BY tag_group_id
;
```

KatyTaylor (commented):
Haven't checked the oligo sequences themselves are correct... happy to do that though if you like.

Have left a few comments/questions.

config/default_records/tag_groups/005_ultima.yml
BenTopping: Yep you are right, think its a copy paste error, thanks for spotting. Updated in: [9688d5e](https://github.com/sanger/sequencescape/pull/5137/commits/9688d5ea3b23b1d35f5c2e6fb6036e6aee8dd979)

config/default_records/tag_groups/005_ultima.yml
andrewsparkes: From what I can tell, adapter_type is used to filter tag groups. So if you want all the relevant ones for the Chromium type of tagging chemistry you can filter to the ones with 'Chromium' adapter type.
Here maybe you could just set it to 'Ultima' for now.

config/default_records/tag_groups/005_ultima.yml
```diff
@@ -0,0 +1,199 @@
+Ultima P1:
```
andrewsparkes, line 1: Little concerned about the group name being mostly the sequencing machine name. If Ultima sequencing machines get used a lot in future, and more tags are created, this name could prove to be way too non-descriptive. Normally the sequencer is not specified for tag groups.
However, there doesn't seem to be much of a standard.
e.g. PF_adapters_UDI_96_A_i7
e.g. PN-1000212 Single Index Kit N, Set A
e.g. pWGS i5 UDI tag group v2 A

andrewsparkes: Also the tag groups might be used for sequencing runs on other sequencers? Perhaps the way they are made is special and unique to Ultima sequencing though, I don't know?

BenTopping: I believe these Tag groups are [Ultima specific](https://eu.idtdna.com/pages/products/next-generation-sequencing/workflow/xgen-ngs-library-preparation/ngs-adapters-indexing-primers/xgen-for-ultima-genomics). 

I double checked the naming with Mike Q and he had already agreed during the research phase that these both be called Ultima P1 / P2.

KatyTaylor: Rich Clark has been advocating for using the exact manufacturer names (e.g. 'PN-1000212 Single Index Kit N, Set A') - but I believe Sabrine suggested this already and they prefer this naming.

KatyTaylor (approved):

PR #5177, opened 2025-08-15
Y25-184: upgrade dataTable bootstrap to v2

sabrine33:


app/frontend/datatables/config.js
```diff
@@ -1,13 +1,19 @@
-// default for bootstrap with added buttons
-const dom =
-  "<'row'<'col-sm-12 col-md-6'l><'col-sm-12 col-md-6'f>><'row'<'col-sm-12'tr>><'row'<'col-sm-12 col-md-5'i><'col-sm-12 col-md-7'pB>>";
```
StephenHulme, line 3: I assume this isn't used anywhere?

sabrine33: Apologies, I meant to include this in the PR. Yes, the arrows do look different in shape and position. I tried a few workarounds by overriding the DataTables Bootstrap styling, but there are still some odd behaviors when sorting. To be fair, I spent about 1h30 on this Friday afternoon and thought it would be better to discuss it with the team—considering the time investment and whether it’s really worth it. If we agree to proceed, we could treat this as a UI enhancement, maybe?

Regarding the `dom` property, the string isn’t needed as there are no visual changes with or without it. That said, the `dom` property is deprecated in v2 and will no longer be supported in v3.
Thank you for having a look into this!

sabrine33:
> I think I found it...
> 
> In `yarn.lock` there are two references to `jquery@`, one is version `3.5.1` the other version `3.7.1`. Guessing that they might be conflicting, and only seeing a reference to `jquery@>=1.7`, I deleted both jquery entries in the lockfile, saved, and rebuilt with `yarn install --dev`. This removed the reference to version `3.5.1` and combined both into a single entry `jquery@>=1.7, jquery@^3.7.1` - which now seems to work! 🤞
> 
> Would you please try that on your side to confirm that it works? I based my test on commit [cd65b4e](https://github.com/sanger/sequencescape/commit/cd65b4e3e390ac4b361f53b0dc970fd3c454858e) `resolve conflicts`.

I don't think we are suppose to touch the lock file

Issue #4962, opened 2025-05-27
Y25-355 - Update Vue from 2.7 to 3.x (Remove Vue and Plate Picks)


StephenHulme:
**Describe the Housekeeping**
Update Vue to version 3, or remove it from Sequencescape (see comments below).

**Blocking issues**
There are a few dependency updates that will need to be made as part of this upgrade.

**Additional context**
See the Limber equivalent https://github.com/sanger/limber/issues/2188

**How to rollback**

Go to https://github.com/sanger/sequencescape/pull/5216 and click Revert and merge the PR.
Release and deploy Sequencescape.

yoldas:
<img width="1152" height="596" alt="Image" src="https://github.com/user-attachments/assets/e9d48089-7685-4e3e-a87e-20e834a962d0" />

PR #5216, opened 2025-09-04
Y25-355 [PR] - Update Vue from 2.7 to 3.x (Remove Vue and Plate Picks)

yoldas:
Closes #

- Remove Vue from entrypoints, eslint config, vite config, and package.json
- Update plate_picks view with a warning message about the removal and help. [Screenshot](https://github.com/sanger/sequencescape/issues/4962#issuecomment-3255885317)
- Remove plate picks and shared vue components
- Remove miragejs, whatwg-fetch, and vitest

Rollback will be the revert of the merge commit and deployment of a new release.


package.json
```diff
@@ -26,26 +26,19 @@
...
   "devDependencies": {
     "@vitejs/plugin-legacy": "^1.8",
-    "@vue/test-utils": "^1.3",
     "eslint": "^9.19",
     "eslint-config-prettier": "^8.10",
     "eslint-plugin-vitest": "^0.5.4",
-    "eslint-plugin-vue": "^9.32",
     "globals": "^15.14",
     "jsdom": "^26.0",
     "miragejs": "^0.1.48",
```
BenTopping: `miragejs` can be removed too.

package.json
```diff
@@ -26,26 +26,19 @@
     "select2": "^4.1.0-rc.0",
     "sortablejs": "^1.15.6",
     "terser": "^5.31.1",
-    "vue": "^2.7.16",
-    "vue-loader": "^15.11.1",
-    "vue-template-compiler": "^2.7.16",
-    "vuex": "^3.6.2",
     "whatwg-fetch": "^3.6.20"
```
BenTopping: `whatwg-fetch` can be removed too

package.json
```diff
@@ -55,8 +48,8 @@
     "@rollup/rollup-win32-x64-msvc": "*"
   },
   "scripts": {
-    "test": "vitest run",
+    "test": "vitest run --passWithNoTests",
```
BenTopping: Do we have no vitest tests now without plate-picks? Can we remove vitest / should we look at adding tests for other components?

package.json
yoldas: "The fetch() function is a Promise-based mechanism for programmatically making web requests in the browser."
No other use of fetch() polyfill in the code, so I have removed the dependency now.

package.json
yoldas: "Mirage JS is an API mocking library"
There are no more imports from mirages such as Server and Model in the code. I have removed the dependency.

package.json
yoldas: Yes, there are no more tests now. I am not sure about whether to remove vitest though.

BenTopping: We also run it in CI but it is redundant so we should probably remove it from there too. We can always re-add vitest at a later date if we find we need it. CC @StephenHulme @KatyTaylor

yoldas: I have removed the vitest dependency and the CI test step.

.github/workflows/js_test.yml
BenTopping, line 1: I think you can delete this whole workflow. Without the yarn test its just doing setup and nothing else.

yoldas: Thanks. I have deleted the workflow now.

PR #5220, opened 2025-09-05
Y25-449 - Flakey plate_purpose test fix 

### LINKED ISSUE #5103: Y25-449 - [BUG] RSpec Unit test intermittently fail with json atom at path "plate_purpose" is missing
**Describe the bug**
Rspec unit test intermittently fail with

```
  1) /api/1/plate_purposes #post supports resource creation
     Failure/Error: expect(JSON.parse(response.body)).to include_json(JSON.parse(response_body))

                 json atom at path "plate_purpose" is missing
                 
     # ./spec/api/plate_purpose_spec.rb:41:in 'block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:73:in 'block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:73:in 'block (2 levels) in <top (required)>'

```
Example of failure (https://github.com/sanger/sequencescape/actions/runs/16467454469/job/46548224781?pr=5021)

**To Reproduce**
1. Run unit tests (bundle exec rspec spec/api/plate_purpose_spec.rb)

BenTopping:
Closes #5103

- Changes test to be non-input plate.

#### Additional Context

This issue is somewhere around https://github.com/sanger/sequencescape/blob/develop/app/api/core/service.rb#L169.
When a input PlatePurpose is created over the v1 api it goes to the plate_purpose endpoint but updates its own class/type to become PlatePurpose::Input. This is messing with the v1 implementation where in the link above the v1 creation uses class to eager load records, which in input plate purpose case changes during creation.
I am not sure why its flaking on CI, it always fails locally.

The solution is to make this plate non-input for the test. This does reduce test coverage but we arent changing any behaviour and this endpoint is redundant due to Limber v2 upgrades so this should be a safe change.

I suspect this starting flaking / broke during a rails upgrade.

### DIFF
--- a/spec/api/plate_purpose_spec.rb
+++ b/spec/api/plate_purpose_spec.rb
@@ -15,7 +15,7 @@ describe '/api/1/plate_purposes' do
         "plate_purpose":{
           "name": "External Plate Purpose",
           "stock_plate": true,
-          "input_plate": true,
+          "input_plate": false,
           "size": 384
         }
       }'
@@ -40,7 +40,7 @@ describe '/api/1/plate_purposes' do
       api_request :post, subject, payload
       expect(JSON.parse(response.body)).to include_json(JSON.parse(response_body))
       expect(status).to eq(response_code)
-      expect(Purpose.last).to be_a(PlatePurpose::Input)
+      expect(Purpose.last).to be_a(PlatePurpose)
     end
   end
 end


yoldas (approved):
The description makes sense and the test is for v1 end point. Thanks for this fix.

PR #5229, opened 2025-09-10
Y25-398 - Transfer pooling info from SS to MLWH

### LINKED ISSUE #5034: Y25-398 - Send pooling info to MLWH
**User story**

The glue between https://github.com/sanger/limber/issues/2427 and https://github.com/sanger/unified_warehouse/issues/795

To be defined further.

Katy

James M and Jamie L, or nominated team members - for the end to end process involving all 3 stories. Or just test within the team, depending on what UAT has been done already on the other stories.

**Acceptance criteria**

- [ ] ...

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- https://github.com/sanger/limber/issues/2427
- https://github.com/sanger/unified_warehouse/issues/795

**Additional context**

**Deployment notes**

sabrine33:
Closes #

- 


app/models/api/messages/comment_io.rb
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
+
+##
+# Api::Messages::CommentIo is responsible for serializing and exposing
+# comment-related metadata linked to PolyMetadatum records. It maps domain
+# attributes (such as batch, position, and tag index) into JSON for API
+# consumption.
```
dasunpubudumal: It's not _really_ for API consumption. The serialisation is for message passing (message passing is more akin to an "interface" rather than an API). The consumer de-serialises and morphs the message into a Rails model that is eventually serialised into an SQL query (which we don't see because of `ActiveRecord` abstractions).

dasunpubudumal (approved):
Looks good to me :)

There are too many ways in which one can invoke messages. That is not good. Hopefully we will be able to unify them with the upcoming [epics](https://github.com/sanger/General-Backlog-Items/issues/609#issuecomment-3269487081).

app/models/api/messages/comment_io.rb
```diff
@@ -0,0 +1,107 @@
...
+        batch_id: batch.id,
+        position: batch_request.position,
+        tag_index: aliquot.aliquot_index_value
+      }
+    end
+
+    def current_entry
+      # Return the first entry, as only one aliquot is expected
+      # for the PolyMetadatum of Request type.
+      @current_entry ||= batch_aliquots.first || {}
```
harrietc52: Is it ok to fail silently if there was more than one aliquot?

app/models/poly_metadatum.rb
```diff
@@ -30,8 +32,22 @@ class PolyMetadatum < ApplicationRecord
...
+  def related_batches
+    return [] unless metadatable_type == 'Request'
+
+    metadatable.submission.batches
+  end
+
+  def broadcast_under_rep_batch_aliquot
+    return unless key == 'under_represented' && related_batches.any?
+
+    Messenger.create!(target: self, template: 'CommentIo', root: 'comment').broadcast
```
harrietc52: What is `root: ‘comment’` for?

spec/models/api/messages/comment_io_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    aliquot
+    batch_request
+    poly_metadatum
+    aliquot_index
+  end
+
+  it 'generates a valid json format' do
+    expect(message.as_json).to eq(expected_json)
+  end
+end
```
harrietc52: Do you want to maybe include multiples of some of the entities? e.g multiple batches in the submission, multiple aliquots for the asset. 

There are a few logic checks in the io code which could be tested too, eg when there are no `matching_aliquots`

harrietc52 (approved):
Just looking at the code, as I don’t have loads of context, but its really clean and readable :) nice!

I’ve noted a few thoughts, and some more tests could be added I would say, but only suggestions - feel free to ignore.

app/models/api/messages/comment_io.rb
StephenHulme: I would agree with this, that API v1 code is scheduled to be removed en-masse in the coming weeks/months with the final removal of where it is used. If this is not part of API v1 and it is possible, it would be good to move this to a different folder.

app/models/poly_metadatum.rb
```diff
@@ -16,6 +16,8 @@ class PolyMetadatum < ApplicationRecord
   # Associations
   belongs_to :metadatable, polymorphic: true, optional: false
 
+  include Api::Messages::CommentIo::PolyMetadatumBatchAliquots
```
andrewsparkes: Polymetadata class is for use in any class, not just aliquots. I don't like the idea of putting Aliquot specific code in this class. Same for other aliquot specific code you've added below.

app/models/poly_metadatum.rb
```diff
@@ -30,8 +32,29 @@ class PolyMetadatum < ApplicationRecord
...
 
+  after_commit :broadcast_under_rep_batch_aliquot
+
   # Methods
   def to_h
     { key => value }
   end
+
+  def related_released_batches
+    return [] unless metadatable_type == 'Request'
```
andrewsparkes: why is this Request and not Aliquot?

andrewsparkes (changes requested):
Not to keen on Aliquot specific code being put in the generic polymetadata class. Can it go in the Aliquot model?

app/models/poly_metadatum.rb
sabrine33: it is been stored within the request - different PR (merged and released) - this PR is for broadcast them to the mlwh

sabrine33: @andrewsparkes  this is the related PR : https://github.com/sanger/limber/issues/2427

andrewsparkes: Ok, but the code should be in Request and Aliquot models, not in the polymetadata class I think.

app/models/poly_metadatum.rb
```diff
@@ -30,8 +32,29 @@ class PolyMetadatum < ApplicationRecord
...
+  #
+  # If the PolyMetadatum has the key 'under_represented', is associated with a 'Request',
+  # and that request is linked to any released batches, a Messenger notification is created.
+  # The template is used to construct the JSON payload for the message,
+  # and the root specifies the type of object the message refers to.
+  # Messenger notification are picked up by a background job which sends the message to the MLWH.
+  def broadcast_under_rep_batch_aliquot
+    return unless key == 'under_represented' && related_released_batches.any?
+
+    Messenger.create!(target: self, template: 'CommentIo', root: 'PolyMetadatum')
```
harrietc52: I don’t know much about this area of the code, but would this rebroadcast all items in the PolyMetadatum? I think my question is related to Andrews, and wondering why broadcasting something specific is in the metadata class

spec/models/api/messages/comment_io_spec.rb
harrietc52: A few more tests could still be added, for each of the methods in the PolyMetadatum model.

app/models/poly_metadatum.rb
dasunpubudumal: Think it's just the model name. The template is still `CommentIo` and hence the data that is pushed would be to the `comments` table. 

This works because she has got a model called `PolyMetadatum` in [`unified_warehouse`](https://github.com/sanger/unified_warehouse/pull/826/files), and the `table_name` in that model is `comments`.

sabrine33: `CommentIo` is the class used for serialization (the to_json method). The root represents the model around which the message is constructed, and the message will be sent with that model’s type as the key (in this case, `PolyMetadatum`).

harrietc52: Ok :)

app/models/poly_metadatum.rb
yoldas: issue: I agree with @andrewsparkes . The responsibility of PolyMetadatum is handling key value pairs. The PR is adding after_commit hook, broadcasting, and other methods for a very specific task and there are not enough guards to avoid side effects.

app/models/poly_metadatum.rb
yoldas: note: root is the name of the top-level key name in the JSON message, which is used by MLWH side to decide which model should handle it. `{ comment: { ... } }` . 

issue: However, on SS side here, this broadcasting is wired to an after_commit hook on any poly metadatum change.

sabrine33: > note: root is the name of the top-level key name in the JSON message, which is used by MLWH side to decide which model should handle it. `{ comment: { ... } }` .
> 
> issue: However, on SS side here, this broadcasting is wired to an after_commit hook on any poly metadatum change.

my understanding, the message is broadcasted when the a 'Messenger' object is created - here it is only created when it is related to 'under_represented' well and needs to be linked to a released batch. Currently we don't support well 'under_rep' status update, but we also would like the message to be broadcast once we start supporting it.

app/models/poly_metadatum.rb
sabrine33: > Ok, but the code should be in Request and Aliquot models, not in the polymetadata class I think.

I am not sure how to move this to Request or Aliquot model - as this is to check if we need to broadcast the poly_metadata to the mlwh or not

app/models/poly_metadatum.rb
```diff
@@ -30,8 +32,29 @@ class PolyMetadatum < ApplicationRecord
   # metadatable_id is the database id of the model instance
   validates :key, uniqueness: { scope: %i[metadatable_type metadatable_id], case_sensitive: false }
 
+  after_commit :broadcast_under_rep_batch_aliquot
```
KatyTaylor: Sorry to be a pain, as I know you made this PR a while ago!

This looks like it will broadcast a message when the polymetadatum is first created. The plan on the story was to broadcast it at a similar time to when the iseq_flowcell messages are created e.g. when the sequencing batch is started. Did you consider this approach? Could discuss this and other comments in a meeting.

KatyTaylor: Drawing to help discussion 😆 Showing the two possible points where we could send to the MLWH. Probably needs me to talk through in person.

Image

sabrine33: > Sorry to be a pain, as I know you made this PR a while ago!
> 
> This looks like it will broadcast a message when the polymetadatum is first created. The plan on the story was to broadcast it at a similar time to when the iseq_flowcell messages are created e.g. when the sequencing batch is started. Did you consider this approach? Could discuss this and other comments in a meeting.

No, I wasn’t aware of the plan. From what I remember, the iseq_flowcell messages are only sent on batch release (perhaps also on status updates?). So, are we expecting users not to go back and mark a well as under_rep again? Maybe that never happens.

Also, when we were UAT-testing in the first round, James and Jamie just went and marked a well from a released batch — if we go with that approach,  may be I’ll need to notify them.

sabrine33: also - is this only releated to illumina pipelines ? ? what about aviti pipeline - populating eseq_flowcell table?

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -0,0 +1,135 @@
...
+    delegate :deleted_at, to: :poly_metadatum
+    delegate :destroyed?, to: :poly_metadatum
+  end
+
+  # Returns all requests related to the batch that include a poly_metadatum entry
+  # with the `UNDER_REPRESENTED` key.
+  #
+  # @return [Array<Request>] an array of requests that contain under-represented metadata.
+  def request_with_under_represented_wells
+    submissions.flat_map(&:requests).select do |r|
```
KatyTaylor, line 47: You can get straight from a batch to its requests without the 'submissions' part. There's a `has_many :requests` association on the batch model.

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -0,0 +1,135 @@
...
+    request.poly_metadata.select { |meta| meta.key == UNDER_REPRESENTED_KEY }
+  end
+
+  ##
+  # Builds comments for a specific poly_metadatum by iterating through
+  # all target aliquots and matching them to lane aliquots by tag map ID.
+  #
+  # The starting point is the request that holds the poly_metadatum for the
+  # under-represented well. This request is typically a LibraryRequest, whose
+  # asset corresponds to the under-represented well.
```
KatyTaylor: Actually for our use case it would be the 'target_asset' that was the under-represented well

app/models/under_rep_well_comments_to_broadcast.rb
KatyTaylor: Ah but this won't work, sorry. The requests associated with the batch will be sequencing requests, and the under-represented wells polymetadata is on the library requests.

We can have a chat about how to retrieve the poly_metadata starting from the batch. We'll have to work backwards through the request chain I think. sequencing_request --> library request

sabrine33: > Ah but this won't work, sorry. The requests associated with the batch will be sequencing requests, and the under-represented wells polymetadata is on the library requests.
> 
> We can have a chat about how to retrieve the poly_metadata starting from the batch. We'll have to work backwards through the request chain I think. sequencing_request --> library request

I am not sure what is the issue of the current implementation - getting the requests from the submission

KatyTaylor: Outcome of our discussion:
`batch.requests` just pulls back the requests attached to the batch, so the sequencing requests.
`batch.submissions` pulls back the submissions attached to the above requests. 
`batch.submissions.flat_map(&:requests)` pulls back the requests attached to the above submissions: the sequencing requests, but ALSO the library prep and multiplexing requests

This works because the submissions in the WGS pipeline include all the requests upfront - library prep, multiplexing, and sequencing - they all share a submission.

For more recent, more complicated pipelines, this would not work, because there can be break points in the pipeline where new submissions must be made to continue - e.g. they just make the library prep submission up front and then subsequent requests are made by  a new manual submission, or an automated one.

Going to leave the code as-is, because this story is a proof-of-concept for the WGS pipeline. If we extend it in the future for other pipelines this will be a fragile point that will need updating.

KatyTaylor (approved):

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -0,0 +1,135 @@
...
+  # The Lane provides access to the `batch_request` (through `lane.source_request`),
+  # which contains the lane position information used in the generated comment.
+  #
+  # @param request [Request] the request that owns the poly_metadatum
+  # @param poly_meta [PolyMetadatum] the metadata record for under-represented wells
+  # @return [Array<UnderRepWellComment>] list of constructed comment objects
+
+  def build_comments_for_poly_meta(request, poly_meta)
+    request.target_asset.aliquots.filter_map do |aliquot|
+      lane = request.asset.descendants.last
```
KatyTaylor, line 118: Another possibly fragile point if used in more recent, more complicated pipelines - not confident whether the `descendants` links would all be intact. OK for now.

yoldas (approved):
praise: Looks good. It is easy to follow and tested as well.
note: asset and target_asset depends on the submissions and the pipeline. For example, in some cases, target_asset may come up as Tube in some others as Lane.

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -0,0 +1,140 @@
...
+      next unless aliquot_matches_lane?(lane, aliquot)
+
+      UnderRepWellComment.new(
+        position: lane.source_request.position,
+        batch_id: id,
+        tag_index: aliquot.tag.map_id,
+        poly_metadatum: poly_meta
+      )
+    end
```
KatyTaylor, line 127: Hi @sabrine33 , we were discussing this code on Friday. Apologies I didn't catch this at code review time, but I think there are several issues with this method.

1. `request.asset.descendants.last` relies on the order of descendants matching the order that the transfers happen in - not sure if this is correct. _Potential consequence: finds the wrong receptacle, leading to incorrect data in the Comment record._
2. `aliquot_matches_lane?` is flawed because aliquot_index_value is not set based on the tag.map_id. Instead, it is an integer from 1 to however many aliquots there are in the lane (e.g. the one I checked had 270 aliquots). _Potential consequence: Quite likely to always return true by accident. Not a useful check but may not have caused any harm._
3. `tag_index: aliquot.tag.map_id` - we need to set tag_index to the same value as the equivalent lane aliquot's aliquot_index_value. _Potential consequence: most data created so far is incorrect - won't be able to successfully join the comment table to the iseq_flowcell table._

I think an alternative approach could be (pseudo-code):
```
# Start with the library prep request that has the polymetadata containining under-represented information (as current)
# Find any related downstream lanes via a more reliable method - perhaps step-wise down the transfer requests, and/or using the sti_type of the receptacle or labware to verify it's a lane
# Find the relevant aliquot within the lane, by comparing with the request target asset aliquot. However, instead of comparing aliquot index value with map id, compare sample + tag id + tag 2 id + tag depth.
# Set the tag_index of the comment to be the same as the lane aliquot's aliquot_index_value.
```

Please let me know if you disagree or if I can help with any of it. This is without thinking about your current story. I think it still works, but the other bit where it finds the relevant requests would need to change (think you've tackled that bit already).

KatyTaylor: P.S. We'll have to do an exercise to check if the data created in prod so far is affected by this.

sabrine33: Hi @KatyTaylor - I changed the whole logic around this already. I just need to update the tests now. I can create a research story to validate prod data

KatyTaylor: Great thanks, look forward to seeing the new version!

PR #5230, opened 2025-09-10
y25-423_Ultima_Sequencing_pipeline_page

wendyyang:
Closes #

- add Ultima model, config & validator


app/validators/ultima_validator.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+class UltimaValidator < ActiveModel::Validator
+  # Used in _pipeline_limit.html to display custom validation warnings
+  def self.validation_info
+    'Batches must contain exactly two requests.'
+  end
```
sabrine33, line 9: I think this is redundant. As there is already  `This pipelines has a limit of 2 requests in a batch. Any requests that would take you over this limit will be disabled and appear faded. `

sabrine33 (commented):
Looks good @wendyyang - Thanks . Just minor request. 
Also, I tested this, and it looks like the fragment size required (from) and (to ) columns they not showing on the requests table - Pipeline page, could you please have a look, thanks

app/models/ultima_sequencing_pipeline.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Specialized sequencing pipeline for Ultima
+class UltimaSequencingPipeline < SequencingPipeline
+  def ot_recipe_consistent_for_batch?(batch)
+    batch.requests.map { |r| r.request_metadata&.ot_recipe }
```
yoldas: This line looks redundant.

app/validators/ultima_validator.rb
yoldas: This should be either removed OR it should provide any validation information for your pipeline that is useful for the user. It is used by the `pipeline_limit` partial ERB with a safe navigation operator to show specific pipeline validation details: `@pipeline.validator_class_name&.constantize&.validation_info`. 

The limit information is already shown in the batch limit section using the `@pipeline.item_limit` but it is a different paragraph. It shows the same info.

config/default_records/request_information_types/003_ultima_request_information_types.wip.yml
```diff
@@ -0,0 +1,21 @@
+---
+ot_recipe:
+  name: OT recipe
+  key: ot_recipe
+  label: OT recipe
+  width: 5
+  hide_in_inbox: 0
+
+fragment_size_required_from:
+  name: Fragment size required(from)
```
yoldas: This needs to have a space character between the word `required` and `(from)` to make it consistent across the application. Apply the same to the label value.

config/default_records/request_information_types/003_ultima_request_information_types.wip.yml
```diff
@@ -0,0 +1,21 @@
...
+  key: fragment_size_required_from
+  label: Fragment size required(from)
+  width: 5
+  hide_in_inbox: 0
+
+fragment_size_required_to:
+  name: Fragment size required(to)
```
yoldas: This needs to have a space character between the word `required` and `(to)` to make it consistent across the code and tests. Apply the same to the label value.

spec/validators/ultima_validator_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+describe UltimaValidator do
+  describe '#validate' do
+    subject(:validator) { described_class.new }
+
+    context 'when batch contains two requests' do
+      let(:record) { create(:batch, request_count: 2) }
+
+      it 'is valid' do
+        validator.validate(record)
+        expect(validator.validate(record)).to be_nil
```
yoldas: This is not a correct assertion. The validation method return value is void (it should not carry meaning for the caller). The assertion needs to be on the errors collection. You could check if record.errors.count is zero or record.errors.full_messages is an empty array. Alternatively, because the error is added on :base, you could check record.errors[:base] is empty.

test/unit/ultima_sequencing_pipeline_test.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class UltimaSequencingPipelineTest < ActiveSupport::TestCase
```
yoldas: Can this be written as an RSpec test please?

spec/models/ultima_sequencing_pipeline_spec.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe UltimaSequencingPipeline, type: :model do
+  let(:pipeline) do
+    described_class.new(
+      workflow: Workflow.new,
+      request_types: [create(:sequencing_request_type)]
```
yoldas: Can this use the ultima_sequencing factory to match the pipeline?

spec/models/ultima_sequencing_pipeline_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    )
+  end
+
+  describe '#ot_recipe_consistent_for_batch?' do
+    it 'returns true when batch is empty' do
```
yoldas: Why is it valid when the batch is empty?

spec/models/ultima_sequencing_pipeline_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    it 'returns false when some requests are missing ot_recipe' do
+      batch = pipeline.batches.build
+      r1 = create(:sequencing_request, request_metadata_attributes: { ot_recipe: 'Free' })
+      r2 = create(:sequencing_request, request_metadata_attributes: {}) # no ot_recipe
+      batch.requests << [r1, r2]
+
+      expect(pipeline.ot_recipe_consistent_for_batch?(batch)).to be false
+    end
+
+    it 'returns true when no requests have ot_recipe' do
```
yoldas: Why would it be valid without OT Recipe? Please check the statement `return true if ot_recipe_list.empty?` in `app/models/ultima_sequencing_pipeline.rb`

yoldas (approved):
Thanks for the changes. Looks good.

PR #5250, opened 2025-09-19
Y25-319 - [PR][Sequencescape] Use sanger-jsonapi-resources version 0.1.2

yoldas:
Closes #

- Use sanger-jsonapi-resources version 0.1.2

Uses the changes in Y25-319 - [PR] [Develop] Add compatibility for obsolete Rack status symbols https://github.com/sanger/jsonapi-resources/pull/5

### DIFF
--- a/app/controllers/api/v2/bait_library_layouts_controller.rb
+++ b/app/controllers/api/v2/bait_library_layouts_controller.rb
@@ -40,7 +40,7 @@ module Api
       private
 
       def respond_with_errors(title, details, status)
-        status_code = Rack::Utils::SYMBOL_TO_STATUS_CODE[status]
+        status_code = Rack::Utils.status_code(status)
 
         errors = details.map { |detail| { title: title, detail: detail, code: status_code, status: status_code } }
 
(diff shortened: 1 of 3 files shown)


app/controllers/api/v2/bait_library_layouts_controller.rb
```diff
@@ -40,7 +40,7 @@ def preview
       private
 
       def respond_with_errors(title, details, status)
-        status_code = Rack::Utils::SYMBOL_TO_STATUS_CODE[status]
+        status_code = Rack::Utils.status_code(status)
```
stevieing, line 43: is there any reason why this is here? I thought that if you are using JSON API resource this would be handle automatically in the resource. Also it is returning a non-standard error.

yoldas: It was added by https://github.com/sanger/sequencescape/pull/4321 . The reason seems to be converting the status symbol into numeric code value in the method. The `respond_with_errors` method unifies the error responses.

However, direct access to `SYMBOL_TO_STATUS_CODE` has a problem that is the same reason we updated the gem.

We use the  `Rack::Utils.status_code(status)` method now, which contains checks against both `SYMBOL_TO_STATUS_CODE` and `OBSOLETE_SYMBOLS_TO_STATUS_CODES` in Rack 3. The method call is the right interface for the conversion of status symbols into numeric value, and access to the other constants in the Rack::Utils should be considered as private.

yoldas: The reason to have custom code in the controller is that it adds a non-standard standard `preview` method, in addition to CRUD methods. It was done for API V1 compatibility. Because it is a custom action, it needs to handle validation and errors. The preview controller method returns a sparsely populated resource, only the layout attribute is included.

PR #5282, opened 2025-10-08
Y25-447 - [PR] Modify Structure of Tecan Cherry-picking Worklists to Reuse Tips for Buffer Addition Steps

yoldas:
…on steps

Closes #

- Add TecanV3 file generation behaviour to reuse tips for buffer addition steps

### DIFF
--- a/app/models/robot.rb
+++ b/app/models/robot.rb
@@ -55,6 +55,7 @@ class Robot < ApplicationRecord
       'Hamilton' => Robot::Generator::Hamilton,
       'Tecan' => Robot::Generator::Tecan,
       'TecanV2' => Robot::Generator::TecanV2,
+      'TecanV3' => Robot::Generator::TecanV3,
       'Beckman' => Robot::Generator::Beckman
     }.fetch(generation_behaviour_property&.value, Robot::Generator::Tecan)
   end
--- a/app/models/robot/generator/tecan_v3.rb
+++ b/app/models/robot/generator/tecan_v3.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+#
+# Handles picking file generation for Tecan robots with reusing tips for buffer
+# addition steps
+class Robot::Generator::TecanV3 < Robot::Generator::TecanV2
+  include Robot::Generator::Behaviours::TecanDefault
+
+  # Groups buffer addition steps by channel, adds 'Comment' lines between steps,
+  # appends a 'Wash' command at the end of each channel group. This enables
+  # reusing tips for buffer addition steps within each channel.
+  #
+  # @param data_object [Hash] the picking data object
+  # @return [String] the buffer addition steps string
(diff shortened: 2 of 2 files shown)


app/controllers/driver_files_controller.rb
```diff
@@ -9,9 +9,20 @@
...
+  # Generates and sends the robot driver file.
+  #
+  # @note Following parameters are required:
+  #   - batch_id: the id of the {Batch} (path parameter)
+  #   - robot_id: the id of the {Robot} (path parameter)
+  #   - barcode: the barcode of the target plate (query parameter)
+  #   - pick_number: the pick number when multiple source plates are used (query parameter)
+  #   - generator_id: the id of the {RobotProperty} to use (query parameter)
+  #
+  # @return [void]
```
StephenHulme, line 21: Should it return `nil` instead of `void`?

app/frontend/entrypoints/robot_behaviours.js
```diff
@@ -0,0 +1,45 @@
...
+ * 7          dropdown-toggle dropdown-item [Download Tecan 6 File]
+ * 8          dropdown-menu
+ * 9              dropdown-item [TecanV2]
+ * 10             dropdown-item [TecanV3]
+ *
+ */
+import $ from "jquery";
+$(function () {
+  $(".dropdown-submenu > button").on("click", function (e) {
+    const $submenuMenu = $(this).next(".dropdown-menu");
```
StephenHulme: What does the dollar sign in `$submenuMenu` signify?

yoldas: What was I thinking? I thought it is jQuery object and qualified with $ to distinguish from DOM elements but I might be wrong. I will remove it to avoid confusion. [update: removed now]

app/controllers/driver_files_controller.rb
yoldas: I use void to mean that the return value has no significance, mostly in nil cases. I will check the conventions for this semantics.

yoldas: The reason I have added yard doc here is because I touched the code in the method and I needed to explain the generator_id param.

See https://rubydoc.info/gems/yard/file/docs/Tags.md (Search for void and find the Literals section). I think it is a good use for a controller action. 

_void — indicates that the type for this tag is explicitly undefined. Mostly used to specify [@return](https://rubydoc.info/gems/yard/file/docs/Tags.md#return) tags that do not care about their return value. Using a void return tag is recommended over no type, because it makes the documentation more explicit about what the user should expect. YARD will also add a note for the user if they have undefined return types, making things clear that they should not use the return value of such a method._

StephenHulme: Thanks for the docs! I didn't realise that void was a thing with Yard.

PR #5285, opened 2025-10-09
Y25-181 - Upgrade to Rails 7.2

StephenHulme:
Closes #4720 after changes made in #5375

#### Changes proposed in this pull request

- Update Rails to 7.2, including related dependencies
- Migrate configs
- Linting
- Remove broken Plate Templates resource from API v2
- Change database cleaning strategy for cucumber tests
- Update comments

#### Instructions for Reviewers

Please see https://github.com/sanger/sequencescape/pull/5410 and #5453 for a subset of the changes made in this PR.

_[All PRs] - Confirm PR template filled_  
_[Feature Branches] - Review code_  
_[Production Merges to `main`]_  
 &nbsp; &nbsp; \- _Check story numbers included_  
 &nbsp; &nbsp; \- _Check for debug code_  
 &nbsp; &nbsp; \- _Check version_

PR #5312, opened 2025-10-27
[dev] Update cucumber-rails

StephenHulme:
Updates config for cucumber rails by running `bundle exec rails generate cucumber:install -f` as part of #5285 

- Update config as suggested by help documentation
- Refactor loading of config items
- Update DatabaseCleaner to recommended package
- Addresses changes to DatabaseCleaner
- Add `!` assertions for submissions not found
- Add checks for the loaded page URL

### DIFF
--- a/features/studies/8447221_data_release_help_text.feature
+++ b/features/studies/8447221_data_release_help_text.feature
@@ -4,7 +4,7 @@ Feature: Update the data release fields for creating a study
   Background:
     Given I am a "manager" user logged in as "user"
     Given a faculty sponsor called "Jack Sponsor" exists
-    And I am on the studies page
+    And I allow redirects and am on the new study page
 
   Scenario Outline: Add help text opposite delay drop down (4044305)
     When I choose "<release strategy>" from "What is the data release strategy for this study?"


features/support/factory_bot.rb
BenTopping, line 1: is this needed? Do we use factory bot in feature tests?

features/support/step_definitions/web_steps.rb
BenTopping, line 1: Is this file generated mostly by cucumber?

features/support/step_definitions/web_steps.rb
```diff
@@ -19,10 +21,20 @@ def with_scope(locator)
 
 Given /^(?:|I )am on (.+)$/ do |page_name|
   visit path_to(page_name)
+  step "I should be on #{page_name}"
+end
+
+Given /^(?:|I )allow redirects and am on (.+)$/ do |page_name|
+  visit path_to(page_name)
```
BenTopping, line 28: Trying to get my head around this... Not sure how this is different from line 22? Is it if we navigate to a given path and then that path redirects us? Does url params count as a redirection? Shouldn't we be going to the correct route in the first place?

features/support/step_definitions/web_steps.rb
StephenHulme: Yes, with later edits by us

features/support/step_definitions/web_steps.rb
StephenHulme: Before there was no assertion that you were on the correct page. This meant that if there were login or authentication errors you wouldn't know about it until you've been debugging for 10 minutes.

Now it checks the URL by default, alerting us of the final destination - unless overridden with the `allow redirects and` regex.

> Does url params count as a redirection?

Thankfully not

> Shouldn't we be going to the correct route in the first place?

Yes... but I suspect in many cases this was just ignored for the tests. I can add some more specifically defined endpoints which would reduce (but not remove) the need for the workaround.

features/support/factory_bot.rb
StephenHulme: Yes, it is used for [user generatation](https://github.com/sanger/sequencescape/blob/12d4a7b5cecf4c9fe3ad8b341dc5e9c5e67c646b/features/support/step_definitions/authentication_steps.rb#L13-L21).

features/support/step_definitions/web_steps.rb
StephenHulme: Some inconsitencies addressed, some remain. It's certainly better now.
Will merge as-is shortly.

PR #5314, opened 2025-10-27
Y25-520 - Fixing model data not being reset for labwhere receptions

dasunpubudumal:
Closes #

- 


app/models/labwhere_reception.rb
```diff
@@ -44,6 +45,10 @@ def user
   # This maintains compatibility with rails
   # rubocop:todo Metrics/MethodLength
   def save # rubocop:todo Metrics/AbcSize
+    if user_code.blank?
```
BenTopping, line 48: Could you add a comment in the code as to why we do the additional check for user_code.blank?

This should never be blank as the form requires this field, but since this can be bypassed this check provides a nicer user error message than the other errors (from the  valid? check). Is that right?

BenTopping (approved):
Tested and works 👍 
Minor comment about doc for the added check, feel free to ignore.

app/models/labwhere_reception.rb
andrewsparkes: Code resets it further down.
Is it to make them present their swipecard for each transaction?
Or prevent double recording?

PR #5317, opened 2025-10-29
Y25-411 - [PR] As a TL (Tris B) I would like SequenceScape to generate a Sample Sheet to load into the Ultima UG 1000 sequencer so that we can run it

yoldas:
Closes #

- Generates Ultima Sample Sheets as separate CSV files per pool in a single Zip file.


app/controllers/ultima_sample_sheet/sample_sheet_generator.rb
```diff
@@ -0,0 +1,221 @@
...
+    # @return [Hash{Tag => Integer}] mapping of tags to index numbers
+    def tag_index_map
+      @tag_index_map ||= batch_tag_groups.flat_map(&:tags).each_with_index.to_h { |tag, i| [tag, i + 1] }
+    end
+
+    # Returns a mapping of tag groups to 1-based index numbers.
+    # @return [Hash{TagGroup => Integer}] mapping of tag groups to index numbers
+    def tag_group_index_map
+      @tag_group_index_map ||= batch_tag_groups.each_with_index.to_h { |tg, i| [tg, i + 1] }
+    end
```
sabrine33, line 197: This is only valid if the user uses ultima P1 tag for the first request and ultima P2 tag for the second request. 
Better to double check with the user first

yoldas: The indexing happens as Tag Group 1: P1(Tag 1: 1, Tag2: 2, ...) and Tag Group 2: P2(Tag1: 97, Tag2: 98). However, each aliquot tag is checked against the tag itself, so request1 and request2 order in the batch does not matter. I will switch their order in the test to show that. This needs updating the test and building proper tag group and tag indexes similar to the generator instead of assumptions.

Update: I have made changes and added tests to make it better, such as predictable sample ID values rather than a counter. The order of the request in the batch does not matter. It will generate the same zip file now.

config/default_records/tasks/003_ultima_tasks.wip.yml
```diff
@@ -1,4 +1,10 @@
 ---
+LoadingUltima:
+  name: Loading
+  workflow: Ultima
+  sorted: 0
+  lab_activity: true
+  sti_type: SetDescriptorsTask
 Opentrons:
   name: Opentrons
   workflow: Ultima
```
sabrine33, line 4: @yoldas - this is can be removed now - as the loading actions are been added within a different story - the PR is already merged

lib/record_loader/task_loader.rb
```diff
@@ -17,18 +17,23 @@ class TaskLoader < ApplicationRecordLoader
...
+      # fallback.
+      options['name'] = name if options['name'].nil?
       workflow_name = options.delete('workflow')
       workflow = find_workflow!(workflow_name, name)
       return unless workflow
 
       options[:pipeline_workflow_id] = workflow.id
-      Task.create_with(options).find_or_create_by!(name: name, pipeline_workflow_id: workflow.id)
+      Task.create_with(options).find_or_create_by!(name: options['name'], pipeline_workflow_id: workflow.id)
     end
```
sabrine33, line 32: Good point! The same behaviour is observed with the descriptor loader as well. My only concern is that this might introduce risks for other pipeline tasks. Since the loading tasks for the Ultima pipeline have already been added and tested, you can remove all the related files and update your PR accordingly.

config/default_records/tasks/002_element_aviti_tasks.yml
```diff
@@ -1,5 +1,5 @@
 ---
-Loading:
+LoadingElementAviti:
   name: Loading
   workflow: Element Aviti
   sorted: 0
```
sabrine33, line 5: There is no need to update the key here - as the unicity is checked by workflow

sabrine33 (changes requested):

lib/record_loader/task_loader.rb
yoldas: Restored to develop version

config/default_records/tasks/003_ultima_tasks.wip.yml
yoldas: Restored to develop version

config/default_records/tasks/002_element_aviti_tasks.yml
yoldas: Restored to develop version

app/controllers/batches_controller.rb
```diff
@@ -8,7 +8,8 @@ class BatchesController < ApplicationController # rubocop:todo Metrics/ClassLeng
 
   before_action :evil_parameter_hack!
 
-  before_action :login_required, except: %i[released]
+  # generate_sample_sheet checks if the download is allowed without login.
+  before_action :login_required, except: %i[released generate_sample_sheet]
```
BenTopping, line 12: Is this to accommodate NPG accessing the sample sheets? I think they no longer need to access them. But probably best keep them accessible without login for the time being in case they do.

app/controllers/ultima_sample_sheet/sample_sheet_generator.rb
```diff
@@ -0,0 +1,232 @@
...
+
+    # Adds the global section to the CSV.
+    # The request parameter is currently unused.
+    # @param csv [CSV] the CSV object to append rows to
+    # @param _request [UltimaSequencingRequest] the request whose global data is to be added
+    def add_global_section(csv, _request)
+      csv << pad(GLOBAL_TITLE)
+      csv << pad(GLOBAL_HEADERS)
+      data = ['WGS native gDNA', 'UG_116cycles_Baseline_1.8.5.2', 'wgs1']
```
BenTopping, line 104: Are these stored anywhere in LIMS? Assume these will need to be flexible in the future. Should 'wgs1' be "WGS1" (from example sample sheet docs)

app/controllers/ultima_sample_sheet/sample_sheet_generator.rb
```diff
@@ -0,0 +1,232 @@
...
+    SAMPLES_TITLE = ['[Samples]'].freeze
+    SAMPLES_HEADERS = %w[
+      Sample_ID
+      Library_name
+      Index_Barcode_Num
+      Index_Barcode_Sequence
+      Barcode_Plate_Num
+      Barcode_Plate_Well
+      application_type
+      study_id
```
BenTopping, line 34: Are we allowed to add extra / non standard headers  (study_id), have we checked whether ultima accepts them?

BenTopping (commented):
Nice one :)
Left a couple questions, mainly due to my lack of understanding, code looks good. I have checked against the provided sample sheet spec. I have not pulled down and tested locally.

app/controllers/ultima_sample_sheet/sample_sheet_generator.rb
yoldas: I have created a story for that (https://github.com/sanger/sequencescape/issues/5327) because I don't have enough info yet; whether application and analysis recipe decided based on pipeline, user input, or latest versions, etc. However, I could create a table and put sequencing recipe in there but I added to the same story. The workaround for the user in case of Ultima updates would be just changing in the sample sheet for now.

app/controllers/ultima_sample_sheet/sample_sheet_generator.rb
yoldas: Additional columns are allowed by 

spec [UG Sample Sheet Reference Guide](https://cdn.sanity.io/files/l7780ks7/production-2024/4ca366aea0f7ac74a70c39897f870479ddeaf3dc.pdf) Section: Examples of Sample Sheets using Optional Fields 

and existing sample sheet examples Confluence [[Ultima UG 100 Sequencer](https://ssg-confluence.internal.sanger.ac.uk/spaces/PSDPUB/pages/271057969/Ultima+UG+100+Sequencer)] Section Ultima Sample Sheets

app/controllers/ultima_sample_sheet/sample_sheet_generator.rb
BenTopping: Ah nice, thanks for clarifying.

sabrine33 (approved):

PR #5322, opened 2025-10-31
Y25-547 Accessioning - added validation on manifest upload to check case of country of origin

### LINKED ISSUE #5218: Y25-547 - Improve validation on country_of_origin column in sample manifest upload
**User story**

As PSD we have had a number of support tickets (e.g. see https://sanger.freshservice.com/a/tickets/39654) where country_of_origin has been uploaded in a sample manifest in a case other than that which EBI recognises. e.g. NAMIBIA vs expected Namibia
This case sensitivity should be checked on upload against the allowed list of countries and stored in the correct case.

Lena (SSR) had to re-upload the manifests for the above ticket and may be interested
Neil (PSD)
Stephen H (PSD)
Andrew (PSD)

PSD can probably test this one. 
Lena might want to help if asked.

**Acceptance criteria**

- [x] check that the country_of_origin column in the spreadsheet displays the correct case-sensitive country list in the select dropdown, so manual selection always works
- [x] check that if you cut and paste a country name into this column it correctly validates that the name matches the allowed country list on upload (error message to user on upload if not)
- [ ] check that if you use a correct country name but incorrect case, that it catches that case discrepancy when saving the manifest and reports the error to the user

**Dependencies**

None

References
None

**Additional context**

Example in the above ticket was that the manifest had value 'NAMIBIA', whereas the expected value for the EBI is 'Namibia'. Same letters but incorrect case. Seems fussy but that's how it works for the API call to the EBI, it rejects on incorrect case.

See `Insdc::Country` in the SS repo for country list.
We do not hardcode the country list, it is fetched and stored.

There is a range defined for the `country_of_origin` column in the SS sample manifest `columns` yaml file.
Column `country_of_origin` is set to be mandatory.
It is a select box with a dropdown in the excel sheet.
Validation in the excel column only checks a value has been chosen, not that it matches one of the countries in the range.
There doesn't appear to be any upload validation other than that the column is mandatory and must have a value. Suggest adding validation on upload to:
- validate the value matches a country in the `Insdc::Country` list, if it does, report back the value from `Insdc::Country` with correct case to the user.

Searching for a specific country name:
`Insdc::Country.find_by(name:)`

**Deployment notes**

Nothing special.

andrewsparkes:
Closes #5218 

Adds validation for country of origin case on manifest upload

### DIFF
--- a/app/sample_manifest_excel/sample_manifest_excel/upload/row.rb
+++ b/app/sample_manifest_excel/sample_manifest_excel/upload/row.rb
@@ -205,7 +205,9 @@ module SampleManifestExcel
       end
 
       def needs_country_of_origin_error?(value)
-        # Using exclude (case sensitive) to check if country matches to Insdc EBI list of countries
+        # Skip the error creation if the value is blank. The column is mandatory by default and the mandatory column
+        # validation will catch this scenario. And if the column is optional, then blank is acceptable.
+        # Using exclude here (which is case sensitive) to check if country matches to Insdc EBI list of countries
         value.present? && Insdc::Country.options.exclude?(value)
       end
 
--- a/spec/sample_manifest_excel/upload/row_spec.rb
+++ b/spec/sample_manifest_excel/upload/row_spec.rb
@@ -523,6 +523,46 @@ RSpec.describe SampleManifestExcel::Upload::Row, :sample_manifest, :sample_manif
         ''
       ]
     end
+    let(:data_empty_value) do
+      [
+        plate.human_barcode,
+        'A1',
+        sample_manifest.sample_manifest_assets.first.sanger_sample_id,
+        'SAMPLE_1',
+        'Destroy after 2 years',
+        'Cohort 1',
+        '',
+        '',
+        '',
+        '',
+        '',
+        '',
+        '',
+        'Nov-16',
+        'Nov-16',
+        '',
+        '',
+        '',
+        '',
+        '',
+        '',
+        '',
+        '',
+        '',
+        '',
+        '1',
+        'Human',
+        'Sample description',
+        '',
+        '',
+        '',
+        '',
+        'Sample phenotype',
+        '',
+        '',
+        ''
+      ]
+    end
 
     before do
       create(:insdc_country, name: 'United Kingdom')
@@ -573,6 +613,12 @@ RSpec.describe SampleManifestExcel::Upload::Row, :sample_manifest, :sample_manif
 
       it 'is valid' do
         expect(described_class.new(number: 1, data: data_valid_no_country_of_origin, columns: columns)).to be_valid
+      end
+    end
+
+    context 'when country of origin column is empty' do
+      it 'is valid' do
+        expect(described_class.new(number: 1, data: data_empty_value, columns: columns)).to be_valid
       end
     end
 


StephenHulme (approved):

app/sample_manifest_excel/sample_manifest_excel/upload/row.rb
```diff
@@ -181,6 +183,41 @@ def validate_sample
...
+      end
+
+      # Format user friendly error message for country of origin value
+      def add_country_of_origin_error(value)
+        suggestion = Insdc::Country.options.find { |c| c.casecmp(value).zero? }
+        message = "#{row_title} Country of Origin value '#{value}' does not match any allowed value " \
+                  '(NB. case-sensitive).'
+        message += " Did you mean '#{suggestion}'?" if suggestion
+        errors.add(:base, message)
+      end
```
KatyTaylor, line 221: Story says "that it corrects it when saving the manifest" - did you want to autocorrect it or just return an error?

spec/data/sample_manifest_excel/manifest_types.yml
KatyTaylor, line 1: Where did the requirement for this new manifest type come from?

KatyTaylor: Sorry just realised it's just the spec version of the file.

spec/sample_manifest_excel/upload/row_spec.rb
KatyTaylor, line 1: Have you also checked where country of origin column is present but not required, and the value is not filled in?

app/sample_manifest_excel/sample_manifest_excel/upload/row.rb
andrewsparkes: On balance I decided better to have the SSR correct it in their manifest file rather than attempt to autocorrect it in the upload. They would have to ditch their copy and redownload a fresh one to pick up the autocorrection, and how would they know they needed to do that.

spec/data/sample_manifest_excel/manifest_types.yml
andrewsparkes: yeah just needed a version without the column for testing, and this was the easiest way.

spec/sample_manifest_excel/upload/row_spec.rb
andrewsparkes: The column is defined as mandatory I think (under conditional formatting `empty_mandatory_cell`?) so is that scenario possible?
On upload I see 'You must set a value for country_of_origin at row: 11'
I don't think it reaches my validation in that scenario.

KatyTaylor: It would be possible for someone to create a new non-mandatory version of the column (in columns.yml), to use in a new manifest type (in manifest_types.yml). In which case, the code shouldn't blow up when the column is not populated. I think it would be OK, as it would (probably) return false for `value.present?` and therefore false for `needs_country_of_origin_error?`.

andrewsparkes: Added a test for that empty scenario. It does indeed handle it ok with that value.present? check. Added a comment there as well to be clear.

KatyTaylor: New test looks good, thanks

PR #5330, opened 2025-11-03
Y25-652 - adds comp phix study to phix stock list

### LINKED ISSUE #5328: Y25-652 - Add new PhiX study to config yml
SciOps need to generate phix tubes with a new study. These will be used for Aviti sequencing.
Study id [Comp PhiX](https://sequencescape.psd.sanger.ac.uk/studies/8250/information)

This new study would need to be added to ss/config/phi_x.yml. 

This is quite urgent as the present Aviti run analysis are being delayed as they require manual intervention.

BenTopping:
Closes #5328 

- Adds Comp PhiX to stock PhiX list
- Adds warning note for future study additions.

### DIFF
--- a/app/views/phi_xes/_stock.html.erb
+++ b/app/views/phi_xes/_stock.html.erb
@@ -15,7 +15,7 @@
     <%= form.label :study_id, 'Study' %>
     <div class="field">
       <%= form.select :study_id, @study_names  %>
-      <div class="form-help">The study with which the PhiX will be associated. <b>N.B. Please contact support to add new PhiX studies.</b></div>
+      <div class="form-help">The study with which the PhiX will be associated. <strong>N.B. Please contact support to add new PhiX studies.</strong></div>
     </div>
   </div>
   <div class='phi-x-form'>


app/views/phi_xes/_stock.html.erb
```diff
@@ -15,7 +15,7 @@
     <%= form.label :study_id, 'Study' %>
     <div class="field">
       <%= form.select :study_id, @study_names  %>
-      <div class="form-help">The study with which the PhiX will be associated.</div>
+      <div class="form-help">The study with which the PhiX will be associated. <b>N.B. Please contact support to add new PhiX studies.</b></div>
```
StephenHulme: [nitpick] suggest use `<strong>` over `<b>`: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/b#usage_notes

StephenHulme: Is it clear what contact support means in this context? (Fresh service ticket/slack/something else?)

BenTopping: Ah cool, learnt something :)

BenTopping: Yes I think its fine as a temporary measure, just helps them understand studies need to be manually added.

StephenHulme: That sounds good to me then, thanks 👌

PR #5353, opened 2025-11-14
Y25-286 - Implement HTTP clients to query taxonomy

StephenHulme:

- Repairs the broken taxonomy lookup functionality using ENA (see the [API guide](https://ena-docs.readthedocs.io/en/latest/retrieval/programmatic-access/taxon-based-search.html))
- Implements the groundwork for an HTTP client library, to be extended with accessioning refactoring
- Starts the migration away from the unsupport `rest-client` to `faraday`
- Pull taxonomy out of the sample controller and into it's own controller
- Adjusts wording, spacing, and highlighting of taxonomy lookup button
- Fixes invalid proxy config

Note: the common name is intentially set to the scientific name - as has been the case in forever...

Useful links:
- [ENA taxonomy overview](https://ena-docs.readthedocs.io/en/latest/faq/taxonomy.html)
- [ENA submitting taxons](https://ena-docs.readthedocs.io/en/latest/faq/taxonomy_requests.html)

How to use the new Taxa Controller:

- http://localhost:3000/taxa?term=human
  ```json
  {
  "taxId": "9606",
  "scientificName": "Homo sapiens",
  "commonName": "human"
  }
  ```

- http://localhost:3000/taxa/9606
  ```json
  {
  "taxId": "9606",
  "scientificName": "Homo sapiens",
  "commonName": "human"
  }
  ```
- http://localhost:3000/taxa/9606
  ```
  400: Missing required parameter: term
  ```

#### Screenshots

**Before:**

Screenshot 2025-11-14 at 12 17 45
<hr>

Screenshot 2025-11-14 at 12 18 02
<hr>
<hr>

**After:**

Screenshot 2025-11-14 at 12 15 38
<hr>

**Success:**

Screenshot 2025-11-14 at 12 40 06
<hr>

Screenshot 2025-11-14 at 12 15 01
<hr>

**Failure:**
Screenshot 2025-11-17 at 11 32 34
<hr>

Screenshot 2025-11-17 at 11 27 50
<hr>

Screenshot 2025-11-17 at 11 22 40
<hr>

### Deployment notes

- Requires `configatron.ena_taxon_lookup_url ` to be defined in deployment project and merged in first: see https://github.com/sanger/deployment/pull/757

### DIFF
--- a/app/controllers/taxa_controller.rb
+++ b/app/controllers/taxa_controller.rb
@@ -16,7 +16,7 @@ class TaxaController < ApplicationController
   rescue ActionController::ParameterMissing
     render plain: 'Missing required parameter: term', status: :bad_request
   rescue Faraday::Error => e
-    Rails.logger.error("Client error in Taxa Controller: #{e.class} - #{e.message}")
+    log_faraday_error(e)
     head :bad_gateway
   end
 
@@ -29,7 +29,7 @@ class TaxaController < ApplicationController
 
     render json: taxon
   rescue Faraday::Error => e
-    Rails.logger.error("Client error in Taxa Controller: #{e.class} - #{e.message}")
+    log_faraday_error(e)
     head :bad_gateway
   end
 
@@ -37,5 +37,9 @@ class TaxaController < ApplicationController
 
   def client
     @client ||= HTTPClients::ENATaxaClient.new
+  end
+
+  def log_faraday_error(error)
+    Rails.logger.error("Client error in Taxa Controller: #{error.class} - #{error.message}")
   end
 end


neilsycamore:
what does http://localhost:3000/taxa?term=hominidae return?
This is a Genus and should return id 9604 && false (not suitable for submission)

app/controllers/taxa_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+class TaxaController < ApplicationController
+  skip_before_action :login_required # there's no sensitive information here
+
+  def index
+    return render plain: 'Missing required parameter: term', status: :bad_request if params[:term].blank?
+
+    # Lookup by term and render results
+    term = params[:term]
+    taxon = client.taxon_from_text(term)
```
yoldas, line 11: question: should we call strip on the term to remove whitespace? The client makes "any-name/term" after url-encoding it, which would add %20 bits. The result would be an empty response.

note: if such a url is tested with browser, it would remove the trailing whitespace and hide this issue. Try it with a preceding space.

suggestion: the term might be enforced using params.require or an explicit check for presence.

app/controllers/taxa_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    return head :not_found if taxon.nil?
+
+    render json: taxon
+  rescue Faraday::Error
+    head :bad_gateway
```
yoldas, line 20: question: Good semantics but it swallows the error details. Should we log this?

app/controllers/taxa_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  def show
+    # Lookup by id and render the taxon
+    id = params[:id] # the given taxon ID
+    taxon = client.taxon_from_id(id)
+
+    return head :not_found if taxon['taxId'].nil?
+
+    render json: taxon
+  rescue Faraday::Error
+    head :bad_gateway
```
yoldas, line 33: question: Swallows the error details. Should we log this?

app/frontend/entrypoints/organism_validation.js
```diff
@@ -1,120 +1,80 @@
...
-$(function () {
-  // Any validate_organism control needs to be setup to update the 'common name' and the 'taxon ID' fields.
-  // It is assumed that the 'common name' and 'taxon ID' fields are paired in order: in other words, the
-  // first 'common name' field goes with the first 'taxon ID' field.  If this is not true then you should
-  // write your own method and call validateSingleOrganism yourself.
-  $(".validate_organism").each(function (pos, control) {
-    $(control).on("click", function (event) {
+document.addEventListener("DOMContentLoaded", function () {
+  // Attach validation to all .validate_organism controls
+  document.querySelectorAll(".validate_organism").forEach(function (button) {
```
yoldas, line 3: note: css class is on the button. OK.

app/frontend/entrypoints/organism_validation.js
```diff
@@ -1,120 +1,80 @@
...
+    button.addEventListener("click", function (event) {
       event.preventDefault();
-      var common_names = $('input[data-organism="common_name"]');
-      var taxon_ids = $('input[data-organism="taxon_id"]');
-
-      common_names.each(function (index, common_name) {
-        validateOrganism(common_name, taxon_ids[index]);
+      const commonNames = document.querySelectorAll('input[data-organism="common_name"]');
```
yoldas, line 6: note: attribute selector is right. OK.

app/controllers/taxa_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class TaxaController < ApplicationController
```
andrewsparkes, line 3: Nice that you've split this out

app/models/ena_accession_service.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-class EnaAccessionService < AccessionService
+class ENAAccessionService < AccessionService
```
yoldas, line 2: note: rails convention for acronyms. Loader can handle file name to class name mapping. See inflections too. OK.

config/locales/metadata/en.yml
```diff
@@ -217,8 +217,8 @@ en:
           edit_info: "ENA tag, Array Express"
 
         sample_taxon_id:
-          label: TAXON ID
-          help: NCBI Taxonomy Identifier. This is appropriate for individual organisms and some environmental samples.
+          label: Taxon ID
+          help: ENA Taxonomy Identifier. This is appropriate for individual organisms and some environmental samples.
```
yoldas, line 221: note: lookup url is changed from NCBI to EBI (ENA) in config. OK.

app/frontend/entrypoints/organism_validation.js
```diff
@@ -1,120 +1,80 @@
...
+    }
 
-        unfound: function () {
-          highlight_field("bad", common_name_field);
-          highlight_field("bad", taxon_id_field);
-        },
-      });
-    },
+    // Set the name and taxon id fields to the returned values
+    commonNameField.value = taxon.scientificName; // yes the field is called common name, yes this is confusing, no I have no idea why we do it this way...
```
andrewsparkes, line 63: Neil may know why. Historical reasons I suspect.

lib/http_clients/base_client.rb
```diff
@@ -0,0 +1,28 @@
...
+module HTTPClients
+  # A generic client that communicates to external services over HTTP.
+  #
+  # Subclass this class to create clients for specific services with custom behaviours.
+  class BaseClient
+    private
+
+    def default_headers
+      # convert HTTPClients::TestHTTPClient to Test HTTP Client
+      client_name = self.class.name.demodulize.underscore.humanize.titleize
```
yoldas, line 14: praise: nice user-agent.

lib/http_clients/base_client.rb
```diff
@@ -0,0 +1,28 @@
...
+      {
+        'User-Agent' => "Sequencescape #{client_name}" # Required by some APIs
+      }
+    end
+
+    def proxy
+      return nil if configatron.disable_web_proxy == true
+      return configatron.proxy if configatron.fetch(:proxy).present?
+      return ENV['http_proxy'] if ENV['http_proxy'].present?
```
yoldas: question: ENV['http_proxy'] looks more specific than configatron.proxy . Is the order right ?

lib/http_clients/ena_taxa_client.rb
```diff
@@ -0,0 +1,64 @@
...
+      response = conn.get("any-name/#{suggestion}")
+      first_taxon = response.body.first
+      return unless first_taxon
+
+      # extract relevant fields and return as a hash
+      {
+        'taxId' => first_taxon['taxId'],
+        'scientificName' => first_taxon['scientificName'],
+        'commonName' => first_taxon['commonName'],
+        'submittable' => first_taxon['submittable']
```
yoldas, line 44: note: boolean is coded as string in the response, e.g ` "submittable" : "true",`

lib/http_clients/ena_taxa_client.rb
```diff
@@ -0,0 +1,64 @@
...
+    # @return [Hash, nil] A hash with 'taxId', 'scientificName', and 'commonName' if found, or nil if not found.
+    def taxon_from_text(suggestion)
+      suggestion = ERB::Util.url_encode(suggestion)
+      response = conn.get("any-name/#{suggestion}")
+      first_taxon = response.body.first
+      return unless first_taxon
+
+      # extract relevant fields and return as a hash
+      {
+        'taxId' => first_taxon['taxId'],
```
yoldas, line 41: note: integer is coded as string in the response, e.g. `"taxId" : "9606",`

Gemfile
```diff
@@ -13,18 +15,18 @@ group :default do
   gem 'mutex_m'
   gem 'syslog'
 
+  # Connections to external HTTP services
+  # See lib/http_clients for examples of usage
+  gem 'faraday'
+  gem 'rest-client' # Deprecated, but still used in some places, replace with Faraday where possible
```
yoldas, line 21: praise: thanks for these comments on dependencies.

app/frontend/entrypoints/organism_validation.js
```diff
@@ -1,120 +1,80 @@
...
-      highlight_field("bad", common_name_field);
-      highlight_field("bad", taxon_id_field);
-    },
+    // Highlight the fields, indicating if they are submittable
+    const submittable = taxon.submittable == "true" ? "success" : "failure";
+    highlightField(commonNameField, submittable);
+    highlightField(taxonIdField, submittable);
+
+    // If the taxon is not submittable, prevent form submission by setting custom validity
+    if (taxon.submittable != "true") {
```
andrewsparkes, line 72: I like these validations. But I think you'll need to discuss with SSRs if you're making it more restrictive for them, e.g. if EBI inaccessible or something

yoldas (changes requested):
Looks great. I have a few questions.

config/config.rb
```diff
@@ -38,7 +38,7 @@
 configatron.r_and_d_division = 'RandD'
 configatron.site_url = 'localhost:3000'
 configatron.sta_plate_purpose_name = 'STA'
-configatron.taxon_lookup_url = 'http://eutils.ncbi.nlm.nih.gov/entrez/eutils/'
+configatron.ena_taxon_lookup_url = 'https://www.ebi.ac.uk/ena/taxonomy/rest/'
```
andrewsparkes, line 41: are these dev URL changes in this file production EBI URLs? are there test EBI URLS for taxonomy?
just getting info I suppose though...

andrewsparkes: Also, what's the difference between the NCBI endpoint and the EBI version? NCBI is a different organisation right? Are you sure it's ok to switch this? (I have no idea)
https://www.ncbi.nlm.nih.gov/

andrewsparkes: I mean, we are submitting to EBI, so their taxon info should be fine. Just wondering why it was NCBI before really.

app/controllers/taxa_controller.rb
StephenHulme: Excellent suggestions - I should have thought of that. WIll implement, thanks!

app/controllers/taxa_controller.rb
StephenHulme: Excellent point, thanks will log too.

lib/http_clients/base_client.rb
StephenHulme: This is [the existing order](https://github.com/sanger/sequencescape/blob/c4972107268a6d052aab76e53962c242648d5530/lib/accession/accession/request.rb#L51-L62), but I agreed, it doesn't make sense for the config to override the envvar.

I'll reorder it to put ENV['http_proxy'] first, then the configatron options as they are.

lib/http_clients/ena_taxa_client.rb
StephenHulme: While somewhat unintuive, this is consistent from ENA all the way to the javascript check preventing repeated casting.

app/frontend/entrypoints/organism_validation.js
StephenHulme: The validations are meant to prevent known incorrect IDs rather than enforce specific input.
If the taxon lookup button is not pressed the validation is not applied. Likewise, if the input is accepted, there's nothing stopping the user from changing the input before submitting the form.

I hope is that these workarounds prevent any hard-blocks from holding up any processes.

When the actual accessioning is performed later, there is a pre-submission check that the taxon-id is populated (but not what with), and I believe that ENA will reject submission within an uknown taxon.

config/config.rb
StephenHulme: > are these dev URL changes in this file production EBI URLs? are there test EBI URLS for taxonomy?
just getting info I suppose though...

They are production URLs, but since it is only a GET request there is no sensitivity here.

> Also, what's the difference between the NCBI endpoint and the EBI version? NCBI is a different organisation right? Are you sure it's ok to switch this? (I have no idea)
https://www.ncbi.nlm.nih.gov/

Having looked into this, [it looks like they use the same underlying NCBI database](https://ena-docs.readthedocs.io/en/latest/faq/taxonomy.html). In addition, the ENA endpoint includes the useful `submittable` parameter for accessioning. Submittability is outlined in the [Taxonomy FAQ](https://ena-docs.readthedocs.io/en/latest/faq/taxonomy_requests.html).

> I mean, we are submitting to EBI, so their taxon info should be fine. Just wondering why it was NCBI before really.

I'm guessing historical reasons - it hasn't been updated in over 15 years and the ENA endpoint is definitely more recent than that.

app/controllers/taxa_controller.rb
StephenHulme: Addressed in 7ecdcff5b1977eb5d15994da3ac9399093d801e5

app/controllers/taxa_controller.rb
StephenHulme: Fixed in c345652edd8cffe53b737b6be0507d36017f2e98

app/controllers/taxa_controller.rb
StephenHulme: Fixed in c345652edd8cffe53b737b6be0507d36017f2e98

lib/http_clients/base_client.rb
StephenHulme: Fixed in bd7f5675dcb674d00db626b1320b564be0a47f16

Issue #4879, opened 2025-04-28
Y25-286 - Accessioning - Refactor duplicated code


andrewsparkes:
**User story**
As PSD we would like the various ways to accession samples to use the same code in the background, rather than two distinct sets of code.

**Who are the primary contacts for this story**
PSD 

**Who is the nominated tester for UAT**
Neil
Internal team testing.

**Acceptance criteria**

To be considered successful the solution must allow:
- [ ] Identify all the accessioning trigger points, and all the accessioning code that they use
- [ ] All trigger points that Accession samples should use the same classes and configuration
- [ ] There should be a single configurable flag that cleanly turns off all accessioning for an environment. (See existing flag `configratron.accession_samples` which works for one code path currently).
- [ ] Unneeded obsolete duplicated code should be removed
- [ ] After completion of the refactoring, [this Confluence how-to article](https://ssg-confluence.internal.sanger.ac.uk/spaces/PSDPUB/pages/290915138/Turn+accessioning+in+Sequencescape+on+or+off) showing how to enable or disable accessioning should be updated.

**Dependencies**
None.

This story is blocked by the following dependencies:
None.

References
This story has a non-blocking relationship with:
Research story here: https://github.com/sanger/sequencescape/issues/4887

**Additional context**
See the comments section on the research story above for the two code sets that perform accessioning.

The code in lib/accession appears more recent (still 8 years old, by Steve I) and better organised than the code starting in app/controllers/sample_controller (14 years old, by Matt D).

Tasks (may need to break this up into a number of stories):
- first identify all the code used in the 3 different accessioning trigger points (list files and methods)
- decide which set of code to keep
   - consider that sample manifest uploads and study 'accession all samples' are asynchronous and use a delayed job, whereas accessioning a single sample via 'Generate accession number' is synchronous
   - note the duplicated configuration of ebi XML tags, duplicated to_XML methods that generate the XML payload for the EBI posts, and all the supporting classes 
   - consider error handling, accessioning must not fail silently if something is wrong on any of the routes, and should report back to the PSD exceptions email system (and slack channel)
   - if needed, write new code and tests to make sure the functionality provided by the current two sets of code is covered by the chosen set of accessioning code (e.g. synchronous single sample accessioning)
- modify the accessioning trigger points to all call the chosen set of code, and test that all 3 accessioning routes work
- remove all the obsolete code, feature and unit tests, and check all 3 accessioning routes still work
  - included in this is
[...]

PR #5362, opened 2025-11-18
Y25-286 - Add accessioning feedback

StephenHulme:
Closes #4879 

Follows on from the accessioning refactoring in #5368 

- Add docs on what is being done and why (see [Accession Status Tracking](https://github.com/sanger/sequencescape/blob/y25-286-add-accessioning-status/app/models/accession/README.md))
- Create `Accession::SampleStatus` model
  - Accession statuses hold individual responses from ENA
- Refactor asynchronus accessioning validation
  - Creates a new high-level `Accession::Error` (to replace deprecated `AccessionService::AccessionServiceError`)
  - Creates a subclassed `Accession::ExternalValidationError` whose error messages are explicitly user-readable
- Add and update statuses on asynchronus (delayed job) accessioning
- Add Accession Statuses tab to Study Information view
- Add Accession Status panel to Sample view
- Improve accessioning test mocking
- Add lots of tests
- Move lib accessioning specs to the appropiate folder to mirror the code directory structure
- Update Rubocop todo

#### Deployment notes

- Includes database schema migration for the creation of two new tables

#### Screenshots

New tab in studies information:

Screenshot 2025-11-19 at 14 18 09

New panel in sample (only shown if accessioning has been attempted, but there is no accession number):
Screenshot 2025-11-25 at 17 41 45

### DIFF
--- a/app/views/studies/information/_accession_statuses.html.erb
+++ b/app/views/studies/information/_accession_statuses.html.erb
@@ -22,42 +22,40 @@
     </tr>
   </thead>
   <tbody>
-    <% @study.sample_progress(@page_elements) do |statistics| %>
-      <% @page_elements.each do |sample| %>
-        <% accession_status = sample.current_accession_status %>
-        <tr>
-          <td><%= link_to(sample.name, sample) %></td>
-          <td>
-              <% if sample.accession_number? %>
-                <% catalogue_url = configatron.accession.view_url + sample.ebi_accession_number %>
-                <%= link_to sample.ebi_accession_number, url_for(catalogue_url), target: :blank %>
-              <% else %>
-                <%= accession_status&.message %>
-              <% end %>
-          </td>
-          <td><%= accession_status&.status&.capitalize %></td>
-          <td><%= accession_status&.updated_at&.in_time_zone&.strftime('%Y-%m-%d %H:%M:%S %Z') %></td>
-          <td>
-            <% if can?(:accession, sample) && configatron.accession_samples %>
-              <%# Add link to perform sample accessioning %>
-              <% if sample.current_accession_status %>
-                <%# If status exists, offer retry %>
-                <%= link_to(
-                  '<i class="fa fa-rotate-right"></i> Retry'.html_safe,
-                  url_for(controller: '/samples', action: 'accession', id: sample.id),
-                  class: "btn btn-sm btn-primary"
-                ) %>
-              <% elsif !sample.accession_number? %>
-                <%# If it has not been accessioned before, offer submit %>
-                <%= link_to(
-                  '<i class="fa fa-upload"></i> Submit'.html_safe,
-                  url_for(controller: '/samples', action: 'accession', id: sample.id),
-                  class: "btn btn-sm btn-outline-primary"
-                ) %>
-              <% end %>
+    <% @page_elements.each do |sample| %>
+      <% accession_status = sample.current_accession_status %>
+      <tr>
+        <td><%= link_to(sample.name, sample) %></td>
+        <td>
+            <% if sample.accession_number? %>
+              <% catalogue_url = configatron.accession.view_url + sample.ebi_accession_number %>
+              <%= link_to sample.ebi_accession_number, url_for(catalogue_url), target: :blank %>
+            <% else %>
+              <%= accession_status&.message %>
             <% end %>
-          </td>
-      <% end %>
+        </td>
+        <td><%= accession_status&.status&.capitalize %></td>
+        <td><%= accession_status&.updated_at&.in_time_zone&.strftime('%Y-%m-%d %H:%M:%S %Z') %></td>
+        <td>
+          <% if can?(:accession, sample) && configatron.accession_samples %>
+            <%# Add link to perform sample accessioning %>
+            <% if sample.current_accession_status %>
+              <%# If status exists, offer retry %>
+              <%= link_to(
+                '<i class="fa fa-rotate-right"></i> Retry'.html_safe,
+                url_for(controller: '/samples', action: 'accession', id: sample.id),
+                class: "btn btn-sm btn-primary"
+              ) %>
+            <% elsif !sample.accession_number? %>
+              <%# If it has not been accessioned before, offer submit %>
+              <%= link_to(
+                '<i class="fa fa-upload"></i> Submit'.html_safe,
+                url_for(controller: '/samples', action: 'accession', id: sample.id),
+                class: "btn btn-sm btn-outline-primary"
+              ) %>
+            <% end %>
+          <% end %>
+        </td>
     <% end %>
   </tbody>
 </table>
--- a/app/views/studies/information/_sample_progress.html.erb
+++ b/app/views/studies/information/_sample_progress.html.erb
@@ -11,30 +11,32 @@
     </tr>
   </thead>
   <tbody>
-    <% @page_elements.each do |sample| %>
-      <% sample_statistics = statistics[sample.id] %>
-      <tr>
-        <td><%= link_to sample.name, sample_path(sample)    %></td>
-        <td><%= sample.updated_at.to_formatted_s(:sortable) %></td>
-        <% @request_types.each do |request_type| %>
-          <% request_type_statistics = sample_statistics[request_type.id] %>
-          <% if request_type_statistics.total > 1 and request_type_statistics.progress > 0 %>
-            <td><%= progress_bar(request_type_statistics.progress) %></td>
-          <% else %>
-            <td>
-              <% if request_type_statistics.passed > 0 %>
-                <span style="display:none">passed</span><%= icon('far', 'check-circle', title: 'Passed') %>
-              <% elsif request_type_statistics.started > 0 %>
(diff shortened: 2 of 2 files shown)


app/models/accession/README.md
```diff
@@ -0,0 +1,37 @@
+# Accession Status Tracking
```
andrewsparkes, line 1: This readme is great, thanks

app/models/accession/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+In an attempt to reduce the number of execution paths for the accessioning of samples specifically we are going to accession both individual samples and all groups of samples (manifests and studies) using the delayed jobs workers. While this has a number of advantages (explained below), there is currently no method to return feedback to the submitting user. To address this, we are going add database-backed records of the state of each item that is being accessioned.
+
+## Accessioning demand
+
+Looking at accessioning logs for the delayed jobs path over the last month shows that 4 workers process sample accessions in parallel at a combined rate of ~1000 samples/second. While the number of accessions submitted at a time can vary, it is not uncommon to receive requests to process ~5000 samples at a time.
+
+## Multiple APIs
+
+Sequencescape currently uses the original v1 accessioning API provided by EBI (https://ena-docs.readthedocs.io/en/latest/submit/general-guide/webin-v1.html). This has been replaced with a v2 API (https://ena-docs.readthedocs.io/en/latest/submit/general-guide/programmatic.html) which has better support for large submissions.
```
andrewsparkes, line 16: Did you decide to switch to API v2 in the end then?

db/schema.rb
```diff
@@ -10,7 +10,16 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2025_08_18_131035) do
+ActiveRecord::Schema[7.1].define(version: 2025_10_30_152601) do
+  create_table "accession_statuses", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
```
andrewsparkes: if sample specific maybe `accession_sample_statuses`?

andrewsparkes (approved):
Looks great, can't see any obvious problems.

app/models/accession/README.md
StephenHulme: No, it's still API v1. I wanted to minimise the number of simultaneous changes to reduce the likelyhood of functionality regressions.

db/schema.rb
StephenHulme: The initial plan was that this could be used for all accession types - but since that isn't the current implementation, and there are no immediate plans to change this, I agree that the more specific naming would be helpful.
This would also change the class name `Accession::Status` to `Accession::SampleStatus` - which again seems very reasonable 👍

StephenHulme: Renamed in 5b2e6c85d0f8f9aafcb9ab24c6ba07101f962bec

app/jobs/sample_accessioning_job.rb
```diff
@@ -32,9 +40,84 @@ def queue_name
...
+      # Finds the most recent accession status by sample id, and marks it as in progress
+      accession_status = Accession::SampleStatus.find_latest_or_create_for_sample(accessionable.sample)
+      accession_status.mark_in_progress
+    end
+
+    # Finds the most recent accession status and removes it
+    def succeed_accession_status
+      # Wrap in a transaction to prevent race conditions
+      Accession::SampleStatus.transaction do
+        accession_status = Accession::SampleStatus.find_latest_or_create_for_sample(accessionable.sample)
```
BenTopping: Should we be creating them if we are going to destroy them straight away? Should this error if it can't find one - same goes for failing and aborting, should there always be a samplestatus present by that point?

BenTopping (commented):
Nice work! Will give it a proper look over tomorrow.

app/controllers/samples_controller.rb
```diff
@@ -158,22 +158,36 @@ def accession # rubocop:todo Metrics/AbcSize, Metrics/MethodLength
 
     @sample.validate_ena_required_fields!
     @sample.accession_service.submit_sample_for_user(@sample, current_user)
+    # TODO: remove this line and replace with reference to @sample.accession
```
BenTopping, line 161: Is this still todo?

app/models/accession/README.md
```diff
@@ -0,0 +1,37 @@
...
+While this is currently intended to be used only with sample accessions, there is no reason why the same pattern could not be used for other accession types in the future.
+
+This is intended to work as follows:
+
+1. A user would submit a sample manifest, accession all samples within a study, or accession individual samples.
+2. Each of these actions would create an individual Accession::SampleStatus record for each item to be accessioned.
+3. If the accessioning process succeeds, the accession number would be updated on the object itself.
+4. If the accessioning process fails, the status record would be updated to 'failed' with the given error message.
+5. All successful statuses would be removed after notification, while failed statuses would be retained for user reference.
+6. Statuses can be deleted after a rolling period of time, say 30 days.
```
BenTopping: I am not sure we need to delete them via a cron or manually - they should automatically get deleted if the accession succeeds? I cant imagine a case where we wont re-accession a sample if its required and has so far failed.

app/models/sample.rb
```diff
@@ -321,6 +321,8 @@ def release
   has_many :requests, through: :assets
   has_many :submissions, through: :requests
 
+  has_many :accession_sample_statuses, class_name: 'Accession::SampleStatus'
```
BenTopping, line 324: Shouldn't a sample only have one instance of an accession status - what use case would there be for more than 1?

lib/accession.rb
```diff
@@ -61,10 +61,16 @@ def <=>(other)
   # Generic high-level accessioning error
   # Usage: raise Accession::Error, "Accessioning failed: #{reason}"
   class Error < StandardError; end
+  class ExternalValidationError < Error; end
 
   CENTER_NAME = 'SC'
   XML_NAMESPACE = { 'xmlns:xsi' => 'http://www.w3.org/2001/XMLSchema-instance' }.freeze
 
+  # See app/models/accession.rb
```
BenTopping, line 69: The file in the comment is empty (minus comments) - is there a future plan here?

app/jobs/sample_accessioning_job.rb
StephenHulme: Good question. It's an attempt to simplify the code, but I think it should probably be relooked at.

I've adjusted the implementation to create a new status for each accessioning attempt - this better matches what users see compared to developers.

app/controllers/samples_controller.rb
StephenHulme: Yes, this is to follow in a later PR

app/models/accession/README.md
StephenHulme: Yes, they should get deleted on successful accession. Agreed, all going well all samples should be accessioned, but I think there might be edge-cases where some samples don't work out and are abandoned.

I haven't implemented a rolling deletion strategy (I don't think there is a need to at this stage), but it could be done, particularly if there is a process change later.

app/models/sample.rb
StephenHulme: Initially I was trying a status to every delayed-job, but have now refactored it so every attempt to accession produces a new status. This will help in the case where there are multiple validation or communication issues affecting a single sample.
On success, all status for a sample will now be deleted.

lib/accession.rb
StephenHulme: Not a future plan, but there is a namespace conflict over `Accession` and where it is defined. The empty file needs to exist, but I don't want to refactor the entire module.

I've updated the comment in 9f9931c6e3650d90e63cc43b72f48c57adf64017 to be clearer.

app/models/accession/README.md
BenTopping: Yeah that makes sense - we can revisit in a couple months if needed.

lib/accession.rb
BenTopping: A joy of rails...

Can we move this whole file to the model file instead of being in lib or does that break other things?

StephenHulme: We probably could... but I would like to keep it in the lib folder, it's most related to the other classes there, rather than being model-related specifically.

app/views/studies/information/_accession_statuses.html.erb
```diff
@@ -0,0 +1,62 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <% @page_elements.each do |sample| %>
+      <% accession_status = sample.current_accession_status %>
+      <tr>
+        <td><%= link_to(sample.name, sample) %></td>
+        <td>
+            <% if sample.accession_number? %>
+              <% catalogue_url = configatron.accession.view_url + sample.ebi_accession_number %>
```
yoldas, line 31: note: I have checked deployment project; view_url ends with a slash.

BenTopping (approved):

app/controllers/samples_controller.rb
StephenHulme: Here's the PR where it is replaced with a feature flag allowing easy switching between the two implementations: https://github.com/sanger/sequencescape/pull/5403

PR #5375, opened 2025-11-24
Y25-181 - Precursor to Rails 7.2

StephenHulme:
Changes in preparation for #5285 

- Replace deprecated proxy object with basic object
- Encapsulate pipeline-related configs into initializers
- Migrate rails configs to yml configs
- Remove plate-template that is no longer used in Gatekeeper
- Fix accessioning tests in response to bugfix `Fixes an issue where validates_associated :on option wasn’t respected
when validating associated records.`
- Fix breaking tests

### DIFF
--- a/config/initializers/animations.rb
+++ b/config/initializers/animations.rb
@@ -1,3 +1,4 @@
 # frozen_string_literal: true
 
-Rails.application.config.disable_animations = ENV.fetch('DISABLE_ANIMATIONS', false).present?
+# Animations are enabled by default, but can be disabled in test environments
+Rails.application.config.disable_animations = false


spec/features/samples/sample_accession_errors_spec.rb
```diff
@@ -18,10 +18,6 @@
       it 'shows informative errors when required fields are missing for accessioning' do
```
BenTopping: Does this test need removal / redoing, its not checking for 'informative errors' any more.

config/initializers/animations.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+Rails.application.config.disable_animations = ENV.fetch('DISABLE_ANIMATIONS', false).present?
```
BenTopping: I think we are only disabling animations for tests - see config/environments/cucumber. In which case they are explicitly setting it, not using an env variable. So think we can remove this entirely?

spec/features/samples/sample_accession_errors_spec.rb
StephenHulme: Good catch, it has value in terms of the page content, but the errors are now handled elsewhere. I'll reword the spec message.

config/initializers/animations.rb
StephenHulme: It is only being used for tests (where it is being used directly), but a value still needs to be set for the non-test case to prevent a `undefined method 'disable_animations' for an instance of Rails::Application::Configuration` error.
I can remove the env.fetch as I don't think it adds anything we can't already do in the test environments, but I think an initializer is still a good place to set this config?

BenTopping: Ah interesting... yes happy to leave it if we cant remove it.

StephenHulme: Updated in 64a6ce0a7afd510e698bed681da85b2f75f755ad

spec/features/samples/sample_accession_errors_spec.rb
StephenHulme: Addressed in 48a505ed8e9e293131d70dea5deab8f374309a84

BenTopping (approved):

PR #5389, opened 2025-11-27
Y25-598 - scRNA - add tube size options

### LINKED ISSUE #2597: GPL-nnn Refactor suppression of well creation
**User story**
As a developer I would like to refactor the way in which plates prevent automatic generation of wells to improve code readability and grep-ability.

**Who the primary contacts are for this story**
James Glover

**Acceptance criteria**
1 - Generation of plates without wells via a more explicit approach (such as PlatePurpose#create_plate_without_wells!)
2 - Log (warn) when the old mechanism is triggered with a stack-trace
3 - Update as many uses of the old mechanism to the new mechanism
4 - Search for logged warning periodically and update corresponding methods
5 - Strip out the supporting code once we're sure its unused.

**Additional context**
Currently the plate_purpose#create! method allows the passing of an optional truthy parameter as the first argument. All other arguments are named arguments/hash. When set, this parameter suppresses the call to wells#construct! either to allow the call to be made asynchronously, or to prevent it completely. In most cases this argument is set to a symbol, such as :do_not_create_wells but theoretically any truthy value will be sufficient.

Known symbols:
:without_wells, :do_not_create_wells, :nope (This one is actually testing the behaviour, and is well commented despite the silly symbol name. Still, sorry about that one)

I don't think there are any other symbols in use, but we may be passing through variables in places.

andrewsparkes:
Closes #2597 

- change to viability threshold value


config/default_records/request_types/014_limber_scrna_core_cdna_prep_request_types.yml
```diff
@@ -15,7 +15,8 @@ limber_scrna_core_cdna_prep_gem_x_5p:
   acceptable_purposes:
     - LRC Bank Seq
     - LRC Bank Spare
-    - LRC Bank Input
+    - LRC Bank Input 1ml
+    - LRC Bank Input 0.8ml
```
StephenHulme: Suggestion: Capacities in purposes are somewhat against our usual pattern I think? Could these be replaced with labware part numbers or Sm/Md/Lg?

spec/uat_actions/generate_tubes_spec.rb
```diff
@@ -19,10 +20,29 @@
...
+      end
+
+      it 'sets the correct sample metadata values' do
+        uat_action.perform
+        tube = Tube.find_by_barcode(report['tube_0'])
+        sample_metadata = tube.aliquots.first.sample.sample_metadata
+        expect(sample_metadata.supplier_name).to eq 'sample_NT2P_0'
+        expect(sample_metadata.collected_by).to eq UatActions::StaticRecords.collection_site
+        expect(sample_metadata.donor_id).to eq 'sample_NT2P_0_donor'
+        expect(sample_metadata.sample_common_name).to eq 'human'
```
StephenHulme: nitpick: suggest the `have_attributes` matcher to condense these four expectations into a single expectation that lists the keys and values, eg: 
```rb
      expect(sample_metadata).to have_attributes(
        supplier_name: 'sample_NT2P_0',
        # transfer_class_name: 'Transfer::BetweenPlates',
        # transfers: nil # not given
      )
```

spec/uat_actions/generate_tubes_spec.rb
```diff
@@ -38,12 +58,27 @@
...
       end
+
+      it 'sets the correct sample metadata values for each tube' do
+        uat_action.perform
+        uat_action.report.each_with_index do |(_key, barcode), i|
+          sample_metadata = Tube.find_by_barcode(barcode).aliquots.first.sample.sample_metadata
+          expect(sample_metadata.supplier_name).to eq "sample_#{barcode}_#{i}"
+          expect(sample_metadata.collected_by).to eq UatActions::StaticRecords.collection_site
+          expect(sample_metadata.donor_id).to eq "sample_#{barcode}_#{i}_donor"
+          expect(sample_metadata.sample_common_name).to eq 'human'
```
StephenHulme: nitpick: suggest the have_attributes matcher as above

spec/uat_actions/generate_tubes_spec.rb
```diff
@@ -55,17 +90,36 @@
...
+      end
+
+      it 'sets the correct sample metadata values' do
+        uat_action.perform
+        tube = Tube.find_by_barcode(expected_report['tube_0'])
+        sample_metadata = tube.aliquots.first.sample.sample_metadata
+        expect(sample_metadata.supplier_name).to eq 'sample_NT6T_0'
+        expect(sample_metadata.collected_by).to eq UatActions::StaticRecords.collection_site
+        expect(sample_metadata.donor_id).to eq 'sample_NT6T_0_donor'
+        expect(sample_metadata.sample_common_name).to eq 'human'
```
StephenHulme: nitpick: suggest the have_attributes matcher as above

StephenHulme (approved):
Looks good,I've added some minor comments

config/default_records/request_types/014_limber_scrna_core_cdna_prep_request_types.yml
andrewsparkes: I agree but this is what Abby requested they be called.
I think they are the same tube model loaded with a different (specific) volume for use in one of the two specific prep methods, so part numbers or vague size descriptions doesn't work. Perhaps she has a name for each path that could be used.

spec/uat_actions/generate_tubes_spec.rb
andrewsparkes: done

spec/uat_actions/generate_tubes_spec.rb
andrewsparkes: done

PR #5397, opened 2025-12-04
Y25-505 bug udi library update

### LINKED ISSUE #5159: Y25-505 - [BUG] uploading UDI library plate manifest fails to update downstream aliquots
Library tags can be updated/corrected by uploading a library manifest after it has been amended. This should update all relevant downstream aliquots not just those in the stock wells.

This functionality has ceased (reported by SSRs 9:38 4/8/25) and requires users to raise service ticket.

Example of steps to test are in FS ticket

**FS Ticket Number**
[SR-36674](https://sanger.freshservice.com/a/tickets/36674)

Main contacts: rcc, jm36

**To Reproduce**

1. Create manifest for library plates
2. Use Template: Library Plate - dual index tag plate
3. Update (see example manifest) and upload
4. Create submission (LB multiplexed is simplest)
5. Step thru Limber creating LB lib pool then LB lib pool norm **_tubes_**
6. Update manifest changing the well id (i.e C1 => F6)
7. Upload manifest (overwriting data button)
8. Stock well tag updates
9. Pool tag does not
10. *Expected behaviour**
pool tags update

Example manifest

[2405stdy_manifest_27643_120825.xlsx](https://github.com/user-attachments/files/21736434/2405stdy_manifest_27643_120825.xlsx)

andrewsparkes:
Closes #5159 

Update downstream aliquots

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -400,7 +400,7 @@ class Sample < ApplicationRecord # rubocop:todo Metrics/ClassLength
   validates_each(:ena_study, on: %i[accession ENA EGA]) do |record, _attr, value|
     record.errors.add(:base, 'Sample has no study') if value.blank?
   end
-  validates_associated :ena_study, allow_blank: true, on: :accession, message: lambda { |record|
+  validates_associated :ena_study, allow_blank: true, on: :accession, message: lambda { |record, _attr|
     "is invalid: #{record.ena_study.errors.full_messages.join(', ')}"
   }
 


app/sequencescape_excel/sequencescape_excel/specialised_field/dual_index_tag_well.rb
```diff
@@ -71,6 +75,38 @@ def tag
...
+      # Aliquots must have a matching sample_id, library_id, tag_id and tag2_id to the given stock_aliquot.
+      def identify_all_matching_aliquots(stock_aliquot)
+        attributes = {
+          sample_id: stock_aliquot.sample_id,
+          library_id: stock_aliquot.library_id,
+          tag_id: stock_aliquot.tag_id,
+          tag2_id: stock_aliquot.tag2_id
+        }
+
+        Aliquot.where(attributes).ids
```
yoldas, line 96: We are not explicitly checking downstream parent-child relationships. Can this cause an update of more aliquots than it should?

andrewsparkes: I wondered this too. But Library id should be unique on it's own, plus here we also match on sample id and combination of tags.
And we are dealing with a manifest re-upload here for library stocks, any use of that library for an aliquot should be updated. I think it's ok unless you can think of a scenario?

I'm more concerned about how / if anything downstream of the final norm tube e.g. flowcells etc. should be updated too?

andrewsparkes: Also, this is how non dual index manifests identify the aliquots they update. I've matched the code used to find the aliquots, found in in the tag substitution code.

app/sequencescape_excel/sequencescape_excel/specialised_field/dual_index_tag_well.rb
```diff
@@ -71,6 +75,38 @@ def tag
       def tag2
         Tag.find_by(tag_group_id: tag2_group_id, map_id: well_index)
       end
+
+      def fetch_single_aliquot(asset)
+        raise StandardError, 'Expecting asset to have a single aliquot' unless asset.aliquots.one?
+
+        asset.aliquots.first
+      end
```
KatyTaylor, line 84: Useful additional error handling, thanks 🙏 . 

Can we not use the outdated 'asset' terminology in the user-facing message though? I assume it's a receptacle. What does the error message look like to the end user? Does it attach the error to the relevant well (for plates) so they know which one is the problem?

app/sequencescape_excel/sequencescape_excel/specialised_field/dual_index_tag_well.rb
```diff
@@ -71,6 +75,38 @@ def tag
...
+
+      # Determine if the tags need to be updated
+      def tags_need_update?(stock_aliquot, new_i7_tag, new_i5_tag)
+        (stock_aliquot.tag != new_i7_tag) || (stock_aliquot.tag2 != new_i5_tag)
+      end
```
KatyTaylor, line 84: If we believe ChatGPT, 

> In Rails, calling update on an ActiveRecord object with attributes that are identical to the existing ones results in no database write, and no update callbacks, but validations are still run.

So I think this method is unnecessary.

app/sequencescape_excel/sequencescape_excel/specialised_field/dual_index_tag_well.rb
StephenHulme: > I've matched the code used to find the aliquots, found in in the tag substitution code

This seems like a good idea - at least it should be consistent... I think it should propogate all the way down, including flowcells, so as not to introduce more inconsistencies.

> And we are dealing with a manifest re-upload here for library stocks

What is the original  _thing_ that is being uploaded in the manifest? Just samples, or samples in combination with something else?

KatyTaylor: Was wondering what library_id was. It's set in `SampleManifest::CoreBehaviour::LibraryAssets` to be the receptacle id. 

So as long as that library_id is propagated corrected down the line of parent-child aliquots I think this is fine. 
The code that does that is in the `TransferRequest` model - `aliquots_for_transfer` and related methods.

Could be worth spot checking a few real lineages of labware from prod data (e.g. one library plate, one library tube + derivatives) to check that the library_id is populated and consistent on all downstream aliquots.

> I'm more concerned about how / if anything downstream of the final norm tube e.g. flowcells etc. should be updated too?

The final receptacles / aliquots will be in Lanes I think. Same data model really (a Lane is just a type of Receptacle) so I'd hope this would extend to that. 

Definitely worth testing if and when the MLWH tables get updated though. I think the well failing functionality behaves differently if the data is already in the MLWH or not.

app/sequencescape_excel/sequencescape_excel/specialised_field/dual_index_tag_well.rb
andrewsparkes: Hmm, I think I copied that error handling from the tag substitution code, so maybe it isn't even visible here at manifest upload, it may need to be done differently. I better test that. And I can change the asset wording. Good spot thx.

app/sequencescape_excel/sequencescape_excel/specialised_field/dual_index_tag_well.rb
andrewsparkes: @StephenHulme 
> What is the original _thing_ that is being uploaded in the manifest? Just samples, or samples in combination with something else?

This is a library manifest, so these are technically Libraries that are being uploaded rather than cruder samples. But I think as far as LIMS is concerned they are represented as samples with tags and other extra metadata. They are samples that have already been through DNA purification, amplification and library preparation steps off-LIMS and are ready for Sequencing.

StephenHulme: That makes sense, thanks for the explanation.

app/sequencescape_excel/sequencescape_excel/specialised_field/dual_index_tag_well.rb
andrewsparkes: Removed, thx.

app/sequencescape_excel/sequencescape_excel/specialised_field/dual_index_tag_well.rb
andrewsparkes: @KatyTaylor 
Wasn't working properly throwing the exception. You get the error page in the GUI (and I added the well to the displayed message), but then the manifest gets stuck in 'processing' state because you've broken out of the code abnormally.

So switched the check to a validation for the single aliquot, which gives you a clean validation error message in the GUI and allows you to re-try later.

I doubt the scenario would every arise in any case, how would you get multiple aliquots in the same stock library well.

andrewsparkes: Screenshot 2025-12-11 at 16 34 16

KatyTaylor: Brilliant thanks for improving that.

PR #5398, opened 2025-12-04
Y25-286 - Rework eventful records and (re-)add accessioning events

StephenHulme:
Closes #4879 

- Update and streamline usage of events recording in Sequencescape on accession
  - Rework and document EventfulRecord module
  - Add new AccessionEvent
  - Replace previous eventing with AccessionEvent
  - Fix sample history rendering errors
  - Add `event_user` (the Rails `current_user` which the event will be recorded against) to tests as required

**Screenshot:**
Screenshot 2025-12-08 at 12 21 02
_The old accession update events and the new update event (in reality only one new update event would occur per sample)._

Screenshot 2025-12-10 at 12 46 24

_With updated wording._

### DIFF
--- a/features/data_release/12400603_update_accession.feature
+++ b/features/data_release/12400603_update_accession.feature
@@ -56,7 +56,7 @@ Feature: object with an accession should be modifiable
     When I update an accession number for sample "sample"
 
     When I am on the event history page for sample "sample"
-    Then I should see "Created sample accession number"
+    Then I should see "Assigned sample accession number"
       And I should see "me"
 
   Scenario: A sample with already an accession number should update itself using its accession number


lib/eventful_record.rb
```diff
@@ -1,31 +1,38 @@
...
-        )
-        # rubocop:enable Layout/LineLength
+      # Dynamically defines a method for constructing, naming, and adding events of a given class.
+      def self.event_constructor(model_event_name, event_class, event_class_method)
+        define_method(model_event_name) do |*args|
+          event = event_class.public_send(event_class_method, proxy_association.owner, *args)
+          self << event if event.eventful.blank?
+          event
+        end
       end
```
StephenHulme, line 24: This is entirely Co-Pilot - I'd appreciate if someone could please confirm that this is a sane and reasonable conversion from meta-coding-as-a-string, to a dynamically generated method?

app/models/accessionable/dac.rb
```diff
@@ -37,9 +37,9 @@ def xml # rubocop:todo Metrics/MethodLength
 
   def update_accession_number!(user, accession_number)
     @accession_number = accession_number
-    add_updated_event(user, "DAC for Study #{@study.id}", @study) if @accession_number
     @study.study_metadata.ega_dac_accession_number = accession_number
     @study.save!
+    @study.events.created_accession_number!('DAC', user)
```
KatyTaylor: Question: in the old code (here and in a couple more examples below), there is an `if @accession_number` check - which is not being replicated in the new code. Why's that?

andrewsparkes: I'm also unsure what's going on here. I assume if an accession number already existed for the sample, it was using the presence or absence of the number to gate certain code and events (i.e. if no number yet then accession the sample and fire creation event, whereas if there was a number we want to update it and fire an update event).

But under what circumstances is an accession number on a sample ever updated to something new? Do they re-accession samples to fetch a new EBI identifier? Seems unlikely to me...

app/models/event/accessioning_event.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+class Event::AccessioningEvent < Event
+  def self.created_accession_number!(eventable, accession_type, user)
```
andrewsparkes: might be worth adding some comments here to explain when an update event is made as opposed to a created event.

app/views/samples/history.html.erb
```diff
@@ -12,13 +12,17 @@
...
+      <thead>
+        <tr>
+          <th>Message</th>
+          <th>Content</th>
+          <th>Created at</th>
+          <th>Created by</th>
+        </tr>
+      </thead>
+      <tbody>
+        <%= render partial: "requests/event", collection: @sample.events.sort{ |a,b| a.id <=> b.id  } %>
```
andrewsparkes, line 24: this is for a single sample yes? not a whole study? cos that might be millions of entries

lib/eventful_record.rb
KatyTaylor: It's still hard to read I think! But looks correct as far as I can tell 👀 and probably out of scope to get rid of the metaprogramming completely!

lib/accession/sample.rb
```diff
@@ -76,9 +76,12 @@ def ebi_alias
       sample.uuid
     end
 
-    def update_accession_number(accession_number)
+    # Updates the accesion number, saving the sample and adding an event to the events table
```
andrewsparkes: typo in comment, 'accession'

app/models/event/accessioning_event.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def self.updated_accession_number!(eventable, accession_type, user)
+    create!(
+      eventful: eventable,
+      message: "Updated #{accession_type} accession data",
+      family: 'accessioning',
+      of_interest_to: 'administrators',
+      created_by: user&.login
+    )
+  end
```
KatyTaylor: What has been updated here exactly? The accession number itself? Or metadata associated with the accessioned entity? Either the method name or the 'message' could be a little misleading.

lib/eventful_record.rb
StephenHulme: Agreed! At least it now has syntax highlighting, and any automated scanning (rubocop, etc.) should pick up any major issues. I think my main concern is I don't want to introduce anything worse than what is already there like memory leaks.

app/models/accessionable/dac.rb
StephenHulme: @KatyTaylor In these `update_accession_number!` functions there is already a check for `accession_number` in the [calling method](https://github.com/sanger/sequencescape/blob/861284b77801e6f0d97e86c01d1b0885ffe0e4ec/app/models/accession_service/base_service.rb#L74-L75).

There are other places elsewhere in the code with `accession_number` checks that I've removed. These checks have now been replaced with errors being raised should a failure occur, rather than frequently checking that we're still on the success path.

StephenHulme: @andrewsparkes This is one of the slightly confusing things with accessioning - there is a primary `::Sample` model which we use everywhere and a dedicated `Accessionable::Sample` for handling accessions. This is referred to as an "accessionable" as it could also be a study, or policy, etc.

What this code is doing is taking the accession number returned by EBI to the Accessionable (which is the model that requested it in the first place) and assigning it to the primary sample or study, etc.

Since there are two ways of accessioning, there is also an `Accession::Sample` (etc) that is handled in a similar way for asynchronous accessioning (manifests, etc.). These will be renamed (and duplicates removed) before the end of the epic.

app/models/event/accessioning_event.rb
StephenHulme: Agreed. And I don't fully know the answer to this (yet). The accessioning API has an ADD command and a MODIFY command. These correlate to creating an initial record of and then updating this record of sample metadata. Unfortunately, our implementation of their API seems to handle these two commands in a (mostly) identical way. This will be cleaned up in a future PR where the updated and created is used in the appropiate manner (and with better wording).

Would it help to make them a single, more generic event for this PR?

StephenHulme: cc @andrewsparkes

app/views/samples/history.html.erb
StephenHulme: Correct, it it is used to display an audit trail for changes made to individual samples individually.

lib/accession/sample.rb
StephenHulme: Sooo many typos, both mine and historical 😆 - I'm quite surprised more haven't crept through! Will fix this one though.

StephenHulme: Thanks for picking up on this - fixed in a1a6f3fe7fa5eace3666707c72da69a250c1155c

PR #5403, opened 2025-12-09
Y25-286 - Combine sample accessioning branches

StephenHulme:
Closes #4879 and follows on from #5398

- Add feature flag to combine sample accessioning branches (`y25_286_accession_individual_samples_with_sample_accessioning_job`)
  - This will put all sample accessioning on the (currently) asynchronous path
  - Samples can be accessioned asynchronously, or synchronously if `SampleAccessioningJob.new(...).perform` is used instead of queueing
  - The old synchronous sample accessioning code will be removed as part of the feature flag removal
  - Study accessioning will stay on the existing synchronous path
- Pull (some) accessioning code out of sample and study models
- Refactor and add tests, moving early accessioning tests from minitest to rspec

#### Deployment notes

Adds the feature flag `y25_286_accession_individual_samples_with_sample_accessioning_job` which can be enabled once #5414 has been merged in.

### DIFF
--- a/spec/factories/study_factories.rb
+++ b/spec/factories/study_factories.rb
@@ -58,7 +58,6 @@ FactoryBot.define do
 
   factory(:open_study, parent: :study) do
     transient { accession_number { nil } }
-    transient { data_release_timing { 'standard' } }
 
     sequence(:name) { |n| "Study#{n}: Open" }
     state { 'active' }
@@ -68,7 +67,6 @@ FactoryBot.define do
     after(:create) do |study, evaluator|
       study.study_metadata.update!(
         data_release_strategy: 'open',
-        data_release_timing: evaluator.data_release_timing,
         study_ebi_accession_number: evaluator.accession_number
       )
     end


app/models/accession_service.rb
```diff
@@ -35,4 +35,45 @@ module AccessionService
...
+    when 'open'
+      AccessionService::ENAService.new
+    when 'managed'
+      AccessionService::EGAService.new
+    else
+      AccessionService::NoService.new(study)
+    end
+  end
+
+  # Return the highest priority accession service that this sample should use.
```
BenTopping, line 55: What does highest priority mean in this case? Why is EGA higher priority than ENA?


Edit: found https://github.com/sanger/sequencescape/pull/5403/files#diff-3a128f74b32158671dd87db5b3e9544d4fa29eef349b3dbab09c9a89edbcb5cfR148. Worth adding this here I think too.

app/models/metadata.rb
```diff
@@ -53,6 +53,7 @@ def #{association_name}
       end
 
       def tags
+        accession_service = AccessionService.select_for_sample(self)
```
BenTopping: Is this tags function specific to samples? Just conscious this is a Metadata metaprogrammed class and is used elsewhere. Maybe this method shouldnt be here in the first place? One for another day maybe..

lib/accession.rb
```diff
@@ -86,4 +86,69 @@ def self.configure
...
+
+  # --- Methods called by external controllers ---
+
+  # Wrapper for sample accessioning with error handling and job management.
+  # Allows accessioning to be triggered from anywhere in the application.
+  # Encapsulates logic for validation, synchronous or asynchronous job execution,
+  # and supports private helper methods for internal workflow.
+  class SampleAccessioning
+    def perform(sample, event_user, perform_now)
+      # TODO: move this validation logic to a validation method somewhere else
```
BenTopping, line 98: Is this still todo?

app/models/accession_service.rb
StephenHulme: There is also a `priority` attribute on the Service classes

lib/accession.rb
StephenHulme: I think so, yes. We're approaching the point where I re-diagram how accessioning works, this time with particular emphasis on validation, so that checks can be moved to the most reasonable place.

app/models/metadata.rb
StephenHulme: Good catch, thank you I think I missed the reuse case... I'll have to have a closer look at this.

StephenHulme: Fortunately it's only used by samples. As such, I've taken it out of Metadata and moved it to Sample in 28ba3c173fd040bd0d50448412c4c293958c861c.
It's slated for removal soon anyway 😌

app/models/accessionable/sample.rb
```diff
@@ -45,7 +45,8 @@ def initialize(sample) # rubocop:todo Metrics/CyclomaticComplexity
 
       # TODO: maybe unify this with the previous loop
```
andrewsparkes, line 46: todo still needed?

app/models/concerns/include_tag.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# Concern to add the include_tag method to samples for accessioning tags
+# (not to be confused with aliquot tags)
```
andrewsparkes, line 4: I am confused, what is an accessioning tag?

app/models/accessionable/sample.rb
StephenHulme: No, this is from old code and is to be removed in #5418 or #5420

app/models/concerns/include_tag.rb
StephenHulme: It's a field/label/key for accessioning, one of `gender`, `phenotype`, `collection_date`, etc. Searching the codebase for `tag` also returns things related to aliquots, oglio, i7, etc. which are not same. The context is _generally_ sufficiently different to not be an issue, but it's an easy trap for the naive (myself included!) to fall into!

lib/accession.rb
StephenHulme: See #5416 and #5419

PR #5410, opened 2025-12-12
tests: attempt to persist seed data and truncate test data

BenTopping:
Attempts at handling cucumber test data consistently. The general approach is to seed the database once on test run start and then rely on tests not mutating it. There have been several cases we have fixed where tests were mutating it and there may be more hidden but we have got to the point it is passing.

We have changed the db cleaner strategy to only clean new data created from the current test (ignore seed data). We have had to go with this approach because transactions were not working correctly after a couple of different attempts. This approach also means we can't parallelise CI using knapsack as the shared db instance causes issues when using truncation as one test may truncate another tests data.

A lot of these tests can be removed when we deprecate V1 API which will reduce the run time. We should also focus on moving off cucumber.

### Proposed changes
- Adds seeded_deletion database cleaner to maintain seed data whilst removing other tests data.
- Updates cucumber database cleaner strategies to use new seeded_deletion.


app/models/sample.rb
```diff
@@ -382,9 +382,7 @@ def rename_to!(new_name)
   validates_each(:ena_study, on: %i[accession ENA EGA]) do |record, _attr, value|
     record.errors.add(:base, 'Sample has no study') if value.blank?
   end
-  validates_associated :ena_study, allow_blank: true, on: :accession, message: lambda { |record|
```
BenTopping: This wasn't valid syntax so was failing the CI tests. Message can only take one argument (a string)

StephenHulme: Resolved in latest merge

features/api/pulldown/submissions.feature
```diff
@@ -50,42 +50,6 @@ Feature: Creating submissions for pulldown
       }
       """
 
-  @create @error
```
BenTopping, line 53: The submission template is generated here https://github.com/sanger/sequencescape/blob/develop/features/support/step_definitions/submission_steps.rb#L21. 
It creates a pulldown_isc request type which has no validators for requiring read_length validation as expected in the test. So I am unsure why it would expect to fail.

features/api/uk10k/submissions.feature
```diff
@@ -72,33 +72,6 @@ And I have a "full" authorised user with the key "cucumber"
       }
       """
 
-  @create @error @asset
```
BenTopping, line 75: Again this tests is v1 api submissions so should be safe to remove. I suspect it is something to do with transactions here https://github.com/sanger/sequencescape/blob/develop/app/api/endpoints/submissions.rb#L17. I don't think this is getting hit though so unsure.

features/api/pipelines.feature
```diff
@@ -107,15 +107,15 @@ Feature: Access pipelines through the API
 
   @create @batch @authorised @error
   Scenario Outline: Attempting to create a batch with invalid request details
-    Given the maximum batch size for the pipeline "Cluster formation PE" is 2
+    Given the UUID for the pipeline "HiSeq 2500 PE (spiked in controls)" is "00000000-1111-2222-3333-444444444445"
```
BenTopping, line 110: Instead of mutating the seed data (Clust formation PE) to have a max_size of 2, we are using an different template with an existing max size of 2 so we dont need to edit seed data.

features/administration/26913815_add_bait_library_editor_to_admin_view.feature
BenTopping, line 1: Changes here prevent editing seed data

StephenHulme (approved):
Great work - thanks so much for tackling this!

features/api/sample_manifests/updating_through_manifest.feature
BenTopping, line 1: This is a restricted / deprecated action and has been for almost 10 years. With removal of API V1 imminent I just removed this to save time.

PR #5430, opened 2025-12-23
Enabling publishing of library asserts to `stock_resource` table.

dasunpubudumal:
Closes #

- Change `stock?` method to return `true` for `LibraryAsset`.
- Update `messenger.rb` to set the `labware_type` in the MLWH message. 


app/models/sample_manifest/library_plate_behaviour.rb
```diff
@@ -5,5 +5,9 @@ module SampleManifest::LibraryPlateBehaviour
   # it sets library_id on aliquots in wells and doesn't generate stock assets.
   class Core < SampleManifest::PlateBehaviour::Base
     include SampleManifest::CoreBehaviour::LibraryAssets
+
+    def stocks?
```
BenTopping, line 9: Could you update this line https://github.com/sanger/sequencescape/blob/develop/app/models/sample_manifest/core_behaviour.rb#L76 instead of overwriting each instance separately?

app/models/library_plate_well.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
```
BenTopping: What does this change do? I cant see the class referenced anywhere.

Is this because of the way we do double messages, e.g [Well, '1'] and then send the full message later on?

app/models/messenger.rb
```diff
@@ -15,7 +15,20 @@ def routing_key
   end
 
   def as_json(_options = {})
-    { root => render_class.to_hash(target), 'lims' => configatron.amqp.lims_id! }
+    message = render_class.to_hash(target)
+    Rails.logger.info("Publishing message: #{message}")
+    { root => process_receptacles(message), 'lims' => configatron.amqp.lims_id! }
```
BenTopping, line 20: Is this messenger just used for receptacles? If its a shared class Im not sure this logic should be here.

app/models/messenger.rb
```diff
@@ -15,7 +15,20 @@ def routing_key
...
+  end
+
+  # Processes the message for receptacle targets, setting labware type if applicable.
+  # @param message [Hash] The message to process.
+  # @return [Hash] The processed message.
+  def process_receptacles(message)
+    return message unless receptacle_target?
+
+    asset_type = fetch_asset_type
+    message['labware_type'] = 'library_plate_well' if library_plate?(asset_type)
```
BenTopping, line 30: is library_plate_well the only affected labware type here, are the other library types okay?

app/models/messenger.rb
dasunpubudumal: No it is not. It is a shared class. The problem is that the `register_stock!` method pushes messages to the broadcast queue with the `Messenger` model (e.g., `[Messenger, 12123]`). It doesn't push messages like `[Well, 12123]` which makes the whole thing a bit weird and difficult. It felt weird having to put this logic here but I couldn't figure out any other way.

app/models/library_plate_well.rb
dasunpubudumal: No it's not needed. I forgot to remove it.

app/models/messenger.rb
dasunpubudumal: I will try to instantiate the wells with the `subject_type = "libarary_plate_well"` upon manifest upload.

PR #5441, opened 2026-01-07
[ux] Show the number of samples in a study

StephenHulme:
Very helpful for finding a study for debugging - as so many have 0 or 2 samples...

- Add a Samples column to the studies page.

Screenshot:
Screenshot 2026-01-07 at 12 54 12

### DIFF
--- a/app/views/studies/_study_list.html.erb
+++ b/app/views/studies/_study_list.html.erb
@@ -19,7 +19,7 @@
         <small><%= link_to 'Details', properties_study_path(study) %></small>
       </td>
       <td><span style="display:none"><%= study.created_at %></span><%= study.created_at.to_formatted_s(:sortable) %></td>
-      <td><%= study.study_samples.size %></td>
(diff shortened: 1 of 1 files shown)


StephenHulme:
> Nice one thanks both! It doesn't instil a lot of confidence in the Rails ORM, if using the study_samples relationship directly makes such a big difference!

The study_samples table is a many-to-many joiner table, so it's more of a Sequencescape implementation issue than a Rails ORM issue. Fortunately we are only looking for a count of samples and not sample-level data itself.

KatyTaylor:
> The study_samples table is a many-to-many joiner table, so it's more of a Sequencescape implementation issue than a Rails ORM issue. Fortunately we are only looking for a count of samples and not sample-level data itself.

I see, the difference is that `.study_samples.size` allows it to just query the join table, whereas `.samples.size` forces it to query both the join table and the samples table.

PR #5444, opened 2026-01-08
Y25-703 - Add ability to modify asynchronous accessions

### LINKED ISSUE #5414: Y25-703 - Add MODIFY action to samples accessioning (present on synchronous path, missing on asynchronous path)
**Description**
Add MODIFY action to samples accessioning (present on synchronous path, missing on asynchronous path).

The asynchronous (batch) path used for manifest uploads and accession-all-samples does not have the ability to update accessioned metadata, only add new data. Attempting to accession (or update) a sample that has already been accessioned results in the error "This sample has already been accessioned". That's great, but unhelpful, as the user knows that the sample has been accessioned, but wants to updated the metadata instead.

The fix is (realtively) straight-forward, tell the accessioning code to use the MODIFY action instead of the ADD action (a line of text in the generated XML file) should the accession number already exist. This should be well documented and tested to prevent continued future confusion.

**Blockers**
#5403 

**Additional context**
Already in progress in `y25-286-confirm-accession-add-modify-behaviour`

StephenHulme:
Closes #5414

This will allow manifest uploads to update accessioned sample data.

- Use MODIFY action when changing samples
- Remove check preventing modifying an already accessioned sample
- Add stack trace to error logs
- Add Save and Accession text to submit button when editing a sample
- Add updated accessioned metadata event to sample history on successful modification

#### Post deployment

Enable feature flag `y25_286_accession_individual_samples_with_sample_accessioning_job`

### DIFF
--- a/app/views/studies/_study_list.html.erb
+++ b/app/views/studies/_study_list.html.erb
@@ -6,6 +6,7 @@
       <th>ID</th>
       <th>Study name</th>
       <th>Date created</th>
+      <th>Samples</th>
       <th>Study Owner</th>
     </tr>
   </thead>
@@ -18,6 +19,7 @@
         <small><%= link_to 'Details', properties_study_path(study) %></small>
       </td>
       <td><span style="display:none"><%= study.created_at %></span><%= study.created_at.to_formatted_s(:sortable) %></td>
+      <td><%= StudySample.where(study_id: study.id).size %></td>
       <td><%= display_owners(study) %></td>
     </tr>
   <% end -%>


lib/accession/sample.rb
```diff
@@ -80,7 +80,7 @@ def ebi_alias
     # for viewing under sample history.
     def update_accession_number(accession_number, event_user)
       sample.sample_metadata.sample_ebi_accession_number = accession_number
-      sample.save
+      sample.sample_metadata.save # prevent an infinite loop due to after_save callbacks on sample.save
```
yoldas, line 83: There is a similar call in app/models/accessionable/sample.rb‎ . However, it does `save!` there. Here it ignores the return value of just `save` .

yoldas (approved):

Issue #5415, opened 2025-12-15
Y25-704 - Relocate small bits of validation left over from previous PRs marked with TODOs


StephenHulme:
**Describe the Housekeeping**
Relocation validation that is in the perform-accessioning path (rather than in the pre-accessioning checks)

Search for TODOs in accessioning lib code (asynchronous path) (`TODO: move this validation logic to a validation method somewhere else`)

**Blocking issues**
#5403 ✅ 

**Additional context**
I think this is the only bit of misplaced validation in the lib code, so it should be a tiny story and could be rolled into something else.

The models on the other hand.....

PR #5451, opened 2026-01-12
Y25-705 - Add validation failures to accession statuses

### LINKED ISSUE #5416: Y25-705 - Send internal validation errors to accessioning statuses instead of exception emails
**Describe the Housekeeping**
Send internal validation errors to accessioning statuses instead of exception emails.

As of #5403, users recieve feedback in the form of Accession Statuses should validation fail on the EBI side. However, many of the accessioning issues are caused by validation failures within Sequencescape (generally us creating a local copy of what EBI requires, as well as enforcing our own requirements). These errors are not communicated to users and cause a significant amount of confusion.

This story would redirect the validation failures (by capturing the `ActiveRecord::RecordInvalid` error) to be added to the accession statuses - or in a similar manner. This would allow users to clearly and easily see why a sample was not able to be accessioned.

In the same way, accessioning errors that currently use the flash-banner feedback mechansim would be recorded in the the accession statuses too. This would allow us to remove the `There are n more errors` message when the old synchronous accession path is removed in #5420.

There should also be a feature-flag added to ALSO send exception notification emails to developers to increase visibility of exception failures. These should be clearly marked with a different subject line or additional header to allow for easy inbox filtering.

**Acceptance criteria**

- [ ] Save validation errors to accession statuses
- [ ] Add feature flag to send exception notification emails for internal and external validation errors

**Blocking issues**
#5403 ✅ 
#5415 

**Additional context**

StephenHulme:
Closes #5416 and #5415 

Builds on and merges into #5444

- Move validation logic to inside delayed job 
- Add validation failures to accession status
- Add feature flags for sending notification exceptions to developers
  - `y25_705_notify_on_internal_accessioning_validation_failures`
  - `y25_705_notify_on_external_accessioning_validation_failures` 
- Change sample saving to use synchronous accessioning path 
- Rework validation failures and messages
- Rework some test expectations

#### Screenshots

Screenshot 2026-01-15 at 12 04 30

_Validation failure recorded with sample._


app/jobs/sample_accessioning_job.rb
```diff
@@ -19,13 +16,12 @@ def self.contact_user
...
       submission = Accession::Submission.new(contact_user, accessionable)
+      accessionable.validate! # See Accession::Sample.validate! in lib/accession/sample.rb
       submission.submit_accession(event_user)
     rescue StandardError => e
       handle_job_error(e, submission)
 
-      # Raising an error to Delayed::Job will signal that the job should be retried at a later time
-      job_failed_message = "#{e.class}: #{e.message}"
-      raise JobFailed, job_failed_message
+      raise # Raising an error signals that the job should be retried at a later time
```
BenTopping, line 24: Is Raise sufficient, do we need to include the error here?

app/controllers/samples_controller.rb
```diff
@@ -104,15 +104,13 @@ def update # rubocop:todo Metrics/AbcSize
...
-    flash.now[:warning] = @sample.errors if @sample.errors.present?
-
     if @sample.update(cleaned_params)
       flash[:notice] = 'Sample details have been updated'
+      flash[:warning] = @sample.errors.full_messages if @sample.errors.present? # show warnings from accessioning
       redirect_to sample_path(@sample)
     else
       flash[:error] = 'Failed to update attributes for sample'
-      render action: 'edit', id: @sample.id
+      redirect_to edit_sample_path(@sample)
```
yoldas, line 114: note: Flash message is shown in the following combinations:
`flash.now` and `render edit`
`flash` and `redirect_to`

BenTopping (approved):
Looks fine to me.

app/controllers/samples_controller.rb
```diff
@@ -176,15 +174,20 @@ def accession # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity,Met
     elsif accession_action == :update
       flash[:notice] = 'Accessioned metadata updated'
     end
-  rescue ActiveRecord::RecordInvalid => e
+
+    # Handle errors for both synchronous and asynchronous accessioning
+    # When the feature flag above is removed, the AccessionService and ActiveRecord errors can be removed
```
yoldas: question: which feature flag?

app/controllers/samples_controller.rb
```diff
@@ -159,14 +157,14 @@ def accession # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity,Met
...
 
-    @sample.validate_ena_required_fields!
-
     accession_action = @sample.accession_number? ? :update : :create
 
     if Flipper.enabled?(:y25_286_accession_individual_samples_with_sample_accessioning_job)
       # Synchronously perform accessioning job
       Accession.accession_sample(@sample, current_user, perform_now: true)
     else
+      # TODO: remove the AccessionService and ActiveRecord errors below when this accessioning path is removed
```
yoldas: nitpick: please add feature flag name and how it is related to the exception handling block.

lib/accession/sample.rb
```diff
@@ -105,7 +124,13 @@ def check_required_fields
...
 
     def check_studies
-      return if exactly_one_study?
+      exactly_one_study?
+      study_requires_accessioning?
+    end
+
+    def exactly_one_study?
+      # Check that sample is linked to exactly one study
+      return true if studies.length == 1
```
yoldas: note: if a sample is associated with more than one study, the error message would be incorrect.

lib/accession.rb
```diff
@@ -62,6 +62,7 @@ def <=>(other)
   # Usage: raise Accession::Error, "Accessioning failed: #{reason}"
   class Error < StandardError; end
   class ExternalValidationError < Error; end
+  class InternalValidationError < Error; end
```
yoldas, line 65: praise: good exception hierarchy with a catch-all Accession::Error root.

lib/accession/sample.rb
```diff
@@ -115,8 +140,11 @@ def check_studies
       end
     end
 
-    def exactly_one_study?
-      studies.length == 1
+    def study_requires_accessioning?
+      # Check if study is present and allowed to be accessioned
+      if sample.ena_study&.accession_required? != true # if true, accession; if false or nil, don't
```
yoldas, line 145: note: if sample has more than one study, because the previous guard is removed in check_studies, this line will be still visited.

app/jobs/sample_accessioning_job.rb
StephenHulme: A lone `raise` will automatically re-raise the triggering error, keeping the original traceback.

The JobFailed error was a wrapper for clarity, but broke the use of the delayed-job code when it was performed synchronously and the error had to be handled rather than just retried.

app/controllers/samples_controller.rb
StephenHulme: In particular, `flash.now` is designed for `render` and was not compatible with the rediect, hiding the warnings.

Redirects are now used in both cases, which is compatible with the warning `flash`.

lib/accession/sample.rb
StephenHulme: The method `ena_study` is a wrapper around `studies.first`. Not ideal, but it does prevent the check from throwing strange errors.

The validation looks for the presense of `self.errors` (see `errors.add`) to fail the validation, so multiple checks can be gracefully handled.

app/controllers/samples_controller.rb
StephenHulme: I've also spotted and fixed a small logic change in 0e2998c34dfdb8e8fa121f89ab0f89633f76049c

app/controllers/samples_controller.rb
StephenHulme: More details added in 5659a75cc562e02d4a8faff09efb4bfe72032c0b

app/controllers/samples_controller.rb
StephenHulme: See 5659a75cc562e02d4a8faff09efb4bfe72032c0b

PR #5456, opened 2026-01-15
Y25-681 Bulk submission and request additional sequencing - add 250 read length for novaseq6000 sp flowcell

### LINKED ISSUE #5355: Y25-681 - Add 250PE Read Length Option for NovaSeq6000 SP Flowcell
**Context**
The NovaSeq6000 SP flowcell supports a 250PE read length kit, but this option is currently missing from Sequencescape (SS), because users cannot select 250PE directly, they rely on manual comments, which can be missed by the loading team. This gap has resulted in a CAPA (25_NC_S_229), requiring the system to formally support 250PE selection.

**User story**
As a sequencing requester, I want the 250PE read length option available for the SP flowcell in Sequencescape, So that I can request this read length directly without relying on manual comments, reducing the risk of loading errors.

Liz C

Liz C

**Acceptance criteria**

- [x] 250PE read length must be added as a selectable option for SP flowcell on:
           - The Request Additional Sequencing page
           - The Bulk Submission upload workflow
- [x] Validation must be added that catches inappropriate use of the 250 read length (only SP flowcell allows it)

**Dependencies**

NA

References
NA

**Additional context**

**Deployment notes**
I added update ability to the request_type and request_type_validators record loaders.
Check counts and update timestamps of rows on those tables before and after deployment to UAT and Training to check that doesn't update something it shouldn't. If it does, possibly need to fix record loader yml files to match prod database.

andrewsparkes:
Closes #5355 

Adds validation to allow 250 read length for NovaSeq 6000 paired end bulk submission and to requests for additional sequencing, when using flowcell type SP.


config/default_records/request_type_validators/default_records.yml
```diff
@@ -9,11 +9,20 @@
 #   other_attribute: 'Value'
 # <%= record_key %> B:
 #   other_attribute: 'Value'
-FlowcellRequested:
+FlowcellRequestedNovaSeq6000PE:
```
yoldas, line 12: note: This is the correct way of doing it, i.e. dedicated flowcell requested config for novaseq 6000, but I am not sure if this would be backward compatible. It needs testing existing novaseq 6000 pipeline batches.

KatyTaylor, line 12: So when deployed to UAT, we would expect this to update the `key` of the existing record (which currently has key `FlowcellRequested`).

config/default_records/request_type_validators/default_records.yml
```diff
@@ -9,11 +9,20 @@
...
     request_type_key: illumina_htp_novaseq_6000_paired_end_sequencing
+ReadLengthRequestedNovaSeq6000PE:
+  request_type_key: illumina_htp_novaseq_6000_paired_end_sequencing
+  request_option: read_length
+  valid_options:
+    - 50
+    - 75
+    - 100
+    - 150
+    - 250
```
KatyTaylor, line 25: When deployed to UAT, we would expect this to update the existing record, which currently has a blank `key`, with:
a) a new `key`
b) added option `250`

lib/record_loader/request_type_loader.rb
```diff
@@ -10,19 +10,30 @@ class RequestTypeLoader < ApplicationRecordLoader
     config_folder 'request_types'
 
     def create_or_update!(key, options)
```
yoldas, line 12: note: the key argument here, the first argument - whatever name it has, is the section name in yaml file, which needs to be unique across all section names within the same record loader folder. If a name or a key specified within that section as well, they are available inside options hash.

config/default_records/request_type_validators/default_records.yml
KatyTaylor: I had a closer look at the `db > seeds` scripts to see if there would be duplication when run locally. It actually works out quite neatly because this is the seed code that currently creates the relevant 'request type validator' record:

```
RequestType.find_each do |request_type|
  read_lengths =
    {
      # By request class
      'HiSeqSequencingRequest' => [50, 75, 100, 150],
      'MiSeqSequencingRequest' => [25, 50, 130, 150, 250, 300],
      'SequencingRequest' => [37, 54, 76, 108]
    }[
      request_type.request_class_name
    ]

  if read_lengths.present?
    RequestType::Validator.create!(
      request_type: request_type,
      request_option: 'read_length',
      valid_options: read_lengths
    )
  end
end
```

This will not create it any more after your change, because we're changing the `request_class_name`.

So all good I think!

lib/record_loader/request_type_validators_loader.rb
```diff
@@ -11,8 +11,34 @@ class RequestTypeValidatorsLoader < ApplicationRecordLoader
...
+    end
+
+    private
+
+    # Finds a RequestType by key and handles missing records based on the environment.
+    def find_request_type!(request_type_key, key)
       rt = RequestType.find_by(key: request_type_key)
-      RequestType::Validator.create_with(options.merge(request_type_id: rt.id)).find_or_create_by!(key:)
+      if rt.nil?
+        message = "RequestType with key '#{request_type_key}' not found for RequestType::Validator with key '#{key}'"
```
yoldas, line 24: note: The associated rake task of the record loader lib/tasks/record_loader/request_type_validators.rake specifies dependency on lib/tasks/record_loader/request_type.rake, which means the request type specified will be available before visiting here.

lib/record_loader/request_type_validators_loader.rb
```diff
@@ -11,8 +11,34 @@ class RequestTypeValidatorsLoader < ApplicationRecordLoader
...
+        if Rails.env.development? || Rails.env.staging? || Rails.env.cucumber?
+          Rails.logger.warn(message)
+          return nil
```
KatyTaylor, line 27: What's the reason for the different behaviour in different environments?

lib/record_loader/request_type_validators_loader.rb
```diff
@@ -11,8 +11,34 @@ class RequestTypeValidatorsLoader < ApplicationRecordLoader
...
+
+    # Creates or finds and updates the RequestType::Validator.
+    def find_or_update_validator!(req_type, key, options)
+      validator = RequestType::Validator.find_or_initialize_by(
+        request_type_id: req_type.id,
+        request_option: options['request_option']
+      )
+      validator.assign_attributes(options.merge(request_type_id: req_type.id, key: key))
+      validator.save!
     end
```
KatyTaylor, line 42: Think this will throw an unhandled exception if previous method returns `nil`? So might as well throw the intentional exception in the first method instead?

lib/tasks/sequencing.rake
```diff
@@ -35,7 +35,7 @@ namespace :sequencing do
               RequestType::Validator.create!(
                 request_type: rt,
                 request_option: 'read_length',
-                valid_options: [150, 50, 75, 100]
+                valid_options: [50, 75, 100, 150, 250]
```
KatyTaylor, line 38: OK so this won't conflict in prod because it will only run if the Request Type doesn't exist. However, it will run when setting up your local environment from scratch. And it duplicates the record loader creation of the request type and request type validator. Not ideal - should ideally get rid of this one. Potentially complicated thought because of all the related records...?

Also, this version of the RequestTypeValidator doesn't set the key.

yoldas (commented):
I think it needs read length request information type record in config/default_records/pipeline_request_information_types/default_records.yml but it is already created in the database.
```sql
select * from sequencescape_production.pipeline_request_information_types prit 
join sequencescape_production.request_information_types rit on prit.request_information_type_id = rit.id
join sequencescape_production.pipelines p on prit.pipeline_id = p.id
where p.name = 'NovaSeq 6000 PE';
```

config/default_records/request_type_validators/default_records.yml
andrewsparkes: Yes, it does this

andrewsparkes: ah, you mean ones in progress? hadn't considered that, but FlowcellRequested is nowhere in our codebase other than one mention in this config file.
could NPG need it you think?

lib/record_loader/request_type_loader.rb
andrewsparkes: yes, this is why I wanted the keys more specific and changed the FlowcellRequested one to include the NovaSeq type name (X or 6000).

lib/record_loader/request_type_validators_loader.rb
andrewsparkes: copied from another one (tag_set_loader), I think maybe it was to prevent blowing up the rest of the config generation while developing/testing? Can remove it, most don't do that.

lib/record_loader/request_type_validators_loader.rb
andrewsparkes: yes I know, but only if someone's remembered to create the request_type in the config, and they've got the keys matching(!).
it blew up poorly when I was testing as I'd mismatched the key names.
not necessary you think? I can remove.

lib/record_loader/request_type_validators_loader.rb
andrewsparkes: yes, see my reply above, it wasn't good when I mismatched the keys in testing

lib/tasks/sequencing.rake
andrewsparkes: hmm, fair point.
and that explains why a lot of keys were null in the DB
I can see if I can remove this setup

config/default_records/request_type_validators/default_records.yml
andrewsparkes: Without the update code, it was creating a second, new record with the key. But then selecting the first, the old one in the code. No good.

config/default_records/request_type_validators/default_records.yml
andrewsparkes: It behaves, tested in UAT and Training now.

config/default_records/request_type_validators/default_records.yml
andrewsparkes: Tested batch creation and all seems good

PR #5486, opened 2026-01-22
Y25-286 - Separate accession eligibility from validity

StephenHulme:
Merges into #5491 

- Do not raise `InternalValidationErrors` if a sample should not be accessioned
- Replace scope `for_sample_accessioning` with Railsy model methods
- Refactor validatation of studies for accessioning
- Removes error messages and does not accession these samples, but records a reason as a rails log instead
  - `Sample is linked to a study that does not require accessioning.`
  - `Sample is not linked to any studies but must be linked to exactly one study.`
  - Removes a lot of email spam...
- Update tests


app/models/sample.rb
```diff
@@ -540,6 +567,11 @@ def ena_study
     studies.first
   end
 
+  def study_for_accessioning
+    # There should only be one study for accessioning, if we want to accession, so we return the only one
```
BenTopping, line 571: Why is this? What if there are more?

app/models/study.rb
```diff
@@ -572,6 +562,18 @@ def accession_number?
...
+  # Returns true if the samples in this study are eligible for accessioning
+  #
+  # A study's samples are eligible for accessioning if:
+  # - the study is not set to never release
+  # - the study has an accession number
+  # - the study requires accessioning
+  #
+  # @return [Boolean] true if the samples in this study are eligible for accessioning, false otherwise
+  def samples_accessionable?
+    !study_metadata.never_release? & accession_number? & accession_required?
```
BenTopping, line 574: Should this be using `&&` instead of bitwise `&`

spec/controllers/studies_controller_spec.rb
```diff
@@ -263,12 +265,36 @@
...
+
+      context 'when samples are part of two accessionable studies' do
+        let(:samples) { create_list(:sample_for_accessioning_with_open_study, number_of_samples) }
+        let(:study) { create(:managed_study, accession_number: 'EGA123', samples: samples) }
+
+        it 'redirects to the accession-statuses tab of the study page' do
+          expect(subject).to redirect_to(study_path(study, anchor: 'accession-statuses'))
+        end
+
+        it 'does not set a flash notice message' do
```
BenTopping: Do users not need to know the samples aren't being accessioned? Seems slightly misleading.

app/models/sample.rb
StephenHulme: I'm going to say "historical reasons". The previous code was `studies.first`.

As a general rule, SS is not designed to handle more than one study per sample, but we allow it some some cases to permit "holding studies". For accessioning purposes, it can be tricky at an ENA/EGA level if a sample is under two studies with different release data criteria.

app/models/study.rb
StephenHulme: It could be, though in Ruby `&&` is a short-circuit operator rather than boolean. It changes to an array and any in the next PR #5499 ([specific place](https://github.com/sanger/sequencescape/pull/5499/files#diff-fb211b1a958eef21d6955a7fb271bcfc4497c3e88090861bc972b5a4a9c0b040))

spec/controllers/studies_controller_spec.rb
StephenHulme: Typo, good catch. There's an expectation, but clearly my copy-pasta skills are not up to scratch!

StephenHulme: Fixed in 6293999676e7ca007f816cda92b0183d059de9fc

BenTopping: Nice. I also meant the users are told the samples are sent to accessioning but in reality they aren't as they fail the validation about belonging to 2 studies so may confuse users thinking its gone through.

StephenHulme: Ah, even better catch! That should be fixed - I"ll look into it...

StephenHulme: Ah there's a subtlety here - the accessioning for the "accession all samples" button happens asynchronously and immediate feedback is not possible.
The current message isn't wrong, but doesn't show the whole picure.
I've added a line to encourage the user to come back to confirm that accessioning succeeded in 278090416db8e432d4c24eda95ed459ddddc21bc.

BenTopping: Ah I see, yes I think that change is sufficient, thanks.

PR #5515, opened 2026-02-03
Y25-706 - Replace accessioning flag config setting with feature flag

### LINKED ISSUE #5417: Y25-706 - There should be a single configurable flag that cleanly turns off all accessioning for an environment
**Describe the Housekeeping**
There should be a single configurable flag that cleanly turns off all accessioning for an environment. (See existing flag `configratron.accession_samples` which works for one code path currently).

Consolidate the existing references to `configratron.accession_samples`. Ideally there should be only one usage of this flag, but two might be required - one for samples accession and one for everything else. This usage should be one of the first bits of code in the accessioning code-path and be easy to find and understand.

**Blocking issues**
#5420

**Additional context**
Be aware of the the `AccessionService::AccessioningDisabledError` which is ideally used in the models (synchronous) accessioning path. A `Accession::AccessioningDisabledError` might need to be created for the lib (asynchronous) accessioning path. Or an alternative approach could be used.

StephenHulme:
Closes #5417 

See related [deployment project PR](https://github.com/sanger/deployment/pull/795).

The [Confluence article on how to turn accessioning on and off](https://ssg-confluence.internal.sanger.ac.uk/spaces/PSDPUB/pages/290915138/Enable+and+disable+accessioning+in+Sequencescape) has been updated to reflect these changes.

- Replaces config `configatron.accession_samples` with Flipper feature flag `y25_706_enable_accessioning`
- Updates tests to match


config/initializers/flipper.rb
```diff
@@ -27,7 +27,7 @@
   # config.flipper.log = true
 
   ## Reconfigure Flipper to use the Memory adapter and disable Cloud in tests
-  # config.flipper.test_help = true
+  config.flipper.test_help = false # Allows spec helpers, see https://github.com/flippercloud/flipper/issues/893
```
andrewsparkes, line 30: what's this doing?

andrewsparkes: ah so you use the flags in tests

andrewsparkes (approved):
looks good

config/initializers/flipper.rb
StephenHulme: It seems that Flipper by default resets all flags prior to test runs - this is good, but happens at the wrong level... 
When I tried to add the `config.around(:each, :accessioning_enabled) do |example|` spec-helper, the helper would set the variable - only for Flipper's test help setting (apparently true by default) to Immediately reset the flag :(
Finding the issue above, and setting `test_help = false` solved the problem and allowed `the :accessioning_enabled spec-helper` to work as intended.

PR #5525, opened 2026-02-06
Y25-016-scrna-submission-pooling-plan

### LINKED ISSUE #5524: Y25-016 - Display proposed pooling plan from successful scRNA Core cDNA Prep pipeline bulk submission
**User story**

As a user of the scRNA Core cDNA Prep pipeline, I would like some changes to the cDNA Prep submission (bulk submission interface only) to display to the user the proposed pooling plan upon successful submission of samples.

Abby, Liz H

Abby, Liz H

**Acceptance criteria**

- [ ] display the study/project/donor group(s) within the submission
- [ ] display the proposed number of samples per pool for each
- [ ] display the requested number of cells per chip well for each

References
- [Y24-512 - Show the result of the donor pooling algorithm sanger/limber#2129](https://github.com/sanger/limber/issues/2129)

**Additional context**

Story referenced above displays similar information at pooling feasibility recheck step so could be helpful.
Similar display as below:

	Study/project/donor group	                Pools (num samples)	Number cells per chip well
	Study X - Project A - Donor group 1	8, 8, 7	                        50,000
	Study Y - Project B - Donor group 2	10, 10	                        90,000

BenTopping:
Closes #5524

- Adds scrna core cdna pooling plan download link to submission show page.
- Adds ScrnaCoreCdnaPrepPoolingPlanGenerator for pooling plan csv content generation

#### Notes

I have tested this against the int suite and confirmed the proposed pooling plan matches the actual pooling seen in LRC PBMC Pools 'Pooling information' section.

I think the next best test will be testing on Training and comparing real test submissions against with the real pooling information as it is difficult to simulate production runs. 

TODO: Update confluence with additional pooling generator content.


app/models/submission/scrna_core_cdna_prep_pooling_plan_generator.rb
```diff
@@ -0,0 +1,52 @@
...
+    # Ideal pool size is just the number of samples divided by the number of pools, but we need to account for
+    # any remainder if the division isn't exact
+    ideal_pool_size, remainder = number_of_samples.divmod(number_of_pools)
+    pools_layout = Array.new(number_of_pools, ideal_pool_size)
+    remainder.times { |i| pools_layout[i] += 1 }
+    pools_layout
+  end
+
+  # Groups the requests associated with a submission by study and project.
+  def self.grouped_requests(submission)
```
BenTopping, line 45: I think requests and labware and synonymous as far as the grouping is concerned at this point. Favoured requests over labware as the study/project data is derived from the submisison rather than related aliquots.

app/models/submission/scrna_core_cdna_prep_pooling_plan_generator.rb
```diff
@@ -0,0 +1,52 @@
...
+      # It would be nice to refactor the scRNA Validator logic here to pull out the pooling plan logic
+      grouped_requests(submission).each do |study_project, subgroup|
+        # Get number_of_pools and cells_per_chip_well requested from the submission
+        number_of_pools = subgroup.first.request_metadata.number_of_pools
+        cells_per_chip_well = subgroup.first.request_metadata.cells_per_chip_well
+        # Build the pools
+        pools_layout = calculate_pools_layout(subgroup.size, number_of_pools)
+
+        # Join the pool sizes into a string for the CSV output
+        number_of_samples_in_pool = pools_layout.join(', ')
```
yoldas, line 26: without additional quotes around the field value, embedded commas between numbers breaks this CSV. Does the output column have quotes?

app/models/submission/scrna_core_cdna_prep_pooling_plan_generator.rb
yoldas: There is a request on each well. This method groups requests, hence wells because of request.asset, by `"#{study} / #{project}"`.

app/models/submission/scrna_core_cdna_prep_pooling_plan_generator.rb
BenTopping: Good point, I have checked this and ues the output has quotes  so the CSV should not break.

app/models/submission/scrna_core_cdna_prep_pooling_plan_generator.rb
BenTopping: Saying this I have realised if they set number_of_lanes in submission to 2 this logic will not work as the submission will have double requests...

But I am not sure if they can / should do this for scRNA, it seemed to break the int suite setup I had. Maybe one for @KatyTaylor or the SSRs?

BenTopping: I have added a uniqueness on assets in case multiple lanes are requested. Changes in [fc5e0a6](https://github.com/sanger/sequencescape/pull/5525/commits/fc5e0a61d93a8a073e1dbbeafc8955aaa7795164)

app/views/submissions/show.html.erb
```diff
@@ -4,6 +4,9 @@
   add :menu, "Print labels for #{order.asset_group.name}" => print_study_asset_group_path(order.study,order.asset_group) unless order.asset_group.nil?
 end %>
 <%- add :menu, "Submissions Inbox" => submissions_path if can? :read, Submission -%>
+<%- add :menu, "Download scRNA Core cDNA pooling plan" => download_scrna_core_cdna_pooling_plan_submission_path(@presenter.submission.id) if
```
KatyTaylor: **suggestion:** Users can tell you what they want, but I would take the word `cDNA` out, because it makes it sound like we're pooling cDNA. I get where it came from, as we're in the cDNA Prep pipeline... but the cDNA hasn't been created yet at the pooling stage.

app/models/submission.rb
```diff
@@ -125,6 +126,10 @@ def multiplexed?
     orders.any?(&:multiplexed?)
   end
 
+  def scrna_core_cdna_prep_gem_x_5p_submission?
+    orders.first.template_name == SCRNA_CORE_CDNA_PREP_GEM_X_5P
```
KatyTaylor, line 130: Ah good find, I didn't know the template name was saved. Might be slightly more robust to check if any of the request types in the submission are the correct type... in case in future the cdna prep request type is chained together with another in a different template. But that could break too if the request type changes.

app/controllers/submissions_controller.rb
```diff
@@ -115,6 +115,29 @@ def study
...
+      redirect_to submissions_path
+      return
+    end
+
+    unless submission.scrna_core_cdna_prep_gem_x_5p_submission?
+      flash[:error] =
+        'This submission does not have the correct template for downloading a scRNA Core cDNA pooling plan'
+      redirect_to submission
+      return
+    end
```
KatyTaylor, line 132: This is just being cautious, right, because the button shouldn't display in this case?

BenTopping: Yes thats right. We have a unique URL that the button invokes so in theory it could be invoked by any submission so we are just guarding against it.

app/views/submissions/show.html.erb
BenTopping: Ah I see, okay ill take it out.

andrewsparkes (approved):

app/views/submissions/show.html.erb
BenTopping: I have removed the term in [42b8ad1](https://github.com/sanger/sequencescape/pull/5525/commits/42b8ad1688fa2aa39aa1578c9d5be0318bf1b9d4)

app/models/submission.rb
BenTopping: Yes good point. I'm worried if we try support future submissions based off request type the logic might not hold up as we are basing this off the requests which might not work if the prep gem x 5p is not the first/primary request type.

app/models/submission/scrna_core_cdna_prep_pooling_plan_generator.rb
yoldas: Note: I can't quite understand how it is related to "multiple lanes", but the change looks OK.

PR #5531, opened 2026-02-09
Y25-286 - Add missing accessioning guard clauses

StephenHulme:
Fixes a bug noticed while documenting accessioning.

- Add missing guard clauses to Sample editting button text
- Add missing guard clauses to Sample Manifest accession buttons
- Refactor accession check for accession statuses buttons
- Move accession all samples logic out of study model and into studies controller
- Add note reminding not add inflate the sample model with controller logic
- Update tests to reflect changes

### DIFF
--- a/spec/helpers/samples_helper_spec.rb
+++ b/spec/helpers/samples_helper_spec.rb
@@ -7,6 +7,10 @@ RSpec.describe SamplesHelper, type: :helper do
 
   let(:sample) { create(:sample) }
 
+  let(:accessioning_enabled) { true }
+  let(:should_be_accessioned) { true }
+  let(:permitted) { true }
+
   before do
     allow(helper).to receive(:accessioning_enabled?).and_return(accessioning_enabled)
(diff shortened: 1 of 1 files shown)


spec/helpers/samples_helper_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  context 'when sample should not be accessioned' do
+    let(:accessioning_enabled) { true }
+    let(:should_be_accessioned) { false }
+    let(:permitted) { true }
+
+    it { is_expected.to eq('Save Sample') }
+  end
+
+  context 'when user is not permitted to accession' do
+    let(:accessioning_enabled) { true }
```
BenTopping: If you move these variables to a parent context and default them to true you should only need to change one variable each time instead of redeclaring them all.

StephenHulme: Good point, thanks!

StephenHulme: Fixed in 03e9f4f74b7e279e43e988cfd3a372e3e4015adf

PR #5542, opened 2026-02-12
Y26-011 - PoolReleased dead letters

BenTopping:
Closes #https://github.com/sanger/General-Backlog-Items/issues/691

- Prevents creation_requests causing an error if none exist on multiplexed_library_tube


yoldas:
Now the asset link creation is synchronous, https://github.com/sanger/sequencescape/pull/4885 , do we still need this PR?

app/models/multiplexed_library_tube.rb
```diff
@@ -41,6 +41,8 @@ def creation_requests
     direct = requests_as_target.where_is_a(Request::LibraryCreation)
     return direct unless direct.empty?
 
-    parents.includes(:requests_as_target).first.requests_as_target
+    # Parents should exist but in the case they don't (e.g. asset_links are yet to be created)
+    # we want to avoid an error and just return an empty array.
```
KatyTaylor, line 45: > just return an empty array

I think it all works, but just to note, this will return `nil` if there are no parents, and `[]` if there is a parent but it has no `requests_as_target`

BenTopping: Good spot. Updated in [399d60b](https://github.com/sanger/sequencescape/pull/5542/commits/399d60b02e40c3b764ed9296593b8327103fb4c7)

PR #5546, opened 2026-02-16
Y26-034-2 - Add request logging middleware

StephenHulme:
Contributes to https://github.com/sanger/General-Backlog-Items/issues/699

While Rails logging is comprehensive, it splits multiple details over many lines making it hard to consolidate logs into a single record. This is most noticable when trying to correlate details from the start of a request (client ip, url) with those in the response (status code, time taken). This PR generates a single log record that can be easily parsed by Kibana to consolidate these details.

I looked at many other similar libraries before writing this middleware, but unfortunately none of them met all our requirements. I tried implementing the two below, but encountered obstacles until deciding that it was easier, simpler, and lighter-weight to write our own.
- [lograge](https://github.com/roidrage/lograge) - No recent support
- [logstasher](https://github.com/shadabahmed/logstasher) - Broken and confusing config

This is currently implemented as two files (plus tests) in Sequencescape, but I would like to extend this to some of our our apps too. My current thought is to just copy-paste this implementation between apps, but maybe making a gem would be better? Different sorts of admin really...

- Create a new Rack middleware for Rails focussed on consolidating details about a request and response into a single log record
  - Inserts before Rails::Rack::Logger (see `bin/rails middleware`)
  - Implements it's own timing using the same clock and algorithm as in the default Rack timing middleware ([Runtime](https://github.com/rack/rack/blob/main/lib/rack/runtime.rb)) for accuracy and consistency
  - Consolidates the multiple metrics into a single log record
  - Allows the log-level of the resulting log to be specified in the initializer
- Current metrics:
  - method: GET
  - url: /samples/1234?foo=bar
  - path: /samples/1234
  - format: html
  - status_code: 200
  - status_message: OK
  - duration_ms: 935
  - client_ip: 172.21.43.210
  - request_id: 9fd18098-dea3-46f0-83c8-c41852441db3
  - tags: ["request","success"]
  - timestamp: 2026-02-12T12:10:50.284+00:00
  - Any additional info passed in at initialisation such as host and version

This will generate a log that looks like:
```log
[INFO] [RequestLogger] {"method":"GET","path":"/samples/1775","format":"html","status_code":200,"status_message":"OK","duration_ms":954,"client_ip":"172.21.12.34","request_id":"cf583876-0138-4e1c-91e5-49a3c6e68781","tags":["request","success"],"@timestamp":"2026-02-16T12:40:19.991+00:00","host":"sss0-uat","version":"v14.87.0"}
```
#### Related:
Add parser to [logstash in deployment](https://github.com/sanger/deployment/pull/804) and [packer](https://github.com/sanger/packer/pull/46) to ingest this log.

### DIFF
--- a/config/initializers/request_logger.rb
+++ b/config/initializers/request_logger.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require_relative '../../lib/deployed'
 require_relative '../../app/middleware/request_logger'
 
 Rails.application.configure do


app/middleware/request_logger.rb
```diff
@@ -0,0 +1,84 @@
...
+  def elapsed_milliseconds
+    start_time = clock_time
+    result = yield
+    end_time = clock_time
+
+    elapsed_ms = ((end_time - start_time) * 1000).round
+    [result, elapsed_ms]
+  end
+
+  def tag_for_status(status_code)
```
yoldas, line 45: A bit paranoid to avoid middleware crash:
```
def tag_for_status(status_code)
  code = status_code.to_i rescue nil
  case code
  when 100..199 then 'informational'
  when 200..299 then 'success'
  when 300..399 then 'redirection'
  when 400..499 then 'client_error'
  when 500..599 then 'server_error'
  else 'unknown'
  end
end
```

StephenHulme: Already tested in [request_logger_spec.rb](https://github.com/sanger/sequencescape/blob/a72cb51b15c2b1078b21b2b9d2aa17694e03d92b/spec/middleware/request_logger_spec.rb#L121-L125).
Returns `nil` which is then removed with `compact!`.
I didn't want to return `unknown`, because as a tag with no context I felt it was more misleading than helpful.

yoldas: I was worried about passing something that can't be cast to integer, but it seems case statement just returns nil in such cases.

Issue #5454, opened 2026-01-14
Y26-012 - Automating buffer addition - wireframes


KatyTaylor:
**Description**
Bit of research into the parent story to enable a better size estimate, and check customer expectations of how the  feature will look.

**Acceptance Criteria**
The following outputs from the research are produced:
- [ ] Create quick wireframes (any method) showing the places in the UI that will change for the parent story - show to the customer for feedback
- [ ] Create description (English, or pseudo-code) of how the cherrypicking algorithm and/or Tecan file generation will need to change for the parent story
- [ ] Split the parent story into subtasks, if it seems appropriate

**Who the primary contacts are for this work**
Katy, Lynda

**Additional context or information**
In case helpful for getting orientated, here are recent(ish) stories done on the cherrypicking code:
- https://github.com/sanger/sequencescape/issues/5101 (Tecan file generation)
- https://github.com/sanger/sequencescape/issues/4413 (Changes to the cherrypicking algorithm, never merged)

PR #5565, opened 2026-02-23
Y26-680 Automated buffer addition for cherrypicking

### LINKED ISSUE #5354: Y25-680 - Automating Buffer Addition to Empty Wells During CP on the Tecan
**Context**
During the Concentration & Purification (CP) step on the Tecan, buffer is only added to wells that have a CP request coming from SequenceScape. However, when a plate is sheared, all 96 wells require 120 µl of buffer, including wells that are empty or not populated via CP. Because the system currently does not fill these empty wells automatically, the laboratory team must manually add buffer to all empty wells.

This is only relevant to cherrypicking and not normalisation.

**User story**
As a laboratory technician, I want the Tecan to automatically add 120 µl of buffer to all wells that are empty during the CP step, so that I can avoid manual buffer addition, reduce the risk of processing errors, and save time by using automation efficiently.

Lynda D

Lynda D

**Acceptance criteria**
- [ ] This functionality only applies when the user ticks a checkbox during the cherrypicking submission process (decided to make this user-specified as were not sure how to distinguish safely between norms and cherrypicks in the system).
- [ ] After the cherrypick has been planned, any empty wells on the target plate(s) should have 120 µl of buffer added to them. This should be reflected in the Tecan driver files and any other relevant places (e.g. the worksheet).
- [ ] All 96 wells must contain 120 µl of liquid after the automated step when processing sheared plates.

**Additional context**
~I hear from the team that Tecan used to fill automatically the plate with buffer when 'empty' from a CP paperwork point of view.~ See comment https://github.com/sanger/sequencescape/issues/5354#issuecomment-3921715146

A working implementation was done in PR https://github.com/sanger/sequencescape/pull/5553 of the research story  https://github.com/sanger/sequencescape/issues/5454 , but it has no tests. 

Note that the TecanV3 driver file generator , which reuses tips, is already in production but database records were not created to enable it.

**Deployment notes**

andrewsparkes:
Closes #[y26-680](https://github.com/sanger/sequencescape/issues/5354)

Optional automated buffer addition for cherrypicking


app/frontend/entrypoints/cherrypick_strategies.test.js
```diff
@@ -0,0 +1,33 @@
...
+    // Re-require the script to attach event listeners
+    jest.resetModules();
+    require("./cherrypick_strategies.js");
+    bufferInput = document.getElementById("buffer_volume_for_empty_wells");
+    autoBufferCheckbox = document.getElementById("automatic_buffer_addition");
+  });
+
+  it("disables buffer input when checkbox is unchecked", () => {
+    autoBufferCheckbox.checked = false;
+    document.dispatchEvent(new Event("DOMContentLoaded"));
```
yoldas, line 25: note: the test may be OK because after DOMContentLoaded event, it registers a change event and then calls `toggleBufferInput` once to do an initial toggle. However, the toggle behaviour could be tested on the registered change event on the checkbox element `automatic_buffer_addition` by a dispatch as in `autoBufferCheckbox.dispatchEvent(new Event("change"));`

note: this PR adds JS testing (package.json) although there is no `yarn test` in CI workflows yet.

app/models/batch/poly_metadata_behaviour.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+module Batch::PolyMetadataBehaviour
+  # Returns whether the Cherrypick automatic_buffer_addition option is enabled
+  # @return [Boolean] whether the automatic_buffer_addition option is enabled
+  def automatic_buffer_addition?
+    %w[1 on].include?(get_poly_metadata(:automatic_buffer_addition))
```
yoldas, line 7: note: check on both **1** and **on** is done to cover both visible and hidden option in the forms of successive pages.

app/models/concerns/has_poly_metadata.rb
```diff
@@ -0,0 +1,36 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :poly_metadata, as: :metadatable, dependent: :destroy, inverse_of: :metadatable
+  end
+
+  # Sets a PolyMetaDatum for the given key and value.
+  # If value is present, it will create or update the PolyMetaDatum with the
+  # given key and value, otherwise it will destroy the PolyMetaDatum with the
+  # given key if that exists.
```
yoldas, line 13: note: this is because PolyMetaDatum validations prevent key duplication and blank values, although there are no such DB constraints.

app/models/well.rb
```diff
@@ -365,4 +365,8 @@ def name
   def library_name
     nil
   end
+
+  def empty?
+    aliquots.blank? || aliquots.empty?
```
yoldas: note: aliquots.empty? is redundant I think because blank? covers both nil and empty.

yoldas (approved):
Looks good. My comments are notes. After UAT, we might consider describing the options and limitations at https://ssg-confluence.internal.sanger.ac.uk/spaces/PSDPUB/pages/232620342/Cherrypicking+options+in+Sequencescape .

app/frontend/entrypoints/cherrypick_strategies.test.js
andrewsparkes: tried changing:
```
autoBufferCheckbox.checked = false;
document.dispatchEvent(new Event("DOMContentLoaded"));
expect(bufferInput.disabled).toBe(true);
```

to:
```
autoBufferCheckbox.checked = false;
autoBufferCheckbox.dispatchEvent(new Event("change"));
expect(bufferInput.disabled).toBe(true);
```

but test fails. expect is false. I'm missing something.

app/models/batch/poly_metadata_behaviour.rb
andrewsparkes: ah ok wondered what that was, added comment

app/models/concerns/has_poly_metadata.rb
andrewsparkes: thx, added to comment

app/models/well.rb
andrewsparkes: thanks, changed to just blank?

app/frontend/entrypoints/cherrypick_strategies.test.js
yoldas: I meant the following, which can be added as well.

```
it("enables/disables buffer input when checkbox is toggled on/off", () => {
   // register the change event handler and initially disable bufferInput 
   // because  autoBufferCheckbox is unchecked 
   
    document.dispatchEvent(new Event("DOMContentLoaded")); 

   // You can set a value and trigger the change event on autoBufferCheckbox, 
    // or just click it instead.
    
    // autoBufferCheckbox.checked = true;
    // autoBufferCheckbox.dispatchEvent(new Event("change"));
    autoBufferCheckbox.click(); // unchecked to checked
    expect(bufferInput.disabled).toBe(false);

    // You can set a value and trigger the change event on autoBufferCheckbox, 
    // or just click it instead.
    
    // autoBufferCheckbox.checked = false;
    // autoBufferCheckbox.dispatchEvent(new Event("change"));
    autoBufferCheckbox.click(); // checked to unchecked
    expect(bufferInput.disabled).toBe(true);
  });
```

KatyTaylor:
There's a lot of code :) any chance of a brief summary / description of changes, and a screenshot or two of the UI changes? Thanks

Edit by Andrew as can't seem to reply: see Abdullah's original [story with the screenshots here](https://github.com/sanger/sequencescape/issues/5454)

app/models/map.rb
```diff
@@ -96,6 +96,8 @@ def self.plate_length(plate_size)
       PLATE_DIMENSIONS[plate_size].last
     end
 
+    # well number counting by columns, length is the number of rows in the plate
+    # e.g. B5 sends this 34 and 8
```
KatyTaylor, line 106: Nice, I remember adding a similar comment elsewhere 😅

app/models/batch/poly_metadata_behaviour.rb
KatyTaylor: Still not understanding this, hopefully will by the time I've finished reviewing...

app/models/concerns/has_poly_metadata.rb
```diff
@@ -0,0 +1,38 @@
...
+  # NB: this is because PolyMetaDatum validations prevent key duplication and blank values,
```
KatyTaylor, line 14:
```suggestion
  # NB: this is because PolyMetaDatum model validations prevent key duplication and blank values,
```

Thanks for the explanation. I still had a minor confusion, hence the suggestion.

KatyTaylor (commented):
Didn't manage to finish reviewing, will look again Monday

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
```diff
@@ -0,0 +1,25 @@
...
+    return unless %w[1 on].include?(params[key])
+
+    # If automatic buffer addition for empty wells is required, check and
+    # set the required buffer volume in the batch polymetadata.
+    key = :buffer_volume_for_empty_wells
+
+    # method valid_float_param? is defined in Cherrypick::Task::PickHelpers
+    unless valid_float_param?(params[key])
+      raise Cherrypick::VolumeError,
+            "Invalid buffer volume for empty wells: #{params[key]}"
```
KatyTaylor, line 33: I assume this exception is caught higher up and displayed nicely in the UI?

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Cherrypick::Task::BufferVolumeForEmptyWellsOption
+  def create_buffer_volume_for_empty_wells_option(params)
+    return unless @batch
+
+    key = :automatic_buffer_addition
+    # The checkbox value is either "1", or nil if not checked.
+    @batch.set_poly_metadata(key, params[key])
```
KatyTaylor: If the value can be 1 or 'on', that seems to be a quirk of the UI, and shouldn't be carried through to the database - can we unify them here and set it to a single value?
(Perhaps I'm not understanding)

app/models/robot/generator/behaviours/tecan_default.rb
```diff
@@ -113,4 +116,88 @@ def sorted_destination_plates
   def description_to_column_index(well_name, plate_size)
     Map::Coordinate.description_to_vertical_plate_position(well_name, plate_size)
   end
+
+  def column_index_to_description(index, plate_size)
+    Map::Coordinate.vertical_plate_position_to_description(index, plate_size)
+  end
```
KatyTaylor: Not important, but I personally wouldn't add an additional method just to rename a method - especially given you've already put a helpful comment on the original method!

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    return unless %w[1 on].include?(params[key])
```
KatyTaylor: You could consider clearing the value in `buffer_volume_for_empty_wells` if `automatic_buffer_addition` has been set to nil.

app/models/robot/generator/behaviours/tecan_default.rb
```diff
@@ -113,4 +116,88 @@ def sorted_destination_plates
...
+  #  {"destination" =>
+  #   {"SQPD-9101" =>
+  #     {"name" => "ABgene 0765",
+  #      "plate_size" => 96,
+  #      "control" => false,
+  #      "mapping" =>
+  #       [{"src_well" => ["SQPD-9089", "A1"], "dst_well" => "A1", "volume" => 100.0, "buffer_volume" => 0.0},
+  #        {"src_well" => ["SQPD-9089", "A2"], "dst_well" => "B1", "volume" => 100.0, "buffer_volume" => 0.0},
+  #        {"dst_well" => "C1", "buffer_volume" => 120.0}]},
+  #        {"dst_well" => "D1", "buffer_volume" => 120.0}]},
```
KatyTaylor, line 138: Neat way to add to the existing structure.

app/models/robot/generator/behaviours/tecan_default.rb
```diff
@@ -113,4 +116,88 @@ def sorted_destination_plates
...
+  def column_index_to_description(index, plate_size)
+    Map::Coordinate.vertical_plate_position_to_description(index, plate_size)
+  end
+
+  # Returns a new data object with buffer entries added for empty destination
+  # wells, if the option is enabled; otherwise returns the original data object.
+  # Only the fields used by the buffer steps are added to the new data object.
+  # @param data_object [Hash] the original data object
+  # @return [Hash] the new data object with buffer entries for empty wells,
+  #   or the original data object if the option is not enabled
```
KatyTaylor, line 113: Worth mentioning here I think you also cut out / do not re-add various bits of data (presumably not needed for the `buffers` method). 

For instance, the `source` object, and the `control` object (which I think is erroneously included in the example output).

app/models/robot/generator/behaviours/tecan_default.rb
```diff
@@ -113,4 +116,88 @@ def sorted_destination_plates
...
+        # Add buffer for empty well
+        mapping << {
+          'dst_well' => dst_well,
+          'buffer_volume' => buffer_volume_for_empty_wells
+        }
+      end
+      obj['destination'][dest_plate_barcode]['mapping'] = mapping
+    end
+    obj
+  end
```
KatyTaylor, line 153: Think this algorithm is good and works. It took me a while to understand so I asked ChatGPT to refactor it. Here's its suggestion, feel free to accept, adapt or reject! My opinion is that methods 1, 2 and 4 are an improvement, and 3 I'm not sure - includes slightly obscure stuff like `filter_map`.

Haven't tested it so may not work. Think returns something slightly different as in doesn't cut out the `source` object (but shouldn't matter).


```
def data_object_for_buffers(data_object)
  buffer_volume = @batch&.buffer_volume_for_empty_wells
  return data_object unless buffer_volume

  destinations = data_object['destination'].transform_values do |plate_details|
    build_destination_plate(plate_details, buffer_volume)
  end

  { 'destination' => destinations }
end

private

def build_destination_plate(plate_details, buffer_volume)
  plate_size = plate_details['plate_size']
  
  {
    'name' => plate_details['name'],
    'plate_size' => plate_size,
    'mapping' => build_mapping(plate_details['mapping'], plate_size, buffer_volume)
  }
end

def build_mapping(existing_mapping, plate_size, buffer_volume)
  plate = Plate.find_by_barcode(plate_details['name'])

  index_to_mapping = existing_mapping.index_by do |entry|
    description_to_column_index(entry['dst_well'], plate_size)
  end

  (1..plate_size).filter_map do |index|
    existing = index_to_mapping[index]
    next existing if existing

    dst_well = column_index_to_description(index, plate_size)
    next unless well_empty?(plate, dst_well)

    {
      'dst_well' => dst_well,
      'buffer_volume' => buffer_volume
    }
  end
end

def well_empty?(plate, dst_well)
  well = plate.find_well_by_name(dst_well)
  well.blank? || well.empty?
end
```

KatyTaylor (approved):
All just suggestions, although don't understand the `1, on` bit. Also I haven't reviewed the tests I'm afraid.

app/frontend/entrypoints/cherrypick_strategies.test.js
andrewsparkes: Thanks, works. Changed.

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
andrewsparkes: it's displayed, albeit twice for some reason:
Screenshot 2026-03-18 at 13 27 28

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
yoldas: In one case, the control is a visible UI while in the other case it is a hidden field, between page transitions. I understand, it can be set to '1' before writing but it would be still sent as 1 or on from the client. We just check if something is set in the value or not.

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
yoldas: The getter check the automatic_buffer_addition flag first to return nil or the buffer_volume_for_empty_wells value but clearing it will be safer as suggested.

app/models/robot/generator/behaviours/tecan_default.rb
yoldas: two types of buffer steps: buffer for empty wells (fill) and buffer for sample wells (top-up). Maybe it should just read: "Only the fields used by the buffer steps are **important** in the new data object"

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
andrewsparkes: probably as have 2 source plates in my batch

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
andrewsparkes: changed, will store '1' as database value

app/models/robot/generator/behaviours/tecan_default.rb
andrewsparkes: original Map method names were confusing, have renamed them to make more sense (I think) so then the renames not needed here

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
andrewsparkes: True, but is that possible? Can you go back to a cherrypick batch and edit parameters like that?

andrewsparkes: Added. Was easy enough to check if it existed and set it to nil.

app/models/robot/generator/behaviours/tecan_default.rb
andrewsparkes: Agree, added to comments.

app/models/robot/generator/behaviours/tecan_default.rb
andrewsparkes: refactored

app/models/cherrypick/task/buffer_volume_for_empty_wells_option.rb
```diff
@@ -5,21 +5,35 @@ def create_buffer_volume_for_empty_wells_option(params)
...
     end
-
     @batch.set_poly_metadata(key, params[key])
   end
+
+  # Most likely it's not set, but if it is, clear it to avoid confusion.
+  def clear_buffer_addition_volume(_params)
+    return if @batch.get_poly_metadata(:buffer_volume_for_empty_wells).blank?
+
+    @batch.set_poly_metadata(:buffer_volume_for_empty_wells, nil)
```
KatyTaylor, line 49: Actually, should we be deleting the record instead? I think you said in a different comment that the model validation prevents `nil` values... so this should blow up?

KatyTaylor: Ah please ignore, I remember that the set_poly_metadata method handles that.

app/models/robot/generator/behaviours/tecan_default.rb
```diff
@@ -110,7 +104,147 @@ def sorted_destination_plates
...
+  # @param opts [Hash] Options including :index_to_mapping, :plate, :index, :plate_size, :buffer_volume_for_empty_wells
+  # @return [void]
+  def mapping_or_buffer_entry(mapping, opts)
+    if opts[:index_to_mapping].key?(opts[:index])
+      mapping << opts[:index_to_mapping][opts[:index]]
+    else
+      buffer_entry = buffer_mapping_for_empty_well(opts[:plate], opts[:index], opts[:plate_size],
+                                                   opts[:buffer_volume_for_empty_wells])
+      mapping << buffer_entry if buffer_entry
+    end
```
KatyTaylor, line 212: Thank you for refactoring (and making the other changes).

I think it's generally more readable. Think this particular method `mapping_or_buffer_entry` is a little abstract and maybe I would just put this logic into the `build_buffer_mapping` method. 

Happy to approve anyway though.

StephenHulme:
I see Jest has been added as a dependency? Should we update the readme and add JS tests to the CI?

PR #5587, opened 2026-03-04
[ux] Add genders to internal accessioning validations

StephenHulme:
Addresses external validation errors like:
```
Sample: 61609112-fdcd-11f0-80eb-0242eb1badd5; Attribute: gender; Reason: must be equal to one of the allowed values: ["female","male","unknown"]; Failed to submit samples to BioSamples; No new BioSample was created for sample with alias 61609112-fdcd-11f0-80eb-0242eb1badd5
```

And prevents repeated accessioning notifications.

- Adds known-EGA-restricted genders: `EGA_GENDERS = %w[Female Male Unknown]`
- Validates their inclusion for EGA accessioned samples
- Should raise `Accession::InternalValidationError` instead of `Accession::ExternalValidationError` allowing faster validation and correction
- Reduces verbosity of accessioning validation error messages

### DIFF
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -25,6 +25,7 @@ class Sample < ApplicationRecord # rubocop:todo Metrics/ClassLength
 
   GC_CONTENTS = ['Neutral', 'High AT', 'High GC'].freeze
   GENDERS = ['Male', 'Female', 'Mixed', 'Hermaphrodite', 'Unknown', 'Not Applicable'].freeze
+  EGA_GENDERS = %w[Female Male Unknown].freeze
   DNA_SOURCES = [
     'Genomic',
     'Whole Genome Amplified',
@@ -193,7 +194,9 @@ class Sample < ApplicationRecord # rubocop:todo Metrics/ClassLength
     end
 
     with_options(on: :EGA) do
+      one_of_ega_genders = EGA_GENDERS.to_sentence(last_word_connector: ' or ', two_words_connector: ' or ')
       validates :gender, presence: { message: 'is required' }
+      validates :gender, inclusion: { in: EGA_GENDERS, message: "must be #{one_of_ega_genders}" }
       validates :phenotype, presence: { message: 'is required' }
       validates :donor_id, presence: { message: 'is required' }
     end
--- a/app/models/sample.rb
+++ b/app/models/sample.rb
@@ -194,7 +194,7 @@ class Sample < ApplicationRecord # rubocop:todo Metrics/ClassLength
     end
 
     with_options(on: :EGA) do
-      one_of_ega_genders = EGA_GENDERS.to_sentence(last_word_connector: ', or ', two_words_connector: ' or ').downcase
+      one_of_ega_genders = EGA_GENDERS.to_sentence(last_word_connector: ', or ', two_words_connector: ' or ')
       validates :gender, inclusion: { in: EGA_GENDERS, message: "must be #{one_of_ega_genders}" }
       validates :phenotype, presence: { message: 'is required' }
       validates :donor_id, presence: { message: 'is required' }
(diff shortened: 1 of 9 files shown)


app/models/sample.rb
```diff
@@ -193,7 +194,8 @@ class Current < ActiveSupport::CurrentAttributes
     end
 
     with_options(on: :EGA) do
-      validates :gender, presence: { message: 'is required' }
+      one_of_ega_genders = EGA_GENDERS.to_sentence(last_word_connector: ', or ', two_words_connector: ' or ').downcase
```
BenTopping: You might be able to use `en.yml` (I18n) to handle this.

lib/accession/sample.rb
```diff
@@ -137,5 +139,15 @@ def check_studies
         errors.add(:sample, 'can only be accessioned if linked to a releasable, accessioned study.')
       end
     end
+
+    def check_sample_for_service
+      service_context = service.ena? ? :ENA : :EGA
+      unless sample.valid?(service_context)
+        sample.sample_metadata.errors.each do |error|
+          # ignore the presence errors as they are already handled by missing_accession_tags
+          errors.add(:sample, "#{error.attribute} #{error.message}") unless error.type == :blank
```
BenTopping, line 148: Slightly confused whats happening here. Is this bubbling up the sample metadata errors to the sample level? If so I think rails should handle this automatically?

app/models/sample.rb
```diff
@@ -25,6 +25,7 @@ class Current < ActiveSupport::CurrentAttributes
 
   GC_CONTENTS = ['Neutral', 'High AT', 'High GC'].freeze
   GENDERS = ['Male', 'Female', 'Mixed', 'Hermaphrodite', 'Unknown', 'Not Applicable'].freeze
+  EGA_GENDERS = %w[Female Male Unknown].freeze
```
BenTopping, line 28: Is EGA case insensitive? The error has lowecase values.

BenTopping (commented):
Nice fix 👌

app/models/sample.rb
StephenHulme: Correct, the user is provided with a dropdown box in titlecase, but the lowercasing happens closer to the API level where the gender is specifically downcased.
Ah, I see what you mean regarding the error versus the input. Good spot, I'll remove the downcasing from the error message 👌

lib/accession/sample.rb
StephenHulme: It is bubbling the errors up to the sample level, but there is a specific handler for missing fields `does not have the required metadata: <missing-field 1>, <missing-field 2>, <missing-field 3>` instead of the default `<missing-field 1> is required, <missing-field 2> is required, <missing-field 3> is required`.

app/models/sample.rb
StephenHulme: It can, but it doesn't handle the list-building as well. I do agree that the to-sentence is a little cumber-some if not very flexible.
I can wrap the existing command in a function if that would be cleaner?

app/models/sample.rb
StephenHulme: Fixed in 382984350bce88e030c9fa52fb2a47aa9669da01

app/models/sample.rb
BenTopping: If its not simple then don't worry, I think this is fine just wanted to check you were aware of the en.yml use cases which you are.

lib/accession/sample.rb
BenTopping: Ah okay so its just a phrasing change for the errors? Happy to leave it if it improves readability for users 👍

StephenHulme: Exactly, yes 👍

BenTopping (approved):
Looks good to me

PR #5598, opened 2026-03-06
restrict tag_sequence to 30 characters

### LINKED ISSUE #5597: Y26-013 - [BUG] Fix flowcell dead letters on `ware-prod`
**Describe the bug**

Following the purging of dead letters in sanger/General-Backlog-Items#603, and the investigation of those dead letters in sanger/General-Backlog-Items#674 - the repeatable dead letter seems to appear on `ware-prod`: `{“flowcell”: …}`.

From some investigation, I believe the cause might be due to the message trying to create 1 flowcell, with many lanes, where there is an identical control in each lane. 

It might be the the position index should be part of the composite key, for a control, so that the duplicate controls would not be classed as duplicate. However, the position is not currently used in the validation of the control data object, within a flowcell message. 

To test this hypothesis, a position could be included in the control data object, or the Unified Warehouse code could be updated to use the lanes position when validating a control. 

It might be worth further investigation into whether the exact same control _should_ be allowed in two different lanes, on the same flowcell. If this is a valid scenario (and there are many other duplicate controls across lanes message which have made it to the MLWH successfully), then it would be worth instead looking into why only these messages have been dead lettered, and not other flowcell messages where duplicate controls exist across multiple lanes. 

Please see more info in the “flowcell” part of the comment [here](https://github.com/sanger/General-Backlog-Items/issues/674#issuecomment-3739420829)

**Associated Stories**
[Y25-399-2 - Investigate dead-letter messages since Y25-399](https://github.com/sanger/General-Backlog-Items/issues/new?template=bug_report.md#top)

**To Reproduce**
1. Visit `ware-prod` RabbitMQ Broker
2. Look in the `psd.multi-lims-warehouse-consumer.dead-letters` queue
3. Find a message with “flowcell” as the high level object

**Acceptance Criteria**
- [ ] Reproduce the dead letter
- [ ] Test the above hypothesis 
- [ ] Create a fix, likely in Sequencescape
- [ ] Resubmit flowcell messages and confirm flowcell data is in MLWH

**Additional context**

sabrine33:
Closes #5597


sabrine33:
I did not do the excel file validation, as that will miss up with the column i7 i5 warning story - validation can support warning or errors, this is should be fine, as the model validation will fail in case when needed

PR #5603, opened 2026-03-10
Accession samples in multiple studies

StephenHulme:
Permits accessioning of samples in multiple studies

- Permit accessioning of samples when belonging to many open (ENA) studies
- Refactor warning dialogs in templates

### DIFF
--- a/spec/models/sample_spec.rb
+++ b/spec/models/sample_spec.rb
@@ -107,11 +107,11 @@ RSpec.describe Sample, :cardinal do
   end
 
   describe '#should_be_accessioned?' do
-    let(:sample) { create(:sample) }
-    let(:study) { create(:study, samples: [sample]) }
+    let(:studies) { create_list(:study, 1) }
+    let(:sample) { create(:sample, studies:) }
 
     context 'when there is exactly one accessionable study' do
-      before { allow(sample).to receive(:studies_for_accessioning).and_return([study]) }
+      before { allow(sample).to receive(:studies_for_accessioning).and_return(studies) }
 
       it 'returns true' do
         expect(sample.should_be_accessioned?).to be true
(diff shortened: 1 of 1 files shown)


spec/models/sample_spec.rb
yoldas: can there be also a separate context where study2 is a managed study for testing the mixed case?

StephenHulme: Yes, good thinking, will add some more tests

StephenHulme: Fixed in d6d4bcb91d63cf58f95b5ea02069ab635575994d

app/helpers/samples_helper.rb
```diff
@@ -18,4 +18,51 @@ def samples_not_accessioned(samples)
...
+    )
+  end
+
+  def multiple_accessionable_studies_warning(sample)
+    # Samples belonging to more than one accessionables study can only be accessioned if all studies are open
+    return nil if sample.all_accessionable_studies_open?
+
+    unopen_studies = sample.studies_for_accessioning.reject { |study| study.study_metadata.open? }
+    accession_warning(
+      'Samples belonging to more than one study can only be accessioned if all accessionable studies are open:',
```
yoldas, line 49: If all of them are managed, they could be accessioned to EGA, because there is no mix of open and managed. This is a discussion point I think.

spec/models/sample_spec.rb
```diff
@@ -126,13 +126,92 @@
...
+        it 'should accession' do
+          expect(sample.should_be_accessioned?).to be true
+        end
+      end
+
+      context 'with two managed studies that are both accessionable' do
+        let(:studies) { [managed_with_accession_1, managed_with_accession_2] }
+
+        it 'should not be accessioned' do
+          expect(sample.should_be_accessioned?).to be false
```
yoldas, line 157: case: if all are managed, we could allow it.

app/helpers/samples_helper.rb
StephenHulme: EGA does have the extra complexity of having different data policies and contacts. But yes, technically there's no reason why not, but it's a changed of business logic that should be discussed.

spec/models/sample_spec.rb
StephenHulme: There are different policy and contact agreements for EGA studies that ENA studies do not have. If this is a desired change to the business logic, it should be relatively easy to implement.

app/helpers/samples_helper.rb
yoldas: We have decided to keep this for now.

spec/models/sample_spec.rb
yoldas: We have decided to keep this for now.

PR #5616, opened 2026-03-17
Y26-056 - Reduce search time

### LINKED ISSUE #5581: Y26-056 - Reduce the loading time of search results
**Describe the Housekeeping**
Reduce the loading time of search results.

**Initial analysis**
Recent user queries from production can be [found in Google sheets](https://docs.google.com/spreadsheets/d/1Wqz6v1L1-0A4jeEiMskSP5VGU67NoH4xXrHrdFqBE8I/edit?usp=sharing).
Page load times of 9+ seconds can be seen for general queries and samples. This is likely because there are a very large number of samples, and by using double-ended wildcard searches, the indexing cannot be used. There are likely to be other issues at play too.
It's also worth noting that the fields searched are not publicised, nor does the result list indicate why a sample was returned, if the search query was not in the sample name. Sample fields searched are: sample name, supplier name, accession number, and id. In the provided search data, no accession numbers are used as queries.

**Blocking issues**
None

**Additional context**
Discovered from Kibana logging

Image

[Selection of user queries](https://docs.google.com/spreadsheets/d/1Wqz6v1L1-0A4jeEiMskSP5VGU67NoH4xXrHrdFqBE8I/edit?usp=sharing)

BenTopping:
Closes #5581

- Adds new search option 'All (wildcard)' to maintain old search logic
- Default search option to 'All' (non leading wildcard search). This dramatically improves performance as we can make use of indexes. This only affects Sample and Labware searches.
- Adds index to labware name to improve search performance.
- Fixes bug where bad user-input types (URL based) causes 500.

### Notes

I have added an unused `_wildcard` to all `for_search_query` methods to keep the signatures the same even though its only used for Sample and Labware. It would be good to make use of this wildcard in the future as we are currently inconsistent with what is being wildcarded but that is out of scope for this story.

### DIFF
--- a/app/controllers/lab_searches_controller.rb
+++ b/app/controllers/lab_searches_controller.rb
@@ -11,10 +11,10 @@ class LabSearchesController < ApplicationController
   private
 
   def perform_search(query)
-    @batches = Batch.for_search_query(query, wildcard: true).to_a
+    @batches = Batch.for_search_query(query, leading_wildcard: true).to_a
     @assets =
       (
-        Labware.for_search_query(query, wildcard: true).for_lab_searches_display.to_a +
+        Labware.for_search_query(query, leading_wildcard: true).for_lab_searches_display.to_a +
           Labware.with_barcode(query).for_lab_searches_display.to_a
       ).uniq
   end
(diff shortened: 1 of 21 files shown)


BenTopping:
After discussion with @StephenHulme we decided not to abstract the wildcard/query layer to the search_behaviour class which would have added leading wildcard to all searches. This is because it would change the behaviour of some existing queries (adding wildcards) and cause issues with other queries (id queries) which is not intended in this PR.

StephenHulme (approved):
Thanks for the discussion and renaming :)

PR #5637, opened 2026-03-23
Y26 031   labware location report by retention instructions

### LINKED ISSUE #5532: Y26-031 - Labware Location Report by Retention Instructions
**Context**
The BioResource team currently has no scalable way to view labware storage requirements based on retention instructions. While retention and destruction data is now recorded, it can only be reviewed per manifest, making it time-consuming and impractical to assess long-term storage capacity needs across the system.

**User story**
As a BioResource team member, I want to generate a labware location report filtered by retention instructions, So that I can understand and plan capacity requirements for long-term sample storage efficiently.

Danni W

Danni W

**Acceptance criteria**
- [ ] A new selectable filter for Retention Instructions is added to the labware location report.
- [ ] The report supports filtering labware based on recorded retention instruction values.
- [ ] The report returns all matching labware with existing location data intact.

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

**Deployment notes**

wendyyang:
Closes #

add retention instruction column  in location report table
add filter by retention instruction on the return record


db/migrate/20260325134057_add_retention_instructions_to_lcation_reports.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddRetentionInstructionsToLcationReports < ActiveRecord::Migration[7.2]
```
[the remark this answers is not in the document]
wendyyang: fixed

app/models/location_report.rb
```diff
@@ -176,14 +176,30 @@ def generate_study_cols_for_row(cur_study)
...
   def search_for_labware_by_selection
-    params = { faculty_sponsor_ids:, study_id:, start_date:, end_date:, plate_purpose_ids:, barcodes: }
+    params = { faculty_sponsor_ids:, study_id:, start_date:, end_date:, plate_purpose_ids:, barcodes:,
+               retention_instructions: }
+    validate_result_size(params)
+
+    # Only plates and tubes are currently supported by this report
+    labwares = Labware.search_for_labware(params).filter { |labware| labware.is_a?(Plate) || labware.is_a?(Tube) }
+
+    filter_labware_by_retention_instructions(labwares)
```
BenTopping, line 186: I think this filter should be moved into the [search_for_labware](https://github.com/sanger/sequencescape/blob/develop/app/models/labware.rb#L158) function as a new scope. This would keep it consistent with the behaviour of the other fields in this selection form.

app/views/location_reports/_location_labware_selection_form.html.erb
```diff
@@ -31,6 +31,11 @@
     </div>
   </div>
 
+  <div class='form-group'>
+    <%= f.label 'retention_instructions', 'Retention instructions' %>
+    <%= f.select :retention_instructions, options_for_select(retention_instruction_option_for_select, location_report_form.retention_instructions), { include_hidden: false }, { class: 'form-control select2', :multiple => true } %>
```
BenTopping, line 36: Do we want users to be able to select multiple at once?

We should be able to change the options here so the labels are the readable versions (Long term storage) but the option values are the normalised long_term_storage. This would prevent us having to format them here https://github.com/sanger/sequencescape/pull/5637/changes#diff-96a48d35086a071258b6b8210f00ccf6a1092d9a4443c8c46d5c1a8765548a5aR228

db/migrate/20260325134057_add_retention_instructions_to_location_reports.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddRetentionInstructionsToLocationReports < ActiveRecord::Migration[7.2]
+  def change
+    add_column :location_reports, :retention_instructions, :text
```
BenTopping: I think string should be fine here as its a small column. If we are only allowing filter by one retention_instruction we could change this to retention_instruction (no 's') and change to an enum.

spec/models/location_report_spec.rb
```diff
@@ -473,6 +475,24 @@
 
           it_behaves_like 'a successful report'
         end
+
+        context 'with retention instructions' do
```
BenTopping, line 480: Test for multiple retention instructions if we are supporting it.

BenTopping (commented):
Couple questions about whether we need to support multiple retention_instructions. 

N.B. Haven't tested locally.

app/models/location_report.rb
wendyyang: refactored

spec/models/location_report_spec.rb
wendyyang: added

app/controllers/location_reports_controller.rb
```diff
@@ -60,8 +60,7 @@ def location_report_params # rubocop:todo Metrics/MethodLength
       :study_id,
       :start_date,
       :end_date,
-      :barcodes,
-      :barcodes_text,
```
StephenHulme, line 64: Are these meant to be removed?

spec/models/location_report_spec.rb
```diff
@@ -473,6 +450,47 @@
...
+            [
+              headers_line,
+              plt_1_line,
+              plt_3_line,
+              tube_1_line
+            ]
+          end
+
+          it_behaves_like 'a successful report'
+        end
```
StephenHulme, line 493: It's difficult to look at these tests and clearly see: "ah, yes there are only the lines returned that are in 2-year storage", etc.

Would it be possible to rename/create labwares like `plate_2_years`, `plate_longterm`, `tube_no_storage` (in a before block or with `let!` - not `let`, since it is lazily evaluated)? Then refactor the tests to clearly show that when all the retention-instructed-labwares are created but only return-after-2-years is added in the filter, that only the return-after-2-years labwares are shown in the report?

StephenHulme (commented):
Looks good in general, but a couple comments and questions.

In the specs, why has `plate_1_custom_metadatum` changed to `plate_1_retention_instruction`?

app/controllers/location_reports_controller.rb
StephenHulme: Ah, I see it is already in the list - but wasn't shown in the preview.

app/views/location_reports/_location_labware_selection_form.html.erb
wendyyang: refactored to remove the normalised.

wendyyang:
> Looks good in general, but a couple comments and questions.
> 
> In the specs, why has `plate_1_custom_metadatum` changed to `plate_1_retention_instruction`?

Because the retention instruction used to live in custom_metadata table, but now move to labware table. so the custom_metadata no longer needed.

spec/models/location_report_spec.rb
wendyyang: The plates are shared in other tests, naming the plate with retention instruction indicator is not good, because initially the plate created has no retention instruction, it is updated later on. 
e.g 
let(:plate_1_set_long_term_storage) { plate_1.update!(retention_instruction: :long_term_storage) }

If creating separate plates just for those 2 relevant contexts, it seems duplicated.  So instead, I rename the update method with retention value, e.g plate_1_set_long_term_storage, plate_3_return_after_two_years, tube_1_set_long_term_storage. Hope it is clearer to show what retention value is set to that plate.

StephenHulme: That is clearer, thanks very much. I can now see that the filter is working as it should be.

StephenHulme:
> > Looks good in general, but a couple comments and questions.
> > In the specs, why has `plate_1_custom_metadatum` changed to `plate_1_retention_instruction`?
> 
> Because the retention instruction used to live in custom_metadata table, but now move to labware table. so the custom_metadata no longer needed.

Was the old column deleted? If so, how did this test pass before the change was made? Or was the column move made in an earlier PR and these tests were not updated?

wendyyang:
> > > Looks good in general, but a couple comments and questions.
> > > In the specs, why has `plate_1_custom_metadatum` changed to `plate_1_retention_instruction`?
> > 
> > 
> > Because the retention instruction used to live in custom_metadata table, but now move to labware table. so the custom_metadata no longer needed.
> 
> Was the old column deleted? If so, how did this test pass before the change was made? Or was the column move made in an earlier PR and these tests were not updated?

It wasn't a column, it is a table called 'custom_metadata' , the table has 'key' & 'value' column, so can store different type of data, the retention instruction was stored there using key 'retention_instruction'. Since the table 'custom_metadata' schema/columns not changed, so the test still works, just not returning any data, I think.

StephenHulme:
Ah that makes a lot more sense, thanks - I'd forgotton that that table used a key-value architecture.

StephenHulme (approved):

PR #5650, opened 2026-03-30
Y26-105 - [PR] [Bug] useq_wafer dead-letters on ware-prod

### LINKED ISSUE #715: Y26-105 - [Bug] useq_wafer dead-letters on ware-prod
There are several dead-letter messages related to the `useq_wafer` table (4 in total, as of 26 March), associated with a couple of batches (batch IDs: 107499, 107608). For each batch, two messages are being sent, as each Ultima batch contains two requests.

**Actions required:**  
- [x] Investigate the root cause of the dead-letter messages  
- [x]  Implement a fix to prevent messages from being routed to the dead-letter queue  
- [x] Re-broadcast the affected messages to MLWH  
- [x] Let NPG know once the messages are fixed and warehouse is updated

yoldas:
Closes #

- Added descriptor input validation in Descriptor#validate_value for required fields and Date values (YYYY-MM-DD, real date, sane year range).
- Updated Tasks::SetDescriptorsHandler::Handler to validate descriptor inputs before processing and return validation errors.
- Updated WorkflowsController#stage redirect logic so users stay on the same task page when validation fails (instead of being sent to the batch page).
- Added/updated tests in descriptor_spec.rb and handler_spec.rb.


app/models/descriptor.rb
```diff
@@ -3,6 +3,9 @@ class Descriptor < ApplicationRecord
   belongs_to :task
   serialize :selection, coder: YAML
 
+  DATE_YEAR_MIN = 1990
```
BenTopping, line 6: Is this to stop cases like the dead letter where we have 62026 as the year?

app/models/tasks/set_descriptors_handler.rb
```diff
@@ -8,6 +8,9 @@ def render
     end
 
     def perform
+      errors = validate_descriptor_inputs
```
BenTopping, line 11: Suggestion: Can we use rails validations here. Make use of the active record errors behaviour and add errors to the task model by running task.valid?. You would need to set validations on descriptor too for this to work I think.

app/models/descriptor.rb
```diff
@@ -11,4 +14,40 @@ def matches?(search)
     search.descriptors.each { |descriptor| return true if descriptor.name == name && descriptor.value == value }
     false
   end
+
+  # Returns an array of validation errors for the submitted descriptor value.
+  # The value comes from the Task Details form for a workflow task on a batch.
+  # @return [Array] An array of error messages, empty if the value is valid
+  def validate_value(submitted_value)
```
BenTopping, line 21: Suggestion: Can you move these to active model validations so you can call descriptor.valid?

BenTopping (commented):
Changes look okay, just a question about moving the validation to the models.

app/models/tasks/set_descriptors_handler.rb
yoldas: Task and Descriptors define the workflow. No instances of them are saved after form submit; the submitted values are serialised into LabEvent. It may look like descriptors or a task is being validated but we are not validating them; we are validating submitted values that are described by them before they are serialised into LabEvent fields in SetDescriptorsHandler's perform method.

descriptor.valid? or task.valid? would mean whether the workflow task definition and descriptor definition (features of parameter) is done right for setting up the pipeline definition.

I am looking into something like DescriptorValue and TaskValue to be able to implement the suggestion.

app/models/descriptor.rb
yoldas: Yes it is a quick sanity check between 1990 and 2100 . I thought a simple check would be enough for this PR. The alternative would be finding out the nature of these expiration dates (there are several different types of expiry fields) and validate them but it may be a blocker if it is not done right.

app/models/descriptor.rb
yoldas: Explained in the comment for the validation added to the perform method.

app/models/tasks/set_descriptors_handler.rb
BenTopping: Oh interesting. Is this something we should validate in LabEvent then? Happy with the current implementation though so up to you.

yoldas: I looked into doing it on LabEvent but it is a generic model which has more than 50 different types of event description values. It would cause if statement to check if it is Opentrons and then somehow find out what the serialised descriptors field content is about.

I think the current place of validations is the cleanest. I would like to keep them.

BenTopping: Sure, sounds good to me. Thanks for looking into it.

BenTopping (approved):

PR #5657, opened 2026-03-31
Y25-687 - Aviti additional information

### LINKED ISSUE #5361: Y25-687 - Capture Additional Information in the Aviti Pipeline
**Context**
The Aviti pipeline does not currently capture key information; specifically the quantification method used and whether a custom primer kit was used. This makes the team to manually track these details in spreadsheets.

**User story**
As a user, I want to record the quantification method used and whether a custom primer kit was used directly within the Aviti pipeline, so that all required information is captured in one system, reducing manual spreadsheets and enabling accurate reporting and cross-charging.

Tris B

Tris B

**Acceptance criteria**
- [ ] A “Quant method used” dropdown is added prior to the loading page in the Aviti pipeline with the options:
       - Tapestation
       - Tapestation & qPCR

- [ ] A “Custom primer kit used” toggle (Yes/No) is added to the Aviti loading page.
- [ ] Both fields must be mandatory before progressing to the next step.

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

**Deployment notes**

BenTopping:
Closes #5361

Warehouse PR: https://github.com/sanger/unified_warehouse/pull/888

- Adds 'Custom primer kit used' to aviti batch tasks.
- Adds 'Quant method used' to aviti batch tasks.


app/models/api/messages/eseq_flowcell_io.rb
```diff
@@ -0,0 +1,77 @@
...
+      with_association(:sample) { map_attribute_to_json_attribute(:uuid, 'sample_uuid') }
+      with_association(:study) { map_attribute_to_json_attribute(:uuid, 'study_uuid') }
+      with_association(:primer_panel) { map_attribute_to_json_attribute(:name, 'primer_panel') }
+      with_association(:library) { map_attribute_to_json_attribute(:external_identifier, 'id_library_lims') }
+      map_attribute_to_json_attribute(:aliquot_type, 'entity_type')
+    end
+
+    # The following methods come from the Aliquot model or the relevant module above.
+    # They are included in the MLWH message under 'controls'.
+    with_nested_has_many_association(:controls) do
```
BenTopping, line 66: Im unsure if element aviti actually has controls? I didnt want to cause regression as the previous flowcell_io template does but I cant see a way to add phix on batch creation for aviti.

config/default_records/descriptors/002_element_aviti_descriptors.yml
```diff
@@ -12,6 +23,17 @@ Final Loading Concentration (pM):
...
+  name: Custom primer kit used
+  task: Loading
+  workflow: Element Aviti
+  kind: Selection
+  value: Yes
+  selection:
+    Yes: Yes
+    No: No
+  required: true
+  sorter: 1
```
KatyTaylor, line 36: Not very important, but I noticed this displays as true/false rather than yes/no

Image

app/models/api/messages/eseq_flowcell_io.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+  end
+
+  renders_model(::Batch)
+
+  map_attribute_to_json_attribute(:id, 'flowcell_id')
+  map_attribute_to_json_attribute(:updated_at)
+
+  with_nested_has_many_association(:requests, as: :lanes) do
+    map_attribute_to_json_attribute(:position, 'lane')
```
KatyTaylor, line 44: Just noting, I see you've added the `'lane'` (compared to flowcell_io) - but also updated unified_warehouse to match so all OK. Got to remember these 2 deployments are very coupled.

app/models/api/messages/eseq_flowcell_io.rb
KatyTaylor, line 1: Did a quick comparison with flowcell_io - removes a lot of stuff - needs careful testing in UAT to make sure nothing necessary was removed, and that the records created in eseq_flowcell look consistent with previous ones.

KatyTaylor (approved):
Haven't checked unified_warehouse side yet

config/default_records/descriptors/002_element_aviti_descriptors.yml
BenTopping: Ah yes good spot. Ill update this config to reflect true/false instead of stringifying the Yes / No as it means we can handle this as a boolean throughout instead of a string.

BenTopping: Made this explicit here: [e41293d](https://github.com/sanger/sequencescape/pull/5657/commits/e41293d7f06f6abdd16428dd790a79492293bf2a). As you said I dont think they will mind using true/false instead of yes/no. I did try it with selection displayed as 'Yes'/'No' with 'True/false' values but it felt like unnecessary obfuscation. Can always implement it if they want it after UAT.

app/models/api/messages/eseq_flowcell_io.rb
BenTopping: Yes we were translating in the warehouse before but that was a hack to let us use the flowcell_io template temporarily.

app/models/api/messages/eseq_flowcell_io.rb
BenTopping: Yep lots of redundant data being sent and ignored from the warehouse. I have done some e2e testing locally and all looks okay. Will do more on UAT.

app/models/api/messages/eseq_flowcell_io.rb
KatyTaylor: I don't feel like I understand the context here, but...
```
SELECT * 
FROM mlwhd_mlwarehouse_proddata.eseq_flowcell
WHERE entity_id_lims = 'control_lane'
ORDER BY id_eseq_flowcell_tmp DESC;
```
returns 0 rows, whereas it returns many for iseq_flowcell.

KatyTaylor: Unless I'm looking at the wrong thing, latest dates for control lanes in iseq_flowcell is 2015, and control requests in SS is 2018. So possibly an outdated concept anyway.

```
SELECT * 
FROM mlwhd_mlwarehouse_proddata.iseq_flowcell
WHERE entity_id_lims = 'control_lane'
ORDER BY id_iseq_flowcell_tmp DESC;
```
```
SELECT * 
FROM sequencescape_snapshot.requests
WHERE sti_type = 'ControlRequest'
ORDER BY id DESC;
```

KatyTaylor: Summary: don't think the controls association here is required.

BenTopping: Yes I came to the same conclusion.
After speaking with Tris they don't have an explicit way of linking controls to lanes for this pipeline (unlike other pipelines that have the action 'added spiked buffer'). They use the descriptor fields to record the barcodes which are sent to the warehouse.

I think its safe to leave this control data here as is as it may be useful in the future and provides no harm at present.

yoldas (approved):

PR #5659, opened 2026-04-01
[ux] Limit of 10 messages per manifest upload

StephenHulme:
Fix for [#INC-59595](https://sanger.freshservice.com/a/tickets/59595?current_tab=details)

- Limit messages arising in manifest uploads

#### Screenshot

Screenshot 2026-04-01 at 16 37 24

### DIFF
--- a/app/controllers/sample_manifest_upload_with_tag_sequences_controller.rb
+++ b/app/controllers/sample_manifest_upload_with_tag_sequences_controller.rb
@@ -43,15 +43,13 @@ class SampleManifestUploadWithTagSequencesController < ApplicationController
   def apply_warning_flash(rows) # rubocop:disable Metrics/AbcSize
     # There are more than 10 messages, truncate the list and add a note about it.
     # This is to prevent the flash from becoming too large and causing issues with cookies.
-    if rows.size > 10
+    row_messages = rows.flat_map { |row| row.warnings.full_messages }.uniq
+    if row_messages.size > 10
       # IfUnlessModifier doesn't improve readability
-      rows = rows.first(10) + ["and #{rows.size - 10} more..."]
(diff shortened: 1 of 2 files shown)


yoldas (approved):
For logging the long flash message items that could not be shown because of cookie error. To give it a try.

yoldas (approved):

app/controllers/sample_manifest_upload_with_tag_sequences_controller.rb
```diff
@@ -40,11 +40,16 @@ def rows_with_warnings
     end
   end
 
-  def apply_warning_flash(rows)
-    flash[:warning] = {
-      'Sample manifest uploaded with warnings!':
-        rows.flat_map { |row| row.warnings.full_messages }.uniq
-    }
+  def apply_warning_flash(rows) # rubocop:disable Metrics/AbcSize
+    # There are more than 10 messages, truncate the list and add a note about it.
```
BenTopping, line 44: Looks like we have a concern for [flash_truncation](https://github.com/sanger/sequencescape/blob/develop/app/controllers/concerns/flash_truncation.rb) don't think it will show the number of hidden messages though...

BenTopping (approved):

app/controllers/sample_manifest_upload_with_tag_sequences_controller.rb
StephenHulme: True... it might be a better solution with some re-working for the dict format and including the hidden messages note. I think we also need to re-measure how much the space the encryption payload takes up as the `flash_trucation` method failed some months ago on a unrelated issue...
Long term, I think a better storage mechanism might be the way forward.

Issue #5052, opened 2025-07-07
Y25-411 - As a TL (Tris B) I would like SequenceScape to generate a Sample Sheet to load into the Ultima UG 1000 sequencer so that we can run it


sabrine33:
**User story**

As a TL (Tris B) I would like SequenceScape to generate a Sample Sheet to load into the Ultima UG 1000 sequencer so that we can run it.
Unlike how it is done in Aviti, the sample manifest needs to be generated **per request** (not per batch).
**Who are the primary contacts for this story**
Mike Q
Tris B

**Who is the nominated tester for UAT**
Mike Q
Tris B

**Acceptance criteria**

- [x] The sample manifest (csv file) is generated on user request per pool.
- [x] sample manifest name should be uniquely identifiable (file name : batchId_NT_number)
- [x] Adding study_id column to the existing column (study_id per sample)
- [x] A button to download the generated file is displayed on the Ultima UG 1000 sequencing batch page.
- [x] In the header section (free text), add the batchId and the pool number
- [x] in the Global section (free text), add the following hardcoded values, Application(WGS native gDNA), sequencing_recipe(UG_116cycles_Baseline_1.8.5.2) and analysis_recipe(wgs1) - The issue is that these values need to be updated following the Ultima software upgrade.
- [x] be able to download the file through an url without auth. (Request from NPG team)

**Additional context**
The Ultima UG Confluence page contains examples that have been used on the sequencer with the tagset we need to support.
Also, please refer to the official Ultima official [documentation](https://cdn.sanity.io/files/l7780ks7/production-2024/4ca366aea0f7ac74a70c39897f870479ddeaf3dc.pdf) 

For the first phase; library type =  Ultima High Throughput PCR Free 96, the application type will always be set to native (WGS native gDNA)

In the future, we may add a feature to send the generated file to SampleDB once the Ultima API is available.

References
z-index related story: 
https://github.com/sanger/limber/issues/2588
This story has a non-blocking relationship with:
- #5041

BenTopping:
Note from Ultima tags:

The sample sheet contains the column `Index_Barcode_Num`
This is currently not an attribute of tags. We believe we can dynamically generate this with a combination of tag.map_id and plate number.
If this proves impractical we can add barcode_num as an attribute of tags and backfill the Ultima tag sets.

yoldas:
[...]

PR #5663, opened 2026-04-02
Y26-106 - [PR] UG200 sample sheet generator

### LINKED ISSUE #5647: Y26-106 - UG200 sample sheet generator
**User story**

The UG200 sequencer software uses an upgraded version (Rev. 02 | March 11, 2026) of the UG100 sequencer software. As a result, the sample sheet format differs slightly. The main differences are in the Global section.

Tris 

Tris

**Acceptance criteria**
Update the Ultima sample sheet generator to support the new version. We also need to continue supporting the current version until the UG100 sequencer software is upgraded.
- [x] For the Ultima UG200 pipeline, the generated file should contain only the Application field in the Global section, set to “WGS Native”
- [x] The Index_Barcode_Num values should run from 193 onwards for the new tag plates (created in https://github.com/sanger/sequencescape/issues/5620), as listed in the tag set files on the referenced Confluence page.
- [ ] ~For the Ultima UG100 pipeline, users should still be able to generate the sample sheet as before (same format as previously)~ Update after meeting with Ultima 1/4/26: no longer necessary.

**References**

Please check this: 
https://github.com/sanger/sequencescape/pull/5370

_Please also refer to the UG200 Confluence page (sample sheet section) for more information._

yoldas:
Closes #

- Added UG200-specific sample sheet generator:
  - Introduces UG200 global section (Application only, value WGS Native).
- Refactored base Ultima generator for configurability
  -   Replaced hardcoded constants with config methods.
  -   Added explicit tag-group config with plate_num and z_start (P1/P2/P3/P4).
  -   Updated tag index mapping to use configured start offsets.
  -   Added sorting by configured plate order and extracted sample_row_for method.
- Updated batch sample sheet downloads: 
  - UG200 can download without login.
  - Added shared ZIP send helper
- Added placeholders for models
  - ultima_ug200_sequencing_pipeline.rb
  - ultima_ug200_sequencing_request.rb
- Tests
  - UG200 global section + tag mapping (Z0193, Z0289, plate mapping).
  - UG200 sample sheet download on batch controller

These changes are backward compatible.


app/controllers/batches_controller.rb
```diff
@@ -389,22 +389,23 @@ def find_batch_by_barcode
   #
   # @return [Boolean] true if download is allowed, false otherwise
   def allow_sample_sheet_download?
-    @batch.pipeline.is_a?(UltimaSequencingPipeline) || logged_in?
+    return true if @batch.pipeline.is_a?(UltimaUG200SequencingPipeline)
+    return true if @batch.pipeline.is_a?(UltimaSequencingPipeline)
+
+    logged_in?
```
KatyTaylor, line 399: Question:

I realise you haven't touched this logic... but I don't get the `logged_in?` part - it seems like we're allowing sample sheet download for non-Ultima pipelines, as long as the user is logged in?

app/controllers/batches_controller.rb
```diff
@@ -389,22 +389,23 @@ def find_batch_by_barcode
...
-      generate_element_aviti_sample_sheet
-    elsif @batch.pipeline.is_a?(UltimaSequencingPipeline)
-      generate_ultima_sample_sheet
-    else
-      flash[:error] = 'Sample sheet generation is not supported for this pipeline.'
-      redirect_to controller: 'batches', action: 'show', id: @batch.id
-    end
+    return generate_element_aviti_sample_sheet if @batch.pipeline.is_a?(ElementAvitiSequencingPipeline)
+    return generate_ultima_ug200_sample_sheet if @batch.pipeline.is_a?(UltimaUG200SequencingPipeline)
+    return generate_ultima_sample_sheet if @batch.pipeline.is_a?(UltimaSequencingPipeline)
```
KatyTaylor, line 414: Because of the inheritance of the pipeline classes, I guess this would break if you swapped around lines 404 and 405 - perhaps a little fragile?

app/controllers/ultima_sample_sheet/sample_sheet_generator.rb
```diff
@@ -196,29 +218,30 @@ def study_id_for(aliquot)
...
-    # the first tag in the second tag group.
+    # numbers. Each group's starting index is taken directly from
+    # ultima_tag_groups_config[:z_start], so numbering is independent of which
+    # other tag groups exist in the database.
     # @return [Hash{Tag => Integer}] mapping of tags to index numbers
     def tag_index_map
-      @tag_index_map ||= begin
-        tags = ultima_tag_groups.flat_map { |tg| tg.tags.sort_by(&:map_id) }
-        tags.each_with_index.to_h { |tag, i| [tag, i + 1] }
+      @tag_index_map ||= ultima_tag_groups.each_with_object({}) do |tg, map|
```
KatyTaylor: Nitpick: rename the `tg` here to `tag_group`, as I read it as an abbreviation of `tag`.

app/controllers/ultima_sample_sheet/ug200_sample_sheet_generator.rb
```diff
@@ -0,0 +1,34 @@
...
+    private
+
+    # Adds the global section to the CSV for UG200.
+    # The request parameter is currently unused.
+    # @param csv [CSV] the CSV object to append rows to
+    # @param _request [UltimaSequencingRequest] the request whose global data is to be added
+    def add_global_section(csv, _request)
+      csv << pad(global_title_config)
+      csv << pad(global_headers_config)
+      data = ['WGS Native'] # Application
```
KatyTaylor, line 30: So the base class / UG100 will use the db record, and this is hardcoded? Seems OK for now but will probably want a refactor in future (when more sample sheet changes come along for future versions / library prep pipelines).

KatyTaylor (approved):
Approach looks good I think. This way we can have 2 versions, one for each pipeline, and switch between them with an easy deployment.

Left a few comments and questions, nothing blocking.

app/controllers/batches_controller.rb
yoldas: One of the requirements of UG100 was allowing NPG to access sample sheets without login (https://github.com/sanger/sequencescape/issues/5052). I have assumed the same for UG200 and added that. I have not touched the others (there was no request to open it for Element Aviti). The method checks if you can call generate_sample_sheet without login (UG100 and UG200); otherwise you must be logged in.

app/controllers/batches_controller.rb
yoldas: I have added a note to the method docs:
```
# @note When checking the pipeline class, always check the most specific
#   subclass first in the inheritance hierarchy, before considering parent
#   classes.
```

app/controllers/ultima_sample_sheet/ug200_sample_sheet_generator.rb
yoldas: Yes, there will be a refactor in the future.

For UG100 we had a database table for easy maintenance of the baseline value mainly. We had all global section values in a table and assumed data-driven customisation. I think, feature upgrades will not have baseline in the globals section and it will be the same as UG200; hence we will not need the table.

For UG200, there is a single value for the Application global value and we set it in the overridden method. Based on the new spec document, there will be preset values corresponding to pipelines. In that case, we would be setting the value based on request metadata, submissions, etc. I think.

app/controllers/batches_controller.rb
KatyTaylor: Thanks for the explanation. I'd mistakenly assumed that this method was controlling whether the link was displayed or not. Makes more sense now.

PR #5688, opened 2026-04-13
fix(labwhere-client): handle request timeouts

### LINKED ISSUE #5685: Y26-128 - [BUG] LabwhereClient handle timeout exception
**Describe the bug**
Several errors for RestClient::Exceptions::OpenTimeout occurred in labwhere_receptions#create during a period where labwhere was inaccessible. This error was not being caught and caused a 500 error for users instead of informing them that labwhere was down.

**To Reproduce**
1. Go to page 'labwhere_receptions'
2. Attempt to scan in labware with a hanging labwhere instance.
    - May be difficult to recreate. Potentially add a binding.pry in labwhere will do this.

**Expected behaviour**
User is shown an error that labwhere is inaccessible

BenTopping:
Closes #5685

- Handle request timeouts to prevent 500 errors when labwhere is unresponsive.

### DIFF
--- a/lib/lab_where_client.rb
+++ b/lib/lab_where_client.rb
@@ -26,13 +26,13 @@ module LabWhereClient
 
     def get(instance, target)
       parse_json(RestClient.get(path_to(instance, target)))
-    rescue Errno::ECONNREFUSED, RestClient::NotFound => e
+    rescue Errno::ECONNREFUSED, RestClient::NotFound, RestClient::RequestTimeout => e
       raise LabwhereException.new(e), 'LabWhere service is down', e.backtrace
     end
 
     def post(instance, target, payload)
       parse_json(RestClient.post(path_to(instance, target), payload))
-    rescue Errno::ECONNREFUSED, RestClient::NotFound => e
+    rescue Errno::ECONNREFUSED, RestClient::NotFound, RestClient::RequestTimeout => e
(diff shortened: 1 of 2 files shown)


lib/lab_where_client.rb
```diff
@@ -26,21 +26,21 @@ def parse_json(str)
 
     def get(instance, target)
       parse_json(RestClient.get(path_to(instance, target)))
-    rescue Errno::ECONNREFUSED, RestClient::NotFound => e
+    rescue Errno::ECONNREFUSED, RestClient::NotFound, RestClient::RequestTimeout => e
```
BenTopping, line 29: RestClient combined with our scan error setup has pretty ugly error handling. Ideally we would catch everything [RestClient::RequestFailed](https://github.com/rest-client/rest-client?tab=readme-ov-file#exceptions-see-httpwwww3orgprotocolsrfc2616rfc2616-sec10html) and utilise the response message but unfortunately this has a few nasty edge cases so leaving it as a simple addition for this bug,

StephenHulme: In previous restclient testing I also had cases where a no-connection resulted in a 500 error being returned. I dug through the code for a couple hours and decided it was easier to convert it to Faraday. I would recommend that path for future work.

StephenHulme (approved):

PR #5698, opened 2026-04-16
ci: only run test ci on push on develop/master

BenTopping:

- Prevents running CI on every non trunk push.
    - This will prevent duplicate actions running when we push and have a PR open. (half our CI usage for SS)

### DIFF
--- a/.github/workflows/cucumber_tests.yml
+++ b/.github/workflows/cucumber_tests.yml
@@ -17,6 +17,9 @@ env:
 
 on:
   push:
+    branches:
+      - develop
+      - master
   pull_request:
     types:
       # defaults
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,6 +2,9 @@ name: Lint
 
 on:
   push:
+    branches:
+      - develop
+      - master
   pull_request:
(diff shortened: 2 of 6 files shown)


KatyTaylor (approved):

StephenHulme:
I agree that there are too many and long checks that are a significant blocker to development.
I was going to object, but seeing how you've implemented removing the `push` checks for feature branches seems like a really good comprise.
(I would suggest we only do this where tests are long, and on an exception basis, rather than make it standard across out repos.)

StephenHulme (approved):

BenTopping:
> I agree that there are too many and long checks that are a significant blocker to development. I was going to object, but seeing how you've implemented removing the `push` checks for feature branches seems like a really good comprise. (I would suggest we only do this where tests are long, and on an exception basis, rather than make it standard across out repos.)

Yes I think this is the standard approach from some other open source repos I checked. There shouldnt be any noticeable changes for devs (unless people push and check CI without a PR?).

KatyTaylor:
> (unless people push and check CI without a PR?).

We've got a team process where we make a draft PR during development - so shouldn't be any need for that.

PR #5717, opened 2026-04-22
Y26-094 (part I) - Improve accession failure handling

StephenHulme:
Refactors and fixes some of the underlying error handling in accessioning to pave the way for more user-friendly notifications.

Requirement for part II: #5718

- Refactor accessioning error notification handling
  - Shorten error messages to prevent repetition
  - Humanise fields mentioned in error messages
  - Create `Accession::InvalidFieldsError` specifically for internal field/tag validation failures
  - Formalise handling of "_No new objects can be added with MODIFY action._" errors as `Accession::ExternalNumberConflictError`
  - Exception notifications are only sent on unexpected external failures
- Prevents retires when accessioning fails due to internal validation errors
- Minor formatting update to sample page
- Replace Faraday stubs with Webmock stubs
- Update feature flag descriptions


app/jobs/sample_accessioning_job.rb
```diff
@@ -69,6 +76,14 @@ def failure(_job)
       abort_accession_status
     end
 
+    # Set attempts to max-reties to prevent further attempts
+    def prevent_retries!
+      return unless @delayed_job
+
+      @delayed_job.attempts = @delayed_job.max_attempts + 1
```
BenTopping, line 83: Are there any helper methods we can use instead or is this the recommended way to force cancel a job.

lib/accession/sample.rb
```diff
@@ -113,11 +120,15 @@ def check_required_fields
       # EBI will still perform its own validation on submission.
       return if Flipper.enabled?(:y25_714_skip_accessioning_tag_validation)
 
-      unless tags.meets_service_requirements?(service, standard_tags)
-        errors.add(:sample, "does not have the required metadata: #{tags.missing.sort.to_sentence.dasherize}.")
+      unless missing_tags.empty?
```
BenTopping: NIT: Wondering if these should be `missing_accession_tags`? to avoid naming confusion with tags/barcodes in our world.

spec/lib/http_clients/ena_taxa_client_spec.rb
```diff
@@ -3,6 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe HTTPClients::ENATaxaClient do
+  # TODO: replace Faraday test stubs (tests middleware) with WebMock (tests endpoints)
```
BenTopping, line 6: Still TODO?

app/jobs/sample_accessioning_job.rb
StephenHulme: After much searching, I couldn't find anything for delayed_job (there are a couple suggestions for active_job, but I don't think they apply). I tried to reduce max_attempts too, but it's not saved in the database and the struct is transient.
Any suggestions welcome.

lib/accession/sample.rb
StephenHulme: That seems like a good idea - thanks, will rename.

spec/lib/http_clients/ena_taxa_client_spec.rb
StephenHulme: Yup, but in a different PR. It's out of scope for this set of changes.

lib/accession/sample.rb
StephenHulme: Fixed in 565f1ed98d547525f3845f9eb0484efeba7f2bad

spec/lib/http_clients/ena_taxa_client_spec.rb
StephenHulme: Fix in https://github.com/sanger/sequencescape/pull/5721

BenTopping (approved):

app/jobs/sample_accessioning_job.rb
yoldas: Isn't there a fail! method, which should set failed_at ? https://github.com/collectiveidea/delayed_job/blob/master/lib/delayed/backend/base.rb

Job selection looks like checking failed_at IS NULL
https://github.com/collectiveidea/delayed_job_active_record/blob/master/lib/delayed/backend/active_record.rb

lib/accession/sample.rb
```diff
@@ -44,17 +44,24 @@ def title
     # If the sample is invalid, logs an error message and raisesAccession::InternalValidationError with
     # details of the validation errors.
     #
+    # @raise [Accession::InvalidFieldsError] if the sample is missing required fields for accessioning
     # @raise [Accession::InternalValidationError] if the sample is not valid for accessioning
-    def validate!
+    def validate! # rubocop:disable Metrics/AbcSize
       return if valid?
```
yoldas, line 50: note: I find these combination of names confusing: return if valid? inside validate! 
validate! is checking errors collection, formatting messages, and raising exceptions with them.

yoldas (approved):

app/jobs/sample_accessioning_job.rb
StephenHulme: Thanks for spotting this @yoldas. There is a `fail!` method that works for failing a job, but it doesn't abort it and still schedules it for retrying.

lib/accession/sample.rb
StephenHulme: It isn't ideal... essentially we are validating an `Accession::Sample` which is validated by checking the original `::Sample` along with associated Studies. The error feedback from the Rails `valid?` is to add messages to a record.`errors` object. This method raises the errors instead.

PR #5734, opened 2026-04-28
repurpose lysate plate with no parents

### LINKED ISSUE #5594: Y26-067 - Bioscan - Update Legacy plate purpose
**User story**
All old Lysate plates that came direct from faculty (as opposed to ones made from LILYS insect plates) should be re-purposed from `LBSN-96 Lysate` to `LBSN-96 Lysate Input` (i.e. any `LBSN-96 Lysate` plate that has no LILYS parents). 

Lena

Should not need UAT.

**Acceptance criteria**

- [ ] All 'LBSN-96 Lysate' plates that have no parents should be repurposed to 'LBSN-96 Lysate Input' plates

**Dependencies**

References
- https://github.com/sanger/limber/issues/2726
(if this story is done then when you view an 'Input' version of a Lysate plate that has previously been through the pipeline, you should see the Bioscan Library prep submission button).

**Additional context**
Background: 
It used to be that Bioscan had only the one type of Lysate plate purpose. This purpose was used whether the plate was created as a child of the `LILYS-96 Stock` plate, or as a direct faculty made Lysate plate. We have since split this into two purposes, `LBSN-96 Lysate` and `LBSN-96 Lysate Input`, similar to other pipelines like scRNA. Both purposes allow for semi-automatic Library prep submission, but this is not working properly for repeat runs where the legacy Lysate plate has no parents.

So this story is about identifying those `LBSN-96 Lysate` plates in the database that have no parents, and changing their purpose to `LBSN-96 Lysate Input`. One with parents (LILYS plate parents) stay as `LBSN-96 Lysate`.

**Deployment notes**
Nothing special

sabrine33:
Closes #5594 

- 

I did not add tests as it is quite simple operation for one run time task


lib/tasks/repurpose_lysate_plate.rake
```diff
@@ -0,0 +1,27 @@
...
+    raise "PlatePurpose 'LBSN-96 Lysate Input' not found" unless lbsn_lysate_input_purpose
+
+    lbsn_lysate_plates = Plate.where(plate_purpose_id: lbsn_lysate_purpose.id)
+
+    plates_without_parents = lbsn_lysate_plates.select { |plate| plate.parents.empty? }
+
+    puts "Found #{plates_without_parents.size} plates to repurpose"
+
+    plates_without_parents.each do |plate|
+      plate.update(plate_purpose_id: lbsn_lysate_input_purpose.id)
```
yoldas: update just returns true or false, which would need checking; changing it to update! would be more suitable which would raise the exception rather than ignoring in case an exception happens.

yoldas (changes requested):

yoldas (approved):

PR #5739, opened 2026-04-29
deprecate LRC Bank input tube purpose

### LINKED ISSUE #5602: Y26-075 - Inactivate Legacy LRC Bank Input Tube Purpose
**Context**
A recent update to the scRNA Core cDNA Prep pipeline introduced two new tube purposes: LRC Bank Input 0.8ml and LRC Bank Input 1ml to distinguish between tube types. The previous tube purpose “LRC Bank Input” is now obsolete but still appears as a selectable option when creating manifests. This has led to incorrect selections, which blocked sample processing in Limber until manual intervention was completed. The legacy option should be removed or inactivated to prevent future errors and ensure correct reporting of library types.

**User story**
As a lab user creating manifests, I want the legacy LRC Bank Input tube purpose to be removed or inactivated, So that only the valid tube purposes (LRC Bank Input 0.8ml and LRC Bank Input 1ml) can be selected, preventing incorrect submissions and processing delays.

Abby C

Abby C

**Acceptance criteria**
- [ ] The LRC Bank Input tube purpose is removed or marked as inactive so it cannot be selected when creating new manifests.
- [ ] Only LRC Bank Input 0.8ml and LRC Bank Input 1ml are available for selection for this workflow.
- [ ] Existing historical records using LRC Bank Input remain unchanged and accessible.
- [ ] The updated tube purpose configuration supports downstream reporting (e.g., MLWH/Tableau) without impacting existing data.

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

**Deployment notes**

sabrine33:
Closes #5602 

- We can filter the acceptable_purposes list by LRC Bank Input in the Ruby model, but updating the schema would be preferable as it enforces data integrity and provides a more scalable foundation for future extensions.
- Did not add tests as not much logic is done here.


lib/tasks/deprecate_lrc_bank_input_purpose.rake
```diff
@@ -0,0 +1,20 @@
...
+# # For reference: Y26-075
+
+namespace :LRC_BANK_INPUT do
+  desc 'Deprecate LRC Bank Input tube purpose'
+  task deprecate_lrc_bank_input_purpose: :environment do
+    lrc_bank_input_purpose = Tube::Purpose.find_by(name: 'LRC Bank Input')
+
+    unless lrc_bank_input_purpose
+      puts "Tube Purpose 'LRC Bank Input' not found. No action taken."
+      next
```
KatyTaylor, line 13: Had to look up to check, but looks like `next` exits from blocks, so would exit from this rake task. 👍

app/models/sample_manifest/sample_tube_behaviour.rb
```diff
@@ -20,7 +20,7 @@ def io_samples
     end
 
     def acceptable_purposes
-      Tube::Purpose.where(target_type: SampleTube)
+      Tube::Purpose.where(target_type: SampleTube, deprecated: false)
```
KatyTaylor, line 23: I tested it locally as I wasn't sure if it would cover all manifest types, but it looks good.

KatyTaylor (approved):

lib/tasks/deprecate_lrc_bank_input_purpose.rake
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    lrc_bank_input_purpose.update!(deprecated: true, deprecated_at: DateTime.now)
+
+    puts 'Done: LRC Bank Input tube purpose is been deprecated.'
```
yoldas, line 18: typo: "is been" should be "has been"

db/migrate/20260429114103_add_deprecated_fields_to_plate_purposes.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+class AddDeprecatedFieldsToPlatePurposes < ActiveRecord::Migration[8.0]
+  def change
+    add_column :plate_purposes, :deprecated, :boolean, default: false, null: false
+    add_column :plate_purposes, :deprecated_at, :datetime
```
yoldas, line 5: suggestion: Our code base has already two separate deprecation conventions:

1. Through `SharedBehaviour::Deprecatable` which has a `deprecated_at` field (with a backing column), and a helper method `deprecated?` and an `active` scope to check deprecation. It is used by the `Product` model.
2. Through a `deprecated` column, an `active` scope, and rails-provided method for the boolean field `deprecated?`. It is used by the `RequestType` model.

This PR goes further and adds them both on the `Purpose` model. It only uses the deprecated column without scoping. We could just pick one of the existing approaches and use it. I would go for the first because it encapsulates the logic and pluggable.

app/models/sample_manifest/sample_tube_behaviour.rb
yoldas, line 23: todo: there are no tests covering the exclusion of the deprecated purposes.
SampleManifest::SampleTubeBehaviour::Core could be 'described' to assert core.acceptable_purposes using a purpose with deprecated as true/false for example.

lib/tasks/deprecate_lrc_bank_input_purpose.rake
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# # One-time task run to deprecate the LRC Bank Input tube purpose.
+# # For reference: Y26-075
+
+namespace :LRC_BANK_INPUT do
```
yoldas, line 6: nitpick: namespace naming is unconventional, it could be `lrc_bank_input` .

db/migrate/20260429114103_add_deprecated_fields_to_plate_purposes.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+class AddDeprecatedFieldsToPlatePurposes < ActiveRecord::Migration[8.0]
```
yoldas, line 2: note (informational; no action necessary): The migration name refers to the plate_purposes table, which backs the Purpose model and has a type column as discriminator. The model is superclass of purposes such as PlatePurpose, Tube::Purpose, and TubeRack::Purpose, etc.

lib/tasks/deprecate_lrc_bank_input_purpose.rake
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# # One-time task run to deprecate the LRC Bank Input tube purpose.
+# # For reference: Y26-075
```
yoldas, line 4: nitpick: double hash looks like a commented out code comment.

yoldas (commented):
Yes, it excludes the LRC Bank Input plate purpose from sample manifests because all Purposes that govern SampleTubes (target_type) are excluded based on the deprecated flag. 

As it is already approved, I am just adding some comments.

app/models/sample_manifest/sample_tube_behaviour.rb
sabrine33: > todo: there are no tests covering the exclusion of the deprecated purposes. SampleManifest::SampleTubeBehaviour::Core could be 'described' to assert core.acceptable_purposes using a purpose with deprecated as true/false for example.

I did think about that, but I chose not to implement it because the logic is fairly minimal and it would likely clutter the tests more than add value.

PR #5763, opened 2026-05-07
Y26-058 - QC report download 

### LINKED ISSUE #5588: Y26-058 - Improve QC Report Download Flexibility in Sequencescape
**Context**
Currently, QC reports in Sequencescape can only be downloaded by Study ID, either for all plates within the study or only plates not previously reported. SSRs perform QC reporting almost daily, and this limitation requires downloading large reports or multiple separate files, followed by manual filtering. The team would like to explore whether QC reports can instead (or additionally) be requested by specific Plate IDs.

**User story**
As an SSR, I want to request QC reports in Sequencescape using specific Plate IDs (in addition to or instead of Study ID), So that I can download only the data relevant to my work and reduce manual filtering and processing time.

Liz C

Liz C

**Acceptance criteria**
- [ ] The system supports requesting QC reports by one or more Plate IDs.
- [ ] The report request interface allows input of multiple Plate IDs (e.g., comma-separated or selectable list).
- [ ] The system validates Plate IDs and provides error messaging for invalid or non-existent IDs.
- [ ] QC reports generated by Plate ID contain the same data fields and format as current Study-based reports.

19Mar2026 Note:

 - Need to look at the scenario where user uses  a combination of the PlateID filter and 'exclude samples with existing reports' option or 'Plate Type' option.

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

**Deployment notes**

BenTopping:
Closes #5588 

- Adds plate_barcodes option to qc_report page.
- Removes requirement for study to be present on qc_reports if plate_barcodes are present.

#### Notes
Clarified with Liz C these plate barcodes do not need to appear in the header section of the report or in the filterable section at the bottom of the qc_reports page.


app/views/qc_reports/_qc_report_form.html.erb
```diff
@@ -11,6 +11,14 @@
...
     <%= f.select :plate_purposes, plate_purposes, { prompt: 'Select plate purpose...', disabled: [''], selected: 'Stock Plate'}, { autocomplete: 'off', class: 'form-control select2', multiple: 'multiple'  } %>
   </div>
+  <div class="form-group">
+    <%= f.label "plate_barcodes_text", 'Plate barcodes' %>
+    <%= text_area_tag 'qc_report[plate_barcodes_text]',
+                      params.dig(:qc_report, :plate_barcodes_text),
+                      cols: 40,
+                      rows: 4,
+                      class: 'form-control' %>
+  </div>
```
StephenHulme, line 22: Is it worth including a label/placeholder/text showing the expected format? Ie: one per line, or comma separated?

spec/models/qc_report_spec.rb
```diff
@@ -199,4 +227,32 @@
       expect(qc_report.qc_metrics.count).to eq(3)
     end
   end
+
+  context 'limit by plate barcodes' do
+    attr_reader :qc_report
+
+    let!(:study) { create(:study) }
```
StephenHulme:
```suggestion
    let(:study) { create(:study) }
```

It's already referenced in `before` so will be called anyway

spec/models/well_spec.rb
```diff
@@ -867,4 +867,187 @@
...
+    end
+
+    it 'returns an empty array if no wells contain samples without empty_supplier_sample_names' do
+      aliquot_2.sample.update(empty_supplier_sample_name: true)
+      expect(described_class.without_blank_samples).to eq([])
+    end
+  end
+
+  describe '#qc_report_in_batches' do
+    let!(:study) { create(:study) }
```
StephenHulme: Same as above `let!` not required

spec/models/well_spec.rb
```diff
@@ -867,4 +867,187 @@
...
+        nil
+      ) { |wells| wells_count += wells.length }
+      expect(wells_count).to eq(3)
+
+      wells_count = 0
+      described_class.qc_report_in_batches(study, false, 'Bespoke RNA', [purpose_2.name, purpose_3.name],
+                                           nil) do |wells|
+        wells_count += wells.length
+      end
+      expect(wells_count).to eq(2)
```
StephenHulme, line 994: What is being checked here?

The `wells_count +=` seems a bit, repetative... Would some of the [rspec yields matchers](https://rspec.info/features/3-12/rspec-expectations/built-in-matchers/yield/) make this easier to understand? 

Repeated below as well.

StephenHulme (changes requested):
Couple comments, but looks good 👌

spec/models/well_spec.rb
BenTopping: We are checking the block yields the expected number of wells. The yield matchers unfortunately don't cover this case.

spec/models/well_spec.rb
BenTopping: Good spot. Updated in [e6d130b](https://github.com/sanger/sequencescape/pull/5763/commits/e6d130b505d009366764a2a7cdb6b6cc1c56e23e). Surprised rubocop doesn't flag that.

spec/models/qc_report_spec.rb
BenTopping: Updated in [e6d130b](https://github.com/sanger/sequencescape/pull/5763/commits/e6d130b505d009366764a2a7cdb6b6cc1c56e23e).

spec/models/well_spec.rb
StephenHulme: Thanks for looking into it

app/views/qc_reports/_qc_report_form.html.erb
BenTopping: Updated in [e6d130b](https://github.com/sanger/sequencescape/pull/5763/commits/e6d130b505d009366764a2a7cdb6b6cc1c56e23e). It can be any tab / whitespace but not commas, I guess we could include commas after splitting...

StephenHulme (approved):
Thanks for the changes 🚀

PR #5769, opened 2026-05-12
Extend 'under-represented wells' feature to other pipelines

sabrine33:
- update the logic behind construct under_rep comments to include all submission types.
- also update the logic behind comparing aliquots
- Tested the fix on training on the batches with wrong data (explained [here](https://github.com/sanger/unified_warehouse/issues/910)) and it returned the expected/right values.

Closes #

- 


andrewsparkes (approved):
Looks good. Clear comments, thorough tests, thanks.

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -39,51 +48,56 @@ def initialize(poly_metadatum:, batch_id:, position:, tag_index:)
...
-  #
+  # Builds all `UnderRepWellComment` objects for the batch by iterating over
+  # each sequencing request and its associated lane.
+  # Aggregates comments across  all lanes and their ancestor plates.
   # @return [Array<UnderRepWellComment>] a flat list of comment objects.
-  #
   def under_represented_well_comments
-    request_with_under_represented_wells.flat_map do |request|
-      build_comments_for_request(request)
+    requests.flat_map do |batch_request|
```
KatyTaylor:
```suggestion
    requests.flat_map do |sequencing_request|
```

Think it's confusing to call this variable `batch_request` since that there is actually a different model / table called `batch_request`, and this is a `request`.
(Also applicable to naming in a couple of other methods).

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -39,51 +48,56 @@ def initialize(poly_metadatum:, batch_id:, position:, tag_index:)
...
+  # library requests in the lane's ancestor plates, then building a comment for
+  # each associated poly_metadatum.
   #
-  # @param request [Request] the request to process
-  # @return [Array<UnderRepWellComment>] comments built for the given request
-  #
-  def build_comments_for_request(request)
-    under_represented_poly_metadata(request).flat_map do |poly_meta|
-      build_comments_for_poly_meta(request, poly_meta)
+  # @param lane [Lane] the lane asset associated with the batch request
```
KatyTaylor:
```suggestion
  # @param lane [Lane] the lane receptacle associated with the batch request
```
What do you get if you do `lane.class`? `Receptacle`? If so, let's use that word here instead of `asset` to avoid using the confusing terminology.

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -39,51 +48,56 @@ def initialize(poly_metadatum:, batch_id:, position:, tag_index:)
...
+  # containing an `under_represented` poly_metadata entry. Covers both
+  # single-submission pipelines (e.g. WGS) and multi-submission pipelines
+  # (e.g. RNA) where sequencing and library prep belong to different submissions.
+  #
+  # @param lane [Lane] the lane whose ancestor plates are traversed
+  # @return [Array<Request>] requests containing under-represented metadata
+  def under_rep_requests_for_lane(lane)
+    lane.ancestors.grep(Plate)
+      .flat_map(&:wells)
+      .flat_map(&:requests)
```
KatyTaylor: Above line will return duplicate Requests, as multiple Plates in a pipeline point to the same Request. Might also get some `nil`s, as some Plates (e.g. first ones in a Pipeline, or ones that don't require Requests) will not point to a Request. Could do a `.uniq.compact`

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -39,51 +48,56 @@ def initialize(poly_metadatum:, batch_id:, position:, tag_index:)
...
-  # @return [Array<UnderRepWellComment>] comments built for the given request
-  #
-  def build_comments_for_request(request)
-    under_represented_poly_metadata(request).flat_map do |poly_meta|
-      build_comments_for_poly_meta(request, poly_meta)
+  # @param lane [Lane] the lane asset associated with the batch request
+  # @param batch_request [Request] the sequencing request whose position is used in the comment
+  # @return [Array<UnderRepWellComment>] comments built for the given lane
+  def build_comments_for_lane(lane, batch_request)
+    under_rep_requests_for_lane(lane).flat_map do |library_request|
```
KatyTaylor, line 96: Could mention in the comment that we are assuming it's a library_request as that is the known use case. The way it's retrieved does not enforce it.

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -96,45 +110,49 @@ def under_represented_poly_metadata(request)
...
   #
-  # @param request [Request] the request that owns the poly_metadatum
-  # @param poly_meta [PolyMetadatum] the metadata record for under-represented wells
-  # @return [Array<UnderRepWellComment>] list of constructed comment objects
-
-  def build_comments_for_poly_meta(request, poly_meta)
-    request.target_asset.aliquots.filter_map do |aliquot|
-      lane = request.asset.descendants.last
-      next unless aliquot_matches_lane?(lane, aliquot)
+  # @param library_request [Request] the library request whose target_asset well holds the aliquots
```
KatyTaylor, line 134: This is a limitation in the original design. Since we are storing the polymetadata on request, not well, it only works for plates that have a request 'finishing' in them. If you try to mark a well on an 'intermediate' plate as under-represented, this will assume that it came from a different plate. Think we need to state this explicitly somewhere and maybe code in some guards against it.

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -96,45 +110,49 @@ def under_represented_poly_metadata(request)
...
+  # @param poly_meta [PolyMetadatum] the under-represented metadata record to associate with the comment
+  # @return [Array<UnderRepWellComment>] comments built for each matching aliquot
+  def build_comments(library_request, lane, batch_request, poly_meta)
+    library_request.target_asset.aliquots.filter_map do |aliquot|
+      matching = find_matching_lane_aliquot(lane, aliquot)
+      next unless matching
 
       UnderRepWellComment.new(
-        position: lane.source_request.position,
+        position: batch_request.position,
```
KatyTaylor: I think this is the `Request`, not the `Batch Request`, so you should pass through the Batch Request instead to get the `position`.

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -96,45 +110,49 @@ def under_represented_poly_metadata(request)
...
-  # Checks if the given aliquot’s tag map ID matches any aliquot in the lane.
+  # Determines whether the given aliquot matches an aliquot on the lane
+  # A match requires equality on:
+  #   - `sample_id`
+  #   - `tag_id`
+  #   - `tag2_id`
+  #   - `tag_depth`
+  #
+  # This comparison is used to associate request aliquots with the
+  # corresponding lane aliquots in order to retrieve the correct tag index for the comment.
```
KatyTaylor:
```suggestion
  # This comparison is used to associate aliquots that were marked as 'under-represented' with their
  # downstream lane aliquots in order to retrieve the correct tag index for the comment.
```

app/models/under_rep_well_comments_to_broadcast.rb
```diff
@@ -96,45 +110,49 @@ def under_represented_poly_metadata(request)
...
   #
   # @param lane [Asset] the lane asset derived from the request
   # @param aliquot [Aliquot] the aliquot being compared
-  # @return [Boolean] true if the aliquot matches a lane aliquot by index value
-  #
-  def aliquot_matches_lane?(lane, aliquot)
-    lane.aliquots.any? { |a| a.aliquot_index_value == aliquot.tag.map_id }
+  # @return [Aliquot, nil] the matching lane aliquot, or nil if no match is found
+  def find_matching_lane_aliquot(lane, aliquot)
```
KatyTaylor, line 167: Could use the `equivalent?` method on `Aliquot`. There's also a `matches?` method, but it allows the upstream aliquot to be untagged, which is less strict.

KatyTaylor (changes requested):
A few things to consider. 

Would be good to test it on at least one plate from each of the mentioned pipelines.

app/models/under_rep_well_comments_to_broadcast.rb
sabrine33: @KatyTaylor  - they are actually batch_requests (which can be/usually are sequencing requests). As we are reading from the model Batch: https://github.com/sanger/sequencescape/blob/fd134ceed3db90ef29b5cd82e526a9001e3738a0/app/models/batch.rb#L44

app/models/under_rep_well_comments_to_broadcast.rb
sabrine33: it does say lane receptacle, I would like to use the term 'lane' as 'receptacle' can be anything

app/models/under_rep_well_comments_to_broadcast.rb
sabrine33: > There are actually batch requests type, that is why we can the read the position

There are actually batch requests type, that is why we can the read the position

sabrine33:
> A few things to consider.
> 
> Would be good to test it on at least one plate from each of the mentioned pipelines.

 I tested on the RNA pipeline but also with different scenarios

app/models/under_rep_well_comments_to_broadcast.rb
KatyTaylor: `Batch.last.requests.first.class` 
=> `HiSeqSequencingRequest` (a type of `Request`)

`Batch.last.batch_requests.first.class`
=> `BatchRequest` (not a type of `Request` - a join table)

app/models/under_rep_well_comments_to_broadcast.rb
sabrine33: position can be accessed directly from the sequencing request as it's joined to batch request,
`batch.requests.first.position `or `batch.request.first.batch_request.position `

KatyTaylor: OK I hadn't noticed that `Request` delegates `position` to `BatchRequest` in `app/models/batch/request_behaviour.rb`, so that's how it works 👍 

Just some variable naming to make a bit clearer then.

KatyTaylor (approved):

PR #5780, opened 2026-05-14
add labware updated_at value and return the latest one

### LINKED ISSUE #5761: Y26-149 - [BUG] Show correct location, when LabWhere has entries under both human and machine barcodes
**Describe the bug**
If a labware has been scanned into LabWhere using both the human and the machine barcodes, Sequencescape does not always show the latest location, but instead shows the latest location associated with the machine barcode. See Fresh Service ticket for example data.

**FreshSerivce/RT Ticket Number**
#SR-62535 Storage locations not matching

**To Reproduce**
1. Go to labware page in Sequencescape for one of the plates in the Fresh Service ticket
2. See incorrect location displayed

**Expected behaviour**
Location displayed should be the most recent location, when taking into account entries under machine and human barcodes.

**Additional context**
```
# app/models/labware.rb
def lookup_labwhere_location
  lookup_labwhere(machine_barcode) || lookup_labwhere(human_barcode)
end
```
Instead of the above, it should lookup location for both machine and human barcodes. If there are locations for both, take the one scanned most recently.

See [Barcodes: 1D vs 2D, and 'machine' vs 'human'](https://ssg-confluence.internal.sanger.ac.uk/spaces/PSDPUB/pages/290916019/Barcodes+1D+vs+2D+and+machine+vs+human) for context around human / machine barcodes (although mostly about tubes).

wendyyang:
Closes #

add labware updated_at in API return and return the latest location

- 


app/models/labware.rb
```diff
@@ -378,9 +388,23 @@ def lookup_labwhere(barcode)
     rescue StandardError => e
       # rescue LabWhereClient::LabwhereException => e
       Rails.logger.error { e }
-      return 'Not found - There is a problem with Labwhere'
+      return { error: true }
```
StephenHulme: Could we maybe return a message as well? I don't think `true` provides enough feedback.

app/models/labware.rb
```diff
@@ -368,8 +368,18 @@ def obtain_retention_instructions
...
+
+    valid_lookups = [machine_lookup, human_lookup].compact.select { |lookup| lookup[:location].present? }
+    return valid_lookups.max_by { |lookup| lookup[:updated_at] || Time.zone.at(0) }[:location] if valid_lookups.any?
+
+    return 'Not found - There is a problem with Labwhere' if [machine_lookup, human_lookup].any? do |lookup|
+      lookup&.dig(:error)
+    end
+
+    nil
   end
```
StephenHulme, line 378: Would it be possible to reduce the complexity of this function (possibly by pulling out some private methods)? 

It's difficult to understand exactly what the new code does.

spec/lib/lab_where_client_spec.rb
```diff
@@ -83,4 +83,26 @@
...
+          'updated_at' => 'Tuesday June  6 2023 16:26'
+        }
+      }
+    end
+
+    it 'captures labware and location updated_at values from the API payload' do
+      labware = described_class.new(params)
+
+      expect(labware.updated_at).to eq('Wednesday May  6 2026 10:46')
+      expect(labware.location.updated_at).to eq('Tuesday June  6 2023 16:26')
```
StephenHulme: This date format is nice for humans, but non-standard and difficult for computers to parse. May I suggest that the updated_at fields are always represented as [ISO 8601](http://en.wikipedia.org/wiki/ISO_8601) standard strings?

spec/models/location_report_spec.rb
```diff
@@ -200,6 +200,10 @@
     end
 
     describe 'report generation' do
+      before do
+        allow(LabWhereClient::Labware).to receive(:find_by_barcode).and_return(nil)
+      end
```
StephenHulme: Is this mocking out the actual code we are testing? Or is it not important to this test?

StephenHulme (commented):
Some comments and suggestions : )

spec/lib/lab_where_client_spec.rb
wendyyang: updated.

spec/models/location_report_spec.rb
wendyyang: re-write to include human barcode

app/models/labware.rb
wendyyang: refactored.

app/models/labware.rb
wendyyang: added error messages in return

StephenHulme (approved):
Thanks for the changes

yoldas (approved):

PR #5821, opened 2026-05-29
Y26-166 - Element aviti adds cost_code to warehouse flowcell message

### LINKED ISSUE #5809: Y26-166 - As Head of Short Read Sequencing (Rich R), I would like the cost code for Aviti Sequencing to be sent to the MultiLIMS Warehouse so that an automated billing report can be generated.
**User story**
As Head of Short Read Sequencing (Rich R), I would like the cost code for Aviti Sequencing to be sent to the MultiLIMS Warehouse so that an automated billing report can be generated.

TW
Rich R
NPG

PSD can test this internally.

**Acceptance criteria**
To be considered successful, the solution must allow:
- [x] The cost code is stored in the eseq_flowcell table in a column named 'cost_code' of type varchar(20).
- [x] The cost_code column should have a column comment 'Valid WTSI cost code'
- [x] The cost_code column should be NULLABLE.
- [x] The cost_code will be populated with data recorded in SequenceScape for each Aviti Element sequencing run.
- [x] Where no cost code is used for a sequencing run NULL will be stored in the cost_code column. [The messages did not include cost_code; hence warehouse side will have nulls for existing records - they will be back-filled if flowcell records are rebroadcasted]

**Additional context**
The cost code for Aviti Element sequencing should be sent to the MLWH in a similar manner to the current Aviti Element data and the Illumina data, which already carries the cost code.
The eseq_flowcell table is already populated by data coming from SequenceScape through the unified_warehouse RabbitMQ messages. These messages will need to now include the cost code and SequenceScape, unified_warehouse, and the database schema updated to support this.

No historical backfill of cost_codes is required.

This is enables /sanger/unified_warehouse/839 which will produce the report once the cost code is present in the MLWH.

NPG should be notified of the changes to the eseq_flowcell table when starting this story.

BenTopping:
Closes #5809 

Relates to unified warehouse PR: https://github.com/sanger/unified_warehouse/pull/913

- Adds cost_code to unified warehouse message for eseq_flowcell.


app/models/api/messages/flowcell_io.rb
```diff
@@ -163,6 +163,7 @@ def control_aliquot_type
   end
 
   # Included in Project model
+  # This is also used in the eseq flowcell message
   module ProjectExtensions
     def project_cost_code_for_uwh
       project_cost_code.length > 20 ? 'Custom' : project_cost_code
```
yoldas, line 169: length call would fail on a null cost code but it is never null on SS side. Update on eseq flowcell will cause backfilling warehouse side.

Update: There is no DB constraint on SS side but there is a validation to enforce cost code in project metadata; so no concern the length call.

yoldas (approved):
There is no DB constraint on SS side but there is a validation to enforce cost code in project metadata; so no concern the length call. Updating existing flow cells will cause backfilling warehouse side.

ProjectExtensions is nicely attached to eseq flowcell IO as well and explained.

PR #5834, opened 2026-06-02
[dev] Add SuperDiff to display differences on rspec failures

StephenHulme:
Because this diff is unhelpful:
```
expected: [["Message", "Content", "Created at", "Created by"], ["Created by Sample Manifest", "Manifest_1", "Mo...010 10:25", ""], ["Updated by Sample Manifest", "Manifest_1", "Monday 12 July, 2010 10:25", "jane"]]
     got: [["Message", "Content", "Created at", "Created by"], ["Created by Sample Manifest", "Manifest_1", "Mo...010 10:25", ""], ["Updated by Sample Manifest", "Manifest_1", "Monday 12 July, 2010 10:25", "jane"]]
```

Interestingly, there is already an old rspec issue for this: https://github.com/rspec/rspec/issues/94

I am a little concerned about the performance impact of superdiff though, will investigate further...

- Add [super_diff](https://github.com/splitwise/super_diff) to display differences on rspec expectation failures
- Is only `require`d and run in the local development environment and not in CI due to performance issues


StephenHulme:
Sadly this might be a non-starter... some unit-test jobs are taking 40-50 minutes which matches up with the 4x increase mentioned in https://github.com/splitwise/super_diff/issues/139

StephenHulme:
Looking at the profiling in the tests, it looks like `it { is_expected.not_to be_valid }` and variations are the primary culprits...

Although `spec/models/work_order_spec.rb` runs quickly on my machine - weird.

BenTopping (approved):
Looks sensible to me.
Its not noticeable slower locally so disabling for CI seems a good compromise. I suspect few people run full test suites locally anyway.

BenTopping:
Also worth noting I think an alternative could be to set the [max_formatted_output_length](https://github.com/rspec/rspec-expectations/blob/main/lib/rspec/expectations/configuration.rb#L70)
 to nil in rspec. But we wouldnt get the nice obvious formatting this gem provides.

PR #5837, opened 2026-06-04
[dev] Move overly nested modules to lib

StephenHulme:
A large amount of excel-related code is stored in deeply-nested duplicate folders and doesn't follow the expected Rails standard of spec locations relative to the files they test, making it difficult to quickly navigate between code and spec.
This follows the [RSpec/SpecFilePathFormat rubocop](https://docs.rubocop.org/rubocop-rspec/latest/cops_rspec.html#rspecspecfilepathformat).

Here is the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=thadeu.vscode-run-rspec-file) I use to quickly switch between code and test (`RSpec: Run Open spec this file`), I'm sure there are similar ones for other IDEs.

- Move `sample_manifest_excel/`, `sequencescape_excel/`, and `bulk_submission_excel/` to `lib/`
- Clean up some minor test issues that became apparent
- Update rubocop todo


BenTopping (approved):
I agree with the sentiment, I have always thought those nested folders were a pain. Nice to properly modularise it.
My only gripe is some of these files have behaviour specific to the app, use of SS models / validations etc, so not sure if that belongs in lib or not.

Good to get another pair of eyes before merging too because I don't often work on this area.

StephenHulme:
> I agree with the sentiment, I have always thought those nested folders were a pain. Nice to properly modularise it. My only gripe is some of these files have behaviour specific to the app, use of SS models / validations etc, so not sure if that belongs in lib or not.
> 
> Good to get another pair of eyes before merging too because I don't often work on this area.

There's an [argument for either direction](https://clayshentrup.medium.com/what-goes-in-rails-lib-92c74dfd955e) of should it be in lib or not: in this case I'm interpreting it as 'not part of core model-view-controller functionality'. If we had (or wanted) an app/importers directory, we could put it under there - but I'm not certain that there would be a clear improvement in structure.

sabrine33 (commented):
Looks okay to me ! thanks

sabrine33 (approved):

PR #5841, opened 2026-06-08
Fix Codecov GPG signature verification failure in GitHub Actions workflows

sabrine33:
The "RSpec tests (4, 0)" CI job was failing at the Codecov upload step (not the test step — all 1855 examples passed). The root cause was a GPG signature verification failure in the Codecov CLI binary download.

## Root Cause

The shared `sanger/.github` codecov action is pinned to `codecov/codecov-action@v5.5.5`. On April 21, 2026, Codecov rotated their GPG signing key after deleting the `codecovsecurity` keybase account. The old action version cannot verify the re-signed binary, causing:

```
gpg: Can't check signature: No public key
==> Could not verify signature. Please contact Codecov if problem continues
```

## Changes Made

All 5 workflow files updated to use `codecov/codecov-action@v7` directly (released June 7, 2026), bypassing the shared action pinned to the broken version:

- `.github/workflows/rspec_unit_tests.yml`
- `.github/workflows/rspec_feature_tests.yml`
- `.github/workflows/rake_tests.yml`
- `.github/workflows/cucumber_tests.yml`
- `.github/workflows/vitest.yml`

Input names updated from hyphen-separated (`disable-search`) to underscore-separated (`disable_search`) to match the action's native input format. `v7` uses the `codecovsecops` account with a working GPG key.

### DIFF
--- a/.github/workflows/cucumber_tests.yml
+++ b/.github/workflows/cucumber_tests.yml
@@ -101,10 +101,11 @@ jobs:
           retention-days: 5
 
       - name: Upload coverage reports to Codecov
-        uses: sanger/.github/.github/actions/tests/codecov@master
+        uses: codecov/codecov-action@v7.0.0
         with:
           name: ${{ github.run_id }}_${{ github.job }}_${{ matrix.ci_node_index }}_${{ github.event_name }}
           token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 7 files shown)


StephenHulme:
Closed in favour of https://github.com/sanger/.github/pull/41

PR #5870, opened 2026-06-18
Lock Sapio Study in SS

### LINKED ISSUE #5824: Y26-172 - (Lock Studies) As an architect (Rich L), I want Sapio Studies to be visible in SequenceScape but not editable there, so that lab work can proceed while Sapio stays the authoritative source — removing the complexity of keeping both systems in sync.
**User story**
As an architect (Rich L), I want Sapio Studies to be visible in SequenceScape but not editable there, so that lab work can proceed while Sapio stays the authoritative source — removing the complexity of keeping both systems in sync.

Tom W
Rich L

PSD can test this internally.

**Acceptance criteria**
This functionality will be feature-flagged,  turn on when Sapio phase 1 goes live. 

- [x] The Study information page for a Sapio study will show a banner "This study is externally managed in Sapio".
- [x] SequenceScape users will **not** be able to edit ANY fields of Sapio study.
- [x] SequenceScape users will **not** be able to change a Sapio study's state e.g. activate it or deactivate it.
- [ ] SequenceScape users will **not** be able to register samples in SequenceScape against a Sapio study.
- [x] SequenceScape users **cannot** create manifests of any kind for a Sapio study.
- [x] SequenceScape users **cannot** create a "Sapio study" in SequenceScape.
- [x] SequenceScape will **not** send or create Study data to the MLWH database for Sapio studies.
- [x] SequenceScape will **not** accession Sapio Study data to the EBI.

**Dependencies**
References
- #5827 

**Additional context**
It is possible that some study fields will be editable for Sapio studies in SequenceScape in the future as Sapio studies only have a subset of SequenceScape Study's fields.

**Deployment notes**

wendyyang:
Closes #

-  add a new field in Study table indicate `mastered_in_sapio`
- block editing from SS UI
- add condition in study controller
- add update control in study model, only integration hub or console can update sapio study
- feature flag with Flipper.


StephenHulme:
I have a few general thoughts and suggestions, rather than comments about individual lines of code:

1. Remove references to Sapio from the code. If these are required - like in the banner message - put them in the config file instead. This prevents our code from going out of date if Sapio (or Integration Hub) is renamed.

2. Rename the `mastered_in_sapio` field to `locked` (or similar). This is shorter, and more descriptive while avoiding potentially-biased language.
There is a movement in software to move away from historically problematic and non-inclusive language. Common examples of this are `master` -> `main` and `black-list` -> `deny-list`.

3. Focus on preventing users from altering the study in the UI, instead of putting validations in the study model. Preventing API alterations is not in the story requirements. The workarounds for the Rails Console and Integration Hub requests can then also be removed.

I hope that all makes sense, please let me know if you have any questions or would like to chat on Zoom.

wendyyang:
> I have a few general thoughts and suggestions, rather than comments about individual lines of code:
> 
> 1. Remove references to Sapio from the code. If these are required - like in the banner message - put them in the config file instead. This prevents our code from going out of date if Sapio (or Integration Hub) is renamed.

good points! 


> 2. Rename the `mastered_in_sapio` field to `locked` (or similar). This is shorter, and more descriptive while avoiding potentially-biased language.

May ask the team for suggestions.

>    There is a movement in software to move away from historically problematic and non-inclusive language. Common examples of this are `master` -> `main` and `black-list` -> `deny-list`.
> 3. Focus on preventing users from altering the study in the UI, instead of putting validations in the study model. Preventing API alterations is not in the story requirements. The workarounds for the Rails Console and Integration Hub requests can then also be removed.
> 
This worths discussions. 

The Acceptance criteria state:
SequenceScape users will not be able to edit ANY fields of Sapio study.
SequenceScape users will not be able to change a Sapio study's state e.g. activate it or deactivate it.
SequenceScape users cannot create a "Sapio study" in SequenceScape.

There are multiple entries to do the above actions. Implementing restrictions on controller/UI level may not be sufficient. It is easy to miss some of them. Future development may add new controllers to update study without realise the need of adding the restrictions.  The consequence of that could cause the studies in Sapio and SS are not consistent.  I think Implementing validation on model level provide a single point of enforcement and reduces the risk of missing an update path. 

Any harm of doing model level validations? One thing I can think of is the validations may only apply for certain fields, not block all fields for update.




> I hope that all makes sense, please let me know if you have any questions or would like to chat on Zoom.

app/controllers/admin/studies_controller.rb
```diff
@@ -9,16 +9,28 @@ def index
     @studies = Study.alphabetical
   end
 
+  # rubocop:todo Metrics/MethodLength, Metrics/AbcSize
   def show
     @study = Study.find(params[:id])
+    if Flipper.enabled?(:y26_171_enable_sapio_mastered_study_restrictions) && @study.mastered_in_sapio
+      flash[:error] = I18n.t('studies.managed_in_sapio.warning_message_1')
```
yoldas: nitpick: if it is showing an error for "mastered_in_sapio", the translation key should read "mastered_in_sapio" in it too; not "managed_in_sapio". 

nitpick: consider using a descriptive key name instead of warning_message_1 , e.g. read_only, cannot_be_edited, editing_not_allowed etc.

app/controllers/admin/studies_controller.rb
```diff
@@ -9,16 +9,28 @@ def index
     @studies = Study.alphabetical
   end
 
+  # rubocop:todo Metrics/MethodLength, Metrics/AbcSize
   def show
     @study = Study.find(params[:id])
+    if Flipper.enabled?(:y26_171_enable_sapio_mastered_study_restrictions) && @study.mastered_in_sapio
```
yoldas: nitpick: consider extracting these conditions into a method to avoid repeating, e.g., sapio_study_restricted?

app/controllers/api/v2/concerns/api_key_authenticatable.rb
```diff
@@ -29,7 +29,8 @@ def authenticate_with_api_key
         private
 
         def validate_api_key(api_key)
-          ApiApplication.find_by!(key: api_key)
+          api_application = ApiApplication.find_by!(key: api_key)
+          Current.api_application = api_application
```
yoldas: praise: this is very useful for the request scope. 

note: We might have add an ApiApplication.integration_hub? to avoid checking against strings everywhere: Current.api_application.integration_hub?

app/models/sample_manifest.rb
```diff
@@ -260,4 +261,14 @@ def find_or_create_qc_assay!
     @qc_assay ||= QcAssay.find_or_create_by!(lot_number: "sample_manifest_id:#{id}")
   end
   # rubocop:enable Naming/MemoizedInstanceVariableName
+
+  def prevent_creation_with_mastered_study
+    return unless Flipper.enabled?(:y26_171_enable_sapio_mastered_study_restrictions)
+    return unless study&.mastered_in_sapio?
```
yoldas: note: if we had a single method to cover the feature and flag on Study model, it would be so useful.

app/models/study.rb
```diff
@@ -634,6 +640,41 @@ def data_release_delay_options(assay_option: false)
 
   private
 
+  # This validation only runs when the value of mastered_in_sapio is changing
+  # It prevents changes to mastered_in_sapio unless the request is coming from Integration Hub
+  # i.e. only Integration Hub can set/change the value of mastered_in_sapio
+  def prevent_mastered_in_sapio_changes_unless_integration_hub
+    return unless Flipper.enabled?(:y26_171_enable_sapio_mastered_study_restrictions)
+    return unless will_save_change_to_mastered_in_sapio?
```
yoldas, line 664: issue: I can't find the `will_save_change_to_mastered_in_sapio? ` method.

app/views/studies/information/show.html.erb
```diff
@@ -17,7 +17,7 @@
 <% add :menu, "Contacts" => collaborators_study_path(@study) -%>
 <% add :menu, "History" => summary_study_information_path(@study) -%>
 <% add :menu, "&nbsp;" => '' -%>
-<% if can? :activate, @study %>
+<% if can?(:activate, @study) && !@study.mastered_in_sapio %>
```
yoldas: question: shouldn't this check the feature flag too?

db/migrate/20260612120000_add_mastered_in_sapio_to_studies.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddMasteredInSapioToStudies < ActiveRecord::Migration[8.0]
+  def change
+    add_column :studies, :mastered_in_sapio, :boolean, default: false, null: false
```
yoldas, line 5: praise: thanks for setting the default.

yoldas (commented):
I have added some comments but avoided repeating the same comment everywhere.

- About "SequenceScape will not send or create Study data to the MLWH database for Sapio studies.": broadcast_with_warren macro will send the study to MLWH independent of how it is updated. Something like this might cover after_commit case: Register a handler instead of using the default:

broadcast_with_warren(handler: SapioAwareStudyHandler.new)

its def <<(message) method then either ignores the message or routes it to Warren.handler << message

where message.record would be  a Study.

- Also to cover after_touch case, override the rebroadcast method of Study.

- I can't see how the story criteria item "not accession Sapio Study data to the EBI." is prevented by this PR.

app/models/current.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+# This class is used to store the current API application for the duration of a request,
+# so that it can be accessed in models and other places where the controller context is not available.
+
+class Current < ActiveSupport::CurrentAttributes
+  attribute :api_application
```
StephenHulme: I would very much like to remove the reliance on `CurrentAttributes` if at all possible. This is because it blurs the line between controller and model making maintenance much harder in the future.

app/models/study.rb
```diff
@@ -634,6 +640,41 @@ def data_release_delay_options(assay_option: false)
...
+    return if integration_hub_request?
+
+    errors.add(:base, I18n.t('studies.managed_in_sapio.warning_message_2'))
+  end
+
+  # This validation prevents any updates to a study that is managed in SAPIO
+  # unless the request is coming from Integration Hub
+  def prevent_updates_when_mastered_in_sapio
```
StephenHulme, line 672: Could we change this to `allow updates from the api`? This solves several naming concerns and also removes the need to check against the request coming from the integration hub itself.

app/views/studies/information/show.html.erb
```diff
@@ -51,7 +51,7 @@
 <% add :menu, "QC Reports" => qc_reports_path -%>
 <% add :menu, "Study QC Reports" => study_reports_study_path(@study) -%>
 
-<% if permitted_to_accession?(@study) & @study.samples_accessionable? %>
+<% if permitted_to_accession?(@study) & @study.samples_accessionable? && !@study.mastered_in_sapio %>
```
StephenHulme: I think the feature-flag check could be added to `def samples_accessionable?`, simplifying the code here and propagating to the rest of the accessioning functionality?

spec/controllers/admin/studies_controller_spec.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
```
StephenHulme, line 1: praise: thanks for adding missing tests

StephenHulme (commented):
Lots of work's gone into this and it's looking good overall : ) 
I've added some suggestions around functions and accessioning.

app/controllers/admin/studies_controller.rb
wendyyang: refactored

app/models/study.rb
wendyyang: Will Integration hub be the only user to use SS study API? What about other  applications/users trying to call SS API, to update the study? How to prevent those cases if not checking if the request from Integration hub?

app/controllers/admin/studies_controller.rb
wendyyang: refactored.

wendyyang: refactored.

app/models/sample_manifest.rb
wendyyang: refactored.

app/models/study.rb
wendyyang: will_save_change_to_attribute?  an ActiveRecord dirty-tracking method.
will_save_change_to_mastered_in_sapio? return true if mastered_in_sapio is being changed and the record is being saved, false otherwise.

app/views/studies/information/show.html.erb
wendyyang: updated.

app/views/studies/information/show.html.erb
wendyyang: updated.

app/models/current.rb
wendyyang: refactored

app/models/study.rb
StephenHulme: Most likely, I think. Although there is a case to be made for [principle of least priviledge](https://en.wikipedia.org/wiki/Principle_of_least_privilege), but Sequencescape doesn't currently support [authorisation, compared to authentication](https://auth0.com/docs/get-started/identity-fundamentals/authentication-and-authorization#what-are-authentication-and-authorization).

app/controllers/sdb/sample_manifests_controller.rb
```diff
@@ -41,6 +41,14 @@ def show
   end
 
   def new
+    @study = Study.find_by(id: params[:study_id])
+
+    if @study&.ui_locked?
```
andrewsparkes, line 46: So to be clear, if the Study is now in Sapio the assumption is all Samples will be entered via Sapio too?

andrewsparkes (approved):
Looks really good!

Suggestion: For UAT ask the SSRs to test it, see if they can find any way to update the study or the samples in it that you might have missed!

app/controllers/sdb/sample_manifests_controller.rb
andrewsparkes: And edited there, instead of via sample manifest re-uploads in SS?

BenTopping:
Can this be closed now as outdated?

KatyTaylor:
> Can this be closed now as outdated?

Did it get pulled into an epic branch or something instead?

wendyyang:
> > Can this be closed now as outdated?
> 
> Did it get pulled into an epic branch or something instead?

Yes. It was pulled into Epic/Sapio_study_integration, and already released.

PR #5893, opened 2026-06-30
Y26-170 - [PR] sequencescape sapio api study search

### LINKED ISSUE #5822: Y26-170 - (Query Studies) As faculty when creating a Study in Sapio I want an API endpoint created in SequenceScape to return existing Study details to Sapio so that duplicate Studies are not created.
**User story**
As a faculty member when creating a Study in Sapio, I want an API endpoint created in SequenceScape to return existing Study details to Sapio so that duplicate Studies are not created.

Rich L (Sapio Architect)
Stephen I (Integration Hub)
Katy, Wendy

UAT testing will be conducted in the relevant Sapio project sprint by nominated SMEs.
However, we should agree the API spec with the Integration Hub team and test internally to PSD prior to this.

**Acceptance criteria**

To be considered successful, the solution must allow:
- [x] Add a new Study Resource API endpoint which 
  - [x] will take as a parameter a UTF-8 string containing a study name
  - [x] optionally with wildcard symbols
- [x] The API will return the set of study entities with matching names.
- [x] For now, make all Study fields available for retrieval (unless there's a good reason to not include any). The Integration Hub should be able to limit the fields returned using ['sparse fieldset selection'](https://www.jsonapi.net/usage/reading/sparse-fieldset-selection.html) 
- [x] Uuid must be in the payload sent to Sapio so that it can store it.
- [x] Both active and inactive studies are returned. The active/inactive field is returned in the payload.
- [x] If the number of matching studies exceeds 20, the request should fail and return an appropriate error message indicating that the result set is too large.
- [x] API documentation and Postman collection are updated

**To refine** (see the answers in comments below)
- What happens if the matching set of studies is large, say > 500? I suspect we should truncate - but need to check Sapio behaviour in this case.
- What happens if Sapio times out on this request?
- SequenceScape studies will be extended to know if they are "mastered" in Sapio. Will this future field be included in the study entity payload?

**Additional context**
Sapio will become the source of truth for all **new** studies once it is live (and any transferred to Sapio).
To avoid large data migrations, Sapio will query SequenceScape for any studies with a similar **name**.
This implies that the study name is globally unique and essentially becomes a primary id of the study entity.

See SapioLIMS Confluence page - ADR 011 Study Creation and Accessioning.

yoldas:
Story: https://github.com/sanger/sequencescape/issues/5822

- /api/v2/sapio/studies endpoint
- index action allows search only `filter[name]=<search>`; no default listing
- there are two search algorithms wildcard_name and contains_name
- search can include wildcards * and ? and specify "exact phrases" using quotes
- wildcards outside quotes trigger wildcard matching
- if number of matches is more than 20, returns RESULT_SET_TOO_LARGE error response
- a custom maxResults parameter can be used to specify higher number of results
- study_metadata, data_release_study_type, program, reference_genome, study_type, and user resources under sapio namespace can only be accessed using include
- show action is allowed for a single study by id, /api/v2/sapio/studies/:id


StephenHulme (commented):
A high-level context question (I hope I'm not misunderstanding something!):

In the [JSON:API specification](https://jsonapi.org/recommendations/#urls) URLs map directly to objects, ie: the SS endpoint `/api/v2/samples/` maps to `app/models/sample.rb`.

With the assumption above, when I look at the `/api/v2/sapio/studies` endpoint, I think I might be accessing a Sapio object, which has a relationship to a SS Study?

Is the intention to create a new Rails model type specifically for Sapio use, or to provide Sapio-specific functionality to existing models?
If it's just the functionality, could the existing study* models be extended with the additional behaviours instead of a new endpoint being created?

yoldas:
> A high-level context question (I hope I'm not misunderstanding something!):
> 
> In the [JSON:API specification](https://jsonapi.org/recommendations/#urls) URLs map directly to objects, ie: the SS endpoint `/api/v2/samples/` maps to `app/models/sample.rb`.
> 
> With the assumption above, when I look at the `/api/v2/sapio/studies` endpoint, I think I might be accessing a Sapio object, which has a relationship to a SS Study?
> 
> Is the intention to create a new Rails model type specifically for Sapio use, or to provide Sapio-specific functionality to existing models? If it's just the functionality, could the existing study* models be extended with the additional behaviours instead of a new endpoint being created?

Our BaseResource enforces explicit field exposure as far as I know. The attributes must be explicitly declared on the resources. It enforces what is fetchable, etc.

/api/v2/sapio/studies allows accessing Study in a customised way. It does not access a "Sapio object".

I don't think there is an intention to create a new model for Sapio use. The existing Study model will provide support for Sapio functionality.

The requirements for the new end point look different to me and I have assumed the existing endpoint and the new one will deviate over time. By exposing a new single endpoint and new resources, I have been be able to decouple them. We will be able to put restrictions on inclusion of other resources through the endpoint, fields, and filters as necessary. However, you may see that in some cases, the related resources just inherit the others, but this allow future customisation. In other cases however there are  details which might get in the way, so I subclassed BaseResource.

The following acceptance criteria item starts with "For now".

"For now, make all Study fields available for retrieval (unless there's a good reason to not include any). The Integration Hub should be able to limit the fields returned using ['sparse fieldset selection'](https://www.jsonapi.net/usage/reading/sparse-fieldset-selection.html)"

I have not defined readonly, writeonly, writeonce etc. because we don't know yet.

StephenHulme:
Thanks for the detailed response 👌 I like the forward thinking and allowing most fields to be accessible by default - will make it much easier to make changes during the initial development stages.

>I don't think there is an intention to create a new model for Sapio use. The existing Study model will provide support for Sapio functionality.
The requirements for the new end point look different to me and I have assumed the existing endpoint and the new one will deviate over time.

Thanks for clarifying this. Do we know of any specific changes from the current behaviour in the new requirements? If it is possible to have a single model that satisfies both sapio and existing requirements I think that would be ideal?

Going through the JSONAPI specifications, there doesn't seem to be any direct guidance on this scenario, but it is suggested that [profiles](https://jsonapi.org/format/#profiles) can be used to show different representations of the same objects, and that [there is no formal requirements for URL structure](https://jsonapi.org/faq/#position-uri-structure-custom-endpoints).

app/controllers/api/v2/sapio/studies_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+
+          super
+        end
+
+        private
+
+        # Checks whether the Sapio studies endpoint feature flag is inactive.
+        #
+        # @return [Boolean] true if the +:y26_170_sapio_studies_endpoint+ flag is disabled
+        def feature_flag_disabled?
```
StephenHulme: suggest: rename to be more descriptive and also in the positive form? (maybe sapio_endpoint_enabled?)

app/controllers/api/v2/sapio/studies_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+        #
+        # @return [Boolean] true if the filter[name] parameter is missing
+        def search_param_missing?
+          params.dig(:filter, :name).blank?
+        end
+
+        # Renders a standardized JSON:API error for a disabled feature flag configuration.
+        #
+        # @return [void]
+        def render_feature_flag_disabled
```
StephenHulme, line 54: nitpick: suggest rename to be in positive form and to be consistent with question mark suffix? (maybe rendering_enabled?)

app/controllers/api/v2/sapio/studies_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+        # @return [void]
+        def render_missing_search_param
+          render_errors(Errors::MissingSearchParam.new.errors)
+        end
+
+        # Returns request context for JSONAPI::Resources, which is available
+        # to the filter method in options[:context].
+        #
+        # For the +index+ action,
+        #   - adds the optional +maxresults
```
StephenHulme: typo: possible duplication?

app/controllers/api/v2/sapio/studies_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+        #    strips the quotes by calling CSV.parse_line on it. We need the
+        #    quotes if user is searching exact phrase.
+        #
+        # @note The +maxResults+ parameter uses the JSON:API naming for custom
+        #   parameters.
+        #
+        # @return [Hash] Context passed to JSONAPI::Resources.
+        def context
+          context = super
+          if action_name == 'index'
```
StephenHulme, line 81: nitpick: use a [guard clause](https://en.wikipedia.org/wiki/Guard_(computer_science)) instead of nested-if (suggest: `return context if action_name != 'index'`)

app/resources/api/v2/sapio/data_release_study_type_resource.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+module Api
+  module V2
+    module Sapio
+      # Provides a JSON:API representation of {DataReleaseStudyType}.
+      #
+      # @note Access this resource using include on {Api::V2::Sapio::StudyResource}
+      #  through the `study_metadata.data_release_study_type` relationship.
```
StephenHulme, line 8: praise: thanks for the helpful guiding comments

app/resources/api/v2/sapio/errors.rb
```diff
@@ -0,0 +1,46 @@
...
+    def errors
+      [
+        JSONAPI::Error.new(
+          status: :not_found,
+          title: 'Not Found',
+          code: 'FEATURE_DISABLED',
+          detail: 'This endpoint is not currently available.'
+        )
+      ]
+    end
```
StephenHulme, line 14: question: does this return a 404 not found, or a jsonapi specific response?

app/resources/api/v2/sapio/study_resource.rb
```diff
@@ -0,0 +1,111 @@
...
+      #    &include=study_metadata.reference_genome
+      #    &fields[studies]=name,uuid \
+      #    &fields[study_metadata]=study_description,study_abstract
+      #
+      # @example GET request for a specific study by ID
+      #  GET /api/v2/sapio/studies/123/
+      #
+      # For more information about JSON:API see the [JSON:API Specifications](https://jsonapi.org/format/)
+      # or look at the [JSONAPI::Resources](http://jsonapi-resources.com/) package for Sequencescape's implementation
+      # of the JSON:API standard.
```
StephenHulme, line 39: praise: thank you for the documentation with examples :)

app/resources/api/v2/sapio/study_search_query.rb
```diff
@@ -0,0 +1,136 @@
...
+          #
+          # Any `%`, `_`, or `\` characters inside literal text are escaped so they
+          # are treated as data, not LIKE metacharacters.
+          #
+          # @param records [ActiveRecord::Relation] The base study relation to filter.
+          # @param query [String] The search string, potentially containing quoted
+          #   phrases and wildcard characters.
+          # @return [ActiveRecord::Relation] A relation filtered by the translated
+          #   LIKE pattern.
+          def wildcard_name_scope(records, query)
```
StephenHulme, line 85: praise: thank you for the full details of how the wildcards and literals work

StephenHulme (changes requested):
Some small suggestions, looks good overall and thank you for the documentation.

app/controllers/api/v2/sapio/studies_controller.rb
yoldas: I have done the block structure on purpose, to isolate the block code explicitly into the index action. It looks semantically correct to me.

app/controllers/api/v2/sapio/studies_controller.rb
yoldas: Either that or negative form of if. I just avoided using unless. OK. I will change that.

app/resources/api/v2/sapio/errors.rb
yoldas: It is used for returning a json:api compliant error in an http 404 response.

app/controllers/api/v2/sapio/studies_controller.rb
yoldas: We are already in Sapio specific studies controller. I don't think there is a need to repeat Sapio.

yoldas:
> Going through the JSONAPI specifications, there doesn't seem to be any direct guidance on this scenario, but it is suggested that [profiles](https://jsonapi.org/format/#profiles) can be used to show different representations of the same objects, and that [there is no formal requirements for URL structure](https://jsonapi.org/faq/#position-uri-structure-custom-endpoints).

I don't agree with this; it changes the content-type header and causes extra complexity. It forces us testing yet another feature of the library, we will have to check whether it is properly implemented for the version we use etc. 

They have a dedicated endpoint now and we can customise the study as they want independently.

app/resources/api/v2/sapio/errors.rb
StephenHulme: Perfect, thanks 👌

StephenHulme:
> > Going through the JSONAPI specifications, there doesn't seem to be any direct guidance on this scenario, but it is suggested that [profiles](https://jsonapi.org/format/#profiles) can be used to show different representations of the same objects, and that [there is no formal requirements for URL structure](https://jsonapi.org/faq/#position-uri-structure-custom-endpoints).
> 
> I don't agree with this; it changes the content-type header and causes extra complexity. It forces us testing yet another feature of the library, we will have to check whether it is properly implemented for the version we use etc.
> 
> They have a dedicated endpoint now and we can customise the study as they want independently.

From a practical standpoint I agree that a custom endpoint is the way to go. Thanks for the discussion, appreciate it.

app/resources/api/v2/sapio/study_search_query.rb
```diff
@@ -0,0 +1,136 @@
...
+          # Escapes SQL LIKE wildcard and escape characters in a literal string.
+          #
+          # @param str [String] The string to escape for use in a LIKE pattern.
+          # @return [String] The escaped string.
+          def sql_escape(str)
+            str.gsub(/[%_\\]/) { |char| "\\#{char}" }
+          end
+        end
+
+        class_methods do
```
andrewsparkes, line 115: is it correct to have multiple 'class_methods' blocks each with a single method inside?

andrewsparkes (approved):
Looks good to me. One minor comment.

app/resources/api/v2/sapio/study_search_query.rb
yoldas: It is technically correct; not a good practice though. I could put all of them together and ignore the rubocop. warning about the method length. Multiple class_methods looked OK to me. Do you think I should change it?

yoldas: I think it is better than disabling the rubocop rule; keeping it.

app/controllers/api/v2/sapio/studies_controller.rb
StephenHulme: I feel like `feature_flag_enabled?` tells me less than `Flipper.enabled?(:y26_170_sapio_studies_endpoint)` and that `end_point_enabled?` would be more descriptive - but not a biggy 👍

StephenHulme (approved):
Thanks for the changes, approved 🚀

PR #5897, opened 2026-07-02
Y26 173 create sapio study

### LINKED ISSUE #5825: Y26-173 - (Create Study) As a faculty user when creating a study in Sapio I want a read-only Sapio Study created in SequenceScape so that lab work can be carried out within it.
**User story**
As a faculty user when creating a study in Sapio I want a read-only Sapio Study created in SequenceScape so that lab work can be carried out within it.

TW
RL
Stephen I

PSD and Integration Hub can test this initially.
SME's will test from Sapio in the relevant Sapio sprint.

**Acceptance criteria**

- [ ] An API endpoint for creating Sapio studies in SequenceScape is created.
- [ ] Calling the API endpoint with a valid Sapio study entity creates a Sapio study within SequenceScape.
- [ ] Calling the API endpoint with a new Sapio study with the same name as an existing SequenceScape or Sapio study will fail and return an agreed error code and message.
- [ ] Allow only the request from Integration hub can call the new API endpoint. e.g check if the Authentication key is from Integration hub.
- [ ] No data is sent to the MLWH when SequenceScape creates the Sapio study.
- [ ] No accessioning of the Sapio study takes place in SequenceScape
- [ ] Sequencescape accepts a uuid in the payload and saves the against the Study. It does not create its own uuid. 
- [ ] A new field 'externally_managed' is set to true on Sapio studies created in Sequencescape
- [ ] New 7/8/26: Only minimal data (study name and uuid) is saved against the Sequencescape study*

**Dependencies**
- #5827 
- #5824 

**Additional context**

*We do not need to persist much data against Sapio studies in Sequencescape as they are essentially empty buckets for registering work against. If we need to bypass validation on the Sequencescape Study model we can do so with conditional statements based on the new `externally_managed` field.

**Deployment notes**

wendyyang:
Closes #

add create API endpoint

- 


yoldas (commented):
issue: In order to create a Study using this API endpoint, there needs to be some agreement with the Integration Hub team. Ideally all metadata or any other associated information should be packed into the payload as *_attributes such as study_metadata_atrributes, so the associations can be established in one go by treating those extra nested structures. The study_params used here is not documented. It is not clear what it is doing. See the relationships
- study
  - study_metadata
    - study_type
    - data_release_study_type
    - reference_genome
    - program
  - user

issue: no API documentation and example to show the agreed payload.
issue: no error listing related to study creation
issue: there are no request tests to cover the success and individual error cases

wendyyang:
> issue: In order to create a Study using this API endpoint, there needs to be some agreement with the Integration Hub team. Ideally all metadata or any other associated information should be packed into the payload as *_attributes such as study_metadata_atrributes, so the associations can be established in one go by treating those extra nested structures. The study_params used here is not documented. It is not clear what it is doing. See the relationships
> 
> * study
>   
>   * study_metadata
>     
>     * study_type
>     * data_release_study_type
>     * reference_genome
>     * program
>   * user
> 
> issue: no API documentation and example to show the agreed payload. issue: no error listing related to study creation issue: there are no request tests to cover the success and individual error cases

Yes, the payload is not decided yet. It will need to be updated later on when we know what attributes are passed in. 
For now, I just focus on making sure the 'create' function can only be called by 'Integration Hub', not anywhere else.

yoldas:
issue: tests were not successful.

app/controllers/api/v2/sapio/studies_controller.rb
```diff
@@ -7,12 +7,16 @@ module Sapio
       # This endpoint is feature-flagged and returns a limited result set (max 20 studies).
       class StudiesController < JSONAPI::ResourceController
         include Concerns::ApiKeyAuthenticatable
+        include Api::V2::Concerns::ApiKeyAuthenticatable
```
andrewsparkes, line 10: Are lines 9 and 10 two different things?
Looks like a duplication as they are named the same.

app/controllers/api/v2/sapio/studies_controller.rb
```diff
@@ -32,15 +36,133 @@ def show
...
+            study_abstract: payload[:abstract],
+            data_release_strategy: payload[:data_release_strategy]
+          }
+        end
+
+        # Resolves Sapio name fields to Sequencescape association IDs.
+        def study_metadata_association_ids
+          payload = sapio_study_payload
+          {
+            faculty_sponsor_id: FacultySponsor.find_by(name: payload[:faculty_sponsor])&.id,
```
andrewsparkes, line 116: What happens if one or more of these find_by's fail?
Who is responsible for setting these up in SS before creation of the study in Sapio?

app/models/study.rb
```diff
@@ -143,7 +143,7 @@ class Study < ApplicationRecord # rubocop:todo Metrics/ClassLength
 
   attr_accessor :approval, :run_count, :total_price
 
-  # used to bypass validation when fixing data in study from console by ticket officer.
+  # used to bypass validation
   attr_accessor :bypass_sapio_validation
```
andrewsparkes, line 147: does this mean it's bypassing Sequencescape study validation because it's coming from Sapio?
or it's bypassing Sapio study validation?

maybe (if the first)
'sapio_created_bypass_validation' ?
or
'external_lims_created_bypass_validation' ?

andrewsparkes (commented):
Might be worth talking through these changes after a standup

app/controllers/api/v2/sapio/studies_controller.rb
wendyyang: removed.

app/controllers/api/v2/sapio/studies_controller.rb
wendyyang: Good question! I think we can give a dummy data to let it through the validations if these fields are not important for downstream lab process. Need discussions.

app/models/study.rb
wendyyang: It is a flag, if set to true, then it is allowed to update the study even it is mastered_in_sapio, mainly used by integration hub request via API, to allow it to update  sapio study. Also, it can also be used in console if needed, e.g ticket officer need to fix data for Sapio study for some reason, then set it true before do the data fix.
Rename as skip_mastered_in_sapio_restriction, hope it is more clear.

PR #5907, opened 2026-07-08
[dev] Add CSP report endpoint and generalise policy

StephenHulme:
Closes https://github.com/sanger/sequencescape/pull/5713

- Add nonce tags to Vite resources
- Create a Content Security Policy controller to log CSP violation reports
- Remove specific exceptions from CSP and use better general settings instead
- Generates logs whenever a policy is violated:
  ```log
  [WARN] [csp-report] {"csp-report":{"document-uri":"http://localhost:3000/admin/accessioning_tools","referrer":"http://localhost:3000/admin","violated-directive":"script-src-elem","effective-directive":"script-src-elem","original-policy":"default-src 'self' https:; font-src 'self' https: data:; img-src 'self' https: data:; object-src 'none'; report-uri /csp-reports","disposition":"report","blocked-uri":"data","status-code":200,"script-sample":""}}
  ```
- Logs can be parsed and analysed in Kibana

> [!WARNING]
> The policy `policy.style_src_attr :unsafe_inline` opens a hole in the CSP layer negating many of the benefits that a content security policy brings.
It is strongly recommended that any instances of inline `style` attributes be converted to Bootstrap classes or stylesheets.
Usages can be found using:
>```sh
>grep -rI --include=\*.html\* --exclude=\*.svg 'style=' app/ > style_src_attr.txt  
>```
>For more detail see https://centralcsp.com/docs/style-src-attr.

##### Official Rails Guide
- https://edgeguides.rubyonrails.org/security.html#content-security-policy-header
##### Useful background
- https://upgraderuby.com/articles/improving-frontend-security-strict-csp-rails-8/
##### Specifications
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints
- https://content-security-policy.com/
- https://w3c.github.io/reporting/#intro

### DIFF
--- a/config/initializers/vite_rails_nonce_patch.rb
+++ b/config/initializers/vite_rails_nonce_patch.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+# Automatically add a nonce to the Vite external resource tags for use with the Content Security Policy.
+module ViteRailsNoncePatch
+  def vite_javascript_tag(*names, **options)
+    options[:nonce] = true unless options.key?(:nonce)
+
+    super
+  end
+
+  def vite_stylesheet_tag(*names, **options)
+    options[:nonce] = true unless options.key?(:nonce)
+
+    super
+  end
+end
+
+ViteRails::TagHelpers.prepend(ViteRailsNoncePatch)
--- a/config/initializers/vite_rails_nonce_patch.rb
+++ b/config/initializers/vite_rails_nonce_patch.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 # Automatically add a nonce to the Vite external resource tags for use with the Content Security Policy.
+#
+# The content security policy itself is defined in config/initializers/content_security_policy.rb
 module ViteRailsNoncePatch
   def vite_javascript_tag(*names, **options)
     options[:nonce] = true unless options.key?(:nonce)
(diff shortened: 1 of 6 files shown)


config/initializers/vite_rails_nonce_patch.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# Automatically add a nonce to the Vite external resource tags for use with the Content Security Policy.
+#
+# The content security policy itself is defined in config/initializers/content_security_policy.rb
+module ViteRailsNoncePatch
+  def vite_javascript_tag(*names, **options)
```
BenTopping, line 7: Is the alternative to this adding `nonce: true` to all `vite_javascript_tag`'s?

BenTopping (commented):
Would it make sense to use `config.content_security_policy_report_only = true` report only to start with in case we have uncaught script executions

StephenHulme:
> Would it make sense to use config.content_security_policy_report_only = true report only to start with in case we have uncaught script executions

Correct, this is already set.

config/initializers/vite_rails_nonce_patch.rb
StephenHulme: Correct, and that does work.

The UJS version is also defined in a central location: https://github.com/sanger/sequencescape/blob/ad7b879564fcc2102a124b4892c3cf2c42e45e62/config/initializers/content_security_policy.rb#L22

BenTopping (approved):

BenTopping (approved):

PR #5923, opened 2026-07-14
feat: adds Iso-Seq_cDNA_amp_primer tag group and tags

### LINKED ISSUE #5755: Y26-146 - Support new tags for Kinnex component samples
**User story**

As a user, I want LIMS to support sub-barcodes (dual indexing) for Kinnex samples and expose this data in MLWH, so that I can accurately trace concatenated fragments without relying on manual tracking and reduce data risk.
To enable this, SS needs to support a new tag group for Kinnex compound sample submissions.

Craig C.

This is can be tested by the developer.

**Acceptance criteria**

- [ ] create a new tag group named “Iso-Seq_cDNA_amp_primer” to accommodate the tag sequences defined in the attached file.

[Iso-Seq_v2_Barcoded_cDNA_Primers.txt](https://github.com/user-attachments/files/27405072/Iso-Seq_v2_Barcoded_cDNA_Primers.txt)

**Additional context**
Please refer to #Y26-037

BenTopping:
Closes #5755

- Adds Iso-Seq_cDNA_amp_primer tag group and tags
- Remove 30 length limit for tag oligos.
   - This remains on the manifests so users should still be protected against copy paste errors. 

#### Additional context
The warehouse limits oligos to 30 characters for flowcell tables. We can't limit oligos at the DB level as prod has tags with oligos of length 99. We can keep the restriction to 30 on manifests and remove it from the model level.


app/models/tag.rb
```diff
@@ -25,7 +25,14 @@ def untag!
 
   broadcast_with_warren
 
-  validates :oligo, length: { maximum: 30 }, allow_blank: true
+  # Length enforcement is not a DB constraint because it was added after the fact
+  # and we don't want to break existing tags. However, at the time of writing
+  # the warehouse has a constaint of 30 characters for oligos for some flowcell tables
+  # so we don't want to allow tags longer than that to be created as they will
+  # dead letter on entrance to the warehouse.
+  # That being said we are starting to get valid tags longer than 30 characters
```
andrewsparkes: I think the problem here is we are conflating multiple things that LIMS is not tracking properly. Oligo Tags for sequencing indexing are not the same thing as primers to aid with amplification (which are longer to ensure uniqueness and good attachment), but we are tracking them in the same field in LIMS. Hopefully this gets tackled better in the new LIMS.

andrewsparkes (approved):
Looks fine. Left a comment.

Have you checked nothing needs changing in the sample manifests for the tag columns and their validations? They might have a limit for the excel sheets for when they upload already tagged samples.

app/models/tag.rb
BenTopping: Ah interesting, I wasn't aware of that. Ill update the comment to include something along those lines.

BenTopping:
> Looks fine. Left a comment.
> 
> Have you checked nothing needs changing in the sample manifests for the tag columns and their validations? They might have a limit for the excel sheets for when they upload already tagged samples.

These tags will only be used in new compound sample manifests being created specifically for this kinnex process. That being said the manifests do currently limit to 30 characters, wondering if that should be 60 to reflect the updated model or if its something that should be removed entirely.

andrewsparkes (approved):

PR #5898, opened 2026-07-03
Epic/sapio study integration

wendyyang:
Closes #

#### Changes proposed in this pull request

include the changes from y26-172 & y26-173
add a new field in Study table indicate mastered_in_sapio
block editing from SS UI
add condition in study controller
add update control in study model, only integration hub or console can update sapio study
add create API endpoint
feature flag with Flipper.

#### Instructions for Reviewers

_[All PRs] - Confirm PR template filled_  
_[Feature Branches] - Review code_  
_[Production Merges to `main`]_  
 &nbsp; &nbsp; \- _Check story numbers included_  
 &nbsp; &nbsp; \- _Check for debug code_  
 &nbsp; &nbsp; \- _Check version_

PR #5925, opened 2026-07-15
Y26-174 - Update studies from Sapio API endpoint

### LINKED ISSUE #5826: Y26-174 - (Update Study) As a faculty user I want changes to Study data in Sapio to be reflected in the Studies within SequenceScape so that data is consistent between systems.
**User story**
As a faculty user I want changes to Study data in Sapio to be reflected in the Studies within SequenceScape so that data is consistent between systems.

This story is linked to the [Sapio team Jira story CI-462](https://jira.sanger.ac.uk/browse/CI-462) (permission might be required to view).

TW
RL
Stephen I

PSD and Integration Hub can test this initially.
SME's will test from Sapio in the relevant Sapio sprint.

**Acceptance criteria**

- [x] An API endpoint for editing Sapio studies in SequenceScape is created.
- [ ] ~Changing the study name is possible (see additional context).~
- [ ] ~Attempting to change the study name to one of an existing study record in SS will fail, no changes will be made and a suitable error code and message returned.~
- [x] Allow only the request from Integration hub can call the new API endpoint.  e.g check if the Authentication key is from Integration hub.
- [ ] The payload includes a uuid, which is used to identify the correct Sequencescape study to update.
- [x] No updates should be accepted if the externally-managed field is false. An error should be returned instead.
- [x] New 7/8/26: The only fields on Study that can be updated are:
  - [ ] ~`externally_managed`. We are expecting a call from Sapio (via Integration Hub) to set it to true when it takes control of a Sequencescape Study. See discussion [here](https://github.com/sanger/sequencescape/issues/5825#issuecomment-5215533748).~ Replaced by [Upsert](https://github.com/sanger/sequencescape/issues/5999).
  - [x] `state`. If a Study is inactivated in Sapio, we would like the corresponding Study in Sequencescape to also be inactivated.

**Dependencies**

- #5825 
- #5824 

**Additional context**
It is agreed that Sapio not allow to change the study name. But SS API keep this option available. We also agree to use uuid as Study identifier across systems, but still enforce the study name to be unique.

**Deployment notes**

StephenHulme:
Closes #5826 

See [API v2 Sapio Study documentation](https://sanger.github.io/sequencescape/Api/V2/Sapio/StudyResource.html)

- Confirm SS Study `name`, `uuid`, and `externally managed` is write-once (then read-only)
- Add full resource specs for Sapio Study and Study Metadata resources
- Make the following fields writable:
  - study.state
- Remove redundant check for changes made to `externally_managed` from integration-hub
- Update study resource usage docs and examples


app/resources/api/v2/sapio/study_metadata_resource.rb
```diff
@@ -17,22 +17,26 @@ class StudyMetadataResource < Api::V2::StudyMetadataResource
...
         # @!attribute [r] reference_genome
         #   @return [ReferenceGenomeResource, nil] The reference genome
         #     associated with this study metadata.
-        has_one :reference_genome, class_name: 'ReferenceGenome', foreign_key_on: :self
+        has_one :reference_genome, class_name: 'ReferenceGenome', foreign_key_on: :self, readonly: true
 
-        # @!attribute [r] program
+        # @!attribute [rw] faculty_sponsor
+        #   @return [FacultySponsorResource] The faculty sponsor associated with this study metadata.
+        has_one :faculty_sponsor, foreign_key_on: :related
```
yoldas: `foreign_key_on: :self` because faculty_sponsor_id is on study_metadata.

yoldas: See also the convenience methods (for IDs and names) at the end of the resource. When a related resource has only a single meaningful field, it can be exposed as though it were a regular attribute. For example, faculty_sponsor_name maps directly to the name attribute of the associated FacultySponsor. Similarly, writing with nested attributes.

yoldas (commented):
I suppose there could be some request tests in spec/requests/api/v2/sapio/ .
I have added a couple of comments about the faculty sponsor.

app/resources/api/v2/sapio/study_metadata_resource.rb
StephenHulme: Ah good point - thanks!

StephenHulme: Decided not include any faculty_sponsor changes in this PR in the end

StephenHulme:
Updated after our discussion from this morning - this is ready for review again.

It probably requires #5898 to be merged in before fully meeting the story acceptance criteria.

app/resources/api/v2/sapio/study_resource.rb
```diff
@@ -34,12 +34,12 @@ module Sapio
       # @example GET request for a specific study by ID
       #  GET /api/v2/sapio/studies/123/
       #
+      # TODO: Add update example(s)
```
BenTopping: Outstanding todo

app/resources/api/v2/sapio/study_resource.rb
```diff
@@ -69,42 +69,46 @@ class StudyResource < Api::V2::BaseResource
 
         # @!attribute [r] name
         #   @return [String] The name of the study.
-        attribute :name
+        attribute :name, readonly: true
 
         # @!attribute [r] uuid
         #   @return [String] The UUID of the study.
-        attribute :uuid
+        attribute :uuid, readonly: true
```
BenTopping: Should UUID be `write_once`? I thought this and `name` were settable by the API?

spec/requests/api/v2/sapio/studies_spec.rb
```diff
@@ -628,4 +629,181 @@
...
+          end
+
+          it 'returns a strict, JSON:API specification compliant error document', :aggregate_failures do
+            expect(json).to have_key('errors')
+            expect(json).not_to have_key('data')
+            expect(json['errors']).to be_an(Array)
+            expect(json['errors'].size).to eq(1)
+          end
+
+          it 'serializes the exact error keys with correct details', :aggregate_failures do
```
BenTopping: Suggestion: Can we return the list of valid states? Knowing Json::API resources I suspect this will not be simple so feel free to ignore if thats the case.

app/resources/api/v2/sapio/study_metadata_resource.rb
BenTopping, line 1: Have inthub/sapio asked for all these attributes to be exposed? Since this is Sapio specific I would only expose the ones they have asked for. For example that blocked attribute is probably redundant?

BenTopping (commented):
Nice work 👌 
A few minor comments. The story asks that this be limited to only the Sapio API user, is that still required? Do we need to add a route guard to check the API key.

app/resources/api/v2/sapio/study_resource.rb
StephenHulme: I don't think so - unfortunately this requirement isn't particularly clear so I can't be certain.
I don't see support for changing or validating UUIDs elsewhere in the codebase at the moment, so I'd rather leave it and we can follow up with a dedicated PR later if required.

app/resources/api/v2/sapio/study_metadata_resource.rb
StephenHulme: I think the latest decision is that they don't need to be updated, but they might need to be viewable. Not certain I'm afraid. Now that this class is `immutable` I have reverted all the read-only changes.

app/resources/api/v2/sapio/study_resource.rb
KatyTaylor: When a Study (that doesn't already exist in SS) is imported into SS from Sapio, Sapio will pass a uuid as part of the payload and we need to save that in SS, rather than create our own one. 

> New record is created automatically in Sequencescape, which includes the uuid from Sapio. Sequencescape does not create its own uuid for this record, but saves the one sent by Sapio. Any required fields are populated with dummy data if necessary to pass validation (which cannot be bypassed or removed).

(https://ssg-confluence.internal.sanger.ac.uk/spaces/SLIM/pages/352944216/Study+Integration+Sapio-Sequencescape-MLWH+-+Design)

StephenHulme: Thanks @KatyTaylor! Is this within the scope of the update story? Or should it be a separate one?

StephenHulme: Looks like this would be covered by https://github.com/sanger/sequencescape/issues/5999

KatyTaylor: I thought it would be already covered by @wendyyang 's 'create' story - https://github.com/sanger/sequencescape/issues/5825 - I didn't review it but I remember discussing the requirement with her.

For the update story - they shouldn't be able to change the uuid - but setting it to 'readonly' looks like it would break the 'create' story as they need to be able to set it.

StephenHulme: Interestingly, it doesn't appear as if this setting is respected by the API request tests for the POST case (it is for the GET). I suspect it's because the jsonapi_resource controller has been overridden, but I'll have to dig into it more to figure out and confirm...

spec/requests/api/v2/sapio/studies_spec.rb
```diff
@@ -772,4 +773,187 @@
...
+              'code' => JSONAPI::VALIDATION_ERROR,
+              'status' => '422'
+            )
+            expect(error.dig('source', 'pointer')).to eq('/data/attributes/state')
+          end
+        end
+      end
+
+      context 'when updating externally_managed' do
+        let(:attributes) { { externally_managed: true } }
```
BenTopping: I think we should be calculating this automatically and not relying on the payload from the INT hub.

Assumption: Any request to this endpoint (create/update) should make the study 'externally_managed' as it now exists in Sapio.

StephenHulme: Agreed, that does make a lot of sense.

app/resources/api/v2/sapio/study_resource.rb
StephenHulme: I found the cause: it's due to special handling within the controller - I've moved it to the resource in #6017

spec/requests/api/v2/sapio/studies_spec.rb
StephenHulme: This actually might be possible - certainly there's an InvalidFieldValue that seems like a good error to raise...

app/resources/api/v2/sapio/study_resource.rb
StephenHulme: > Should UUID be write_once? I thought this and name were settable by the API?

Yes, they should be - now changed in 6e365a427fdca2af7e1ffda5012e1a906f46e92b

StephenHulme:
@BenTopping 
> The story asks that this be limited to only the Sapio API user, is that still required? Do we need to add a route guard to check the API key.

Confirmed that we do, this has now been implemented in 9e16648437cf2dab72810bd403561abdc66081e6 and tested in b5777ccae949c849507d550686a189afa4f436b3.

spec/requests/api/v2/sapio/studies_spec.rb
StephenHulme: The actual change was made in #6017, but it is now tested in 384088a952cb7df0fbc622f21aadfb1e2ded1e51

spec/requests/api/v2/sapio/studies_spec.rb
StephenHulme: This would have to be a custom exception - I'm not sure it's worth it. The docs do include the possible states.

app/resources/api/v2/sapio/study_resource.rb
StephenHulme: Thanks for highlighting this 🙈 

Added in df1d0b52de0a5e640d5d7d82f03f0691393f9e22.

StephenHulme:
Thanks for the comments, I've updated the PR (including changes discussed this morning) and it's ready for review again please.

spec/requests/api/v2/sapio/studies_spec.rb
BenTopping: Im not sure I understand this test. The study is already `externally_managed: true` based on the defaults and when we are sent a payload with it in it remains true? Should this be testing when its attempted to be set to false?

spec/resources/api/v2/sapio/study_metadata_resource_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  let(:resource_model) { build_stubbed(:study_metadata) }
+
+  # Mutability
+  it { expect(described_class.mutable?).to be(false) }
+
+  # Model Name
+  it { is_expected.to have_model_name 'Study::Metadata' }
+
+  # Attributes
+  it { is_expected.to have_readwrite_attribute :old_sac_sponsor }
```
BenTopping, line 18: If its immutable shouldn't these all be readonly?

BenTopping (approved):
Looks good to me, couple minor comments, feel free to ignore

spec/requests/api/v2/sapio/studies_spec.rb
StephenHulme: Good catch. This is an old test that no longer makes sense. I've changed it to test the false case in 3c0d61ba79a8842d39c22d6018c0c0277ca63f04 and updated the spec to confirm that nothing changes when a change is attempted in 7e9e93fb1e293be64f856efc2850c3004b271d3b.

spec/resources/api/v2/sapio/study_metadata_resource_spec.rb
StephenHulme: You would think so, but no. Immutability is applied at the class level, but read-only-ness is applied at the attribute level.
This is a JSONAPI-Resources quirk and is consistent with our implementation.

I added the immutability check to the spec to highlight this oddness.

PR #5930, opened 2026-07-17
Y26-195 - Remove redundant UG200 batch descriptors

### LINKED ISSUE #5857: Y26-195 - Modify UG200 sequencing pipeline fields
**User story**

As a member of the sequencing team, I would like the fields on the sequencing pipeline pages adjusted to match what's needed for the Ultima UG200 pipeline, so that I am not asked to fill out irrelevant information.

Tris, Sabrine, Katy

Tris

**Acceptance criteria**

- [ ] Opentrons section contains just:
  - [ ] Pipette carousel
  - [ ] Opentrons Inst. Name
- [ ] The AMP section can be removed.

**Deployment notes**

BenTopping:
Closes #5857

- Remove redundant UG200 descriptors

#### TODO
- Move script to single_use directory instead of migration.✅ 
- Confirm what to do with existing production data ✅ 
    - It doesn't look like UG200 is used yet at time of writing so no existing descriptors in lab events.
- Confirm useq_wafer messaging still works with missing descriptors. ✅ 
    - Tested both with existing descriptors and without.


yoldas (commented):
I think we need to remove the record loader entries as well.

```
ls -1 config/default_records/{descriptors,tasks}/*ultima_ug200*
config/default_records/descriptors/005_ultima_ug200_descriptors.yml
config/default_records/tasks/005_ultima_ug200_tasks.yml
```

BenTopping:
> I think we need to remove the record loader entries as well.
> 
> ```
> ls -1 config/default_records/{descriptors,tasks}/*ultima_ug200*
> config/default_records/descriptors/005_ultima_ug200_descriptors.yml
> config/default_records/tasks/005_ultima_ug200_tasks.yml
> ```
Yes you are right. I had forgotten to commit the changes as was toggling them on/off while testing the rake task 🤦‍♂️ 
Added in [d5d416b](https://github.com/sanger/sequencescape/pull/5930/commits/d5d416b1cb9e52ff3f3c0d38e6cf87ea808bf25e)

yoldas (approved):
Looks good.
- LabEvents part may be necessary for UAT cleanup.
- Thanks for applying the single_use pattern for rake tasks here as well.

lib/tasks/single_use/README.md
sabrine33, line 1: This is great! thank you for applying the single use pattern. 
The changes looks okay to me, just one question, did you check if any data is been registered against the to-be-deleted descriptor in prod?

BenTopping: Yep good question, there are currently no UG200 batches at the time of writing.
If there were, it shouldn't be affected by the removal of these descriptor records as they are not directly linked. LabEvents (where descriptors are stored) create and store the values as hash with the name of the descriptor as the key and the field value as the value, they don't foreign key link to the records - (https://github.com/sanger/sequencescape/blob/develop/lib/acts_as_descriptable.rb#L37)

sabrine33 (approved):

PR #5935, opened 2026-07-20
add data_release_prevention_reason attribute to mlwh

### LINKED ISSUE #5867: Y26-200 - As an Architect (Rich L) I would like the study data export to the MLWH database to include the reason for preventing data release so that there is consistency between SS data and the MLWH.
**User story**
As an Architect (Rich L) I would like the study data export to the MLWH database to include the reason for preventing data release so that there is consistency between SS data and the MLWH.

Rich L
NPG
Tom W

PSD can internally test this.

**Acceptance criteria**
- [x] A new varchar(255) nullable  column is added to the MLWH study table named data_release_prevented_reason
- [ ] The new data_release_prevented_reason has a descriptive COMMENT associated with it, that describes when it is non-blank and what possible values it may take.
- [x] Any new studies or modifications to studies in SequenceScape export the value entered into "What is the reason for **preventing** data release?" on the SequenceScape study page to the data_release_prevented_reason column in MLWH.
- [x] This column is backfilled for existing study data where the data_release_prevented_reason value is not blank.

**Additional context**
This story is a result of Fresh Service ticket #SR-61557.

In the SequenceScape UI for creating studies we have the following two fields of interest:

| UI Field Name | Displayed/Required When | Exported to MLWH study table column |
|------------|-----------------|-------------------------------|
|  Reason for **delaying** release | When "How is data release to be timed" is set to "delayed". Otherwise is blank.  | data_release_delay_reason  |
| What is the reason for **preventing** data release? | When "What is the data release strategy" =  'Not Applicable' AND "How is the data release to be timed" is set to "never". Otherwise is blank. | Is not currently exported |

The fact that "What is the reason for **preventing** data release?" is not exported to the MLWH study table appears to be an oversight - this story is to address this.

Please inform NPG and Rich on the proposed new column.

**Deployment notes**
We will need to deploy a new version of SequenceScape and unified warehouse. In addition we will need to export historical data from SeqeunceScape to the MLWH.

Add any details that should be noted when deploying this feature. For example, is a database migration needed, a specific rake task or other script that needs to be run, a dependency on another application or service that needs to be released first etc?

wendyyang:
Closes #

export data_release_prevention_reason to MLWH

- 


andrewsparkes (approved):
Looks Ok to me.
Rebroadcasting all the studies at once may slow, test this in case you need to do it in smaller batches. Hopefully fine.

lib/tasks/support/rebroadcast_studies_with_data_release_prevention_reason.rake
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+# This is a rake task to rebroadcast studies to backfill data release prevention reason in MLWH.(story Y26-200)
+# To run this task, use the following command:
+# bundle exec rails support:rebroadcast_with_data_release_prevention_reason [ DRY_RUN=true ]
```
StephenHulme: I could be wrong, but I think the environment variable `DRY_RUN=true` would need to be at the start of the `bundle ...` command in order to apply and not at the end?

I made this toy example to test:

File:
```sh
# test.sh
echo "> $TEST"
```

Terminal:
```cli
$ bash test.sh TEST=hi 
$ >
$ TEST=hi bash test.sh
$ > hi
```

StephenHulme (approved):
Small comment, but I think it looks good 👍

StephenHulme:
Weird that the patch coverage seems low, despite having the missing lines covered by a test 🤔

lib/tasks/support/rebroadcast_studies_with_data_release_prevention_reason.rake
wendyyang: Thank you for pointing out this! But:
bundle exec rails support:rebroadcast_with_data_release_prevention_reason DRY_RUN=true
does work though!

AI explains:
Rake has some special handling for arguments that look like:
KEY=value

For Rake tasks, these can be made available through the environment/argument handling, so the task ends up seeing: 
ENV['DRY_RUN']
# => "true"

However, 
DRY_RUN=true bundle exec rails support:rebroadcast_with_data_release_prevention_reason

is less confusing! I have updated the doc.

wendyyang:
> Weird that the patch coverage seems low, despite having the missing lines covered by a test 🤔

That is what confuses me too. But going to ignore that.

lib/tasks/support/rebroadcast_studies_with_data_release_prevention_reason.rake
StephenHulme: Fascinating! I learnt something today, thank you 🙂

StephenHulme:
> > Weird that the patch coverage seems low, despite having the missing lines covered by a test 🤔
> 
> That is what confuses me too. But going to ignore that.

Are studies being correctly created at the right point int he test process? How many studies are printed by `puts "Found #{studies.count} studies"`? The code within `studies.find_each do |study|` does not appear to be covered by coverage, so I wonder if it isn't finding any studies?

It could also be codecov errors? The website report says 27 files changed - but clearly that isn't the case...

Issue #5903, opened 2026-07-07
Y26-209 - Disable MLWH Broadcasts for Sapio-managed Studies


wendyyang:
**User story**
As an architect (Rich L), I want Sequencescape to stop broadcasting update messages for Sapio-managed Studies to MLWH.

**Who are the primary contacts for this story**
Tom, Katy, Wendy

**Who is the nominated tester for UAT**
PSD can test this internally.

**Acceptance criteria**

This functionality will be feature-flagged, turn on when Sapio phase 1 goes live. [No need for this - Sequencescape will never broadcast Sapio studies]

To be considered successful the solution must allow:

- [x] SequenceScape will not send or create Study data to the MLWH database for Sapio studies. [AY: Filtered out Sapio Studies in the PR, however we need to know if associations will be passed by Sapio to Sequencescape for sync. If so, we need a contract similar to mastered_in_sapio? flag probably as delegate].

**Dependencies**

This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
Some discussion is [here](https://github.com/sanger/sequencescape/pull/5870#pullrequestreview-4626558926)

broadcast_with_warren macro will send the study to MLWH independent of how it is updated. Something like this might cover after_commit case: Register a handler instead of using the default:
broadcast_with_warren(handler: SapioAwareStudyHandler.new)

its def <<(message) method then either ignores the message or routes it to Warren.handler << message

where message.record would be a Study.

Also to cover after_touch case, override the rebroadcast method of Study.

**Deployment notes**

Add any details that should be noted when deploying this feature. For example, is a database migration needed, a specifc rake task or other script that needs to be run, a dependency on another application or service that needs to be released first etc?
These notes are intended to ensure the developers releasing the feature know what needs to be done.

yoldas:
Notes:

signature:
`Warren::Callback::broadcast_with_warren(handler: Warren.handler)`

registers:
`after_commit BroadcastWithWarren.new(handler: handler)`

signature:
`BroadcastWithWarren. initialize(handler:, message_class: Warren::Message::Short)`

Warren.handler is a singleton initialised by the gem according to the config file config/warren.yml

```
ubuntu@sss0-prod-jammy:/var/www/sequencescape/current$ bundle exec rails c
Loading production environment (Rails 8.0.5)
[...]

PR #5954, opened 2026-07-30
Update MLWH broadcasts for externally managed

yoldas:
Closes #

This is a patch for already released story https://github.com/sanger/sequencescape/issues/5903 to remove mastered_in_sapio and sapio references. 

- Rename `broadcast_with_warren_except_sapio` to `broadcast_with_warren_except_externally_managed`
- Rename `broadcast_except_sapio` to `broadcast_except_externally_managed`


app/models/study.rb
```diff
@@ -603,7 +603,8 @@ def subject_type
   end
 
   def rebroadcast
-    broadcast_except_sapio
+    # Replace broadcast.
```
StephenHulme: clarification: is this a todo item? Or does replace mean rebroadcast in this context?

StephenHulme (approved):
One small question, but looks good - thanks for the change 👌

app/models/study.rb
yoldas: It is not a to-do. It is highlighting that we using `broadcast_except_externally_managed` instead of `broadcast` coming just after the comment line. I see how it may be confusing. I will update the comment. There is one more comment like this. I will change that too. Thanks.

yoldas:
previous test notes: https://github.com/sanger/sequencescape/issues/5903#issuecomment-4997251715

yoldas:
Test: broadcast_with_warren_except_sapio (after_commit)

```ruby
# Broadcasts if the method does not exist.
study = Study.find_by_name('Y26-209 UAT Test Study')
study.respond_to?(:externally_managed?) # false
study.name += ' u1'
study.save

# Broadcasts if the method exists and returns false.
def study.externally_managed?
  false
end
study.respond_to?(:externally_managed?) # true
study.externally_managed? # false
study.name += ' u2'
study.save

# Does not broadcast if the method exists and returns true.
def study.externally_managed?
  true
end
study.respond_to?(:externally_managed?) # true
study.externally_managed? # true
study.name += ' u3'
study.save
```

Test: (rebroadcast) broadcast_except_sapio (after_touch)

```ruby
# Broadcasts if the method does not exist.
study = Study.find_by_name('Y26-209 UAT Test Study')
study.respond_to?(:externally_managed?) # false
study.touch

# Broadcasts if the method exists and returns false.
def study.externally_managed?
  false
end
study.respond_to?(:externally_managed?) # true
study.externally_managed? # false
study.touch

# Does not broadcast if the method exists and returns true.
def study.externally_managed?
  true
end
study.respond_to?(:externally_managed?) # true
study.externally_managed? # true
study.touch
```

PR #5982, opened 2026-08-12
Y25-263 - Bug - fix to various classes to facilitate loop back to XP plate

andrewsparkes:
To fix bug preventing creation of additional Multiplexing submissions from XP plate.

Closes #5981 

The story description explains how to replicate the original issue.

Essentially you run a pipeline once, then want to go back to the XP plate and create a new Multiplexing and Seq submission.

It was failing because it was using the submission id from the first run (via the outer request) when trying to create transfer requests, instead of filtering to the newer, active submission with it's requests in well.requests_as_source. This causes it to try to re-use the MX pool tube from the first run, and triggers a tag clash error as it tries to put new aliquots (with the same tags) into that previously used tube. 

The fixes in this PR look for that new active submission, and defaults to the current way of looking in outer request if not present.


StephenHulme (approved):
It looks good, a fair bit of complexity there but thank you for the good doc-strings they make it much easier to understand.

app/models/transfer/from_plate_to_tube_by_submission.rb
```diff
@@ -7,16 +7,70 @@ class Transfer::FromPlateToTubeBySubmission < Transfer::BetweenPlateAndTubes
...
+  # Iterates over each well-to-tube transfer, yielding the source well, the destination MX library
+  # tube, and the active submission_id for the corresponding source plate well.
+  # Overrides the default behaviour in Transfer::ControlledDestinations to pass the correct
+  # submission_id to TransferRequest, preventing stale submission IDs being used when a well
+  # has requests from both a completed and a new active submission.
+  def each_transfer
+    well_to_destination.each do |source, destination_and_additional_information|
+      destination, *extra_information = Array(destination_and_additional_information)
+
+      # source here is the parent well for this transfer.
```
KatyTaylor, line 19: The above confused me because if `source` is the parent well, why would you then need to pull `source_wells` from `extra_information`. But I think from reading the `well_to_destination` method, that the above comment is wrong, and `source` is actually the "stock" well, not the source/parent well?

app/models/transfer/from_plate_to_tube_by_submission.rb
```diff
@@ -7,16 +7,70 @@ class Transfer::FromPlateToTubeBySubmission < Transfer::BetweenPlateAndTubes
...
+      # extra_information[0] contains the source_wells array going into the destination tube, where
+      # usually multiple parent plate source wells are pooled into the same tube.
+      source_wells = extra_information.first || []
+      source_well = source_wells.first
+      # We want the active submission chosen to go forward for the source well, if one exists
+      submission_id = active_submission_id_for(source_well || source)
+      yield(source, destination, submission_id)
```
KatyTaylor, line 26: Haven't looked yet at where the above submission_id gets used.

app/models/transfer/from_plate_to_tube_by_submission.rb
```diff
@@ -7,16 +7,70 @@ class Transfer::FromPlateToTubeBySubmission < Transfer::BetweenPlateAndTubes
...
+
+  # Builds a hash mapping each stock well to its destination MX library tube and the array of
+  # source plate wells that trace back to that stock well.
+  # Overrides the parent implementation to pass the source plate well through to
+  # locate_mx_library_tube_for, so the active submission can be resolved from the current
+  # plate rather than from the stock well (which may carry a stale submission reference).
+  def well_to_destination
+    source
+      .stock_wells
+      .each_with_object({}) do |(stock_well, source_wells), store|
```
KatyTaylor, line 39: Confused as to why you've renamed these `stock_well`, `source_wells` arguments, compared to the version in `Transfer::BetweenPlateAndTubes` - they seemed correct before.

If that is wrong, it would suggest a bug, because you're using `source_wells` in the code as if it's source wells not stock wells.

app/models/transfer/from_plate_to_tube_by_submission.rb
```diff
@@ -7,16 +7,70 @@ class Transfer::FromPlateToTubeBySubmission < Transfer::BetweenPlateAndTubes
...
+        next if tube.nil? || should_well_not_be_transferred?(stock_well)
+
+        store[stock_well] = [tube, source_wells]
+      end
+  end
+
+  # Returns the MX library tube that the given well should be transferred into.
+  # Resolves the correct tube by looking up the active submission on the source plate well
+  # (passed as source_well). Falls back to using the stock well itself if no source well is
+  # provided, preserving backwards-compatible behaviour for callers that do not supply one.
```
KatyTaylor, line 53: Wondering about the effectiveness of this backwards compatibility (like the idea though). `source_well` will only be `nil` if `source_wells.first` in the calling method is empty - not sure if that should ever happen? Did you find a particular scenario? Thanks.

KatyTaylor (changes requested):
Sorry for slowness, finding the code quite dense (not very familiar with this area) and getting distracted by other things. I've only looked at the first class in detail, but have a few concerns. Let me know if you want to talk through. Obviously seems to be behaving well in your tests though 🤔 Thanks.

app/models/transfer/from_plate_to_tube_by_submission.rb
andrewsparkes: yielded submission_id is consumed by Transfer.create_transfer_requests callback, to become TransferRequest.submission_id.
It's so the TransferRequest gets the correct submission_id rather than source.pool_id which is the old/previous submission_id from a previous run.

app/models/transfer/from_plate_to_tube_by_submission.rb
andrewsparkes: source is the parent well.

however. source_wells is poorly named, it's the array of ancestor stock wells.

Plate stock_wells gives a hash like:
{
  source_plate_well => [ancestor_stock_well_1, ancestor_stock_well_2]
}

I'll rename and update the comments.

app/models/transfer/from_plate_to_tube_by_submission.rb
andrewsparkes: yep source_wells poorly named, renaming

app/models/transfer/from_plate_to_tube_by_submission.rb
andrewsparkes: good spot, not needed.
think I did that before overriding well_to_destination maybe...

app/models/transfer/from_plate_to_tube_by_submission.rb
```diff
@@ -7,16 +7,67 @@ class Transfer::FromPlateToTubeBySubmission < Transfer::BetweenPlateAndTubes
...
+
+  # Returns the submission_id of the first active (pending or started) request on the well's
+  # requests_as_source. Falls back to the well's first submission_id when no active request
+  # is found, handling the normal single-submission case and the aggregation case where the
+  # outer request is nil.
+  def active_submission_id_for(well)
+    well
+      .requests_as_source
+      .find { |r| %w[pending started].include?(r.state) }
+      &.submission_id || well.submission_ids.first
```
yoldas, line 64: asset_cache walkthrough

```
Well | submission_id | asset_cache
A      | 100                   | miss -> compute, cache
B      | 100                   | hit -> cached
C      | 200                  | miss -> compute, cache
```

For example, `cache[100] = Submission.find(100).multiplexed_labware`

If `active_submission_id_for` returns `nil`,
`cache[nil] = Submission.find(nil).multiplexed_labware` - blows up

This issue was not introduced in this PR. The existing code had this already.

app/models/transfer/from_plate_to_tube_by_submission.rb
```diff
@@ -7,16 +7,67 @@ class Transfer::FromPlateToTubeBySubmission < Transfer::BetweenPlateAndTubes
...
+    well_to_destination.each do |source_plate_well, destination_and_additional_information|
+      destination, *extra_information = Array(destination_and_additional_information)
+
+      # The hash key is the well on the transfer's source plate.
+      # The additional information contains the ancestor stock wells pooled into the destination tube.
+      ancestor_stock_wells = extra_information.first || []
+      submission_well = ancestor_stock_wells.first || source_plate_well
+      # Use the active submission associated with the first ancestor stock well, when available.
+      submission_id = active_submission_id_for(submission_well)
+      yield(source_plate_well, destination, submission_id)
```
yoldas, line 25: note: this is the actual fix, passing a resolved submission_id in yield, allowing creation of transfer requests with an "active" submission id.

app/models/transfer_request.rb
```diff
@@ -172,13 +175,14 @@ def outer_request
   end
 
   # A sibling request is a customer request out of the same asset and in the same submission
-  def sibling_requests # rubocop:todo Metrics/AbcSize
+  def sibling_requests
```
yoldas, line 178: note: this is used from `on_started`  and `outer_request_candidates `

config/default_records/plate_purposes/011_scrna_core_cdna_prep_plate_purposes.yml
```diff
@@ -34,9 +34,10 @@ LRC GEM-X 5p cDNA PCR CITE:
   stock_plate: false
   cherrypickable_target: false
 LRC GEM-X 5p CITE SUP Input:
-  input_plate: true # Passed when submission is built.
-  stock_plate: true # Uploaded via sample manifest.
+  input_plate: true
+  stock_plate: true
   cherrypickable_target: false # Not destination in cherrypicking.
+  default_state: passed # Default state for the input plate so no submission needed
```
yoldas, line 40: note: record loader won't change this attribute for the existing purpose; it is already passed in production. This is matching the production in development environment.

yoldas (approved):
There are a few naming suggestions in other comments. The tests cover the path and the scenario from the ticket (old completed request + new active one). I have highlighted a few hot spots with notes. I think this PR resolves the issue.

app/models/transfer/from_plate_to_tube_by_submission.rb
```diff
@@ -7,16 +7,67 @@ class Transfer::FromPlateToTubeBySubmission < Transfer::BetweenPlateAndTubes
 
   private
 
-  def locate_mx_library_tube_for(well, _stock_wells)
-    asset_cache[well.submission_ids.first]
```
KatyTaylor, line 11: The confusion continues...

Just played around in training env by:
1. finding a 'LB Cap Lib PCR-XP'
2. creating a new multiplexing submission on it
3. Creating the next plate, 'LB Cap Lib Pool'
4. Inspecting the wells & aliquots on that plate

Where `p` is the 'LB Cap Lib Pool' plate...

```
sequencescape(training):026> p.wells.first.submission_ids.count
=> 1
sequencescape(training):027> p.wells.first.submission_ids.first
=> 200111
sequencescape(training):028> Submission.find(200111).created_at
=> 2026-08-28 08:35:24.000000000 BST +01:00
```

Sure enough, this is the wrong submission. But why is it the wrong submission, when this is a new plate?

I think we're 'fixing' the wrong step here. The submission id in the well, and the request id in the aliquots, should refer to the new submission, because when you create the second 'LB Cap Lib Pool' from the 'LB Cap Lib PCR-XP', it should be using the new submission to do that.

andrewsparkes: Not quite sure what you're expecting, the fix is not in Training.
The XP will already have one submission on it.
You created a 2nd, but when you create a new child it uses the 1st submission id in the Transfers, rather than the new, active one.
Lets discuss in morning.

app/models/transfer/from_plate_to_tube_by_submission.rb
andrewsparkes: I can't think how you'd get to a tube pooling step in a pipeline without a submission.
We could add a catch for nil and throw a useful exception message?

PR #5987, opened 2026-08-13
add unique key for external_id in uuid table

### LINKED ISSUE #5975: Y26-227 - Sapio Study - enforce study uuids uniqueness in Sequencescape
The use of uuids to uniquely identify studies (and other entities) relies on the uuids being unique. They should be unique already, because 
a) it's extremely unlikely we would generate the same uuid twice, and 
b) it has been enforced at a database level in MLWH (will be removed by https://github.com/sanger/unified_warehouse/issues/939)
, but we can make sure they are by adding a uniqueness constraint in Sequencescape at the database level.

This requirement came up because we decided to use uuids to link records between systems for the Sapio integration - https://ssg-confluence.internal.sanger.ac.uk/spaces/SLIM/pages/352944216/Study+Integration+Sapio-Sequencescape-MLWH+-+Design

**Acceptance Criteria**
- [x] Study uuids in Sequencescape should be enforced unique. 
- [x] Consider making all uuids unique, regardless of whether they are related to studies or not.

**Additional Context**

Potential point of confusion:
- In the above design, we decided that the MLWH may contain 2 study records with the same uuid in some cases (where a record is transferred from Sequencescape to Sapio)
- However, Sequencescape will still only hold one study record, so uuids should be unique within Sequencescape.

Initially, this story was named "Constraint in SequenceScape for one study uuid per study object." 
I don't think this is as important, as it's already enforced at the model level. The developer can consider whether it's also worth adding as a constraint at the database level.

wendyyang:
Closes #

add db migration file to change external_id column to be unique in uuid table

- 


StephenHulme (approved):
Test in training to make sure there are no conflicts and that it doesn't take too long to run, but looks good 👍

PR #5988, opened 2026-08-13
implement bulk validation and insertion for component samples

### LINKED ISSUE #5595: Y26-068 - Bioscan - Investigate reported slowness in the Traction Export process
**Description**
As Jamie L representing the lab staff I would like the perceived slowness of the Bioscan export process investigated and if possible improved. The lab staff report it can take up to an hour after clicking the export button for the new Tube and sample to appear in Traction.

**Acceptance Criteria**
The following outputs from the research are produced:
- [x] Breakdown of time taken for the export process, using realistic sample numbers
- [x] Identify any slow points and if there are possible improvements
- [x] Implement the fix if time allows (or create separate story)

Jamie L

Andrew, Abdullah, Katy

The process does not take very long when done via the integration suite. Perhaps the volume of samples is an issue.

Process steps as I understand it:

- Click Export button in Limber
- Limber calls Sequencescape api endpoint `/api/v2/bioscan/export_pool_xp_to_traction`
- This creates a job called ExportPoolXpToTractionJob
- The job creates a compound sample
- And creates a RabbitMQ message to send the compound sample and tube details to Traction
- A Traction consumer picks up the message and the tube and compound sample details are imported into Traction

sabrine33:
Closes #

- 


app/models/sample_compound_component.rb
```diff
@@ -24,4 +24,18 @@ def nested_component_samples_validation
...
   end
+
+  # Bulk validation to be called before insert_all the component samples in CompoundSampleHelper#create_compound_sample
+  # Avoids running N queries for 9k samples by doing set-based checks.
+  # No need to check if the compound sample is already a compound sample, as it is created in the same transaction and
+  #   will not have any component samples yet.
+  def self.bulk_validate_component_samples(component_samples)
+    errors = []
+    invalid_component_ids = SampleCompoundComponent.where(compound_sample_id: component_samples.map(&:id))
+      .pluck(:compound_sample_id).uniq
```
BenTopping, line 35: `.distinct` before the pluck instead of `.uniq` might be more performant as it keeps it in the SQL. I haven't tested that though.

BenTopping (commented):
Worth another set of eyes but I think this looks good. How long does the process take now based on training data?

sabrine33:
> Worth another set of eyes but I think this looks good. How long does the process take now based on training data?

I put some info on the story's comments section, this is some performance results 

```
Current implementation:

get_message_data: 1 hour 14 minutes - This includes the check if exists and then creation (the compound sample and population the compound_component sample)
get_message_data: 11.6 seconds - When no creation is required (compound sample exists )

With bulk validation and insertion:

get_message_data: 13.34 seconds
```

app/models/sample_compound_component.rb
sabrine33: > we need to do the uniq on the compound_sample_id as we are expecting multiple component_sample_id have the same compound_sample_id.
I don't think adding distinct will do much in this case - as no indexes on the compound_sample_id and
component_sample_id presents.
(I did not add indexes here as the check/find is not bad the latency is more from the insertions - I put more details on the story comments)

BenTopping (approved):
Wow - Major performance win! Nice work.

PR #5996, opened 2026-08-18
[ux] Accession samples with latest data

StephenHulme:
Closes #5587 where error notifications are sent _after_ the metadata has been fixed.

- Create the SampleAccessioningJobs by reference instead of value, passing the sample and user IDs instead of a payload of serialised data. This means that each sample accessioning attempt will use the latest data from the database.
- Allow `build_accessionable` to be accessed outside of an instance


spec/jobs/sample_accessioning_job_spec.rb
```diff
@@ -254,6 +254,35 @@
...
+        # Run job the first time, which fails due to 2 missing fields
+        perform_and_expect_validation_error(job)
+        expect(Rails.logger).to have_received(:warn)
+          .with(/Sample does not have the required metadata: sample common name and sample taxon/)
+
+        # Populate one of the missing fields
+        sample.sample_metadata.update(sample_common_name: 'Updated common name')
+
+        # Run the job for a second time, which fails due to only one missing field - the metadata was updated
+        job = described_class.new(sample.id, event_user.id) # Create a new job instance to simulate a retry
```
BenTopping, line 280: For a 'retry' does it create a new job like this or re-process the existing job object? Just checking whether we need to make a new job here or re-attempt the existing one

spec/lib/accession_spec.rb
```diff
@@ -12,6 +12,7 @@
 
       it 'does not add an accession number' do
         described_class.accession_sample(sample, event_user)
+        sample.reload
```
BenTopping, line 17: Maybe worth a comment saying reload is needed to properly test the change because it re-fetches the sample using its id instead of using the same cached version of the sample.

spec/jobs/sample_accessioning_job_spec.rb
StephenHulme: It creates a new object in memory, which is reloaded from the database each run. Reusing the existing `job` would mean that the sample would need to be reloaded inside the job for testing purposes.

spec/lib/accession_spec.rb
StephenHulme: Fixed in 7ec08831e62141e8a86d972281b3c828562291de

BenTopping (approved):

PR #6007, opened 2026-08-21
Y26-229 -  Refactor externally managed study validations

### LINKED ISSUE #5980: Y26-229 - Refactor Study Validations for Externally Managed Studies
**User story**

As a developer, I want Study validations to distinguish between existing Sequencescape managed and externally managed studies, so that validations for fields that are not applicable to externally managed studies can be skipped.  When an externally managed study is created, only the name and uuid fields are provided. Therefore, validations for other fields should not apply to externally managed studies.

Wendy, Katy, Stephen

PSD

**Acceptance criteria**

- [ ] Review existing Study validations and identify those that should not apply to externally managed studies.
- [ ] Skip those validations when externally_managed is true.
- [ ] Existing study creation and update behaviour remains unchanged for SS managed studies.
- [ ] Add or update tests covering both SS managed and externally managed studies.

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

**Deployment notes**

BenTopping:
Closes #5980 

- Adds externally_managed? checks to all study validations.
- Removes abstracted associations in favour of explicit associations.

#### Additional context
I preferred in-class validations instead of adding a separate validator module for only externally_managed false studies as it keeps the code in one place. There are quite a few custom_attribute's that have validation that couldn't be extracted to a separate module.


app/models/study.rb
```diff
@@ -793,15 +821,6 @@ def study_type_valid?
       errors.add(:study_type, 'is not specified') if study_type.name == 'Not specified'
     end
 
-    # When a study is mastered in Sapio (externally_managed), all required-field
-    # validation errors are suppressed after validation. Sapio is the source of
-    # truth for these studies and will provide field values over time via updates.
-    after_validation :clear_externally_managed_errors, if: -> { externally_managed? }
```
BenTopping, line 799: I removed this after cleaning up the abstracted validations.

spec/models/study_spec.rb
```diff
@@ -992,69 +992,84 @@
...
-    let(:study) { create(:sapio_study) }
-
-    # test that updates are prevented when the feature flag is enabled and the update is from the UI
-    context 'when feature flag is enabled and updated from the UI', :externally_managed_restrictions_enabled do
-      it 'prevents updates' do
-        study.name = 'New Name'
+  context 'when a study is externally managed' do
+    describe '#validation' do
+      it 'is valid with only the study name' do
+        study = described_class.new(name: 'Externally Managed Study', externally_managed: true)
```
BenTopping, line 998: This should be sufficient to catch future regressions, e.g. someone tries to adds a validation that doesn't filter out externally_managed?

app/models/study.rb
```diff
@@ -156,6 +155,8 @@ class Study < ApplicationRecord # rubocop:todo Metrics/ClassLength
   role_relation(:collaborated_with, 'collaborator')
 
   belongs_to :user
+  # TODO: This is stored at study and study_metadata level, we should remove it from one of them
+  belongs_to :reference_genome
```
StephenHulme, line 159: There's a note in the JSONAPI study resource that the metadata version is preferred. Probably depends on the underlying data in the DB too.

app/models/study.rb
```diff
@@ -669,7 +698,6 @@ def prevent_externally_managed_changes_unless_integration_hub
   # unless the request is coming from Integration Hub
   def prevent_updates_when_externally_managed
     return unless externally_managed_restrictions_enabled?
-    return unless externally_managed?
```
StephenHulme, line 672: suggestion: consider renaming if externally_managed? is no longer a guard.

app/models/study.rb
yoldas: I added the note at https://github.com/sanger/sequencescape/blob/develop/app/resources/api/v2/sapio/study_resource.rb#L20

because there are more reference genomes  set in study_metadata than study.
```sql
with rgs as (
  select s.name as study_name, sr.name as study_reference_genome, smr.name as metadata_reference_genome
  from studies s join study_metadata sm on sm.study_id = s.id
    join reference_genomes sr on s.reference_genome_id = sr.id
    join reference_genomes smr on sm.reference_genome_id = smr.id
)
select 'studies' as thing, count(*) as count from rgs where coalesce(study_name, '') != ''
union select 'study_reference_genomes', count(*) from rgs where coalesce(study_reference_genome, '') != ''
union select 'metadata_reference_genomes', count(*) from rgs where coalesce(metadata_reference_genome, '') != '';
```
```
thing,count
studies,8264
study_reference_genomes,410
metadata_reference_genomes,5773
```

yoldas (approved):

app/models/study.rb
andrewsparkes: Isn't reference genome a required field? From those numbers it looks like 2000+ studies don't have one?
Maybe it's not known in advance for some projects? (ToL? Bioscan?)

app/models/study.rb
BenTopping: Ah good catch that should still be a guard, there is a missing test that should have caught this! Otherwise when externally_managed_restrictions_enabled? is true normal studies will get this error...

I was going to move this check to the validate call of this method but think it better belongs here. I have restored this and added a test [95793cd](https://github.com/sanger/sequencescape/pull/6007/commits/95793cd0609ae1a9533cdd46e706bfb837b68f27)

PR #6043, opened 2026-09-09
Y25-299 - [PR][BUG] Fix inconsistent behaviour when adding columns for bulk submission sheet

### LINKED ISSUE #4900: Y25-299 - [BUG] Fix inconsistent behaviour when adding columns for bulk submission sheet
**Describe the bug**
When updating the bulk submission spreadsheet through our code (e.g., adding a new column), the `is_error` condition is applied inconsistently across the columns in the spreadsheet.

This needs a fix for new updates to the bulk submission sheet.

**RT Ticket Number**
If applicable

**To Reproduce**
1. Add `is_error` condition to one of the columns that currently does not have the condition in its `conditional_formattings` in [`config/bulk_submission_excel/columns.yml`](https://github.com/sanger/sequencescape/blob/develop/config/bulk_submission_excel/columns.yml).
2. Start Sequencescape.
3. Go to "Start new bulk submission" page, select a template and generate the Excel sheet **without filling out any details**[^1].
4. Open the downloaded sheet with Excel and see columns have got red coloured that have no relationship to the column that  `is_error` was added as a conditional formatting.
**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**

- It might be helpful to have a look at the logic behind `is_error` at [`app/sequencescape_excel/sequencescape_excel/formula.rb`](https://github.com/sanger/sequencescape/blob/develop/app/sequencescape_excel/sequencescape_excel/formula.rb#L51).
- Also, it might have to do something with columns having the same condition for validations and conditional formattings. For example, having `:greaterThanOrEqual` for `validation` options and having `number_greater_than_one` for `conditional_formattings`.

[^1]: Note that this occurs only when no information for the labware is filled (i.e., no barcode is given).

yoldas:
Closes #

- Fix the misconfigured range name
- Add test for misconfiguration error


spec/data/bulk_submission_excel/columns.yml
```diff
@@ -256,7 +256,7 @@ requested_flowcell_type:
       error: "You must enter a flowcell type from the list provided."
       promptTitle: "Flowcell Type"
       prompt: "Select a flowcell type from the approved list"
-    range_name: :requested_flowcell_type
```
andrewsparkes, line 259: Can you explain the reason for the name change here? requested_flowcell_type is mentioned in a lot of places in the code base, including in extract_request_options in bulk_submission. is it maybe a filtered list?

yoldas: This is the spec data file in /spec/data/bulk_submission_excel/column.yml and it was not matching the ranges.yml but it was being silently ignored by the tests.

If you look at the actual config file config/bulk_submission_excel/columns.yml , **range_name**  (not column name) value flowcell_type matches the entry in config/bulk_submission_excel/ranges.yml . Otherwise, configured range_name cannot be found.

The alternative is changing the spec data file ranges.yml to read requested_flowcell_type for **range_name** instead but it will not match the actual config values ; we should change the actual config values to read requested_flowcell_type for range_name as well. Do you think we should do it?

andrewsparkes: ah it's the spec file.
the spec config files for columns and ranges should be copies of the real files in my opinion, or ideally be the real files themselves. not sure why it's done this way. Maybe don't worry about it.

andrewsparkes (approved):

PR #6049, opened 2026-09-12
Y26-245 - [PR] 'Upsert' endpoint to create study or update it to be externally managed

### LINKED ISSUE #5999: Y26-245 - 'Upsert' endpoint to create study or update it to be externally managed
**User story**

When a study is created in Sapio, and propagated to mlwarehouse and Sequencescape via the Integration Hub, we would like the Integration Hub to act the same whether the study exists in Sequencescape or not. Therefore, the Integration Hub will just call a Sequencescape 'upsert studies' endpoint, and the Sequencescape will behave differently depending on whether it already knows about the study or not.

Katy, Steve I

Test within PSD, and with the Integration Hub.

**Acceptance criteria**

- When Sapio study creation API endpoint is called:
  - If Sequencescape **does not** already have a study with that uuid:
    - [x] Study is created as per https://github.com/sanger/sequencescape/issues/5825
  - If Sequencescape already has a study with that uuid:
    - [x] Study record is updated so that externally_managed = true
    - [x] Study record is broadcast to mlwarehouse to mark is_current = false

**Additional context**

See Slack conversations:
- https://psd-team.slack.com/archives/C0BDXSBQ4BD/p1786441826131799
- https://psd-team.slack.com/archives/C0BDXSBQ4BD/p1787146745160879 

[Sequence diagram](https://ssg-confluence.internal.sanger.ac.uk/plugins/gliffy/viewer.action?inline=false&pageId=351371301&name=Create%20Study%20in%20Sapio%20Copy&version=2&lastPage=%2Fpages%2Fviewpage.action%3FpageId%3D351371301&imageUrl=%2Fdownload%2Fattachments%2F351371301%2FCreate%2520Study%2520in%2520Sapio%2520Copy.png%3Fversion%3D2%26modificationDate%3D1787059795000%26api%3Dv2&gonUrl=%2Fdownload%2Fattachments%2F351371301%2FCreate%2520Study%2520in%2520Sapio%2520Copy%3Fapi%3Dv2%26version%3D2) - relevant portion copied below:

Image

**Deployment notes**

yoldas:
Closes #

## Summary

Implements Y26-245: `POST /api/v2/sapio/studies` now behaves as an upsert based on the supplied uuid, so the Integration Hub can call the same "create study" endpoint whether or not Sequencescape already knows about the study.

- Adds a new feature flag, `y26_245_sapio_study_upsert`, gating all of the below.
- If the uuid is unknown, behaviour is unchanged (Y26-173's existing create flow).
- If the uuid already belongs to an existing Study, that Study is returned instead of raising a 409 Conflict, and is marked `externally_managed = true` if it isn't already (no-op, no broadcast, if it's already managed - see note on idempotency below).
- On the actual `false -> true` transition, the Study is broadcast to mlwarehouse via the same mechanism `Study#rebroadcast` already uses (`Warren::Message::Full`, immediate/synchronous), bypassing the usual
  `broadcast_except_externally_managed` suppression for externally managed studies.
- `Api::StudyIo` (controls the warehouse message payload for Study) now includes `is_current`, derived as `!externally_managed?`
- When upsert is enabled, a uuid is now required - a missing uuid returns a new `MissingUuid` (400) error rather than silently falling back to an auto-generated uuid, since upsert has nothing to match against without one.
- Updated the endpoint's docs (`Api::V2::Sapio::StudyResource` and the controller's `create` action) to describe the new dual behaviour, including for when the flag is disabled.

## Design notes

- The broadcast only fires on the real `false -> true` transition, not on every repeated call. Once a Study is externally managed, Sequencescape is no longer the authority on its data, so re-broadcasting on every subsequent upsert call risked overwriting data mlwarehouse has since received directly from Sapio.
- A repeated upsert call on an already-managed Study is a no-op success (200, unchanged), not an error - chosen for safety under retries/at-least-once delivery from Integration Hub, rather than rejecting it outright.
- Chose a direct/synchronous broadcast over the `Warren::Message::Short` + `QueueBroadcastConsumer` async pattern . See comparison at https://github.com/sanger/sequencescape/issues/5999#issuecomment-5618984860

## Open questions

- Confirm the client can handle a `200` vs `201` response depending on whether the uuid was new or already known.
- Confirm the `is_current` mapping (`is_current = !externally_managed`) matches what mlwarehouse expect (Test it in UAT)
- Does IH always give a UUID, i.e. should we create a new Study instead of returning a missing uuid error?
- Does the message need a different `id_lims` for broadcasting?

### DIFF
--- a/app/controllers/api/v2/sapio/studies_controller.rb
+++ b/app/controllers/api/v2/sapio/studies_controller.rb
@@ -199,13 +199,33 @@ module Api
         end
 
         # Wraps the existing Study resource so it renders exactly like a
-        # freshly created resource.
+        # freshly created resource, marking it as externally managed if it
+        # is not already.
         #
         # @param id [Integer] the primary key of the existing Study.
         # @return [JSONAPI::ResourceOperationResult] the existing Study resource
         def existing_resource_result(id)
           resource = resource_klass.find_by_key(id, context:)
+          mark_externally_managed(resource._model)
           JSONAPI::ResourceOperationResult.new(:ok, resource)
+        end
+
+        # Marks +study+ as externally managed, unless it already is.
+        #
+        # @note This does not yet broadcast the change to mlwarehouse (see Y26-245).
+        #
+        # @param study [Study] model to mark as externally managed
+        # @return [void]
+        def mark_externally_managed(study)
+          return if study.externally_managed?
+
+          # Bypasses Study#prevent_updates_when_externally_managed, which would otherwise block this
+          # save as soon as externally_managed becomes true below. Api::V2::Sapio::StudyResource's own
+          # before_create/before_update hook sets this for the normal create/update flow, but this
+          # method saves the model directly, so it must be set here too.
+          study.skip_externally_managed_restriction = true
+          study.externally_managed = true
+          study.save!
         end
 
         # Looks up the id of the existing Study with the supplied uuid.
--- a/app/controllers/api/v2/sapio/studies_controller.rb
+++ b/app/controllers/api/v2/sapio/studies_controller.rb
@@ -32,9 +32,11 @@ module Api
           super
         end
 
-        # Creates a new externally managed Study.
-        #
-        # It should NOT be used to transfer an existing Sequencescape study to Sapio.
+        # Creates a new externally managed Study, or, when the
+        # +y26_245_sapio_study_upsert+ feature flag is enabled and the
+        # supplied uuid already belongs to an existing Study, marks that
+        # Study as externally managed instead
+        # (see StudyProcessor#create_resource).
         #
         # @return [void]
         def create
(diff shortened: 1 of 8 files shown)


app/controllers/api/v2/sapio/studies_controller.rb
```diff
@@ -145,28 +147,117 @@ def context
...
+        #   subsequent calls risks overwriting data mlwarehouse has received.
+        #
+        # @param study [Study] model to mark as externally managed
+        # @return [void]
+        def mark_externally_managed(study)
+          return if study.externally_managed?
+
+          # Bypasses Study#prevent_updates_when_externally_managed, which
+          # would otherwise block this save as soon as externally_managed
+          # becomes true below. Api::V2::Sapio::StudyResource's own
```
BenTopping, line 230: Praise: Nice comment here

BenTopping (approved):
Really good work. Great comments explaining the thought process and the various functions.

PR #6053, opened 2026-09-15
Y26-231 - Ultima library conversion submissions

### LINKED ISSUE #5984: Y26-231 - Ultima Library Conversion - bulk submissions
**User story**

As Ultima conversion SSR users we will need to be able to make submissions for Illumina-->Ultima library conversion, plus multiplexing and sequencing.

SSR user Liz C
Mike Q
Andrew S
Katy T
Abdullah Y

SSR user Liz C

**Acceptance criteria**

- [ ] As an SSR, I can make the following submissions on the final plate of any Illumina library prep (submission template and request type names are suggestions only):
  - [ ] `Limber-Htp - Ultima Conv - Ultima UG100 sequencing` (made up of request types: ultima htp conversion + ultima htp multiplexing + ug100 seq)
  - [ ] `Limber-Htp - Ultima Conv - Ultima UG200 sequencing` (made up of request types: ultima htp conversion + ultima htp multiplexing + ug200 seq)
  - [ ] `Limber-Bespoke - Ultima Conv - Ultima UG100 sequencing` (made up of request types: ultima ltp conversion + ultima ltp multiplexing + ug100 seq)
  - [ ] `Limber-Bespoke - Ultima Conv - Ultima UG200 sequencing` (made up of request types: ultima ltp conversion + ultima ltp multiplexing + ug200 seq)
- [ ] In a bulk submission for any of the above, I can set values for a column ‘ultima application’. 
  - [ ] The above column It is a dropdown list, with the values coming from the list of applications created in https://github.com/sanger/sequencescape/issues/5983
  - [ ] The above column is mandatory for the Ultima conversion submission templates
  - [ ] On upload, it errors unless all rows have the same application
- [ ] On upload, it sets the values for the corresponding UGA and UGB primer names for the choice of application, from the data structure created in https://github.com/sanger/sequencescape/issues/5983
  - [ ] The application and the primer data is persisted in the Ultima conversion submission requests in new request metadata fields (for use later in validations of tag plate selection, batch creation and in writing to useq_wafer table (handled in other stories)).

**Dependencies**
None

References
[Y26-230 - Ultima Library Conversion - Persist applications and their applicable primer set](https://github.com/sanger/sequencescape/issues/5983)
Can use placeholder values if that story is not yet completed.

**Additional context**
To make column validation dependant on submission template see the existing scRNA example in method apply_additional_validations_by_template_name in app/models/submission/validations_by_template_name.rb

This will be a manual submission on XP plates from any Illumina library pipeline. Acceptable purposes will include all those XP plate purposes (and the XP Norm plate purpose for HiC).

Example drafts of Submission template and request config can be found in Abdullah's [draft PR](https://github.com/sanger/sequencescape/pull/5960)
NB. do not use this PR as is, it has not been updated to the latest plan for 2 conversion pipelines (manual and high throughout). But it helps as a reference. And the HiC pipeline update will be handled in another story.

And where possible separate the config into WIP'd files so we can control release.

**Deployment notes**
Nothing special

BenTopping:
Closes #5984

- Adds high throughput and bespoke submission templates for ultima conversion.
- Adds ultima application request_metadata relationship to store submitted ultima application

#### Additional notes
- The new submissions fail on building if no ultima_application is specified. This is because the requests fail because they validate it exists. It is done this way for a couple of other submission-template specific fields so I have left it as acceptable. Ideally this would be better done on upload, i.e. checking the field isn't empty but that would be custom validation.


config/default_records/request_types/027_ultima_conversion_request_types.wip.yml
```diff
@@ -0,0 +1,31 @@
+limber_ultima_htp_conversion:
+  name: Limber Ultima HTP Conversion
+  asset_type: Well
+  order: 2
+  request_class_name: UltimaConversionRequest
+  for_multiplexing: false
+  billable: true
+  product_line_name: Ultima
+  acceptable_purposes:
```
BenTopping, line 9: Im not clear on what 'XP plates' need adding here, this is a non-exhaustive list, probably one for UAT / stakeholders.

config/default_records/request_types/027_ultima_conversion_request_types.wip.yml
```diff
@@ -0,0 +1,31 @@
...
+    - LB Lib PCR-XP
+    - LTN Lib PCR XP
+    - RVIG Lib PCR XP
+    - RVI Lib PCR XP
+    - LBB Lib PCR-XP
+    - LBB Lib PCR-XP Norm
```
yoldas, line 31: suggestion: except for the key and name, all fields are repeated. They are likely to stay the same. The whole section can be aliased and reused.

```
limber_ultima_htp_conversion: &limber_ultima_conversion
name: Limber Ultima HTP Conversion
...

limber_ultima_ltp_conversion:
  <<: *limber_ultima_conversion
  name: Limber Ultima LTP Conversion
```

app/models/attributable.rb
```diff
@@ -50,6 +50,11 @@ def field_infos
     attribute_details.map { |detail| detail.to_field_info(self) }
   end
 
+  # Field infos for both custom attributes and associations, used for read-only display
+  def display_field_infos
+    field_infos + association_details.map { |detail| detail.to_field_info(self) }
```
yoldas, line 55: suggestion: `to_field_info` here also builds a list of selection options as well. This will cause loading of every row of the associated table (ultima_application, bait_library, primer_panel, target_purpose) for all requests which then will be discarded to show a plain value. I think to_field_info is designed for editable views. We might avoid full table selects if we do the following:

1. add new method on `Association` something like this to drop the **selection**,

```
def to_display_field_info(*_args)
  FieldInfo.new(display_name: display_name, key: assignable_attribute_name, kind: kind)
end
```

2. change `display_field_infos` here to call that instead:
```
field_infos + association_details.map { |detail| detail.to_display_field_info(self) }
```

View only call from app/views/shared/metadata/show/_request.html.erb (changed in this PR)
Partial rendering in app/views/requests/show.html.erb

app/models/bulk_submission.rb
```diff
@@ -253,7 +253,8 @@ def process # rubocop:todo Metrics/CyclomaticComplexity
     '% phix requested',
     'low diversity',
     'ot recipe',
-    'wafer size'
+    'wafer size',
+    'ultima application'
```
yoldas, line 257: These fields can be validated on upload using the ValidationsByTemplateName approach but there are several of them only one of them is added by PR. Unlike the scRNA Core case, there are many submission template names to handle. I think the validation should be part of another story. I agree with the comment on the PR description.

app/models/ultima_conversion_request.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Used for Illumina to Ultima library conversion.
+class UltimaConversionRequest < CustomerRequest
+  has_metadata as: Request do
+    belongs_to :ultima_application, optional: false
```
yoldas, line 6: note: every request class that calls has_metadata gets its own distinct inner Metadata class. All mapped to the same physical request_metadata table but each is a separate class object with independent validations. So, no other request type's metadata class sees these validators, in this case PresenceValidator. This is OK.

app/models/ultima_conversion_request.rb
```diff
@@ -0,0 +1,12 @@
...
+
+    association(:ultima_application, :name, required: true)
```
yoldas, line 8: note: This creates an Attributable::Association configuration, translates ultima_application_name to a name lookup to associate with the UltimaApplication record. Interestingly, the required parameter looks like setting CSS class 'required' in label_options. This is OK.

app/views/shared/metadata/show/_request.html.erb
```diff
@@ -1,7 +1,7 @@
 <%# NOTE[xxx]: Kind of a hack because we're not actually building a form %>
 <%= fields_for(request) do |form| %>
   <%= form.fields_for(:request_metadata, builder: Metadata::ViewBuilder) do |metadata_fields| %>
-    <%- request.request_metadata.field_infos.each do |field_info| %>
+    <%- request.request_metadata.display_field_infos.each do |field_info| %>
```
yoldas, line 4: note: this partial is the place my comment about `display_field_infos` relevant.

config/bulk_submission_excel/columns.yml
```diff
@@ -386,3 +386,22 @@ wafer_size:
     range_name: :wafer_size
   conditional_formattings:
     empty_cell:
+ultima_application:
```
yoldas, line 389: suggestion: please add this configuration to the test data as well in spec/data/bulk_submission_excel/columns.yml , so it is tested for range and conditional formatting consistency.

config/bulk_submission_excel/ranges.yml
```diff
@@ -32,3 +32,6 @@ wafer_size:
     - "5TB"
     - "10TB"
     - "20TB"
+ultima_application:
```
yoldas, line 35: suggestion: please add this configuration to the test data as well in spec/data/bulk_submission_excel/ranges.yml , so it is tested.

config/default_records/submission_templates/023_ultima_library_ltp_conversion_templates.wip.yml
```diff
@@ -0,0 +1,15 @@
+# These submission templates are used to enter the illumina to ultima library ltp conversion pipeline
```
yoldas, line 1: thanks for separating LTP and HTP submission configurations as wip files.

config/default_records/plate_purposes/005_limber_purposes.yml
```diff
@@ -87,3 +87,14 @@ RVIG Lib PCR XP:
...
+# Final plates for several pipelines including Hi-C
+# Required in sequencescape to enable illumina to ultima library conversion submissions
+LBB Lib PCR-XP:
+  input_plate: false
+  stock_plate: false
+  cherrypickable_target: false
+LBB Lib PCR-XP Norm:
+  input_plate: false
+  stock_plate: false
+  cherrypickable_target: false
```
yoldas, line 100: note: I would put these purpose configurations into a separate file with a new serial number so it can be ripped separately and history of record additions becomes visible and also identifiable with file naming but it is OK.

db/migrate/20260915080340_add_ultima_application_id_to_request_metadata.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddUltimaApplicationIdToRequestMetadata < ActiveRecord::Migration[8.1]
+  def change
+    add_column :request_metadata, :ultima_application_id, :integer
```
yoldas, line 4: nitpick: The SQL type is technically not correct here (integer). We have not specified the data type for the UltimaApplications table ID column, defaulted to Rails 8 bigint. Because add_column requires a type (it can't infer), it needs a bigint for consistency but integer will work too.

config/locales/metadata/en.yml
```diff
@@ -82,6 +82,9 @@ en:
         wafer_size:
           label: Wafer Size
 
+        ultima_application_name:
+          label: Ultima application name
```
yoldas, line 86: suggestion: This label should read `Ultima Application` to have a consistent naming for the UI, matching the column name in the spreadsheet.

yoldas (approved):
Looks good and tested. I have several suggestions but nothing blocking.

sabrine33 (approved):
looks good to me once addressed Abdullah's comments

config/locales/metadata/en.yml
sabrine33: The key is not matching in the request -  it should be  'ultima_application'

sabrine33 (changes requested):

app/models/bulk_submission.rb
BenTopping: Yes it is a bit of an awkward area of submissions. Ideally we would only generate the fields relating to the submission on creation so we don't confuse users with lots of other fields.
I agree that it could be improved for Ultima fields, one for another story.

app/models/ultima_conversion_request.rb
BenTopping: Yes I think it is CSS only, which is slightly redundant here but not harmful.

config/default_records/plate_purposes/005_limber_purposes.yml
BenTopping: I see what you mean but I think this makes sense in limber purposes because it is a Limber only purpose. Hopefully plate purpose created_at timestamps and git history should be sufficient for record tracking.

db/migrate/20260915080340_add_ultima_application_id_to_request_metadata.rb
BenTopping: Good spot, addressed in [ff1912a](https://github.com/sanger/sequencescape/pull/6053/commits/ff1912ae1a950cf8488771a2a22f604d62664ccf)

config/bulk_submission_excel/columns.yml
BenTopping: Ah interesting thanks. I had forgotten about these. Do you know why it can't just use the real values?

Updated in [ff1912a](https://github.com/sanger/sequencescape/pull/6053/commits/ff1912ae1a950cf8488771a2a22f604d62664ccf)

app/models/attributable.rb
BenTopping: Good spot, I have updated in [ff1912a](https://github.com/sanger/sequencescape/pull/6053/commits/ff1912ae1a950cf8488771a2a22f604d62664ccf)

config/default_records/request_types/027_ultima_conversion_request_types.wip.yml
BenTopping: Updated in [ff1912a](https://github.com/sanger/sequencescape/pull/6053/commits/ff1912ae1a950cf8488771a2a22f604d62664ccf)

config/locales/metadata/en.yml
BenTopping: Its a bit convoluted but this needs to be Ultima Application Name because it is the name of the association we set on the request for rendering. This en.yml is just used to grab the label given the attribute: https://github.com/sanger/sequencescape/blob/develop/app/models/attributable/attribute.rb#L134

In this case this is ultima_application_name because of https://github.com/sanger/sequencescape/pull/6053/changes#diff-59bad938a7f40fc3e263f675060bdaafd8c7ed339f9e4a4d5192bb2547600947R8

sabrine33: I don't think it is that smart though, as does not seem to work 

Screenshot 2026-09-23 at 15 43 19

config/bulk_submission_excel/columns.yml
yoldas: Refactoring to use real config from tests would be another piece of work.

config/locales/metadata/en.yml
BenTopping: Thanks, this is from the submissions/new page.
I have updated it in [70a7f9e](https://github.com/sanger/sequencescape/pull/6053/commits/70a7f9e3e1223d0038076ea90f6e54672df0aaf4) to also include a label for this case.

I wonder if I should just store ultima_application_name instead of ID in the request metadata instead so we don't have to convert it across to name in a few places. ID is probably more solid as name doesn't have a uniqueness constraint at the moment...

PR #6062, opened 2026-09-21
Y26-264 - [PR] Register Solaris 2.0 P1 tag group for use in Ultima PCR Free pipelines

### LINKED ISSUE #6050: Y26-264 - Register Solaris 2.0 P1 tag group for use in Ultima PCR Free pipelines
**User story**

As an Operations Manager in Genomics Platforms, I would like a new Ultima tag group/plate registered, which has different adapter sequences (apart from the barcodes/indexes) to the existing 'P1' plate. This is so that we can track which one was used in the lab, to help avoid / troubleshoot mistakes.

Katy, Neil, Tom, Andrew

Can probably test in PSD, but Sara W if needed.

**Acceptance criteria**

- [x] Create tag group, set, tag layout template as described in ticket #SR-74639 *
- [x] Add this tag layout template to the allowed list of tag_layout_templates in Limber pipeline config for `UPF Lib` and `UPF2 Lib` (TBC - perhaps new one only for UG200 and old one only for UG100).
- [x] Map this new tag group to the same plate number and z indices as the existing 'P1' group, in the sample sheet generator (`UltimaSampleSheet::SampleSheetGenerator > ultima_tag_groups_config`) **

**Dependencies**

N/A

**Additional context**

*Andrew and/or Neil already did this once... Katy undid it since it looked like a duplicate. We now understand that the difference between this and the existing 'P1' tag group is not in the oligos in the database (which are purely the barcode/index part of the adapter), but in the other parts of the adapter, e.g. the primers. These parts are not tracked in the database, except through using the tag group 'name'.

**It can just be added to the hash for now. We will create another issue to make this more scalable for the future / allow self service.

**Deployment notes**

yoldas:
Closes #

- Add Solaris P1 tag group and tag layout template records.
- Add tag set records for all Ultima tag groups for completeness and UI support
- Add support for Solaris P1 tag group in sample sheet generation
- Add test to check plate numbers when Solaris P1 and existing P4 are used in the same batch


config/default_records/tag_groups/009_ultima_solaris_p1.yml
KatyTaylor, line 1: Checked the oligos against existing `Ultima P1` default record yml - same apart from name

KatyTaylor: Suggestion: could consolidate existing ultima tag groups into one file, since separation into `005_ultima` and `006_ultima_ug200` is a little misleading now.

config/default_records/tag_sets/005_ultima_tag_sets.yml
KatyTaylor, line 1: Nitpick: Name of file should be more generic as does not just contain 'p1'

KatyTaylor (approved):
Approved with a couple of minor suggestions. Thanks for getting it out so quickly.

config/default_records/tag_sets/005_ultima_tag_sets.yml
andrewsparkes: yeah '005_ultima_tag_sets.yml' maybe

andrewsparkes (approved):
Look fine

config/default_records/tag_groups/009_ultima_solaris_p1.yml
yoldas: Naming reflects the feature and order of record addition I think, except tag sets, which I will fix shortly.

We were asked to enable all Ultima tag groups for both UG100 and UG200 . I thought it would not matter to call it just ultima in the record loader file name.

As for naming of the Solaris P1 tag group itself, I have searched the commercial name but I could not find an exact match. The naming pattern matches both the Fresh Service ticket (except parentheses  and comma before P1) and the existing pattern for P4.

